PR #2, opened 2018-03-04
Transpile CJS and UMD builds

BTMPL:
I believe those two should be transpiled.

- most webpack config disencourage transpiling node_modules content
- the market share of browsers without (proper) support for ES6 is significant (IE, Safari < 10, Opera Mini, PhantomJS)


markerikson:
Yep, just hadn't gotten that far yet.  There's a reason why the initial version numbering was 0.0.x :)

PR #4, opened 2018-03-06
Enable Redux devtools by default (fix #3)

### LINKED ISSUE #3: Enable Redux devtools by default
Unlike a production build of something like React that has much more overhead for developer tooling, Redux devtools has very little overhead especially because it's mostly disabled until the Redux developer tools extension is open. I think it would be best to keep this on by default, but allow users 
 to manually set it to `NODE_ENV !== production` if it's necessary for some reason. 

https://medium.com/@zalmoxis/using-redux-devtools-in-production-4c5b56c5600f

nickmccurdy:
#3


markerikson:
Sweet.

PR #8, opened 2018-03-06
Switch from yarn to npm

nickmccurdy:
Redux recently switch back to npm 5 from yarn, so I'd prefer to be consistent
with their source in case we merge back into the main repository. I'm also
working on add examples based on the existing Redux examples, which are also
using npm lock files.


markerikson:
Eh, I'm completely on the Yarn train myself at this point, but okay.

PR #9, opened 2018-03-06
Pass error/meta to case reducers, support non-FSAs

hnordt:
I would like to suggest that we pass error/meta to case reducers:

```js
function myCaseReducer(state, payload, { error, meta }) {
    // ...
}
```

It might be useful.

Also I've added support to non-FSA.

Building is now failing, I don't know how to add support for rest spread in Rollup. Alternatively I could change the code to use `omit()`:

```js
const {type, payload} = action;
const rest = omit(action, ["type", "payload"])
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,33 +2,36 @@
 
 **A simple set of tools to make using Redux easier**
 
-[NPM: @acemarke/redux-starter-kit](https://www.npmjs.com/package/@acemarke/redux-starter-kit)
+[NPM:  @acemarke/redux-starter-kit](https://www.npmjs.com/package/@acemarke/redux-starter-kit)
 
 ### Purpose
 
 The `redux-starter-kit` package is intended to help address three common complaints about Redux:
 
-* "Configuring a Redux store is too complicated"
-* "I have to add a lot of packages to get Redux to do anything useful"
-* "Redux requires too much boilerplate code"
+- "Configuring a Redux store is too complicated"
+- "I have to add a lot of packages to get Redux to do anything useful"
+- "Redux requires too much boilerplate code"
 
 We can't solve every use case, but in the spirit of [`create-react-app`](https://github.com/facebook/create-react-app) and [`apollo-boost`](https://dev-blog.apollodata.com/zero-config-graphql-state-management-27b1f1b3c2c3), we can try to provide some tools that abstract over the setup process and handle the most common use cases, as well as include some useful utilities that will let the user simplify their application code.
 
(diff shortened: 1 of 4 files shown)


nickmccurdy:
Nice, I like that we're using the FSA package for consistency with other tooling.

> I would like to suggest that we pass error/meta to case reducers:

Does this change actually cause the FSA package to pass error and meta to the reducer or is this something the user would set up manually in action creators?

> Building is now failing, I don't know how to add support for rest spread in Rollup.

Rest spread is supported in Node 8.6.0 (active LTS), so I'd recommend updating Node and not worrying about it. We could tweak babel-preset-env to compile against older versions of Node when we need to, or you can do it if you can't update Node for some reason.

hnordt:
@nickmccurdy the FSA package is just used for checking if the action is a FSA.

I'm passing the error and meta to the case reducer so you can use that info to transform state inside the case reducer:

```js
const addTodoSuccess = todo => ({ type: ADD_TODO, payload: todo })

const addTodoFailure = error => ({ type: ADD_TODO, payload: error, error: true })

// here we don't need to read error or meta, just payload
const addTodoCaseReducer = (state, todo) => state.push({ ...todo, completed : false })

const myErrorManagerCaseReducer = (state, payload, { error }) => error ? state.push(payload) : state
```

The reason is that when `error` is `true`, `payload` could be anything, `new Error()` is just a convention, but might not be followed. Also the user might want to read meta.

nickmccurdy:
That makes sense, would you mind documentation that this convention can be optionally used in the readme?

hnordt:
@nickmccurdy I've updated the `createReducer` section in `README`. Please review it, you might want to improve the wording.

nickmccurdy:
I like your wording and the code examples, but I find the difference between the first two examples a little confusing. They both use FSAs now so maybe they should be merged since there's not much different between them. We should also stay consistent and use immer-style mutating reducers (unlike the FSA docs) because that's unrelated to how we implement reducers with or without FSAs.

hnordt:
@nickmccurdy I've merged first two examples. I removed the returns, I'm not familiar with immer, so you might want to review.

markerikson:
Eh, I'm inclined to say we just switch back to a standard `(state, action)` signature rather than `(state, payload)`.  I guess if people want to grab just the `payload` in their reducer, they can always destructure it like `function myReducer(state, {payload})`

markerikson:
I'll close this PR, but please go ahead and submit one that changes `createReducer` to just pass the whole action instead.

PR #11, opened 2018-03-06
Port Redux's todos example

### LINKED ISSUE #29: Add example
Issue to track #11 (adding at least one of Redux's examples here, adapted to use the starter kit)

nickmccurdy:
Fixes #29.

I wanted to add an example both for documentation and dogfooding. I think todos is a great example because it's already used in many places including the basic Redux docs, this package's readme, reselect's readme, and Redux's automated tests.

I think we should do the async example next since it's also from the Redux docs and it leverages redux-thunk. The more similar we keep this to Redux's original examples, the easier it is for users to compare setup with this package and for us to potentially merge this back into Redux's examples.

- [x] Copy example from Redux to make a potential upstream merge easier
- [x] Use this package instead of using Redux directly
- [x] Use mutating reducers with immer
- [x] ~Use FSA~
- [x] Use selectors
- [x] Keep usage consistent with readme (now it's based more on the original example)


nickmccurdy:
I think this is ready!

spsaucier:
It's unfortunate that this doesn't include any examples of `createSlice` or `createAction`.

markerikson:
Yeah, this PR is way stale.  Thanks for starting it, @nickmccurdy , but we probably oughta just close it and restart from scratch.

Issue #5, opened 2018-03-06
Why immer inside a starter kit?


ematipico:
This is a genuine question. I have been using `redux` for long time and one of the pillars is having pure reducers. And I also know what `immer` does.

As this is a starter kit, so a bulk of code for people that want start from scratch (newbies and pros), what are the reasons for putting `immer` inside the starter package? 

A new comer that is trying to learn `redux` cold have difficulties or be confused. 

Thanks in advance for the feedback!

nickmccurdy:
Good question. One of the reasons we decided to include immer is to address the issue of users having accidental store mutations because they don't know that Redux requires immutable updates, or they don't know how to do immutable updates. We discussed this in more detail in https://github.com/reactjs/redux/issues/2858 and https://github.com/reactjs/redux/issues/2859#issuecomment-369469762. I was honestly against immer at first for similar reasons to yours, but Mark convinced me this is a good idea because it safeguards impure reducers to force them to be pure while still allowing reducers to be pure when desired (so immer's addition should be totally back compatible, there's probably some performance overhead at least). My only concern is that this could teach users starting with only the starter kit that Redux allows immutable updates. I don't think there's a better way to solve this because store mutation checks will not be added to Redux's core (see https://github.com/reactjs/redux/issues/2858), so I'm happy with immer personally. I'm curious if you have other ideas though.

markerikson:
Yeah, my one concern with using Immer is that someone who's not familiar with it might look at reducers and think "oh sure, I can write `state[5].completed = true` in any Redux app and it's fine".  

But, the overall goal of the package is to simplify app setup and cut down on the overall amount of code users need to write.  One of the biggest pain points with Redux is having to write immutable update logic by hand, and Immer is by far the best solution I've seen to that.

nickmccurdy:
I'm also concerned about that. I think when this package is ready we should update the Redux docs to only show how to use Redux with the starter, that way it would be unusual or even "advanced" for a developer to use Redux without the starter kit, similar to how CRA is recommended in React's docs. Then there could be an advanced "usage without starter kit" section that discusses immutable updates in more detail. In other words this package would have the recommended usage of Redux, and Redux core would be a more advanced lower level abstraction for edge cases.

ematipico:
Yeah I admit that using manipulation inside a reducer is a big win especially for people that don't know how to create a pure function. 

I wonder if it would be possible to expose a second function called `createPureReducer` which won't use `immer` under the hoods? 

[...]

PR #13, opened 2018-03-07
docs(readme): add immutable example, add link to immer

Gabri3l:
This PR addresses the discussion at #5 . I noticed in the conversation was used a `.map` to update the state immutably, if you guys prefer that instead of slicing let me know and I will update the PR.


Gabri3l:
Alright let me update that!

ematipico:
Top notch!

PR #15, opened 2018-03-07
Replace createDefaultMiddleware with getDefaultMiddleware

### LINKED ISSUE #14: createDefaultMiddleware Semantics
I would like to suggest a change in the way we support custom middleware because `createDefaultMiddleware`, at least for me, sounds confusing. You are not actually creating a default middleware, you are applying your own middleware while keeping the library defaults.

Instead of:

```js
const middleware = createDefaultMiddleware(logger);
const store = configureStore({ middleware });
```

We could support passing a function to middleware and receiving the default middleware as a param:

```js
const middleware = defaultMiddleware => [defaultMiddleware, logger];
const store = configureStore({ middleware });
```

This also allows the consumer to control the order of defaultMiddleware in the chain, which might be useful for some people.

`defaultMiddleware` would be an object, so in the future we if add more middlewares we could let the user choose the ones he wants:

```js
const middleware = ({ thunk, saga }) => [thunk, saga, logger];
const store = configureStore({ middleware });
```

The implementation for `middleware` argument would be something like:

```js
export function configureStore(options = {}) {
    const {
        reducer,
        devTools = true,
        preloadedState,
        enhancers = [],
    } = options;

    let { middleware } = options;
    middleware = parseMiddleware(middleware);

    // ..
}

const defaultMiddleware = { thunk, saga }

function parseMiddleware(middleware) {
    if (middleware === undefined) {
      // The order for our own defaultMiddleware doesn't matter, so that's not an issue.
      return Object.values(defaultMiddleware)
    }

    if (Array.isArray(middleware)) {
        return middleware
    }

    if (typeof middleware === "function") {
      // Here we need to check for plain objects. The reason is that defaultMiddleware is
      // an object to support destructuring when the consumer wants to select specify
      // middlewares from the defaultMiddleware.
      // But if the consumer uses defaultMiddleware as is, we need to convert it
      // to an array.
      // The order for our own defaultMiddleware doesn't matter, so that's not an issue.
      return middleware(defaultMiddleware)
        .reduce(
          (result, middleware) =>
             isPlainObject(middleware)
               ? [result, ...Object.values(middleware)]
               : [result, middleware],
          []
        )
    }

    throw new Error("`middleware` should be an array or function")
}
```

There is a potential issue where users could think that they would be allowed to pass middleware as objects:

```js
const apiMiddleware = { todoMiddleware, userMiddleware, ... }

  middleware: defaultMiddleware => [defaultMiddleware, apiMiddleware]
})
```

It's not really an issue as long as you don't expect the order of `apiMiddleware` to be maintained. But that's something I didn't like, I don't know how to improve this. Help is welcome.

hnordt:
Closes #14

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ const store = configureStore({ reducer: rootReducer });
 Full example:
 
 ```js
-import {configureStore, createDefaultMiddleware} from "@acemarke/redux-starter-kit";
+import {configureStore, defaultMiddleware} from "@acemarke/redux-starter-kit";
 
 // We'll use redux-logger just as an example of adding another middleware
 import logger from "redux-logger";
@@ -80,7 +80,7 @@ const reducer = {
     visibility : visibilityReducer
 };
 
-const middleware = createDefaultMiddleware(logger);
+const middleware = [...defaultMiddleware, logger];
 
 const preloadedState = {
     todos: [{
@@ -110,9 +110,9 @@ const store = configureStore({
 ```
 
 
-#### `createDefaultMiddleware`
+#### `defaultMiddleware`
 
-Adds redux-thunk to the given array of middlewares. Useful if you need to add custom middlewares without removing redux-thunk.
+`defaultMiddleware` is useful if you need to add custom middlewares without removing redux-starter-kit's default middleware. Currently it is an array that includes only `redux-thunk`.
 
 
 #### `createReducer`
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ const store = configureStore({ reducer: rootReducer });
 Full example:
 
 ```js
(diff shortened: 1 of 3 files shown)


nickmccurdy:
This looks great to me. Based on https://github.com/markerikson/redux-starter-kit/issues/14#issuecomment-371258114 it might be a good idea to either free the array or turn it into a function to prevent accidental mutations of the defaults. What do you think?

hnordt:
@nickmccurdy pushed the changes based on https://github.com/markerikson/redux-starter-kit/issues/14#issuecomment-371258114

nickmccurdy:
@markerikson If you approve of this, could we get this merged and published? It would be a breaking change technically.

PR #16, opened 2018-03-08
Update selectorator to support the new argIndex syntax

nickmccurdy:
Unfortunately even if I `npm update selectorator` after installing `redux-starter-kit`, I still get stuck with version 3.2.0 when the package-lock.json says 3.3.0. It seems like in order to update re-exported dependencies, they have to be updated manually in redux-starter-kit and then the user has to update to the new version. In my opinion we should drop the re-exports in the long run, but this is a temporary workaround so I can use 3.3.0's new features in the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,9 +159,9 @@ For more specifics, see the [`selectorator` usage documentation](https://github.
 ```js
 function createSelector(
     // Can either be:
-    //    - An array containing selector functions and string keypaths
+    //    - An array containing selector functions, string keypaths, and argument objects
     //    - An object whose keys are selector functions and string keypaths
-    paths : Array<Function | string> | Object<string, string | Function>
+    paths : Array<Function | string | Object> | Object<string, string | Function>
 )
 ```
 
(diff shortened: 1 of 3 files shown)


nickmccurdy:
I also added docs since the updated version of selectorator makes it easier to write selectors using props,

markerikson:
Can you recap what steps you followed where you weren't seeing the expected upgrade results?

nickmccurdy:
I did `npm rm @acemarke/redux-starter-kit; npm install @acemarke/redux-starter-kit` to get the latest subdependency versions so I could update selectorator. Even though I saw selectorator version 3.3.0 in my `package-lock.json` diff, importing `createSelector` from `@acemarke/redux-starter-kit` still gave me syntax errors from selectorator version 3.2.0. I then changed `import { createSelector } from '@acemarke/redux-starter-kit'` to `import createSelector from 'selectorator'` and the new selectorator syntax immediately started working.

To be honest I don't understand what caused this, I just wanted to work around this issue so I could show off the new prop selector syntax in the example.

PR #20, opened 2018-03-15
Use Redux's ESLint and Prettier settings

nickmccurdy:
I also formatted the codebase to match Redux. If you don't like this, remove `.prettierrc` to use Prettier's default formatting settings which are more similar to what we have so far (mainly semicolons and double quotes, which Redux avoids).

 Here's the non-formatting part of the PR that just adds config files: https://github.com/markerikson/redux-starter-kit/pull/20/files/c790a3d545fbb4f6f3423bbc86e4e5d220a101be


nickmccurdy:
@markerikson Could you review these settings and let me know if you're comfortable using Redux's or if you'd rather use a more similar Prettier style? Here's the non-formatting part of the PR that just adds config files: https://github.com/markerikson/redux-starter-kit/pull/20/files/c790a3d545fbb4f6f3423bbc86e4e5d220a101be

nickmccurdy:
I'll merge this to keep the code style consistent for newer PRs, but if you'd like to change the code style adjust `.prettierrc` (or remove it to use defaults) and run `npm run format`.

PR #22, opened 2018-03-15
Tweak Rollup

nickmccurdy:


hswolff:
Can you give a high level description of what this is doing? I'm not entirely clear about the intent for the two configs for rollup. How are they intended to be used differently? Also will this require a change to `package.json` as well to point to the right entry point?

nickmccurdy:
Sorry for the confusion, I squashed a bunch of commits together and it got messy.

I made the following changes, mostly inspired by https://github.com/rollup/rollup-starter-lib/blob/babel/rollup.config.js:
- Drop empty external (which are defaults) while explicitly listing externals in the umd bundle to hide the noisy missing module warnings
- Extract the babelrc to another file so we're not duplicating it and so it can be used by other tools (Jest, ESLint, etc.)
- Use the external-helpers as recommended by Rollup
- Tweak package.json to fix missing repository warning and disable sideEffects so bundlers know that this is a helper library with pure functions (it makes tree shaking this package when used as a dependency of another package more effective)
- Combine the second and third configs since Babel recommends using the same config for multiple formats when possible for performance. Not that this reverts part of f3623f96bdad2412146e2b8a8888694e9f7c48dc because Rollup recommends not running dependencies on Babel unless absolutely necessary. We might want to double check this doesn't break anything in case it was necessary, but I'm not completely sure how

hswolff:
SGTM

PR #23, opened 2018-03-16
Add unit tests with Jest

nickmccurdy:


nickmccurdy:
I'll merge this so the tests can be used in other PRs, we can improve them later if necessary.

PR #31, opened 2018-04-22
Remove re-exports

### LINKED ISSUE #19: Consider removing re-exports
This package currently imports and re-exports selectorator without using it internally, while also exporting parts of the redux and immer packages that are used internally. In my opinion maintenance for both this package itself and dependent apps/packages would be easier if we removed our re-exports. This would include removing the dependency on selectorator, though in my opinion it's a very useful package and we should still recommend it in the same way in our readme.

## Advantages
- Less maintenance work involved in updating dependencies, and more freedom for dependent packages to update dependencies without waiting for us to update them or installing newer upstream versions. For example: #16 
- Less package bloat for users that don't want the packages we recommend. For example if someone uses their own memoization library that isn't compatible with reselectorator, they won't have to set up tree shaking to remove the extra dependency if we make it optional.
- Explicit understanding of where dependencies come from. Users would be aware of the packages they're using directly, but not implementation details of this package such as redux-thunk.
- Closer to the intention of package.json files only including dependencies that are required internally (ignoring optionalDependences and peerDependencies which I don't think would be useful in this case).
- Dependencies of dependencies don't satisfy the peer dependencies of other dependencies of a package. For example, if we kept our current behavior and someone installed a new package that peer depends on selectorator, they may have to install selectorator manually just to satisfy the peer dependency. This could cause confusion.

## Disadvantages
- Slightly more installation work. This could be remedied by simply adding other packages to the install script in our readme, for example `npm install @acemarke/redux-starter-kit selectorator`. This is a common practice in documentation for popular npm packages.

nickmccurdy:
Fixes #19, please refer to it for my rationale.

I removed the dependency on selectorator and some re-exports of Immer and Redux functions. The goal is not to discourage their use, but encourage users to install companion packages separate for more flexible versioning and so they're not locked into the versions bundled with this package. I tried to preserve the docs for the formerly re-exported functions, but I'm not sure about the placement within the readme and wording, so I'm open to suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,3 +181,58 @@ const todosReducer = createReducer([], {
   TOGGLE_TODO: toggleTodo
 })
 ```
+
+### Recommended Packages
+
+While `redux-starter-kit` automatically sets up a store for you with recommended middlewares and some additional helper functions, we have recommendations for optional packages that you may find useful in your Redux application code.
+
+#### [`selectorator`](https://github.com/planttheidea/selectorator)
+
+It acts as a superset of the standard `createSelector` function from [Reselect](https://github.com/reactjs/reselect). The primary improvements are the ability to define "input selectors" using string keypaths, or return an object result based on an object of keypaths. It also accepts an object of customization options for more specific use cases.
+
+For more specifics, see the [`selectorator` usage documentation](https://github.com/planttheidea/selectorator#usage).
+
+```js
+function createSelector(
+    // Can either be:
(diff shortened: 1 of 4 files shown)


nickmccurdy:
@xDae I saw you upvote #19, I'd appreciate feedback if you have time

markerikson:
I'm still really not sold on this idea.  I think I can agree with not re-exporting the stuff from Redux, because we expect the user to have that installed already, but I really want a selector library included in here.

Issue #28, opened 2018-04-21
Support Redux 4


nickmccurdy:
Redux 4 is out. The tests already pass using it, ~~but `redux-devtools-extension` needs to update its peer dependency (blocked by https://github.com/zalmoxisus/redux-devtools-extension/pull/490). I'd like wait until this is fixed before adjusting our peer dependencies to reduce confusion and limit peer dependencies seen by users,~~ this issue is just here to remind us to adjust the peer dependency after `redux-devtools-extension` updates.

nickmccurdy:
The devtools extension was updated for Chrome and npm, but not Firefox. Do we want to continue waiting?

nickmccurdy:
@markerikson The devtools extension was updated for Chrome and npm, but not Firefox. Do we want to continue waiting?

nickmccurdy:
True, we've waited long enough, and most devs are on Chrome anyway. Firefox users can always stay on older Redux versions, install the new extension from source, or instantiate the devtools manually.

PR #32, opened 2018-04-24
WIP: TypeScript support

### LINKED ISSUE #30: Typing support
A lot of Redux users use TypeScript or Flow, and with Redux 4 out, I think it would be a good idea to provide optional types for this module. I would personally prefer to start by porting the codebase to TypeScript, but we could use separate definition files if desired. I may give this a shot but I want to wait for feedback and fix #28 first.

nickmccurdy:
Fixes #30, it seems like a lot of extra work to also support Flow so I'm ignoring it for now.

- [x] Fix `finalCompose` type error
- [x] Port sources to TypeScript to strongly type internal values
- [ ] Port tests to TypeScript
- [ ] Use `"strict": true`
- [ ] Declarations for [selectorator](https://github.com/planttheidea/selectorator) (I'm working on an upstream PR)
- [ ] Consider waiting on #28 (Redux 4 will likely require adjusting types)


nickmccurdy:
I underestimated how much work this would be, I'm going to give up on this for now. Hopefully it'll be easier to get back to this in the future after we change our API.

Issue #17, opened 2018-03-08
Consider using packages from  redux-utilities


nickmccurdy:
The [redux-utilities](https://github.com/redux-utilities/flux-standard-action) org has some pretty popular packages, it might make sense to depend on them partially. I haven't really used them that much so please voice your opinions.

## [redux-actions](https://github.com/redux-utilities/redux-actions)
This seems like a cool abstraction to handling actions, similar to the one we already have. Though if it requires FSAs it's probably not a great idea unless we can get them to patch it upstream to support plain actions.

## [reduce-reducers](https://github.com/redux-utilities/reduce-reducers)
This is a handy way to compose reducers together. I don't think we need it yet but it could help us write more complex high order reducers in the future.

## [redux-promise](https://github.com/redux-utilities/redux-promise)
This is one of the more popular async packages along with thunk and saga. I'm fine with redux-thunk, but could there be an advantage to switching to promises or supporting both thunks and promises? Can the thunk package alone support promises and completely replace the need to use redux-promise? I think that's an especially important thing to consider since many async HTTP request libraries use promises now.

nickmccurdy:
Hm, it seems like redux-thunk supports manual promise handling on the return values of action creators, so I'm fine with not adding redux-promise it it provides a lot of syntax sugar and abstraction over errors and FSAs that we don't want to necessarily force users to use.

markerikson:
I've definitely been considering adding something _like_ `redux-actions` to this lib.  I'm not sure whether it's better to DIY that, or pull in the lib.

One reason we might want to DIY it is that the `createReducer` utility we've already got is basically the same as `handleActions()` from `redux-actions`, I think.

For `reduce-reducers`, we could probably just copy it rather than adding a package.  Hardly worth the extra dependency :)

nickmccurdy:
Hmm, I guess `redux-actions` is the only package that would make sense to consider using? It does sound like a good idea, are you aware if it works without FSA?

I was suggesting `reduce-reducers` more in terms of using it internally than re-exporting it, but I guess we probably won't need it and we could always pull it in if/when we do.

markerikson:
Another related example or source of inspiration is https://github.com/christiangenco/createReducerActions , and the article [Namespacing Actions for Redux](https://kickstarter.engineering/namespacing-actions-for-redux-d9b55a88b1b1) is also relevant.

markerikson:
Other similar util libs (there's many of them, just listing a couple):

[...]

PR #37, opened 2018-08-22
initial work on create slice

neurosnap:
Extracted from this comment: https://github.com/markerikson/redux-starter-kit/issues/17#issuecomment-414543588

#17

### DIFF
--- a/src/createSlice.js
+++ b/src/createSlice.js
@@ -1,0 +1,45 @@
+import createNextState from 'immer';
+
+const defaultReducer = (state) => state;
+const getType = (slice, action) =>
+  slice ? `${slice}/${action}` : action;
+
+export default function createSlice({
+  slice = '',
+  actions = {},
+  initialState,
+}) {
+  const actionKeys = Object.keys(actions);
+
+  const reducerMap = actionKeys.reduce(
+    (map, action) => {
+      map[getType(slice, action)] = actions[action];
+      return map;
+    },
+    {},
+  );
+
+  const reducer = (state = initialState, { type, payload }) => {
+    const actionReducer = reducerMap[type] || defaultReducer;
+    const produce = (draft) => actionReducer(draft, payload);
+    return createNextState(state, produce);
+  };
+
+  const actionMap = actionKeys.reduce(
+    (map, action) => {
+      map[action] = (payload) => ({
+        type: getType(slice, action),
+        payload,
+      });
+
+      return map;
+    },
+    {},
+  );
+
+  return {
+    actions: actionMap,
+    reducer,
+    slice,
+  };
+}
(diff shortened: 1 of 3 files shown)


neurosnap:
I updated the PR with the requested changes.  I also added some tests.  Happy to make more changes as recommended.

nickmccurdy:
Thanks. So the goal compared to createReducer is to bundle in actions instead of just returning a merged slice reducers, right? Do you think we could make the API more like createReducer with initialState, actionsMap, and slice as ordered args? Also, why is the slice property in the retuend bundled?

neurosnap:
So to clarify you would like to see: 

```
const counter = createSlice({
  initialState,
  actionsMap,
  slice,
});
```

I added slice as a way to get the slice back as a const if a string literal was used inside the `createSlice` function.  I can remove that.

nickmccurdy:
Ah, so it's just for namespacing actions? That sounds like it could still be useful.

@markerikson What do you think about this API for wrapping createReducer in reducer and action bundles with optional slice based namespacing?

markerikson:
Uh... yes, that was exactly the point in the first place :)  Like `autodux`, but using our own `createReducer` internally.

nickmccurdy:
Cool, I'm in favor of merging as long as we can get the order in https://github.com/markerikson/redux-starter-kit/pull/37#issuecomment-419762825

markerikson:
I think I'd prefer a single "options object" rather than multiple positional arguments for this API.

nickmccurdy:
`createReducer` already has a signature of `(initialState, actionsMap)`, so I'd prefer to stay consistent by either following the order in https://github.com/markerikson/redux-starter-kit/pull/37#issuecomment-419762825 or also converting `createReducer` to an options object signature.

neurosnap:
options object makes sense to me personally, but I don't care either way.

This library also has nothing to do with creating selectors like autodux does, so it is not at perfect parity.  I personally think that auto-generating selectors will really bloat this code and beyond a root selector for each slice I wouldn't add anything more than that.

neurosnap:
Also, it might be useful to write a little helper for creating actions that anyone can use.  Example:

```js
export default function createAction(type) {
  const action = (payload) => ({
    type,
    payload,
  });

  action.toString = () => `${type}`;
  return action;
}
```

This will allow us to stringify the action returned from `createSlice` and is something very similar to what autodux does.  This will allows `createSlice` to work with libraries like `redux-saga`

markerikson:
I'd actually like to clone the "create selectors automatically" functionality if we could.

neurosnap:
Ok, if it's alright with you I would prefer to add that in as a separate PR.  It seems distinct enough to be its own feature.

markerikson:
Superseded by the work done on `robodux` and ported directly to `redux-starter-kit`, as discussed in #17 .  @neurosnap , thanks for writing `robodux` and allowing us to port that over!

PR #38, opened 2018-08-29
Typescript support

### LINKED ISSUE #28: Support Redux 4

### LINKED ISSUE #30: Typing support

nickmccurdy:
I would appreciate code reviews but please do not merge yet.

- [x] Strictly type with TypeSscript 3 (does not support TypeScript 2)
- [x] Use Babel 7 to retain similar build tooling
- [x] Attempt to make createReducer Action types safe (requires complex mapped types and inference)
- [x] Set up declaration publishing
- [ ] Finish typing [selectorator](https://github.com/planttheidea/selectorator)

- Fixes #28
- Fixes #30


nickmccurdy:
I think I finally got the action types in a decent state. 😄 The disadvantage is that in TS files createReducer now requires both its generic arguments (state type and union of action types), but it maps each slice reducer's key to the appropriate action type so slice reducer arguments can be inferred. Usage in JS files should be unaffected, though I'm not 100% sure if I got the Babel 7 config working with external helpers properly.

Declaration files are also added to dist in the install script, so types should be published automatically. The normal JS builds still work without the TS compiler or Babel preset.

I'd appreciate a code review, but I still need to finish selectorator before this can be merged.

nickmccurdy:
@neurosnap By the way thanks for your advice on createReducer types, here's what I have so far: https://github.com/nickmccurdy/redux-starter-kit/blob/typescript/src/createReducer.ts

nickmccurdy:
I'd be happy to update this if we can switch from Selectorator to Reselect

markerikson:
Remind me what the specific difficulties are in adding types to Selectorator?

markerikson:
@nickmccurdy : *ping*

What are the specific problems typing Selectorator?  Is it primarily the multiple-parameters issue?

What would be your thoughts on cloning Selectorator into the lib, and maybe enforcing _only_ passing the input selectors as an array instead of multiple args?

markerikson:
Superseded by #73 .

PR #42, opened 2018-10-18
Provide example of combining createAction with createReducer

doxick:
The power of action creators returned by createAction lies in the `toString()` override when battling boilerplate for action constants.

This provides an example of using an action creator as the key in the `actionsMap` provided to `createReducer`


markerikson:
Nice, thanks!

PR #43, opened 2018-10-18
Correct example code in readme for createSlice

doxick:
The readme has two code examples for `createSlice` which have two minor errors in them.


markerikson:
Yup, left over from me copy-pasting this example from `robodux`, which currently passes just `payload` :)

PR #44, opened 2018-10-22
createAction with payloadCreator argument

doxick:


markerikson:
This PR is stale and obsolete now that we've converted to TS.  I'd appreciate it if you could re-created using current master as a baseline.

markerikson:
Closing this on the grounds that it's stale.  I still like the idea, but it would need to be reimplemented.

PR #45, opened 2018-10-22
createReducer with optional slice parameter

doxick:
Adding the slice as an optional parameter to `createReducer` allows the slice to be defined once (in the reducer) and reused within `combineReducers` and selectors.


markerikson:
This PR is stale and obsolete now that we've converted to TS.  I'd appreciate it if you could re-created using current master as a baseline.

markerikson:
Closing this on the grounds that it's stale.

Issue #67, opened 2019-01-04
Question: Why is null treated as a non-serializable value?


Dudeonyx:
`JSON.stringify(null)` or `JSON.stringify({ value: null })` work perfectly fine

markerikson:
Excellent question, and I'd say that's a bug.  I swiped the code for `isPlain` from a Stack Overflow question, and apparently `null` isn't handled.

Could you submit a PR to add a check for `null` in here?

https://github.com/reduxjs/redux-starter-kit/blob/3f8b779d6c276993514261e5260b1dbf209adea4/src/serializableStateInvariantMiddleware.js#L3-L12

markerikson:
Fixed in v0.3.1 .

PR #69, opened 2019-01-05
Fix: null is no longer flagged as non-serializable

Dudeonyx:
Fixes issue #67 

Adds a strict equality check for null in isPlain


markerikson:
Thanks!  I'll try to get a new version out today.

PR #70, opened 2019-01-06
Basic Typescript typings

denisw:
This PR is a less ambitious alternative to #38. It is not a rewrite in TypeScript, but simply adds a separate type definition file (`index.d.ts`), similar to how TypeScript support is provided in the core Redux repo. Also like in the case of Redux, the typings come with a set of test TypeScript files run through [typings-tester](https://www.npmjs.com/package/typings-tester) to ensure that the typings work as intended.

I am sure that some of the typings could be made even more precise. I'd say, though, that this is a decent first step that we can iterate on later, and it should be definitely better for TypeScript users than not having typings at all.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,7 +1,83 @@
-import { Action, ActionCreator } from 'redux'
+import {
+  Action,
+  ActionCreator,
+  Reducer,
+  AnyAction,
+  Middleware,
+  StoreEnhancer,
+  ReducersMapObject,
+  Store
+} from 'redux'
 
 // The `redux-starter-kit` typings are a superset of the `redux` typings.
-export * from 'redux'
+export {
+  Action,
+  ActionCreator,
+  AnyAction,
+  Middleware,
+  Reducer,
+  ReducersMapObject,
+  Store,
+  StoreEnhancer
+} from 'redux'
+
+/* configureStore() */
+
+/**
+ * A configuration object that can be passed to `configureStore()`.
+ */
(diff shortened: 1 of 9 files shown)


markerikson:
This looks interesting.  But, how much effort would it be to either push #38 forward, or just convert the codebase to TS from scratch?  `createSlice` in particular shouldn't be too hard given that it's really the code from https://github.com/neurosnap/robodux but compiled to JS.  `selectorator` v4 is now written in TS and has typings, so that issue seems to be resolved.

denisw:
I took the definition file approach because it seemed less intrusive, and more it line with what has been done for core Redux. But I'd also be happy to continue the TypeScript rewrite work started in #38 instead. From the perspective of a redux-starter-kit user, there is no perceivable difference; it is a pure manner of what is preferable from a maintainers' point of view. What do you think?

Regarding `createSelector()`, this PR already re-exports selectorator's upstream typings. Same for `createNextState()` from immer's `produce()`. I didn't know `createSlice()`. I didn't know that `createSlice()` came from a TS library, but robodux's `createSlice()` doesn't seem to have an explicit return type defined, so in that regard the typing in this PR is better because it doesn't rely on possibly changing rules of TypeScript's type inference.

denisw:
Alright! I'm going to pick up the TypeScript rewrite work then. As a first step, I'm going to open a PR with an upgrade to Babel 7 so that we can use `@babel/preset-typescript`.

denisw:
I opened the Babel 7 PR as #71.

Closing this PR.

PR #71, opened 2019-01-07
Update to Babel 7

denisw:
This is a pre-requisite to using @babel/preset-typescript later.

This change required upgrading `rollup-plugin-babel`, which in turn required upgrading Rollup.js itself. Fortunately, the upgraded `rollup-plugin-babel` automatically takes care of Babel helpers deduplication and turning off module transpilation in Babel's env preset, so I was able to drastically simplify the Babel configuration.

For the same reason, we don't need `rollup-plugin-replace` anymore.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,18 +1,3 @@
 {
-  "presets": [
-    [
-      "env",
-      {
-        "modules": false
-      }
-    ]
-  ],
-  "env": {
-    "development": {
-      "plugins": ["external-helpers"]
-    },
-    "test": {
-      "plugins": ["transform-es2015-modules-commonjs"]
-    }
-  }
+  "presets": ["@babel/preset-env"]
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -316,15 +316,6 @@
       "integrity": "sha512-0LyEcVlfCoFmci8mXx8A5oIkpkOgyo8dRHtxBnK9RRBwxO2+JZPNsqtVEZQ7mJFPxnXF9lfmU24mHOPI0qnlkA==",
       "dev": true
     },
-    "@babel/plugin-external-helpers": {
-      "version": "7.2.0",
-      "resolved": "https://registry.npmjs.org/@babel/plugin-external-helpers/-/plugin-external-helpers-7.2.0.tgz",
-      "integrity": "sha512-QFmtcCShFkyAsNtdCM3lJPmRe1iB+vPZymlB4LnDIKEBj2yKQLQKtoxXxJ8ePT5fwMl4QGg303p4mB0UsSI2/g==",
-      "dev": true,
-      "requires": {
-        "@babel/helper-plugin-utils": "^7.0.0"
-      }
-    },
     "@babel/plugin-proposal-async-generator-functions": {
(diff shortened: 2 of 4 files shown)


Dudeonyx:
Hi, I'm the guy who offered to help convert rsk to typescript and I'd like to work together with you.
I've already created a typescript version of it in https://github.com/Dudeonyx/redux-ts-starter-kit, with some other internal changes, especially in `createSlice()`,  most of the changes were merged into the original `robodux` last night.

Dudeonyx:
> I see that your fork contains some changes on top of the conversion to TypeScript. I think the best way to go forward here would be to do a "straight" conversion without API or behavior changes first, and then look at the rest (such as the changed return value of `configureStore` or the `selectorator`-`reselect` swap) in separate PR's.

I was thinking the same, let's do the conversion first and in subsequent PR's we can discuss the benefits (or lack of) then.

~But when we're done, I strongly believe we should move away from `selectorator` though because from a practical stand point it isn't type safe, the returned selectors always have a generic `Function` type. No indication of what input it accepts or what output it gives.  At that point you might as well use `any`.~

~Due to the fact that `selectorator` accepts strings as input, This won't ever be remedied unless Typescript as a whole is overhauled.~

Scratch that, we can simply alter the typing so the user can specify an input & output type when creating the selector, along the lines

```js
import { createSelector } from "redux-starter-kit";

 interface State {
  foo: {
  bar: string;
  };
  baz: string;
 }
                            // State is input type, string is output type
const getBarBaz = createSelector<State, string>(["foo.bar", "baz"], (bar, baz) => {
  return `${bar} ${baz}`;
});

// getBarBaz() has signature: (state: State) => string;

const state = {
  foo: {
    bar: "bar"
  },
  baz: "baz"
};

console.log(getBarBaz(state)); // "bar baz"
```

Dudeonyx:
made the PR https://github.com/planttheidea/selectorator/pull/13

Early signs from the maintainer are positive so If it gets merged we can use selectorator else use a custom fork of it

markerikson:
So I just tried checking out this branch, merged `master` over to pick up a couple tweaks to the UMD output, did a build, and compared the build output from `master` and this branch.

The output is similar, but all of the Babel 7 files actually wind up bigger by some amount.  Inspecting the actual output files, it's a variety of small changes.  I'm not sure how much of that is changes in Babel 7, and what might be a change in the Babel config.  For example, string concatenations went from `camelize('get ' + slice)` to `camelize("get ".concat(slice))`

denisw:
> For example, string concatenations went from camelize('get ' + slice) to camelize("get ".concat(slice))

This is apparently a change made for spec compliance. The [docs for @babel/plugin-transform-template-literals](https://babeljs.io/docs/en/babel-plugin-transform-template-literals) explain this:

> When [the `loose` option is] false or not set, all template literal expressions and quasis are combined with `String.prototype.concat`. It will handle cases with `Symbol.toPrimitive` correctly and throw correctly if template literal expression is a `Symbol()`. See [babel/babel#5791](
https://github.com/babel/babel/pull/5791).

markerikson:
Superseded by #73 .

PR #73, opened 2019-01-07
TypeScript rewrite

denisw:
Based on the Babel 7 PR (#71).

The codebase will be ported here incrementally based on #38, #70 and @Dudeonyx's TypeScript fork.

__To do:__

- [x] ~`createAction.js`~
- [x] ~`createReducer.js`~
- [x] ~`configureStore.js`~
- [x] ~`isPlainObject.js`~
- [x] `createSlice.js`
- [x] `index.js`
- [x] ~`serializableStateInvariantMiddleware.js`~
- [x] `sliceSelector.js`
- [x] ~Add type tests using [typings-tester](https://www.npmjs.com/package/typings-tester) (also used in the core Redux repo)~


denisw:
@Dudeonyx I just finished turning everything to TypeScript. I would really appreciate a review by you, especially regarding the interfaces and type arguments used (they are a bit different from the ones you used in your fork).

I also added some [typings-tester](https://www.npmjs.com/package/typings-tester) tests that make it easy to verify that e.g. type argument inference works the way we envision.

denisw:
@markerikson In this PR I added doc comments to all functions so that they show up during auto-completion in Visual Studio Code and other editors. These comments are taken from the current docs, but are modified and extended a bit. I would be happy to backport these improvements to the docs pages (likely in a separate PR) if they look good to you.

tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "compilerOptions": {
```
Retsam: With no `target` specified, the default is `"ES3"`, meaning TS will transpile everything down to ES3 syntax.  It'd probably be better to set the target higher (I think even "TSNext" would be safe) and leave the transpiling to `babel-preset-env`.

denisw: Thanks for the hint! Apparently, the `target` doesn't seem to make any difference when using Babel's TypeScript plugin, though. The build output was exactly the same no matter whether I set `"target": "es2018"` or not.

.babelrc
```diff
@@ -1,18 +1,3 @@
...
-  ],
-  "env": {
-    "development": {
-      "plugins": ["external-helpers"]
-    },
-    "test": {
-      "plugins": ["transform-es2015-modules-commonjs"]
-    }
-  }
+  "presets": ["@babel/preset-env", "@babel/preset-typescript"]
```
resir014: You might also need `@babel/plugin-proposal-class-properties` and `@babel/plugin-proposal-object-rest-spread` if you use these features.

tsconfig.json
resir014: It's still safer to explicitly set the target, IMO. Since we're compiling using Babel, it's safe to set it as `esnext`.

denisw: Done, although I chose `es2018` (as `@babel/preset-env` wouldn't compile down non-standardised features).

.babelrc
denisw: Yeah, I'd prefer not to add them until we need them. We will get a nice error message telling us which plugin to add, should the need arise in the future.

src/createAction.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Returns the action type of the actions created by the passed
+ * `createAction()`-generated action creator (arbitrary action creators
+ * are not supported).
+ *
+ * @param action The action creator whose action type to get.
+ * @returns The action type used by the action creator.
+ */
+export function getType<P>(actionCreator: PayloadActionCreator<P>): string {
```
Kovensky: This `P` is just a secret `any` because it doesn't show up in output positions.

`getType<T extends string>(actionCreator: PayloadActionCreator<T, any>): T`

src/createReducer.ts
```diff
@@ -0,0 +1,56 @@
...
+ * `undefined` state to determine the initial state. Instead, the initial
+ * state is explicitly specified as an argument to `createReducer()`.
+ *
+ * In addition, a case reducer can choose to mutate the passed-in `state`
+ * value directly instead of returning a new state. This does not actually
+ * cause the store state to be mutated directly; instead, thanks to
+ * [immer](https://github.com/mweststrate/immer), the mutations are
+ * translated to copy operations that result in a new state.
+ */
+export type CaseReducer<S = any, A extends Action = AnyAction> = (
```
Kovensky: Are redux _states_ required to be objects, or can they be primitive types too? If they have to be objects, `S extends object`.

src/createReducer.ts
```diff
@@ -0,0 +1,56 @@
...
+ */
+export function createReducer<S = any, A extends Action = AnyAction>(
+  initialState: S,
+  actionsMap: CaseReducersMapObject<S, A>
+): Reducer<S, A> {
+  return function(state = initialState, action): S {
+    return createNextState(state, draft => {
+      const caseReducer = actionsMap[action.type]
+      return caseReducer ? caseReducer(draft as S, action) : draft
+    }) as S
```
Kovensky: Why are these `as S` assertions necessary? Is the type of the `<Draft>` argument being inferred incorrectly in `createNextState`? This can probably be fixed by just adding a return type annotation to the function (`(draft): S | void => { ... }`).

(Also, I just noticed immer's type definitions naively assume class types are named / unique; they are not. Any object with a `_` key is a `Nothing`; they need a `unique symbol` type for that)

src/createSlice.ts
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * The slice's reducer.
+   */
+  reducer: Reducer<S, A>
+
+  /**
+   * Action creators for the types of actions that are handled by the slice
+   * reducer.
+   */
+  actions: { [type in AT]: ActionCreator<A> }
```
Kovensky: After you fix the type of `PayloadAction`:

```ts
type ActionTypes<A extends Action> = A extends { type: infer T } ? T : never
type ActionWithType<A extends Action, T extends string> = Extract<A, { type: T }>
```

```ts
actions: { [T in ActionTypes<A>]: ActionCreator<ActionWithType<A, T>>
```

The `AT` generic argument is unneeded / redundant.

src/createSlice.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  /**
+   * Selectors for the slice reducer state. `createSlice()` inserts a single
+   * selector that returns the entire slice state and whose name is
+   * automatically derived from the slice name (e.g., `getCounter` for a slice
+   * named `counter`).
+   */
+  selectors: { [key: string]: (state: any) => S }
```
Kovensky: Why is this `any` here? Why does this _return_ `S`? Shouldn't `S` be the input?

This is one object that probably should actually be a generic argument, but annoyingly that kills function argument inference.

src/index.ts
```diff
@@ -0,0 +1,23 @@
...
+// types using the `export { ... } from` syntax. Because it compiles
+// modules, independently, it has no way of knowing whether an identifier
+// refers to a type or value, and thus cannot strip the type re-exports
+// out of the generated JS.
+//
+// https://github.com/babel/babel/issues/8361
+//
+// As a workaround, the root of this repository contains an `index.d.ts`
+// that contains all type re-exports. Whenever adding a new public function
+// or type, remember to export it in `index.d.ts` as well.
```
Kovensky: You can reeport them through aliases, but that's kind of a dirty workaround as well, and gets horrible if the type you're reexporting is generic.

```ts
import { FooType } from './foo'
export type FooType = FooType
```

src/isPlainObject.ts
```diff
@@ -2,7 +2,7 @@
  * @param {any} obj The object to inspect.
  * @returns {boolean} True if the argument appears to be a plain object.
  */
-export default function isPlainObject(obj) {
+export default function isPlainObject(obj: any): obj is object {
```
Kovensky: `obj: unknown`

src/serializableStateInvariantMiddleware.ts
```diff
@@ -1,6 +1,14 @@
...
+import { Middleware } from 'redux'
+
+/**
+ * Returns true if the passed value is "plain", i.e. a value that is either
+ * directly JSON-serializable (boolean, number, string, array, plain object)
+ * or `undefined`.
+ *
+ * @param val The value to check.
+ */
+export function isPlain(val: any) {
```
Kovensky: `isPlain(val: unknown): val is any` 😓

With negation types this probably would be `: val is not symbol`, but those are only in TS 3.4 or 3.5.

src/serializableStateInvariantMiddleware.ts
```diff
@@ -25,10 +33,10 @@ const NON_SERIALIZABLE_ACTION_MESSAGE = [
...
 
 export function findNonSerializableValue(
-  obj,
-  path = [],
-  isSerializable = isPlain
-) {
+  obj: any,
+  path: string[] = [],
+  isSerializable: (value: any) => boolean = isPlain
+): any {
```
Kovensky: `obj: unknown`, `path: ReadonlyArray<string> = []`, and you don't need an annotation for `isSerializable` (the correct annotation would be `: typeof isPlain` anyway).

(implementation comment: replace `for..in + hasOwnProperty` with `Object.keys`)

src/serializableStateInvariantMiddleware.ts
```diff
@@ -61,9 +69,28 @@ export function findNonSerializableValue(
...
+/**
+ * Options for `createSerializableStateInvariantMiddleware()`.
+ */
+export interface SerializableStateInvariantMiddlewareOptions {
+  /**
+   * The function to check if a value is considered serializable. This
+   * function is applied recursively to every value contained in the
+   * state. Defaults to `isPlain()`.
+   */
+  isSerializable?: (value: any) => boolean
```
Kovensky: `isSerializable?: typeof isPlain`

src/sliceSelector.ts
```diff
@@ -0,0 +1,30 @@
+export type Selector<S, R> = (state: S) => R
+
+export function createSliceSelector<S = any>(): Selector<S, S>
+export function createSliceSelector<
+  S extends { [key: string]: any } = any,
+  R = any
+>(slice: string): Selector<S, R>
```
Kovensky: what is `R`?

tsconfig.json
```diff
@@ -0,0 +1,10 @@
...
+    "allowSyntheticDefaultImports": true,
+    "lib": ["es2015"],
```
Kovensky: This `lib` is already provided by the `target` so you shouldn't really need it.

tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
```
Kovensky: Also `esModuleInterop: true`

denisw:
@Kovensky Thank you a lot for the very thorough review! I'm still learning capture and maintain as much type information as practical, so your input is extremely appreciated.

I'll comment on your points one-by-one. On a general note, this PR is deliberately about a one-to-one translation to TypeScript without any changes to the API design as such (which surely could be made more typing-friendly at some places). Improving the API for type-ability is then the next step after this has been merged.

package.json
```diff
@@ -9,10 +9,15 @@
   "author": "Mark Erikson <mark@isquaredsoftware.com>",
   "license": "MIT",
   "devDependencies": {
-    "babel-core": "^6.26.0",
-    "babel-eslint": "^7.2.3",
-    "babel-plugin-external-helpers": "^6.22.0",
-    "babel-preset-env": "^1.6.1",
+    "@babel/core": "^7.0.0",
+    "@babel/plugin-external-helpers": "^7.0.0",
```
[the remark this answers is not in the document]
denisw: Thanks for the hint - I'll look into that.

package.json
```diff
@@ -48,7 +56,16 @@
     "selectorator": "^4.0.1"
```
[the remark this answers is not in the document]
denisw: The situation is better now than before, but I agree that `reselect` would be vastly superior because it only accepts functions, and is typed so that types can be inferred automatically.

That being said, changing the library (and thus the public API) is out of this PR's scope, which is only about translating the existing API.

rollup.config.js
```diff
@@ -1,12 +1,13 @@
 import resolve from 'rollup-plugin-node-resolve'
 import babel from 'rollup-plugin-babel'
 import commonjs from 'rollup-plugin-commonjs'
-import replace from 'rollup-plugin-replace'
 import pkg from './package.json'
 
-const input = 'src/index'
+const input = './src/index.ts'
```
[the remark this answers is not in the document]
denisw: There shouldn't, but at least in-between it wouldn't build without `.ts`. I'll try to remove it again and see what happens.

src/createSlice.ts
Dudeonyx: I fixed this in my version, The `selectors` has a fixed name `getSlice` instead of generating one which messes up auto-complete.

```js
/** 
 * // ...
 * @template SS - [SliceState] - The slice of state the reducer handles
 * @template S - [State] - The redux state tree.
 * // ...
 */
interface Slice<A = any, SS = any, S = any> {
selectors: {
    getSlice: (state: S) => SS;
  };
// ...
}
```
where `S` in

src/serializableStateInvariantMiddleware.ts
Dudeonyx: can't we use `isPlain(val: unknown): val is undefined
  | string
  | boolean
  | number
  | any[]
  | null
  | { [x: string]: any }`

for now?

src/sliceSelector.ts
Dudeonyx: I'm guessing it's the type of the return value of the selector

src/configureStore.ts
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Returns any array containing the default middleware installed by
+ * `configureStore()`. Useful if you want to configure your store with a custom
+ * `middleware` array but still keep the default set.
+ *
+ * @return The default middleware used by `configureStore()`.
+ */
+export function getDefaultMiddleware(
+  isProduction = IS_PRODUCTION
+): Middleware[] {
```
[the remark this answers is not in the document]
Dudeonyx: Doing it like this

```js
const IS_PRODUCTION = process.env.NODE_ENV === 'production';

export function getDefaultMiddleware(isProduction = IS_PRODUCTION) {
  const middlewareArray = [thunk,];
  let middlewareArrayPlus;

  if (!isProduction) {
    middlewareArrayPlus = [
      createImmutableStateInvariantMiddleware(),
      thunk,
      createSerializableStateInvariantMiddleware(),
    ];
  }

  return middlewareArrayPlus === undefined
    ? middlewareArray
    : middlewareArrayPlus;
}
```
preserves the types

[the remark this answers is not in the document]
denisw: Hm, yeah, this could be `[ThunkMiddleware<S, A>, ...Middleware[]]` for instance. However I have no idea how we would preserve the extension type in `configureStore`. Would we need overrides with a one-element `middleware` tuple, two-element `middleware` tuple, etc., so that we can write `Store<S,A> & Ext1 & Ext2 & ...`?

src/configureStore.ts
```diff
@@ -0,0 +1,140 @@
...
+ * A friendly abstraction over the standard Redux `createStore()` function.
+ *
+ * @param config The store configuration.
+ * @returns A configured Redux store.
+ */
+export function configureStore<S = any, A extends Action = AnyAction>(
+  options: ConfigureStoreOptions<S, A>
+): Store<S, A> {
+  const {
+    reducer = undefined,
```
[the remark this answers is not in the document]
denisw: It's for documentation purposes, because the type of `reducer` is actually non-nullable. There is still a "is reducer defined" runtime check for JS users (and TS users using `any` 😄).

src/configureStore.ts
```diff
@@ -0,0 +1,140 @@
...
+      // its type definition file yet:
+      //
+      // https://github.com/zalmoxisus/redux-devtools-extension/pull/624
+      trace: !IS_PRODUCTION
+    })
+  }
+
+  const storeEnhancers = [middlewareEnhancer, ...enhancers]
+
+  const composedEnhancer = finalCompose(...storeEnhancers) as StoreEnhancer
```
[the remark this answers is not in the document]
denisw: Yep. Again, I'm not sure how to arrive at the final `StoreEnhancer` type considering that the array could have any amount of elements.

src/createAction.ts
```diff
@@ -0,0 +1,52 @@
+import { Action } from 'redux'
+
+/**
+ * An action with a string type and an associated payload. This is the
+ * type of action returned by `createAction()` action creators.
+ *
+ * @template P The type of the action's payload.
+ */
+export interface PayloadAction<P = any> extends Action<string> {
```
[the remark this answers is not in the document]
denisw: > Action<string> prevents reducers from being type-safe 

That's a bit strongly-worded - it's still type-safe in the sense that only strings are allowed. But good idea to use `T extends string` to allow literal types like `Action<'increment'>` being inferred!

src/configureStore.ts
```diff
@@ -0,0 +1,140 @@
...
+  devTools?: boolean
+
+  /**
+   * The initial state. You may optionally specify it to hydrate the state
+   * from the server in universal apps, or to restore a previously serialized
+   * user session. If you use `combineReducers()` to produce the root reducer
+   * function (either directly or indirectly by passing an object as `reducer`),
+   * this must be an object with the same shape as the reducer map keys.
+   */
+  preloadedState?: DeepPartial<S>
```
Dudeonyx: This `preloadedState?: DeepPartial<S>` is gonna cause issues when the `reducer` is an object and a `preloadedState` is given, because then S is inferred from the preloadedState `DeepPartial<S>` and not the reducer `ReducersMapObject<S, A>`.
This is an issue because the preloaded state is can be a partial application of the state, instead it is treated as the entire state.

i.e 
```js
export const store = configureStore({
  reducer: {
    hi: hiReducer,
    auth: authReducer,
    ords: ordersReducer,
  },
  preloadedState: {
    auth: {},
  },
});
```

will throw a type error saying `hi` and `ords` don't exist in `S`

src/createReducer.ts
denisw: Redux states can be anything, not just strings.

src/createReducer.ts
denisw: I fixed it by typing the `draft` parameter as `S`. (It cannot be `undefined` because of `state = initialState`.)

src/createSlice.ts
denisw: > Why is this `any` here? Why does this _return_ `S`? Shouldn't `S` be the input?

The input is the state that _contains_ the state of type `S`. So if the slice name were a type parameter `N` (not sure if it should), the selector type would be `(state: { [key in N]: S }) => S`.

src/index.ts
denisw: Yeah, the ugliness of re-exporting generic types was the reason I found the `index.d.ts` to be a better trade-off.

src/isPlainObject.ts
denisw: Thanks, I'm not used to the new `unknown` type yet.

src/serializableStateInvariantMiddleware.ts
denisw: To be honest, I think `isPlain(val: any): boolean` is totally fine. I don't think these more complex types offer any practical benefit.

src/configureStore.ts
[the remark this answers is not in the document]
Dudeonyx: yeah, we would need to write an exponential amount of overloads since not only is the number of `middlewares` variable the number of `enhancers` is as well.
So we'd end up with overrides for one-element `middleware` tuple and one-element `enhancer` tuple, one-element `middleware` tuple and two-element `enhancer` tuple, two-element middleware tuple and one-element `enhancer` tuple, two-element `middleware` tuple and two-element `enhancer` tuple, etc

tsconfig.json
denisw: Don't think it does anything because compilation happens through Babel not `tsc`, but I set it anyway.

src/configureStore.ts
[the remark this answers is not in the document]
denisw: Yes, I think this is a point where we need to favor practicality over maximum type precision. If we cannot find an economical way to get the type just right, we'll probably get by just fine by having the configured store's `dispatch` accept a relatively generic type like `unknown` for now. We can later see if we can tweak the API to make typing easier.

package.json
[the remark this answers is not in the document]
denisw: I checked again and `@babel/plugin-external-helpers` is actually not used. I installed it because it was in the Babel 6 setup, but realized it shouldn't be there and forgot to remove the package. Will remove.

src/createSlice.ts
denisw: This didn't work. TypeScript inferred the key type of `slice.actions` as `string` in one of the typing tests, so it was possible to write `slice.actions.nonExistingAction`.

src/configureStore.ts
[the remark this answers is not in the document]
denisw: Ah, good catch. Your fix looks good to me @Dudeonyx, even though you could _theoretically_ specify an explicit type for `PS` that has properties not in `S`. No harm in that, though, as there is no real way to do this accidentally.

I'll adopt this version and add a typing test to make sure this doesn't break in the future. Thanks @Dudeonyx !

src/configureStore.ts
```diff
@@ -72,7 +76,7 @@ export interface ConfigureStoreOptions<S = any, A extends Action = AnyAction> {
    * function (either directly or indirectly by passing an object as `reducer`),
    * this must be an object with the same shape as the reducer map keys.
    */
-  preloadedState?: DeepPartial<S>
+  preloadedState?: DeepPartial<PS>
```
Dudeonyx: What you said about users specifying an explicit type for `PS` that has properties not in `S` got me thinking about how to get rid of the `PS` type, and I figured out that I could lower the "inference priority", so to speak, of the preloaded state by adding a bit of indirection with a ternary that will always return `S`.

```js
export function configureStore<S = any, A extends Action = AnyAction>(  // no more `PS` type
  options: ConfigureStoreOptions<S, A>
): Store<S, A> {
//...
}

//...

export interface ConfigureStoreOptions<S = any, A extends Action = AnyAction> {  // no more `PS` type
// ...
preloadedState?: DeepPartial<S extends any ? S : S>;  // <- this ternary
//...
}
```

The `S extends any ? S : S` ternary will always result in `S` but it prevents `preloadedState` from overiding the state inferred from the reducer object.

It looks janky, but no more `PS` type allowing users to potentially shoot themselves in the foot.

Dudeonyx: Also it means you no longer have to type cast `preloadedState as DeepPartial<S>,` in `createStore`

denisw: @Dudeonyx Nice hack! I applied that version (including an explanatory comment).

src/configureStore.ts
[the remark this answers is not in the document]
denisw: For now, I addressed the problem by letting `configureStore()` return an `EnhancedStore extends Store` type that is hardcoded to allow dispatching thunks in addition to plain actions. This at least unblocks the use of redux-thunk. Let's improve on this after this PR (perhaps with the help of an API change).

denisw:
@markerikson I feel this PR is now in a good-enough shape for final review and merging. Thanks to everyone who contributed input and TypeScript advice!

It's not perfect, but I'm sure that with some API tweaks and TypeScript wizardry we can improve things afterwards.

markerikson:
@denisw, @Kovensky  : I pushed a commit with a behavior change I'd been working on the last couple days over on `master` (only importing the mutation check middleware in development).

However, right now the typings test is exploding with errors like this:

```
    Semantic: D:/Projects/redux/redux-starter-kit/src/createReducer.ts (51, 5) Type 'Immutable<S>' is not assignable to type 'S'.
      Type 'S | (S extends AtomicObject ? S : S extends ReadonlyArray<any> ? S[number][] extends S ? { [P in keyof S]: ReadonlyArray<Immutable<S[number]>>; }[keyof S] : ImmutableTuple<S> : { readonly [P in keyof S]: Immutable<S[P]>; })' is not assignable to type 'S'.
        Type 'S extends AtomicObject ? S : S extends ReadonlyArray<any> ? S[number][] extends S ? { [P in keyof S]: ReadonlyArray<Immutable<S[number]>>; }[keyof S] : ImmutableTuple<S> : { readonly [P in keyof S]: Immutable<S[P]>; }' is not assignable to type 'S'.
          Type '(Function & object & S) | (String & object & S) | (Number & object & S) | (Boolean & object & S) | (RegExp & object & S) | (Date & object & S) | (Promise<any> & object & S) | ... 4 more ... | (S extends ReadonlyArray<...> ? S[number][] extends S ? { [P in keyof S]: ReadonlyArray<...>; }[keyof S] : ImmutableTuple<......' is not assignable to type 'S'.
            Type 'S extends ReadonlyArray<any> ? S[number][] extends S ? { [P in keyof S]: ReadonlyArray<Immutable<S[number]>>; }[keyof S] : ImmutableTuple<S> : { readonly [P in keyof S]: Immutable<S[P]>; }' is not assignable to type 'S'.
              Type '(S[number][] extends S ? { [P in keyof S]: ReadonlyArray<Immutable<S[number]>>; }[keyof S] : ImmutableTuple<S>) | { readonly [P in keyof S]: Immutable<S[P]>; }' is not assignable to type 'S'.
                Type 'S[number][] extends S ? { [P in keyof S]: ReadonlyArray<Immutable<S[number]>>; }[keyof S] : ImmutableTuple<S>' is not
assignable to type 'S'.
                  Type '{ [P in keyof S]: ReadonlyArray<Immutable<S[number]>>; }[keyof S] | ImmutableTuple<S>' is not assignable to type 'S'.
                    Type 'ReadonlyArray<Immutable<S[number]>>' is not assignable to type 'S'.
```

I suspect this may be related to https://github.com/mweststrate/immer/issues/289.

Thoughts?

denisw:
@markerikson I was able to reproduce this issue after upgrading `immer` to version 1.10.3 (it's currently still locked to 1.9.3 in `package-lock.json`).

Reading the issue you linked, my understanding is that `produce()` is now explicitly reflecting in its type signature that it's returning a fully read-only version of `S` (`Immutable<S>`). So trying to implement a reducer of type `(S, A) => S` with it, where the type `S` is unconstrained, creates a type error because the returned `Immutable<S>` is immutable while `S` might also be a mutable type.

I now applied the fix that most accurately mirrors the runtime behavior: `createReducer()` now returns a `Reducer<Immutable<S>, A>` instead of a `Reducer<S, A>`. The signature now looks like this:

```typescript
export function createReducer<S = any, A extends Action = AnyAction>(
  initialState: Immutable<S>,
  actionsMap: CaseReducersMapObject<S, A>
): Reducer<Immutable<S>, A>
```

You can pass a mutable state in (`S` is assignably to `Immutable<S>` because the latter type expects less), but you will always get a `readonly`-typed `S` out (as `immer` freezes the result). See the following type test I added:

```typescript
// `initialCounters` has a mutable type ...
const initialCounters = {
  a: 0,
  b: 0
}

// ... but I can still pass it as initial state ...
const reducer = createReducer(initialCounters, {
  incrementA: state => {
    state.a += 1
  },
  incrementB: state => {
    state.b += 1
  }
})

// ... and as state argument to the reducer ...
const newCounters = reducer(initialCounters, { type: 'incrementA' })

// ... but states generated by the reducer are immutable.
// typings:expect-error
newCounters.a++
```

Dudeonyx:
@denisw @markerikson  The immer issue was considered a breaking change and was removed as of `immer v1.10.5` 

I think it's better to remove the fix because from what I read on the [issue](https://github.com/mweststrate/immer/issues/289) the `Immutable` type can cause issues with arrays, see this [comment](https://github.com/mweststrate/immer/issues/289#issuecomment-454022548)

phryneas:
neat, but in the current release on ts-next, the index.d.ts references src/... files, which are not distributed (only dist included)

could you please fix that?

markerikson:
Cool.  I'll put up a new `@ts-next` release this evening.

markerikson:
@denisw : I don't think the `index.d.ts` changes are correct.  Right now the typings tests are failing because it can't import things into `index.d.ts`.  We do single-file builds, so I'm pretty sure there's no `dist/createActions.js`, etc.

denisw:
Just noticed that my declaration file approach broke the typing tests. I'll reverted the change and resorted to simply adding `src/` to `"files"` in `package.json`.

markerikson:
Awright, no one's saying "it works", but no one's whining about this either.  Let's do it.

markerikson:
Published as https://github.com/reduxjs/redux-starter-kit/releases/tag/v0.4.0 .  Thanks to everyone for working on this!

jerolimov:
It works! Thanks @denisw and @markerikson

vincentjames501:
Thanks @denisw and @markerikson , this is great!

From the looks of the original PR message, is the `createSlice` type definitions completely finalized? I've been using it the last few days and bumping into a few issues with TypeScript (comments inline).

```typescript
import { configureStore, createSlice } from 'redux-starter-kit';
import { combineReducers } from 'redux';
// Is there a better way to import/use this type? Pulling from `/src` doesn't feel right
import { PayloadAction } from 'redux-starter-kit/src/createAction';

const user = createSlice({
  slice: 'user',
  initialState: { name: '', age: 100 },
  reducers: {
    setUserName: (state, action: PayloadAction<string>) => {
      state.name = action.payload;
    },
  },
});

  reducer: combineReducers({
    user: user.reducer,
  }),
});
// This results in a compilation issue
// const store2 = configureStore({
//   reducer: {
//     user: user.reducer,
//   },
// });

// Could the resolved type be (action: string) => PayloadAction<string, string>
// instead of (...args: any[]) => PayloadAction<any, string> as calls like this work
// user.actions.setUserName(1, 'foo', true);
user.actions.setUserName('eric');

const state = store.getState();

// No autocompletion here and things like this compile and blow up at runtime:
// console.log(user.selectors.getUser2(state));
// I realize the selectorName is dynamically generated here and can make things tricky
console.log(user.selectors.getUser(state));
```

Dudeonyx:
@vincentjames501 this is wrong

```ts
const user = createSlice({
  slice: 'user',
  initialState: { name: '', age: 100 },
  reducers: {
    setUserName: (state, action: PayloadAction<string>) => {
      state.name = action.payload;
    },
  },
});

```

The case reducers do not receive the entire action object, just the payload. So it should be

```ts
const user = createSlice({
  slice: 'user',
  initialState: { name: '', age: 100 },
  reducers: {
    setUserName: (state, payload: string ) => {
      state.name = payload;
    },
  },
});
```

markerikson:
@Dudeonyx : in RSK, they do receive the whole action:

https://github.com/reduxjs/redux-starter-kit/blob/5b38860ff4e068ee9898f9a8b6700366f04a5689/src/createReducer.ts#L56

Dudeonyx:
@markerikson my mistake, thought it was just the payload like in `robodux`

markerikson:
Yeah, I debated back and forth on that one, and decided to pass the entire action for consistency with the rest of the Redux world.  Users can always destructure `{payload}` if they want.

phryneas:
The problem brought up by @vincentjames501 seems to still persist - action creators are not really getting a valid payload type. 
Are you still working on this or did that maybe just go unnoticed?

denisw:
@phryneas I created a payload-action-related fix for `createSlice` as #110 a while ago. Could you test if that fixes your issue?

phryneas:
@denisw will do later, thanks :)

PR #74, opened 2019-01-09
Rewrite docs for createReducer()

denisw:
This version tries to state more clearly the motivation behind the function and how the immer-powered state mutation works.


markerikson:
Great!  Merging now.

PR #77, opened 2019-01-11
Rewrite createAction() docs

denisw:
Not sure if the _Non-String Action Types_ section is really needed, but it seems like a non-obvious caveat that it easy to miss.


markerikson:
Forgot this was here.  Merging, thanks!

PR #79, opened 2019-01-13
createReducer doc: Fix intention and add missing payload extraction

jerolimov:
Hi, small changes in the createReducer documentation code example


markerikson:
Good catch.  Thanks!

PR #84, opened 2019-01-21
Re-export compose and combineReducers from redux for convenience

biernacki:
Just to mach what's in the [docs](https://redux-starter-kit.js.org/api/other-exports) and for convenience this PR re-exports `compose` and `combineReducers` from the original `redux` package.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -6,7 +6,9 @@ export {
   Middleware,
   Reducer,
   Store,
-  StoreEnhancer
+  StoreEnhancer,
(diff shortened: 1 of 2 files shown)


markerikson:
Thanks.  Looks like that was in the JS code up through 0.3.3, but got missed in the TS conversion:

https://github.com/reduxjs/redux-starter-kit/blob/f2af5811215b2e8c4a05e9d9e3962f363e0fbb0f/src/index.js#L11

PR #86, opened 2019-01-22
Alternate: Add ability for slices to listen to other actions

Dudeonyx:
Implemented my proposed changes from https://github.com/reduxjs/redux-starter-kit/pull/83#issuecomment-456379590

@denisw  @markerikson your inputs are appreciated

*Note: this PR is for the `feature/other-slice-action` branch not `master`*

* ~~Removed `getType()` utility~~

* ~~`slice` is no longer attached to `actionsMap` and `reducerMap`~~

* ~~Removed `reducerMap` and directly forward `reducers` to `createReducer`~~

- [x] `createAction()` action creators `type` property returns a `string literal` for better type saftey

- [x] Fixed tests

- [x] Added tests

### DIFF
--- a/src/createAction.ts
+++ b/src/createAction.ts
@@ -39,7 +39,7 @@ export function createAction<P = any, T extends string = string>(
     return { type, payload }
   }
 
-  actionCreator.toString = () => `${type}`
+  actionCreator.toString = (): T => `${type}` as T
 
   actionCreator.type = type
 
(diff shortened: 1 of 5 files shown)


markerikson:
Per my comment in the other thread, this appears to be removing functionality that I want to keep.  I don't think this is the direction I want to go.

markerikson:
Okay.  one other question.  @jessidhia mentioned something in Reactiflux about `&`-ing the action creator with a string, possibly in the return value of `createAction`:

> [7:13 PM] **Jessie** : acemarke to the return type of `createAction`; it will probably need an `as` cast
> [7:13 PM] **Jessie** : `const result = ...; return result as typeof result & string`

Would that allow the action creator itself to be used directly as a key in an object, without a TS2464 error like https://github.com/Microsoft/TypeScript/issues/14207 ?

PR #88, opened 2019-01-22
`configureStore` -> `createSlice`

modernserf:
Am I misunderstanding something, or should this be `createSlice` here?


denisw:
Thanks! However, someone else was faster than you and created #87. 😉

PR #94, opened 2019-01-24
Improve slice action creator types

### LINKED ISSUE #93: Parameters of actions are not infered
After trying out your hello world example I noticed, that the parameters of the actions are not infered.
Here's what I got:

```typescript
export const counter = createSlice({
	slice: 'counter', 
	initialState: 0,
	reducers: {
		increment: (state) => state + 1,
		decrement: (state) => state - 1,
		multiply: (state, action: PayloadAction<number>) => state * action.payload
	}
});

counter.actions.multiply() // should error out
counter.actions.multiply("2") // should error out

```
The call to counter.actions.multiply without any or the wrong arguments shouldn't be possible since it will generate a runtime error. Is this intended behaviour or am I doing something wrong / it's not possible?

denisw:
Instead of `...args: any[]`, each action creator now expects the payload inferred from the corresponding case reducer.

Fixes #93.

### DIFF
--- a/src/createSlice.ts
+++ b/src/createSlice.ts
@@ -1,12 +1,17 @@
-import { Action, AnyAction, ActionCreator, Reducer } from 'redux'
+import { Action, AnyAction, Reducer } from 'redux'
 import { createAction, PayloadAction } from './createAction'
 import { createReducer, CaseReducersMapObject } from './createReducer'
 import { createSliceSelector, createSelectorName } from './sliceSelector'
 
+/**
+ * An action creator atttached to a slice.
+ */
+export type SliceActionCreator<P> = (payload: P) => PayloadAction<P>
+
 export interface Slice<
   S = any,
   A extends Action = AnyAction,
-  AT extends string = string
+  AP extends { [key: string]: any } = { [key: string]: any }
 > {
   /**
    * The slice name.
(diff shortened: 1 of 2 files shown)


markerikson:
@denisw : I'm seeing a TS compilation failure in Travis, and also showing up locally:

```
src/createSlice.test.ts:38:35 - error TS2345: Argument of type 'Action<any>' is not assignable to parameter of type 'PayloadAction<any, string>'.
38         expect(reducer(undefined, actions.increment())).toEqual(1)
                                     ~~~~~~~~~~~~~~~~~~~
src/createSlice.test.ts:78:33 - error TS2345: Argument of type 'Action<any>' is not assignable to parameter of type 'PayloadAction<any, string>'.
  Property 'payload' is missing in type 'Action<any>' but required in type 'PayloadAction<any, string>'.
78       expect(reducer(undefined, actions.increment())).toEqual(1)
                                   ~~~~~~~~~~~~~~~~~~~
  src/createAction.ts:12:3
    12   payload: P
         ~~~~~~~
    'payload' is declared here.
```

denisw:
@markerikson Fixed now.

markerikson:
Great, thanks!

jerolimov:
This was a great improvement when using the actions. 👍

Issue #102, opened 2019-01-26
typings issue between createSlice and configureStore's reducer prop ?


Sharlaan:
I'm trying to integrate RSK in a minimal Typescript project, but i stumbled upon what looks like a typing conflict :
![](https://i.imgur.com/9ARfMAm.png)

Apparently configureStore.reducer expects to use `AnyAction` interface while createSlice reducer uses `PayloadAction`.
I'm a bit new in TS i don't see how to solve this conflict.

deps :
- redux-starter-kit 0.4.2
- typescript 3.2.4

markerikson:
Paging the TS crew: @denisw @Dudeonyx @jessidhia .

(I will say this is the kind of thing I was kinda concerned about with switching the code to TS: questions I have no way to help answer myself.)

denisw:
The type error here isn't about `PayloadAction` vs. `AnyAction`; the former is a subtype of the latter, so the types are compatible in this respect. But looking at the error message, the problem seems to be that we are not handling the "pass an object instead of a reducer function" correctly, as it says:

> `{ counter: Reducer<number, ...> }` not assignable to `Reducer<{ counter: number }, ...>`

We should be accepting the object and inferring from the object shape what the store's state type is, but that doesn't seem to be quite working. I'll investigate.

denisw:
Sorry, you were right @Sharlaan - it's really the action type that was the issue. I failed to realise that the reducer's action type parameter must be a _supertype_ of the store action type (it has to be able to accept every store action), which is not the case if the reducer takes only `PayloadAction`.

I created #104 to fix this.

Sharlaan:
Ok nice ! Glad it helped.

At the moment i'm trying to find an elegant way to type State.

Would definitely need an example of typing integration along react-redux.

markerikson:
Just published this as 0.4.3.  Thanks!

TidyIQ:
I'm getting the same error in v0.5.1

TidyIQ:
```
// slices/snackbar.ts

import { createSlice, PayloadAction } from "redux-starter-kit";

export interface SnackbarState {
  message: string;
  open: boolean;
  undo: null | (() => void);
  variant: "success" | "warning" | "error" | "info";
}

const initialSnackbarState: SnackbarState = {
  message: "",
  open: false,
  undo: null,
  variant: "info"
};

type OpenSnackbarAction = PayloadAction<SnackbarState>;
[...]

PR #104, opened 2019-01-27
Return Reducer<S, AnyAction> from createReducer

denisw:
This is more correct as the generated reducer does, in fact, support passing actions other than the ones explicitly handled. Also, this fixes an incompatibility with `configureStore` (#102).


markerikson:
Published as 0.4.3.

vishnugopal:
I'm on 0.4.3 and still encountering this issue.

Seems to be very similar to OP, and am in a similar situation. I don't have enough knowhow to fix this.

Screenshot 2019-04-20 at 5 52 34 PM

PR #110, opened 2019-02-05
Type payload-less slice action creators to take no arguments

### LINKED ISSUE #108: Type issue with ActionCreators from createSlice with no payload
I am creating a slice for my state, and one of the actions I want to create does not take a payload. However, when mapping the dispatch to the props in `connect`, I get a weird type error.
image
`
Argument of type 'typeof MyComponent' is not assignable to parameter of type 'ComponentType<Matching<{ loading: boolean; } & { setLoaded: SliceActionCreator<void>; }, MyProps>>'.
  Type 'typeof MyComponent' is not assignable to type 'ComponentClass<Matching<{ loading: boolean; } & { setLoaded: SliceActionCreator<void>; }, MyProps>, any>'.
    Types of parameters 'props' and 'props' are incompatible.
      Type 'Matching<{ loading: boolean; } & { setLoaded: SliceActionCreator<void>; }, MyProps>' is not assignable to type 'Readonly<MyProps>'.
        Types of property 'setLoaded' are incompatible.
          Type 'SliceActionCreator<void>' is not assignable to type '() => void'.
`

Here's a minimal reproduction of the issue I'm having:

```Typescript
import React, { PureComponent } from 'react';
import { createSlice } from 'redux-starter-kit';
import { connect } from 'react-redux'

interface MyState {
    loading: boolean;
}

const INITIAL_STATE: MyState = {
    loading: true,
}

const slice = createSlice({
    slice: 'mySlice',
    initialState: INITIAL_STATE,
    reducers: {
        setLoaded: (state) => { state.loading = false; },
    }
})

interface MyProps {
    loading: boolean;
    setLoaded(): void;
}

class MyComponent extends PureComponent<MyProps> {
    componentDidMount() {
        // perform some check for loading
        const setLoaded = this.props.setLoaded;

        setLoaded();
    }

    render() {
        const loading = this.props.loading;
        if (loading) {
            return (
                <p>Loading...</p>
            );
        } else {
            return (
                <p>Loaded!</p>
            )
        }
    }
}

export default connect(
    (state: MyState) => ({
        loading: state.loading,
    }),
    {
        setLoaded: slice.actions.setLoaded,
    }
)(MyComponent)

```

If I change the type of the action property in `MyProps` to take a void payload, then the error goes away, but that's obviously not the correct type:

```Typescript
interface MyProps {
    loading: boolean;
    setLoaded(payload: void): void;
}
```

Am I doing something wrong, or is there a bug in the typing?

denisw:
Previously, it was not possible to assign a slice action creator without payload to `() => any` because the type was inferred as `(payload: void) => any`. Now we use a conditional type for `SliceActionCreator` that special-cases `void`.

Fixes #108.

### DIFF
--- a/src/createSlice.ts
+++ b/src/createSlice.ts
@@ -6,11 +6,12 @@ import { createSliceSelector, createSelectorName } from './sliceSelector'
 /**
  * An action creator atttached to a slice.
  */
-export type SliceActionCreator<P> = (payload: P) => PayloadAction<P>
+export type SliceActionCreator<P, T extends string = string> = P extends void
+  ? () => Action<T>
+  : (payload: P) => PayloadAction<P, T>
 
 export interface Slice<
   S = any,
-  A extends Action = AnyAction,
   AP extends { [key: string]: any } = { [key: string]: any }
 > {
   /**
@@ -21,13 +22,15 @@ export interface Slice<
   /**
    * The slice's reducer.
    */
-  reducer: Reducer<S, A>
+  reducer: Reducer<S>
 
   /**
    * Action creators for the types of actions that are handled by the slice
    * reducer.
    */
-  actions: { [type in keyof AP]: SliceActionCreator<AP[type]> }
+  actions: {
+    [type in keyof AP]: SliceActionCreator<AP[type], Extract<type, string>>
+  }
 
   /**
    * Selectors for the slice reducer state. `createSlice()` inserts a single
@@ -101,9 +104,7 @@ export function createSlice<
   S = any,
   A extends PayloadAction = PayloadAction<any>,
   CR extends CaseReducersMapObject<S, A> = CaseReducersMapObject<S, A>
->(
-  options: CreateSliceOptions<S, A, CR>
-): Slice<S, A, ExtractPayloads<S, A, CR>> {
+>(options: CreateSliceOptions<S, A, CR>): Slice<S, ExtractPayloads<S, A, CR>> {
   const { slice = '', initialState } = options
   const reducers = options.reducers || {}
   const extraReducers = options.extraReducers || {}
--- a/type-tests/files/createSlice.typetest.ts
+++ b/type-tests/files/createSlice.typetest.ts
@@ -1,4 +1,5 @@
 import {
+  Action,
   AnyAction,
   createSlice,
   PayloadAction,
@@ -63,6 +64,10 @@ import {
       multiply: (state, action: PayloadAction<number>) => state * action.payload
     }
   })
+
+  const increment: () => Action<'increment'> = counter.actions.increment
+  const multiply: (payload: number) => PayloadAction<number, 'multiply'> =
+    counter.actions.multiply
 
   counter.actions.increment()
   counter.actions.multiply(2)


phryneas:
Works for me nicely!

I could go from
```typescript
export const slice = createSlice<
    State,
    any,
    {
        increment: CaseReducer<State, PayloadAction<number>>;
        decrement: CaseReducer<State, PayloadAction<number>>;
        navigation: CaseReducer<State, NavigationAction<{ counter: string }>>;
    }
>({
    slice: 'home',
    reducers: {
        increment: (state, action) => {
            state.counter += action.payload;
        },
        decrement: (state, action) => {
            state.counter -= action.payload;
        },
        navigation: (state, action) => {
            state.counter = Number.parseInt(action.payload.routeParams.counter);
        }
    },
    initialState: { counter: 0 }
});

```
to
```typescript
export const slice = createSlice({
    slice: 'home',
    reducers: {
        increment: (state, action: PayloadAction<number>) => {
            state.counter += action.payload;
        },
        decrement: (state, action: PayloadAction<number>) => {
            state.counter -= action.payload;
        },
        navigation: (state, action: NavigationAction<{ counter: string }>) => {
            state.counter = Number.parseInt(action.payload.routeParams.counter);
        }
    },
    initialState: { counter: 0 } as State
});
```

The only thing that is not working is defining the State type as a Generic like this (but this is only a nice-to-have):
```typescript
export const slice = createSlice<State>({
    slice: 'home',
    reducers: {
        increment: (state, action: PayloadAction<number>) => {
            state.counter += action.payload;
        },
        decrement: (state, action: PayloadAction<number>) => {
            state.counter -= action.payload;
        },
        navigation: (state, action: NavigationAction<{ counter: string }>) => {
            state.counter = Number.parseInt(action.payload.routeParams.counter);
        }
    },
    initialState: { counter: 0 } 
});
```

phryneas:
Actually I am noticing right now that example B seems to be working with the current version without this patch, too, so I might have never had a problem & been the wrong one to test this o_O

Sorry!

pat-son:
Is there an ETA on this getting merged? When can we expect a new version of RSK?

markerikson:
Woops, forgot this was here - sorry!  I've been completely occupied with work on React-Redux v7 the last couple months.

@denisw : is there any chance we can make a no-param action creator actually _not_ try to assign `payload` at all?  Right now, if I do something like `onClick={increment}`, the React event object always gets passed through as the `payload`, and that creates problems (React doesn't like the event object being kept around, and RSK's "no non-serializable values" middleware yells at you).

denisw:
@markerikson I guess we could add a function which explicitly creates an action creator taking no parameters. Perhaps there could be separate `createAction` and `createPayloadAction` functions.

And alternative approach is to keep a single `createAction` function, but require a chained call to a method (perhaps called `withPayload`) if you want the action creator to take a payload.

```js
const increment = createAction('increment')
// increment takes no payload

const multiply = createAction('multiply').withPayload()
// multiply does take a payload
```
In TypeScript, you would then be able to specify the payload type with the `withPayload` call:

```ts
const multiply = createAction('multiply').withPayload<number>()
```

denisw:
Superseded by #133.

PR #120, opened 2019-03-07
Fix link to api breaking

j33n:
Apparently, this footer link [API Reference](https://redux-starter-kit.js.org/api) breaks, whereas this works [API Reference](https://redux-starter-kit.js.org/api/configurestore)


markerikson:
Thanks!

PR #128, opened 2019-03-26
Create declaration files with typescript

### LINKED ISSUE #125: Error in typescript with project references
So I have a monorepo with multiple typescript libs. I wanted to try redux-starter-kit in one of my lib. But after adding that as a dependency I can not build without error because I see 
```
error TS6307: File '/mnt/c/dev/github/sensenet/sn-client/node_modules/redux-starter-kit/src/configurestore.ts' is not in project file list. Projects must list all files or use an 'include' pattern.
```
The redux-starter-kit works great in a repo where project references are not turned on. The source of the problem might be the index.d.ts file in the root of the project because it points to the files located in src.

B3zo0:
## What I did in this pr
- removed index.d.ts file from root and from package.json
- changed createSerializableStateInvariantMiddleware to named export
- added a typings.ts file with the reexports and the package files. This can be used as a typing file when built with typescript
- changed the tsconfig.json to emit declaration files only for files under src

Fixes #125


markerikson:
Seems reasonable to me.  Thanks!

PR #132, opened 2019-04-10
Upgrade to Immer 2.x

ezzatron:
This PR is a simple upgrade of the `immer` dependency. Version 2 of Immer only introduces a single BC break that shouldn't affect `createReducer` as it has to do with returning promises from a producer function.

From the [migration](https://github.com/mweststrate/immer#migration) section of the Immer README:

> ## Migration
>
> **Immer 1.\* -> 2.0**
>
> Make sure you don't return any promises as state, because `produce` will actually invoke the promise and wait until it settles.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3103,7 +3103,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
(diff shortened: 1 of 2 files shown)


markerikson:
Simple enough.  Thanks!

PR #133, opened 2019-04-10
Improve type inference of case reducers

### LINKED ISSUE #131: Type errors with createSlice in TypeScript 3.4.1
When the reducers in a `createSlice` have mixed payload action types, I get weird type errors as of TypeScript 3.4.1 (Maybe 3.4.0 as well, I didn't check).

image

It's like it's assuming that all `PayloadAction`s it sees should be the same type as the first one.

Code for reproduction:

```TypeScript
interface MyState {
    name: string;
    loading: boolean;
    count: number;
}

const INITIAL_STATE: MyState = {
    name: '',
    loading: true,
    count: 0,
}

const slice = createSlice({
    slice: 'mySlice',
    initialState: INITIAL_STATE,
    reducers: {
        setCount : (state, action: PayloadAction<number>) => { state.count = action.payload },
        setName: (state, action: PayloadAction<string>) => { state.name = action.payload },
        setLoaded: (state) => { state.loading = false; },
    }
})
```

denisw:
Previously, the TypeScript compiler would reject case reducer maps (such as the ones passed to `createReducers` and `createSlice`) with different incompatible `PayloadAction` types. I restructured the types a bit to improve type inference in these cases.

Fixes #131

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,6 @@
 module.exports = {
   extends: 'react-app',
-  parser: 'typescript-eslint-parser',
+  parser: '@typescript-eslint/parser',
 
   rules: {
     'jsx-a11y/href-no-hash': 'off',
--- a/package-lock.json
+++ b/package-lock.json
@@ -975,6 +975,47 @@
         "redux": "^3.6.0 || ^4.0.0"
       }
     },
+    "@typescript-eslint/parser": {
+      "version": "1.6.0",
+      "resolved": "https://registry.npmjs.org/@typescript-eslint/parser/-/parser-1.6.0.tgz",
+      "integrity": "sha512-VB9xmSbfafI+/kI4gUK3PfrkGmrJQfh0N4EScT1gZXSZyUxpsBirPL99EWZg9MmPG0pzq/gMtgkk7/rAHj4aQw==",
+      "dev": true,
+      "requires": {
+        "@typescript-eslint/typescript-estree": "1.6.0",
+        "eslint-scope": "^4.0.0",
+        "eslint-visitor-keys": "^1.0.0"
+      },
+      "dependencies": {
+        "eslint-scope": {
+          "version": "4.0.3",
+          "resolved": "https://registry.npmjs.org/eslint-scope/-/eslint-scope-4.0.3.tgz",
+          "integrity": "sha512-p7VutNr1O/QrxysMo3E45FjYDTeXBy0iTltPFNSqKAIfjDSXC+4dj+qfyuD8bfAXrW/y6lW3O76VaYNPKfpKrg==",
+          "dev": true,
+          "requires": {
+            "esrecurse": "^4.1.0",
+            "estraverse": "^4.1.1"
+          }
+        }
+      }
+    },
+    "@typescript-eslint/typescript-estree": {
+      "version": "1.6.0",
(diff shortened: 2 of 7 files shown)


denisw:
I added a commit that upgrades to TypeScript 3.4.3, which I needed to do to reproduce the problem in #131. As a consequence, I also had to upgrade the ESLint TypeScript parser.

wldcordeiro:
@denisw I had an overlapping dependency bump but this will solve an error case for me!

kroogs:
Thank you for your work on this, @denisw. It's exactly what I had hoped to find.

kroogs:
After more research, I've discovered the issue stems from the distributive behavior of conditional types. By encapsulating both the relevant type parameter and its match each in a single-element tuple, it deactivates the distributive nature of the conditional and preserves unions passed to PayloadAction. I've created a pull request for the change here https://github.com/denisw/redux-starter-kit/pull/1.

kroogs:
Apologies for multiple notifications; I got ahead of myself when I first thought I had a solution.

The behavior I described comes from how TypeScript performs distributive conditional inference on a union of contra-variant types. By wrapping each side of the match of `P` in `SliceActionCreator<P>` with a single-element tuple, it prevents the distributive behavior and preserves a union type when inferred in `CaseReducerActionPayloads`. I have an example at https://github.com/kroogs/redux-starter-kit/commit/cc3c894ca4658e101b0f3c8914b7fd843d751056 that works as I describe and passes the full test suite. I'd like to submit a PR unless there's a better solution.

markerikson:
@denisw : does this supersede #110 , or do we need both of them?

denisw:
@markerikson Pretty sure we only need this PR. I closed the other one.

markerikson:
@denisw : great, that simplifies things :)

markerikson:
Awright, green lights, and I don't know what I'm doing types-wise anyway :)

markerikson:
@kroogs : not quite sure I'm following what you're saying there.  If you think there's still typings issues after this PR was merged, please go ahead and file a separate issue / PR.

markerikson:
Hmm. @denisw : looking at the diffs for this PR and #110 , the slice action types do look different.

```ts
// #110
export type SliceActionCreator<P, T extends string = string> = P extends void
  ? () => Action<T>
  : (payload: P) => PayloadAction<P, T>

// #133
export type SliceActionCreator<P> = P extends void
  ? () => PayloadAction<void>
  : (payload: P) => PayloadAction<P>
```

Looks like 110 is returning a plain `Action`, while 133 is returning a `PayloadAction`. Does that make any meaningful difference?

kroogs:
@markerikson The typing issue I encountered with `PayloadAction` remains. I'll create a new PR tomorrow with a detailed description and a simple fix.

farah:
@tydira hi there, I know this is a very old pr, but was curious with your fix in #138 about the distributed union behaviour. I was unable to replicate the type errors you mentioned in your [comment](https://github.com/reduxjs/redux-toolkit/pull/133#issuecomment-484819765).

> I'm creating a reducer with an action type of PayloadAction<T | T[]> and the action creator returned by createSlice() wants a payload of the type **T & T[]**.

I was unable to reproduce this error v3.2 & 3.4 at he commits prior to pr 138. The test cases you added also would have passed prior to your change. I'm not exactly sure whats happening here, any ideas?

tydira:
@farah That's odd. Is it possible you're using a non-local TS install? As evident in surrounding comments, it was an issue that a few people were able to confirm at the time.

Issue #134, opened 2019-04-15
Discussion: Improve Typings


hedgerh:
I want to help improve the typings for `redux-starter-kit` and get some documentation written up around TS usage.

I have an ideal usage/API in mind, based on `typesafe-actions`.

**Creating actions with `createTypesafeAction` and creating reducers with `createReducer`:**
```ts
import { createTypesafeAction, createReducer, ActionType } from 'redux-starter-kit'

// createTypesafeAction is an optional API addition,
// but gives a cleaner way to do this than `createAction`
const actions = {
  login: createTypesafeAction('LOGIN')<ILogin>(),
  loginSuccess: createTypesafeAction('LOGIN_SUCCESS')<IUser>(),
}

interface IState {
  isLoading: boolean;
  user?: IUser;
}

const initialState: IState = {
  isLoading: false,
}

// borrowed from typesafe-actions,
// lets us get types from a map of action creators and keep type literals intact
type Action = ActionType<typeof actions>

export const reducer = createReducer<IState, Action>(initialState, {
  [actions.login.type]: (state, action) => {
    // since we have type literals in tact,
    // action should be inferred to PayloadAction<"LOGIN", ILogin>
    state.isLoading = true
  },
  [actions.loginSuccess.type]: (state, action) => {
    // action inferred to PayloadAction<"LOGIN_SUCCESS", IUser>
    state.isLoading = false
    state.user = action.payload
  },
})
```

I have everything working down to actually being able to infer `action` based on the caseReducer key.  I'm hitting a bit of a limit to my skills with Typescript to infer type based on key.

**createSlice**

I use redux-saga.  With `typesafe-actions`, I'm able to type the `action` parameter in my sagas by doing:

```ts
const actions = { login: createStandardAction('LOGIN')<ILogin>() }

function* loginSaga(action: ActionType<typeof actions.login>) {}
```

I'm not sure how you'd go about doing something like this with `createSlice`.  I'm trying to think of some way that you could type the action param in your case reducers, and also come out with actions that have the correct literals for `type` and the correct payload shapes.

Would love to hear other people's thoughts on how we could improve the typings.

markerikson:
Obligatory paging of the TS Experts:  @denisw, @Dudeonyx , @jessidhia

tim-hm:
[...]

PR #136, opened 2019-04-29
Improve typings: Added ability to typecast action types in `createSlice`

Dudeonyx:
Referencing issue #134 about improving typings while not changing the core api just for typescript users.

This PR adds the ability to typecast action types in createSlice, works as follows

```ts
const formSlice = createSlice({
  slice: 'form',
  reducers: {
    setName: (state, action: PayloadAction<string, 'form/setName'>) => {
      state.name = action.payload;
    },
    setSurname: (state, action: PayloadAction<string, 'form/setSurname'>) => {
      state.surname = action.payload;
    },
    setMiddlename: (
      state,
      action: PayloadAction<string, 'form/setMiddlename'>,
    ) => {
      state.middlename = action.payload;
    },
    resetForm: (state, _: PayloadAction<undefined, 'form/resetForm'>) =>
      formInitialState,
  },
  initialState: formInitialState,
});
```

giving the following action creator type signatures in formSlice.actions

image

image

image

Note: payload types `undefined`, ~~never~~ , and `void` have action creator signature types requiring no payloads
image

Caveat: Possibility of user error in typecasting exists but until typescript supports string literal concatenation not much can be done.

Bonus: I 'fixed' the typescript issue in `createReducer`, so no more need to `// @ts-ignore` that part, Fix has no impact on exposed api's or types.

Bonus 2: the `slice` argument and return in `createSlice` is now a string literal, hooray for slightly barely better type safety


Dudeonyx:
Pinging @denisw, @Jessidhia, @hedgerh

hedgerh:
Will take a look when I get home later 👀

markerikson:
For the still-mostly-TS-clueless: can I get a comparison of both the general usage and the type-checking status before and after this PR?  What do things look like now, and what do they look like as a result of this?

Dudeonyx:
> For the still-mostly-TS-clueless: can I get a comparison of both the general usage and the type-checking status before and after this PR? What do things look like now, and what do they look like as a result of this?

@markerikson The changes introduced are not breaking for TS users, in fact unless told about it most people won't notice it.

Previously you would typecast your reducers actions like this

```ts
const formSlice = createSlice({
  slice: 'form',
  reducers: {
    setName: (state, action: PayloadAction<string>) => {
      state.name = action.payload;
    },
  },
});
```

formSlice.actions.setName would have the following type signature:

image

With this PR, first of all, the `type` prop in the action creator is exposed in the type signature, becoming like this:

image


Second, users can now optionally typecast the action `type` to a string literal instead of a generic string i.e

```ts
const formSlice = createSlice({
  slice: 'form',
  reducers: {                                       // Talking about \|/
    setName: (state, action: PayloadAction<string, 'form/setName'>) => {
      state.name = action.payload;
    },
  },
});
```

So `formSlice.actions.setName`'s type signature becomes

image

Notice that the `type` prop is no longer just a generic `string` but now a specific string literal `form/setName`

hedgerh:
What happens if you intentionally misspell the action type in the case reducer args annotation?  Sorry didn’t have a chance to look at this last night. About to take a look

hedgerh:
What does current usage look like?  when creating a slice, do you just assert `action: PayloadAction<PayloadShape>`?

> Second, users can now optionally typecast the action type to a string literal instead of a generic string

So we don't lose the ability to assert the shape of just the payload if we aren't concerned with the action type, correct?

@markerikson this is useful for someone who needs to make use of these actions elsewhere and doesn't want to create a whole new interface for it.  I would use this like so:

```
const { actions, reducer } = createSlice({
  slice: 'auth',
  reducers: {
    login: (state, action: PayloadAction<ILogin, 'auth/login'>) => ...
  }
})

export function* loginSaga (action: ActionType<typeof actions.login>) {
  // rather than creating some extra interface that isn't strongly tied to the actual action,
  // i can type the saga's arg based on the actual action I'm using.
}

export function* watchSaga () {
  yield takeLatest(actions.login.type, loginSaga)
}
```

@Dudeonyx we may need some `ActionType` type that can be used to type actions by the action's `type`.  agree?  If not, is there another way to do this without one?  Going off what `typesafe-actions` does.

Also, do we need to align with https://github.com/reduxjs/redux-starter-kit/pull/109 at all? @denisw

markerikson:
I need some advice from you TS experts on this one.  You think it's worth merging / ready to merge?

@hedgerh @Dudeonyx @denisw @jessidhia

hedgerh:
Hopefully @Dudeonyx can get back about my last comment.  I'll pull this down tonight and see how it's looking, though.

markerikson:
I'd still appreciate some further feedback on this PR.

markerikson:
Following up again.  Is this still relevant after all the other recent changes?

phryneas:
This is still relevant, as the recent changes did not affect the `type` of an action - but given all the recent changes, it would have to be rewritten. (Also, this PR contains quite a few commits not directly correlating with the PR itself, so those might be better off in separate PRs?)

It still bears the problem of user-specified types potentially not matching the real, generated type. But as TS has no type-string-concatenation, this is the best way to currently do this and TS cannot check those. And at least, an error would be located better directly in the slice, close to the `slice` name definition and the reducer's name than somewhere else. Potential errors would be more easy to spot that way.

markerikson:
@Dudeonyx, @phryneas : can someone rebase / reimplement this on latest master?

markerikson:
Well, this has been sitting around for months.  I'll continue to leave it open for now, but unless someone can take the time to fix up the conflicts, I don't see it getting into 0.8 or 1.0.

markerikson:
Actually, looking at the current source, is this already covered?

```ts
export type PayloadAction<
  P = any,
  T extends string = string,
  M = void
> = WithOptionalMeta<M, WithPayload<P, Action<T>>>
```

Meh.  I'm going to go ahead and close this.  If someone can take the time to review the current behavior in comparison, I'm still open to trying to improve it if possible.

PR #138, opened 2019-04-30
Fix PayloadAction inference

kroogs:
By wrapping the `P` match in `SliceActionCreator<P>` with a single-element tuple, it prevents the conditional from being checked distributively, which otherwise has the effect of converting a union into an intersection when used with contra-variant types.

Without this change, the modified test emits the following error:

`Argument of type '2' is not assignable to parameter of type 'number & number[]'.
      Type '2' is not assignable to type 'number[]'.`

 `Argument of type 'number[]' is not assignable to parameter of type 'number & number[]'.
      Type 'number[]' is not assignable to type 'number'.`

My original two comments on this issue are here: https://github.com/reduxjs/redux-starter-kit/pull/133#issuecomment-484819765 and here: https://github.com/reduxjs/redux-starter-kit/pull/133#issuecomment-485247351. More information about distributive conditionals can be found here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html. They're very useful and typically match our intent when combining unions with other types, but in cases like these they end up transforming the resulting type into something entirely useless.

### DIFF
--- a/src/createSlice.ts
+++ b/src/createSlice.ts
@@ -5,6 +5,10 @@ import { createSliceSelector, createSelectorName } from './sliceSelector'
 
 /**
  * An action creator atttached to a slice.
+ *
+ * The `P` generic is wrapped with a single-element tuple to prevent the
+ * conditional from being checked distributively, thus preserving unions
+ * of contra-variant types.
  */
 export type SliceActionCreator<P> = [P] extends [void]
   ? () => PayloadAction<void>
--- a/src/createSlice.ts
+++ b/src/createSlice.ts
@@ -6,7 +6,7 @@ import { createSliceSelector, createSelectorName } from './sliceSelector'
 /**
  * An action creator atttached to a slice.
  */
-export type SliceActionCreator<P> = P extends void
+export type SliceActionCreator<P> = [P] extends [void]
   ? () => PayloadAction<void>
   : (payload: P) => PayloadAction<P>
 
--- a/type-tests/files/createSlice.typetest.ts
+++ b/type-tests/files/createSlice.typetest.ts
@@ -60,12 +60,16 @@ import {
     reducers: {
       increment: state => state + 1,
       decrement: state => state - 1,
-      multiply: (state, action: PayloadAction<number>) => state * action.payload
+      multiply: (state, { payload }: PayloadAction<number | number[]>) =>
+        Array.isArray(payload)
+          ? payload.reduce((acc, val) => acc * val, state)
+          : state * payload
     }
   })
 
   counter.actions.increment()
   counter.actions.multiply(2)
+  counter.actions.multiply([2, 3, 4])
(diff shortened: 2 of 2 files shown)


markerikson:
I'll similarly ask for help on this one. You think it's worth merging / ready to merge?

@hedgerh @Dudeonyx @denisw @jessidhia

hedgerh:
Looks like @denisw may have more context here.  I'll take some more time to look over this tonight if he's not available, though.

denisw:
I didn't know about distributive conditional types - thanks for the pointer! The PR looks good to me. The only thing I'd add is a comment at the definition of `SliceActionCreator` explaining why the single-element tuple type is used, as it's far from obvious. :)

kroogs:
Cheers, @denisw. I didn't know about distributive conditionals either until I ran into this and wanted to figure out what was happening.

Does anyone have an explanatory comment in mind? I'm thinking something appended to the bottom of the `SliceActionCreator<P>`'s original comment that says:

> The P generic is wrapped with a single-element tuple to prevent the
> conditional from being checked distributively, thus preserving unions
> of contra-variant types.

I'll commit and push this if there are no objections.

markerikson:
Sure, seems good.  (I still don't quite grok the meaning, but sounds plausible :)  )

renatoruk:
Thank you for this library, it really improves the workflow with redux. I am currently using it on a react typescript project for a short time, but can already see the benefit. 👏 

I ran into the same issue as @kroogs with the first reducer I wrote using createSlice.

Slice instance looks like this:

```
export const module = createSlice({
    initialState: {
        visible: true,
    },
    reducers: {
        setVisible(state: ModuleState, action: PayloadAction<boolean>) {
            state.visible = action.payload;
        },
    },
});

```

Trying to dispatch that action from outside resulted with error:
TS: 2345 Argument of type 'boolean' is not assignable to parameter of type 'false & true'.

This is code in the container:
```
const opened = true;
dispatch(module.actions.setOpened(opened));
```

I just modified the dist node_modules for the time being as this PR fixes the problem (thank you @kroogs !), but it's only a temporary solution.

Is this PR ready to be merged? If not, what can we do to improve it?

Retsam:
As perhaps an alternate way of reaching the same end, perhaps:

```ts
// Since TS only distributes conditionals when the checked type is a naked type parameter
// intersecting with an empty interface will prevent it from distributing, without changing the contract of T
type NoDistribute<T> = T & {};

export type SliceActionCreator<P> = NoDistribute<P> extends void ? //...
```

If you wanted to link the docs on distributive conditional types for context in the comment, they're here: https://www.typescriptlang.org/docs/handbook/advanced-types.html#distributive-conditional-types.

markerikson:
In the absence of further comments, I'm going to assume this is okay and go with it.

PR #139, opened 2019-05-07
Fix: SerializableStateInvariantMiddleware: Use `isSerializable` when checking state

ali-rantakari:
SerializableStateInvariantMiddleware: Actually use the supplied
`isSerializable` function when searching for non-serialisable values
in the state.

### DIFF
--- a/src/serializableStateInvariantMiddleware.test.ts
+++ b/src/serializableStateInvariantMiddleware.test.ts
@@ -158,4 +158,43 @@ describe('serializableStateInvariantMiddleware', () => {
     expect(value).toBe(badValue)
     expect(actionType).toBe(ACTION_TYPE)
   })
+
+  it('Should use the supplied isSerializable function to determine serializability', () => {
+    const ACTION_TYPE = 'TEST_ACTION'
+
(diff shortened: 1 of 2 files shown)


ali-rantakari:
Pinging @markerikson for review please

src/serializableStateInvariantMiddleware.test.ts
```diff
@@ -112,18 +112,14 @@ describe('serializableStateInvariantMiddleware', () => {
     expect(action).toBe(dispatchedAction)
   })
 
-  it('Should log an error when a non-serializable value is in state', () => {
-    const ACTION_TYPE = 'TEST_ACTION'
-
+  function dispatchActionWithValueInTestStore(middleware: Middleware, actionType: string, badValue: any) {
```
kgregory: Not sure you need to introduce this.  You can get by setting up the store in each test (like all the other tests).

ali-rantakari: OK — I got rid of it and updated the PR.

markerikson:
Sure. Thanks!

Issue #140, opened 2019-05-07
Supporting Immutable.JS in `serializable-state-invariant-middleware`


kgregory:
I really like the problem that `serializable-state-invariant-middleware` is meant to solve, but I use `immutable` structures in my store, which will always be seen as non-serializable by the current version of this middleware.  

I think a minor change would allow me to augment this middleware to allow the use of `immutable`.

Add a new option that will retrieve the entries for a given value:

```js
export interface SerializableStateInvariantMiddlewareOptions {
  /**
   * The function to check if a value is considered serializable. This
   * function is applied recursively to every value contained in the
   * state. Defaults to `isPlain()`.
   */
  isSerializable?: (value: any) => boolean,
  /**
   * The function that will be used to retrieve entries from each
   * value.  If unspecified, `Object.entries` will be used. Defaults
   * to `undefined`.
   */
  getEntries?: (value: any) => [string, any][],
}
```

Consumers that use `immutable` can write their own functions:

```js

// immutable version 3.8.2 assumed for this demo!
const isSerializable = (value: any): boolean =>
  Iterable.isIterable(value) || isPlain(value);

const getEntries = (value: any): [string, any][] =>
  Iterable.isIterable(value) ? value.entries() : Object.entries(value);

const middleware = createSerializableStateInvariantMiddleware({
  isSerializable,
  getEntries,
});
```

We can accept these functions as part of the options provided at creation:

```js
export function createSerializableStateInvariantMiddleware(
  options: SerializableStateInvariantMiddlewareOptions = {}
): Middleware {
  const { isSerializable = isPlain, getEntries } = options

  return storeAPI => next => action => {
    const foundActionNonSerializableValue = findNonSerializableValue(
      action,
      [],
      isSerializable,
      getEntries,
    )
    // ...
```

We can use `getEntries` in `findNonSerializableValue` (if provided):

```js
export function findNonSerializableValue(
  value: unknown,
  path: ReadonlyArray<string> = [],
  isSerializable: (value: unknown) => boolean = isPlain,
  getEntries?: (value: unknown) => [string, any][]
): NonSerializableValue | false {
  // ...
```

Switch to `Object.entries` when evaluating nested values:

```js
  let entries;
  if (getEntries != null) {
    entries = getEntries(value);
  } else {
    entries = Object.entries(value);
  }

  for (const [property, nestedValue] of entries) {
    const nestedPath = path.concat(property)
    // ...
```

[...]

PR #141, opened 2019-05-07
Allow consumer to augment middleware to tolerate certain structures (e.g. Immutable)

kgregory:
In Issue #140 I said I would wait before submitting a PR, but here we are.

This pull request will enhance `serializable-state-invariant-middleware` to accept an additional option at creation: `getEntries`.  Additionally, it will use the `isSerializable` when checking action and state.

These two options will allow a consumer to augment the middleware to tolerate certain structures, such as ImmutableJS.  In the example below, the consumer is configuring the middleware to permit `immutable` structures and consider them to be serializable:

```js
// immutable version 3.8.2 assumed for this demo!
const isSerializable = (value: any): boolean =>
  Iterable.isIterable(value) || isPlain(value);

const getEntries = (value: any): [string, any][] =>
  Iterable.isIterable(value) ? value.entries() : Object.entries(value);

const middleware = createSerializableStateInvariantMiddleware({
  isSerializable,
  getEntries,
});
```

### DIFF
--- a/docs/api/otherExports.md
+++ b/docs/api/otherExports.md
@@ -20,6 +20,7 @@ Accepts an options object with `isSerializable` and `getEntries` parameters.  Th
 Example:
 
 ```js
+import { Iterable } from 'immutable';
 import {
   configureStore,
   createSerializableStateInvariantMiddleware,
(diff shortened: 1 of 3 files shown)


kgregory:
Additional commit to amend docs for new middleware option

mayteio:
Anything holding this back from being merged?

markerikson:
@mayteio : me, mostly :)  I've been swamped with work, spent the couple weeks prepping for a conference talk, just gave that, and will be doing more traveling the next few days.  I haven't had time to sit down and play maintainer for RSK.

Hopefully I'll have some time in the next couple weeks.

kgregory:
@markerikson haven't pushed on this because of exactly those reasons.  You do a crazy amount of good work that helps so many developers.  Thanks for that!

markerikson:
Okay, seems reasonable.  Merging.

PR #144, opened 2019-05-22
Docs: Fix createReducer example switch statement

krisaoe:
The createReducer example included invalid switch statement syntax.


markerikson:
Thanks!

PR #145, opened 2019-05-22
Docs: Fix mutating state example.

krisaoe:
The comparison example on how to "mutate" state was not targeting the correct state property.


markerikson:
Thanks!

Issue #148, opened 2019-06-21
createAction doesn't support a `meta` attribute


jmccarrell-lmi:
And as far as I can tell there is no way to possibly insert it.

not a big deal, any action that has meta attributes, you can just manually create.

If I were to make a proposal, it would be that it would be a second argument.

markerikson:
Yep, and it doesn't support customizing `payload` either, other than passing it in directly.  This is something I'd like to handle in some way before calling this 1.0.  See #82 .

markerikson:
The two main questions are how `createAction()` should accept callbacks for `payload` and `meta` customization, and how `createSlice()` should accept them.

The obvious inspirations here would be:

https://redux-actions.js.org/api/createaction#createactiontype-payloadcreator-metacreator

https://github.com/ericelliott/autodux#action-creators

Based on that, I'm thinking roughly:

```ts
type PayloadCreator = (...args) => Payload;
type MetaCreator = (...args) => Meta;

createAction(
    type : String,
    payloadCreator? : PayloadCreator ,
    metaCreator? : MetaCreator ,
)

createSlice(
    reducers: {
        simpleReducer: ReducerFunction,
        complexReducer: {
             payload? : PayloadCreator ,
             meta? : MetaCreator ,
             reducer : ReducerFunction
        } 
    }
)
```

Actual TS types to be figured out, of course, but you get the idea.

phryneas:
I'm just exploring options right now. Would something like
```typescript
type PrepareAction<OriginalPayload, Payload, Meta> = (originalPayload: OriginalPayload) => {payload: Payload; meta? : Meta}

createAction(
    type : String,
    prepare?: PrepareAction,
)

createSlice(
    reducers: {
        simpleReducer: ReducerFunction,
        complexReducer: {
             prepareAction: PrepareAction,
             reducer : ReducerFunction
        } 
    }
)
```

also be acceptable? That could be way easier to type.

TidyIQ:
How do we currently workaround the inability to customise the payload with arguments? This seems like an extremely crucial missing feature. Almost every reducer I have needs the payload to be adjusted based on the function parameters.

For example, a reducer that opens an error message snackbar and displays a message. The message can't be hardcoded into the action creator as obviously the error message changes based on what the actual error is.

Or a very common case - saving form input values to state.

markerikson:
Fixed by #149 and out as 0.6.3 .  Please let me know if that works for you!

[...]

PR #149, opened 2019-06-22
add prepareAction option to createAction

phryneas:
This is a rough WIP for the feature requested in #148. Essentially, it allows to pass a `prepare` function with the signature `(originalPayload: OriginalPayload) => {payload: Payload; meta? : Meta}` to modify the Payload and potentially add a Meta to the created action. You can see examples of use in the type tests I added.

This is working, but needs tuning around the types - but before that I would like some feedback if I'm going in the right direction with this.

@markerikson  @tanhauhau what do you think?

### DIFF
--- a/src/createAction.ts
+++ b/src/createAction.ts
@@ -18,8 +18,8 @@ export type PayloadAction<
 export type Diff<T, U> = T extends U ? never : T
 
 export type PrepareAction<P> =
-  | ((...args: any) => { payload: P })
-  | ((...args: any) => { payload: P; meta: any })
+  | ((...args: any[]) => { payload: P })
+  | ((...args: any[]) => { payload: P; meta: any })
 
 /**
  * An action creator that produces actions with a `payload` attribute.
@@ -30,7 +30,7 @@ export type PayloadActionCreator<
   PA extends PrepareAction<P> | void = void
 > = {
   type: T
-} & (PA extends (...args: any) => any
+} & (PA extends (...args: any[]) => any
   ? (ReturnType<PA> extends { meta: infer M }
       ? (...args: Parameters<PA>) => PayloadAction<P, T, M>
       : (...args: Parameters<PA>) => PayloadAction<P, T>)
@@ -78,6 +78,9 @@ export function createAction(type: string, prepareAction?: Function) {
   function actionCreator(...args: any[]) {
     if (prepareAction) {
       let prepared = prepareAction(...args)
+      if (!prepared) {
+        throw new Error('prepareAction did not return an object')
+      }
       return 'meta' in prepared
         ? { type, payload: prepared.payload, meta: prepared.meta }
         : { type, payload: prepared.payload }
--- a/src/createAction.ts
+++ b/src/createAction.ts
@@ -7,33 +7,50 @@ import { Action } from 'redux'
  * @template P The type of the action's payload.
  * @template T the type used for the action type.
  */
-export interface PayloadAction<P = any, T extends string = string>
-  extends Action<T> {
+export type PayloadAction<
+  P = any,
+  T extends string = string,
+  M = void
+> = Action<T> & {
   payload: P
-}
+} & ([M] extends [void] ? {} : { meta: M })
 
-export type Diff<T, U> = T extends U ? never : T;
+export type Diff<T, U> = T extends U ? never : T
+
+export type PrepareAction<P> =
+  | ((...args: any) => { payload: P })
+  | ((...args: any) => { payload: P; meta: any })
 
 /**
  * An action creator that produces actions with a `payload` attribute.
  */
-export type PayloadActionCreator<P = any, T extends string = string> = { type: T } & (
-  /*
-  * The `P` generic is wrapped with a single-element tuple to prevent the
-  * conditional from being checked distributively, thus preserving unions
-  * of contra-variant types.
-  */
-  [undefined] extends [P] ? {
-    (payload?: undefined): PayloadAction<undefined, T>
-    <PT extends Diff<P, undefined>>(payload?: PT): PayloadAction<PT, T>
-  }
-  : [void] extends [P] ? {
-    (): PayloadAction<undefined, T>
-  }
-  : {
-    <PT extends P>(payload: PT): PayloadAction<PT, T>
-  }
-);
+export type PayloadActionCreator<
+  P = any,
+  T extends string = string,
+  PA extends PrepareAction<P> | void = void
+> = {
+  type: T
+} & (PA extends (...args: any) => any
+  ? (ReturnType<PA> extends { meta: infer M }
+      ? (...args: Parameters<PA>) => PayloadAction<P, T, M>
+      : (...args: Parameters<PA>) => PayloadAction<P, T>)
+  : (/*
+     * The `P` generic is wrapped with a single-element tuple to prevent the
+     * conditional from being checked distributively, thus preserving unions
+     * of contra-variant types.
+     */
+    [undefined] extends [P]
+      ? {
+          (payload?: undefined): PayloadAction<undefined, T>
+          <PT extends Diff<P, undefined>>(payload?: PT): PayloadAction<PT, T>
+        }
+      : [void] extends [P]
+      ? {
+          (): PayloadAction<undefined, T>
+        }
+      : {
+          <PT extends P>(payload: PT): PayloadAction<PT, T>
+        }))
 
 /**
  * A utility function to create an action creator for the given action type
@@ -44,18 +61,35 @@ export type PayloadActionCreator<P = any, T extends string = string> = { type: T
  *
  * @param type The action type to use for created actions.
  */
+
 export function createAction<P = any, T extends string = string>(
   type: T
-): PayloadActionCreator<P, T> {
-  function actionCreator(payload?: P): PayloadAction<undefined | P, T> {
-    return { type, payload }
+): PayloadActionCreator<P, T>
+
+export function createAction<
+  PA extends PrepareAction<any>,
+  T extends string = string
+>(
+  type: T,
+  prepareAction: PA
+): PayloadActionCreator<ReturnType<PA>['payload'], T, PA>
+
+export function createAction(type: string, prepareAction?: Function) {
+  function actionCreator(...args: any[]) {
+    if (prepareAction) {
+      let prepared = prepareAction(...args)
+      return 'meta' in prepared
+        ? { type, payload: prepared.payload, meta: prepared.meta }
+        : { type, payload: prepared.payload }
+    }
+    return { type, payload: args[0] }
   }
 
-  actionCreator.toString = (): T => `${type}` as T
+  actionCreator.toString = () => `${type}`
 
   actionCreator.type = type
 
-  return actionCreator as any
+  return actionCreator
 }
 
 /**
--- a/src/createSlice.ts
+++ b/src/createSlice.ts
@@ -98,6 +98,17 @@ type CaseReducerActions<CR extends SliceCaseReducers<any, any>> = {
         : PayloadActionCreator<void>)
 }
 
+type NoInfer<T> = [T][T extends any ? 0 : never];
+type SliceCaseReducersCheck<S, ACR> = {
+    [P in keyof ACR] : ACR[P] extends {
+        reducer(s:S, action?: { payload: infer O }): any 
+    } ? {
+        prepare(...a:never[]): { payload: O }
+    } : {
+
+    }
+}
+
 function getType(slice: string, actionKey: string): string {
   return slice ? `${slice}/${actionKey}` : actionKey
 }
@@ -110,6 +121,9 @@ function getType(slice: string, actionKey: string): string {
  *
  * The `reducer` argument is passed to `createReducer()`.
  */
+export function createSlice<S, CR extends SliceCaseReducers<S, any>>(
+  options: CreateSliceOptions<S, CR> & { reducers: SliceCaseReducersCheck<S, NoInfer<CR>> }
+): Slice<S, CaseReducerActions<CR>>
 export function createSlice<S, CR extends SliceCaseReducers<S, any>>(
   options: CreateSliceOptions<S, CR>
 ): Slice<S, CaseReducerActions<CR>> {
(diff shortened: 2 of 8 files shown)


markerikson:
Skimming the code, I have two immediate thoughts.

First, I don't think this is quite what I had in mind.  What I'm really looking for is the ability to define a function that takes whatever parameters I declare, and turns those into the payload.  A couple hypothetical examples:

```ts
const payloadFromArgs = (a : string, b : number) => ({a, b});

const payloadIdGenerator = (name : string) => {
     const id = uuid();
     return {id, name};
}
```

Second: ow, the types are starting to hurt my eyes :(

tanhauhau:
I personally think that redux-action's `createAction` api is flexible and great, but man , when I look at [their types from DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/redux-actions/index.d.ts), it's kind of pain to read.

phryneas:
> Skimming the code, I have two immediate thoughts.
> 
> First, I don't think this is quite what I had in mind. What I'm really looking for is the ability to define a function that takes whatever parameters I declare, and turns those into the payload. Typically, folks who write action creators turn multiple arguments into a single object payload, or do some calculations in the action creator like generating an ID.

Ah, that's very valid, I didn't think of that. So if the prepareAction method would take multiple arguments, the general direction for the API would be okay?

> Second: ow, the types are starting to hurt my eyes :(

Yup :( Mine too :( 

Problem is: We can't really touch any exported pre-existing type other than adding more generics at the end because people might be using them. And I might have broken that rule a few times anyways. That's why I meant this still needs a lot of clean-up.
That leads (in addition to this already being very exhaustive types anyways) to weird generic ones, or new types that do almost the same. :/

Also, the distinction between PayloadAction and SliceAction doesn't really help. Do you know why that isn't one type?

phryneas:
> I personally think that redux-action's `createAction` api is flexible and great, but man , when I look at [their types from DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/redux-actions/index.d.ts), it's kind of pain to read.

Hm, yeah, that might also be an option. I think the types might only hurt that much because of backwards compatiblity to older TS versions. A problem would be getting a good API for `createSlice` that doesn't end up in a combinatoric explosion of types (only reducer, reducer with payloadCreator, reducer with metaCreator, reducer with payloadCreator and metaCreator - that might lead to a LOT of types :( )

markerikson:
I have no idea why any particular types exist in this lib atm - I didn't do the TS conversion.

As a general observation: breakages are never good, but this lib is pre-1.0 for a reason.  If it's necessary to break existing types to get things to work better, go ahead.

phryneas:
@markerikson  @tanhauhau 
So here's a second definition with the API initially described in #148.

https://github.com/reduxjs/redux-starter-kit/compare/master...phryneas:prepareAction2?expand=1

The TypeScript doesn't look much more readable :/

So, assuming adding the ...args in this branch wouldn't require much effort - what do you think? Which branch is worth continuing?

@tanhauhau - do you want to give it a try so we have a third alternative to choose from?

tanhauhau:
🤔 let me try..

markerikson:
Anyone had a chance to think through this further?

markerikson:
API-design wise: would it be better to have separate creator functions for `payload` and `meta`, or somehow have one function that _must_ return `payload` and _could_ return `meta`?

phryneas:
> Anyone had a chance to think through this further?

I was still waiting if @tanhauhau comes up with something. But if you decide to go either route I'd be happy to try and simplify one of mine solutions as well.

> API-design wise: would it be better to have separate creator functions for `payload` and `meta`, or somehow have one function that _must_ return `payload` and _could_ return `meta`?

Typing-wise, I still think it would be easier to have one function there instead of two, as there's already enough different conditions flying around.

There could also be a third alternative - what do you think of something like this? Not really sure if I like it myself but may be worth a thought.
https://codesandbox.io/s/createaction-api-experiment-fz0p1

tanhauhau:
@markerikson @phryneas sorry for keep you waiting, i've come up nothing. 😢
man its hard to type a flexible api

phryneas:
> @markerikson @phryneas sorry for keep you waiting, i've come up nothing. cry
> man its hard to type a flexible api

@tanhauhau no worries, this one is really hard to type.

@markerikson - have you come up with a decision on which API to settle? I'll happily try to simplify (or at least streamline a bit) the typings for any approach.

markerikson:
Mmm... yeah, let's go with something like `(...args) => ({payload, meta?})` for the `prepare` callback.

phryneas:
Okay, as so much changed with #157 , I completely redid the PR.

I also added a lot more tests, but there are still some cases where I'll need more tests.

I'm afraid I couldn't find a way to type it so that https://github.com/reduxjs/redux-starter-kit/pull/149/files#diff-87542dc1981a518809a25a787c658ee4R75 throws a type error (prepare result & reducer input not compatible). Might try a few other things in the next days though.

markerikson:
Hmm.  FWIW, the "prepare" thing shouldn't be part of `createReducer()` itself, as it's really a question of what the action creator accepts and returns.  So, just changes to `createAction` and `createSlice` should be necessary.

phryneas:
Good point. What would you think about adding something like https://github.com/joonhocho/tsdef as a dependency? That might make the types a lot more readable. (although that's no guarantee, might have to try)

markerikson:
I have no problems with _using_ some of those typedefs if they make sense.  But, I'd hesitate on adding it as a specific dependency.  Seems like that kind of thing would be easily copy-pastable into our own code instead.

markerikson:
@phryneas : think you have time to push this ahead?  I'd love to get it in relatively soon.

phryneas:
@markerikson had a chaotic week but I have some spare hours this evening so I'll put some time in.

markerikson:
No worries, just checking!

phryneas:
@markerikson okay, by now this should be fairly complete.

Now there's just the question of that one failing test. I'd love to have that behaviour of an error being thrown there, but I don't know how to type it - if the `reducer` and `prepare` methods have differing Payload types, TS just seems to widen the type to `any` instead of showing an error.

The only idea to solve that would be to export another `enhancedCaseReducer` method that essentially looks like

```typescript

function enhanceCaseReducer<S, P>(
  reducer: CaseReducer<S, PayloadAction<P>>,
  prepare: PrepareAction<P>
): EnhancedCaseReducer<S, PayloadAction<P>> {
  return { reducer, prepare };
}

```

and people would use it like this:

```typescript

createSlice({
  slice: 'test',
  initialState: 0;
  reducers: {
    incrementMultiple: enhanceCaseReducer(
      (state, action: PayloadAction<number>) => state + action.payload,
      (a: number, b: number)  => ({payload: a+b})
    )
  }
})
```

that would be typesafe.

What do you think?

markerikson:
I'd really rather not have any special exports just for TS usage.

Stupid TS question: is it somehow possible to throw an error if a type has turned into `any` or something?

phryneas:
> Stupid TS question: is it somehow possible to throw an error if a type has turned into `any` or something?

It couldn't distinguish between loosening an inferred type vs. manually typing as any. So if we did something like that, it wouldn't be possible to manually any-type enhanced reducers any more. 

So I'd suggest against that.

Maybe I'm just missing something with the typings there - but I've tried about 10 different versions and I'm pretty much out of ideas. Are there other people you can ping on that? Maybe a second opinion could really help.

markerikson:
Tweeted for help, and also paging the usual suspects:  @Dudeonyx , @denisw , @jessidhia

nickmccurdy:
I read the thread and I’ll try to help, but I don’t think I’m understanding the goal here. Why not just do the ...arts form as @markerikson originally suggested and let users figure out how that maps to their action creator type? And what’s an example of code that would have a false positive in the typechecker with the current type implementation (where two different payloads match and infer to any when they shouldn’t match)?

markerikson:
@nickmccurdy : see this non-working types test:

https://github.com/reduxjs/redux-starter-kit/pull/149/files#diff-d3f6ba32c4f3b495b630095111a617e9R162-R186

The `prepare` callback is returning a number, while the corresponding reducer is expecting a string.  Instead of erroring, this is falling back to `any`.

nickmccurdy:
That link seems to be broken (just goes to top of PR diff), is this what you're referring to? https://github.com/reduxjs/redux-starter-kit/pull/149/files#diff-d3f6ba32c4f3b495b630095111a617e9R71-R76

markerikson:
bah, that link doesn't work either :)

Direct file link:

https://github.com/reduxjs/redux-starter-kit/blob/8ec73956be2e685a2b7ce0a92544188c88a3b3c0/type-tests/files/createSlice.typetest.ts#L162-L186

nickmccurdy:
It's not widening the type. The types don't pass generic type parameters properly, which disables the inference.

The type of the reducer property is not given in this test code, so it is inferred and matched against the type `CaseReducer<S, A>`.  With no generic parameter type given for `A`, it defaults to `PayloadAction` because of  `A extends PayloadAction`. With no generic parameter type given to `PayloadAction`, it defaults to `any` because of `P = any`.

I believe you should be able to do something like this: `type EnhancedCaseReducer<S, A extends PayloadAction<P>, P> = {`

nickmccurdy:
On an unrelated note regarding maintenance, the `types.test.ts` implementation makes it difficult to troubleshoot what type error is happening in the tests. Instead of running TypeScript declaration tests dynamically in a test framework, I'd recommend simply running `tsc` on them. It will give the appropriate status code to fail the build if there are type errors or assertion failures in comments.

phryneas:
> It's not widening the type, it's using the default type parameter value of `any` because no type parameter is given to be inferred.
> 
> The type of the reducer property is not given in this test code, so it is inferred and matched against the type `CaseReducer<S, A>`. With no generic parameter type given for `A`, it defaults to `PayloadAction` because of `A extends PayloadAction`. With no generic parameter type given to `PayloadAction`, it defaults to `any` because of `P = any`.
> 
> `type EnhancedCaseReducer<S, A extends PayloadAction<P>, P> = {`

Hm, while that makes sense, it doesn't seem to be working for me - could you share a working example?

nickmccurdy:
Sorry, this isn't nearly as simple as I thought. I've been trying to pass all these types in `createSlice.ts` so they wouldn't default to any, and I seem to have the same issue.

phryneas:
@dragomirtitian solved this - https://twitter.com/TitianCernicova/status/1155549716777709570

Thank you so much Titian!

markerikson:
Yay!  Busy with some other stuff atm, but I'll have to try it out in the next day or two to see how it behaves.

phryneas:
No hurry :)

src/createAction.ts
```diff
@@ -7,33 +7,50 @@ import { Action } from 'redux'
...
-  }
-);
+export type PayloadActionCreator<
+  P = any,
+  T extends string = string,
+  PA extends PrepareAction<P> | void = void
+> = {
+  type: T
+} & (PA extends (...args: any[]) => any
+  ? (ReturnType<PA> extends { meta: infer M }
```
markerikson: Just so you know, my eyes are still bleeding here :)

src/createSlice.ts
```diff
@@ -72,12 +77,36 @@ export interface CreateSliceOptions<
...
   [T in keyof CR]: CR[T] extends (state: any) => any
-  ? void
-  : (CR[T] extends (state: any, action: PayloadAction<infer P>) => any
-    ? P
-    : void)
+    ? PayloadActionCreator<void>
+    : (CR[T] extends (state: any, action: PayloadAction<infer P>) => any
+        ? PayloadActionCreator<P>
+        : CR[T] extends { prepare: PrepareAction<infer P> }
+        ? PayloadActionCreator<P, string, CR[T]['prepare']>
```
markerikson: wat

src/createSlice.ts
```diff
@@ -72,12 +77,36 @@ export interface CreateSliceOptions<
...
+        : PayloadActionCreator<void>)
+}
+
+type NoInfer<T> = [T][T extends any ? 0 : never];
```
markerikson: voodoo

markerikson:
Welp, I seriously have no "real" feedback to offer on this PR.  The type shenanigans y'all are capable of are well and truly beyond me.

But, the examples and the tests look pretty comprehensive, so I'm going to assume this actually works at this point.

Thank you _so_ much for putting this one together!  Lots of folks have been asking for this.

markerikson:
Smacking the button!

src/createSlice.ts
phryneas: Yeah, I'm still "wat" here, too :)

src/createAction.ts
phryneas: Yeah, that one is wild. But it defines many different behaviours :D

src/createSlice.ts
phryneas: Oh, come on, THAT one isn't even so much different than before =D

Big difference is that it builds the complete ActionCreator type instead of just the payload, and also it now has 4 cases instead of 3.

src/createAction.ts
matthew-gerstman: FWIW here's how I understand this code. 

PayloadActionCreator always has a `type` property of _type_ `T`

If PA is a function we check if it has a `meta` field. If yes the final type is 
```
{
  type: T extends string,
  meta: M,
  payload: P
}
  ```

If the return value does not have a `meta` field it's
```
{
  type: T extends string,
  payload: P
}
  ```

Now the else, if PA is `not` a function we default back to the older behavior which terrifies me.

@phryneas as some practical feedback, it might be helpful to make it clear to the reader
1. where the if/else statements are
2. what the final type will be in each scenario.

phryneas: Almost, but no.

All Action creators share the `{ type: T }`, but the rest of the functionality differs.

If a `PA` argument is passed, the first branch will be taken.

Within that branch, if PA returns something with `{ meta }`, again take the first branch. Resulting type is a function `(...args: Parameters<PA>) => PayloadAction<P, T, M>` (with the additional `type` from above -  all that follow will, too). Otherwise it's a function `(...args: Parameters<PA>) => PayloadAction<P, T>)`.

Now, on to the other branch - no PA given. This has to handle some special cases first and then falls back to the most used case.

If P is undefined, it returns a method that might take 0 or 1 argument. If P is void, it's a method with 0 arguments. Otherwise it is a method with 1 argument. All these methods infer their argument and return that inferred argument as the payload of the returned action.

Formatting this is in a language that knows only ternaries and in a file that will be shuffled around with prettier is unfortunately something I did not achieve :/

Issue #157, opened 2019-07-15
Action creators from `createSlice` are not inferred as `PayloadActionCreator`


markerikson:
I'm finally getting a chance to use RSK myself in a real project, and I just ran into an interesting issue.

In `createAction()`, we specifically attach the generated type to the generated action creator so that we can use the action creator itself as the type value, by both overriding `toString()` and adding it as `actionCreator.type = type`.

I've got some code I just wrote that looks like:

```ts
const someSlice = createSlice({
  /* omit */
});

const {doSomething} = someSlice.actions;

// later, in a middleware
switch(action.type) {
    case doSomething.type: { // error: Property 'type' does not exist on type '(payload: SomePayload) => PayloadAction<SomePayload, string>'

    }
}
```

So, even though we're using `createAction()` inside of `createSlice`, there's a disconnect in the types.

In our TS code, we specifically have [a `PayloadActionCreator` type](https://github.com/reduxjs/redux-starter-kit/blob/2eb5b4bf49a44e64757e54d64d9593e7ce7d74da/src/createAction.ts#L18-L22) that appears to define itself as "a function with this payload, plus the `type` field".

However, I don't see that `PayloadActionCreator` type being used in `createSlice` anywhere.  Instead, there's [a `SliceActionCreator` type](https://github.com/reduxjs/redux-starter-kit/blob/2eb5b4bf49a44e64757e54d64d9593e7ce7d74da/src/createSlice.ts#L13-L15) that reuses the `PayloadAction` definition.

Seems like `SliceActionCreator` ought to be referring to `PayloadActionCreator` somehow?

markerikson:
/cc @denisw, @jessidhia, @Dudeonyx , @phryneas :  

any ideas?

markerikson:
I just tried specifically hand-editing the typings for `SliceActionCreator` in my local app codebase `node_modules` to be:

```ts
export declare type SliceActionCreator<P> = P extends void ? PayloadActionCreator<void> : PayloadActionCreator<P>;
```
And that seems to compile okay.

phryneas:
Those two ActionCreators seem to have diverged on the zero-argument-case - even though they actually are exactly the same internally. 
Drives me nuts since forever, actually :) (See my comment here: https://github.com/reduxjs/redux-starter-kit/pull/149#issuecomment-504729239)

I guess I'll try to make those behave similarly when I touch them for #149 anyways :)

markerikson:
[...]

PR #158, opened 2019-07-15
infer action creators from `createSlice`  as `PayloadActionCreator`

phryneas:
see #157 

This changes the createSlice types to use PayloadActionCreator and modifies the behaviour of PayloadActionCreator to account for `P = void` and `P = Something | undefined`.

Here are some examples:

```typescript

const someNumber: number = 5;
const someString: string = "";

const voidActionCreator = createAction<void>('a');
voidActionCreator(); // Action<'a'>
voidActionCreator(undefined); // Action<'a'>

const stringActionCreator = createAction<string>('b');
stringActionCreator('x'); // PayloadAction<'x', 'b'>

const optionalStringActionCreator = createAction<string | undefined>('c');
optionalStringActionCreator(); // Action<'c'>
optionalStringActionCreator(undefined); // Action<'c'>
optionalStringActionCreator('x') // PayloadAction<'x', 'c'>

const mixedActionCreator = createAction<string | number>('d');
mixedActionCreator('x') // PayloadAction<'x', 'd'>
mixedActionCreator(someNumber) // PayloadAction<number, 'd'>
mixedActionCreator(5) // PayloadAction<5, 'd'>
mixedActionCreator(someString) // PayloadAction<string, 'd'>

const anyActionCreator = createAction('e');
anyActionCreator() // Action<'e'>
anyActionCreator(someNumber) // PayloadAction<number, 'e'>
anyActionCreator(5) // PayloadAction<5, 'e'>
anyActionCreator(someString) // PayloadAction<string, 'e'>
```

so payload types are now interfered much more accurate, which is especially useful for the `<any>` case.

There is one open question open regarding the typing of the `undefined` argument: if `undefined` is not omitted, but explicitly specified, should we return an Action or a PayloadAction<undefined>? (See https://github.com/reduxjs/redux-starter-kit/pull/158/files#diff-437575fb1d5c39e3edd424ab38899f8eR27 )

PS: I could easily do away with the excessive type interference, but that would hurt the  `<any>` case a bit in my eyes and I think it won't hurt, so I'm looking forward to your opinion on it ;)

### DIFF
--- a/type-tests/files/createSlice.typetest.ts
+++ b/type-tests/files/createSlice.typetest.ts
@@ -81,7 +81,7 @@ import {
 
 
 /*
- * Test: Slice action creator types properties are correct
+ * Test: Slice action creator types properties are "string"
  */
 {
   const counter = createSlice({
@@ -97,7 +97,12 @@ import {
     }
   })
 
-  const x: "increment" = counter.actions.increment.type;
-  const y: "decrement" = counter.actions.decrement.type;
-  const z: "multiply" = counter.actions.multiply.type;
+  const s: string = counter.actions.increment.type;
+  const t: string = counter.actions.decrement.type;
+  const u: string = counter.actions.multiply.type;
+
+  // typings:expect-error
+  const x: "counter/increment" = counter.actions.increment.type;
+  // typings:expect-error
+  const y: "increment" = counter.actions.increment.type;
 }
--- a/type-tests/files/createSlice.typetest.ts
+++ b/type-tests/files/createSlice.typetest.ts
@@ -77,3 +77,27 @@ import {
   // typings:expect-error
   counter.actions.multiply('2')
 }
+
+
+
+/*
+ * Test: Slice action creator types properties are correct
+ */
+{
+  const counter = createSlice({
+    slice: 'counter',
+    initialState: 0,
+    reducers: {
+      increment: state => state + 1,
+      decrement: state => state - 1,
+      multiply: (state, { payload }: PayloadAction<number | number[]>) =>
+        Array.isArray(payload)
+          ? payload.reduce((acc, val) => acc * val, state)
+          : state * payload
+    }
+  })
+
+  const x: "increment" = counter.actions.increment.type;
+  const y: "decrement" = counter.actions.decrement.type;
+  const z: "multiply" = counter.actions.multiply.type;
+}
--- a/src/createAction.ts
+++ b/src/createAction.ts
@@ -12,14 +12,43 @@ export interface PayloadAction<P = any, T extends string = string>
   payload: P
 }
 
+export type Diff<T, U> = T extends U ? never : T;
+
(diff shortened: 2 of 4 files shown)


markerikson:
I vote that `createAction()` should always return a `PayloadAction`, for consistency (and on the grounds that we'll also probably end up throwing `meta` in there too in the near future too).

type-tests/files/createSlice.typetest.ts
```diff
@@ -77,3 +77,27 @@ import {
...
+      increment: state => state + 1,
+      decrement: state => state - 1,
+      multiply: (state, { payload }: PayloadAction<number | number[]>) =>
+        Array.isArray(payload)
+          ? payload.reduce((acc, val) => acc * val, state)
+          : state * payload
+    }
+  })
+
+  const x: "increment" = counter.actions.increment.type;
```
markerikson: Why are these not `"counter/increment"`?  That's really confusing to me.

markerikson:
One other question to throw out there.  If we do eventually tack on `meta`, is `PayloadAction` really the best term?  Should it maybe be `FSA` instead or something?

type-tests/files/createSlice.typetest.ts
phryneas: Good catch.
Because I typed them wrongly and those tests only check for types at compile time, not validate that it is the real output.
That said, I'll have to remove that and only type the `type` as string then - can't do concatenation with types :(

phryneas:
> One other question to throw out there. If we do eventually tack on `meta`, is `PayloadAction` really the best term? Should it maybe be `FSA` instead or something?

I'm all for that. :+1:

phryneas:
PS:  Generally, I'm for a renaming, but having changed it to FSA everywhere in the code, it looks really more like a generic than a type to me, with it being only capital letters. Not really sure if that would be good for readablility. Do we have other options?

I was thinking along the lines of:

* FSAction
* ExtendedAction
* SpecifiedAction
* DetailedAction

But reading all those above, I have to say that I'd also be fine with keeping PayloadAction and just adding a third parameter for `meta`... Not 100% happy with any of those.

markerikson:
Eh, let's keep `PayloadAction` for now.  Less churn.

markerikson:
Cool.  All right, let's get this out.  Thank you so much for working on this!

phryneas:
Happy to help  :sunglasses: 

Just because I noticed, in case you forgot (I don't know your release workflow):

While the 0.6 is available on npm, the commit & tag are still missing on GitHub. Did you forget to push?

markerikson:
Not quite.  I actually did the release from work, but I'm behind a corporate proxy.  I use https://github.com/sindresorhus/np to do my releases, but it seems that doesn't work great behind proxies.  It was able to hit NPM to publish, but not do the Github work.

I'll get those up tonight after I get home.

PR #160, opened 2019-07-16
allow enhancing of dispatch method type

phryneas:
As per https://twitter.com/phry/status/1151252237403992064

What do you think?

### DIFF
--- a/src/configureStore.ts
+++ b/src/configureStore.ts
@@ -94,10 +94,13 @@ export interface ConfigureStoreOptions<S = any, A extends Action = AnyAction> {
  * A Redux store returned by `configureStore()`. Supports dispatching
  * side-effectful _thunks_ in addition to plain actions.
  */
-export interface EnhancedStore<S = any, A extends Action = AnyAction>
+export interface EnhancedStore<S = any, A extends Action = AnyAction, ED extends EnhancedDispatch = UseThunkDispatch>
   extends Store<S, A> {
-  dispatch: ThunkDispatch<S, any, A>
+  dispatch: ED extends UseThunkDispatch ? ThunkDispatch<S, any, A> : ED;
 }
+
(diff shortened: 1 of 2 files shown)


phryneas:
I want to add that in most cases this would not be very useful, because the `connect`-injected Dispatch and the `useDispatch`-returned Dispatch will be the normal Dispatch by default.

This can only get useful in combination with the tips mentioned in
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-redux/index.d.ts#L482
for `useDispatch` and not at all for `connect` calls (I don't see a way on how to inject Dispatch there).



### A completely different way
of solving the problem would by augmenting the original `Dispatch` type definition. This would be done by creating a `.d.ts` in a folder referenced by `tsconfig.json`'s `typeRoots` or in a file that is included or referenced by a triple-slash directive.

```typescript
import { Action, AnyAction, Dispatch } from "redux";
declare module "redux" {
  interface Dispatch<A extends Action = AnyAction> {
    (action: Action<"baz">): Promise<"bar!">;
  }
}
```

markerikson:
Was this something you were actually envisioning being merged in, or just a proof of concept for something?  (Yes, I remember the question and discussion I originally asked that led to you working on this, just not sure what the goal was for this PR specifically.)

phryneas:
Not 100% while it might be nice (especially because it is an easy way to *remove* the thunk dispatch if one doesn't actually use the thunk middleware, like I do most of the time), I guess it would not be used by many and add another level of complexity to the types. 

Maybe adding the "A completely different way" part somewhere in a "Typescript Tricks" chapter of the documentation and call it a day instead?

markerikson:
Yeah, something to note down the road.  I'll go ahead and close this, but thanks for putting it together.

PR #162, opened 2019-07-22
Suggestion : Improve current error message

Taranys:
I lost 15min on my first try of this nice library just because I made a typo ("reducers" instead of "reducer") and I didn't understand why my reducer was reject by configureStore 😢 

I suggest two error messages with the configureStore to help newbie like me 😄 

I'm not a good english speaker so don't hesitate to suggest a better message if you think this PR can be helpful

### DIFF
--- a/src/configureStore.test.ts
+++ b/src/configureStore.test.ts
@@ -71,6 +71,16 @@ describe('configureStore', () => {
   describe('given no reducer', () => {
     it('throws', () => {
       expect(configureStore).toThrow(
+        '"reducer" argument is mandatory. Check that you provide it and you do not make a typo.'
+      )
+    })
+  })
+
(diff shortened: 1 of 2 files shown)


markerikson:
Which specific error message did you see before this PR?

Taranys:
I saw "Reducer argument must be a function or an object of functions that can be passed to combineReducers". So I thought I made a mistake on my configuration, I try to add combineReducer, to create my own method... I understood after 15m that the reducer received by RSK was undefined 😅

markerikson:
Swinging back around to this.  The PR actually adds a second error message so that the first one never gets thrown, so I'm going to close this.  However, I'll tweak the wording and clarify that `reducer` is required.  Thanks!

PR #164, opened 2019-07-28
Add Basic tutorial

markerikson:
I've been saying for way too long that I was going to start writing some tutorials for RSK.  I've done some prep work intermittently over the last few weeks, and tonight I finally sat down and wrote a first "Basic Tutorial" page.

The planned outline:

- "Basic": quick intro to the APIs using a counter app
- "Intermediate": show how to convert an existing Redux app to RSK, by converting the standard Redux "todos" example.  This will also demonstrate use of Immer-powered reducers to "mutate" state, and a basic example of `createSelector`.  I may try to squeeze in a thunk usage here too, even though the original doesn't have any.  I originally recreated it from scratch at https://github.com/markerikson/rsk-todos-example , but I'll probably redo that to start from the original Redux example as a starting point so I can show the conversion process instead.
- "Advanced": I rewrote @dceddia 's "Github Issues" example app using RSK and React-Redux hooks (see https://github.com/markerikson/rsk-github-issues-experiment ).  This will demonstrate larger-scale use of slices and usage with TypeScript.

### DIFF
--- a/docs/tutorials/basic-tutorial.md
+++ b/docs/tutorials/basic-tutorial.md
@@ -1,0 +1,8 @@
+---
+id: basic-tutorial
+title: Basic Tutorial
+sidebar_label: Basic Tutorial
+hide_title: true
+---
+
+# Basic Tutorial: Building a React-Redux Todo App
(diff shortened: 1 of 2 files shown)


markerikson:
In the interest of getting this out sooner rather than later, I'm gonna merge it in now, and do the other tutorials as follow-ups when I write them.

PR #168, opened 2019-08-03
refactor typings for readability

phryneas:
This should make the typings more readable for someone just diving into this lib's typings, as we already discussed on twitter :)

### DIFF
--- a/src/createAction.ts
+++ b/src/createAction.ts
@@ -6,20 +6,37 @@ import { Action } from 'redux'
  *
  * @template P The type of the action's payload.
  * @template T the type used for the action type.
+ * @template M The type of the action's meta (optional)
  */
 export type PayloadAction<
   P = any,
   T extends string = string,
   M = void
-> = Action<T> & {
-  payload: P
-} & ([M] extends [void] ? {} : { meta: M })
-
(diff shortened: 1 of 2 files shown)


src/createAction.ts
```diff
@@ -6,51 +6,55 @@ import { Action } from 'redux'
...
+export type ActionCreatorWithoutPayload<T extends string = string> = WithTypeProperty<T, () => PayloadAction<undefined, T>>;
+
+export type ActionCreatorWithPayload<P, T extends string = string> =
+  WithTypeProperty<T, <PT extends P>(payload: PT) => PayloadAction<PT, T>>;
+
 /**
  * An action creator that produces actions with a `payload` attribute.
  */
 export type PayloadActionCreator<
   P = any,
```
phryneas: I could go and rename P to payload, T to Type and PA to PrepareActionFn here, but I'm not really sure if it makes things better or worse in the usages of these below.
Do you have an opinion there?

phryneas:
@markerikson what's your status on this? From my side this would be ready if you don't have any additional comments.

phryneas:
Okay :) Just wanted to make sure this didn't get buried ^^

markerikson:
Sure, let's do this.

Issue #165, opened 2019-07-29
Typescript errors when passing null as payload to actions returned from createSlice


marko911:
When using actions created from `createSlice`, Typescript complains with: 

`Argument of type 'null' is not assignable to parameter of type '{}'` .
But if you just use actions from `createAction` , it compiles. 

[Sandbox example](https://codesandbox.io/embed/black-firefly-uhweh)

markerikson:
tagging @phryneas @Dudeonyx @denisw @jessidhia

phryneas:
Okay, this seems to come from the fact that you are using TS 3.3 - in TS 3.5, the action type infers to a `PayloadActionCreator<unknown, string, void>` which in fact accepts `null`. Older versions of TS default to `{}` instead of `unknown`.

I'll try and see if we can have it default to `unknown` in older TS versions. But adding a test for this will be impossible if we don't want to downgrade TS for all tests (although it might be a good idea to run the tests in different TS versions in a separate ticket - what do you think @markerikson ?)

phryneas:
see #166 - could you please test this?

marko911:
Ok I updated my TS versio from 3.4 to 3.5.3 and it solves the problem yes.

markerikson:
Hopefully resolved by #166 .  Merged, and will probably get that change out in 0.8 .

PR #174, opened 2019-08-14
Use `void` instead of `any` for undefined payloads

Krisztiaan:
Additional layer of type security, don't let the user implicitly opt out of type checking

### DIFF
--- a/type-tests/files/createAction.typetest.ts
+++ b/type-tests/files/createAction.typetest.ts
@@ -19,7 +19,7 @@ function expectType<T>(p: T): T {
 }
 
 /*
- * Test: PayloadAction type parameter is optional (defaults to `any`).
+ * Test: PayloadAction type parameter is required.
  */
 {
   // typings:expect-error
@@ -61,12 +61,10 @@ function expectType<T>(p: T): T {
       payload
     }),
     { type: 'action' }
-  ) as PayloadActionCreator<number>
+  ) as PayloadActionCreator<number | undefined>
 
   expectType<PayloadAction<number>>(actionCreator(1))
-  // typings:expect-error
   expectType<PayloadAction<undefined>>(actionCreator())
-  // typings:expect-error
   expectType<PayloadAction<undefined>>(actionCreator(undefined))
 
   // typings:expect-error
@@ -118,12 +116,9 @@ function expectType<T>(p: T): T {
  * Test: createAction() type parameter is required, not inferred (defaults to `void`).
  */
 {
-  const increment = createAction<number>('increment')
+  const increment = createAction('increment')
+  // typings:expect-error
   const n: number = increment(1).payload
-  // typings:expect-error
-  const s: string = increment('1').payload
-  // typings:expect-error
-  const t: string = increment(1).payload
 }
 /*
  * Test: createAction().type is a string literal.
--- a/type-tests/files/createAction.typetest.ts
+++ b/type-tests/files/createAction.typetest.ts
@@ -22,8 +22,11 @@ function expectType<T>(p: T): T {
  * Test: PayloadAction type parameter is optional (defaults to `any`).
  */
 {
+  // typings:expect-error
   const action: PayloadAction = { type: '', payload: 5 }
+  // typings:expect-error
   const numberPayload: number = action.payload
+  // typings:expect-error
   const stringPayload: string = action.payload
 }
 
@@ -31,7 +34,7 @@ function expectType<T>(p: T): T {
  * Test: PayloadAction has a string type tag.
  */
 {
-  const action: PayloadAction = { type: '', payload: 5 }
+  const action: PayloadAction<number> = { type: '', payload: 5 }
 
   // typings:expect-error
   const action2: PayloadAction = { type: 1, payload: 5 }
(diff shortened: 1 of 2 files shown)


markerikson:
Hmm. A bunch of type tests fail because of this change.  I'm not clear what improvement this is supposed to make.

Going to close this for now.  If you still feel it's a point of concern, please let me know and we can discuss further.

Krisztiaan:
@markerikson thanks for taking a look, we've made this exact patch in our project using this library, and it revealed a bunch of missed or mismatched action payload typings, while forcing us to specify payload where it was expected. I would look into fixing the tests if we can agree this would be beneficial to have merged.

markerikson:
Sure, sounds good.

markerikson:
@phryneas : since you're consulting, any thoughts on this one?

markerikson:
So looking at the type errors, some of them from "raw" uses of `PayloadAction` like this:

```ts
  const action: PayloadAction = { type: '', payload: 5 }
  const numberPayload: number = action.payload
```

Also:

```ts
  const increment = createAction('increment')
  // Error: Argument of type '1' is not assignable to parameter of type 'void | undefined'
  const n: number = increment(1).payload
```

Those are fixable by adding generic to `PayloadAction<number>` and `createAction<number>`.  Potentially reasonable.  But, I'm not entirely sure that's "right" here.

phryneas:
Well, it would definitely change the defined behaviour.

Especially [this test](https://github.com/reduxjs/redux-starter-kit/blob/master/type-tests/files/createAction.typetest.ts#L22) defines that a PayloadAction should fall back to `any` if there's no type defined.  And #165 shows that people are using it like this.

I guess all other of those tests are building on this "acceptance criteria" that was formulated in this first test.

As for me, I'd like this to change, as those types being more strict will prevent bugs like people accidentally forgetting the type parameter and then doing werid stuff. 
But this will be breaking for people.

Your call if you want this :)

Krisztiaan:
> Sure, sounds good.
> 
> Hmm. I don't suppose this is related to #211 at all?

This does not seem related, as the PR would only cover the `default` payload type that is used when nothing is supplied

markerikson:
I guess it would help if I better understand the ramifications of the type change, in terms of end-user experience, actual code usage, and tests that would have to change.

phryneas:
I think basarat describes this a whole lot better than me: https://basarat.gitbooks.io/typescript/docs/options/noImplicitAny.html

The point is: he describes `noImplicitAny`, which is a compiler option the user could turn on or off (they should, TS becomes pretty much useless without it). But we're a lib here, we can't leave that decision for the user and we can't detect if the user uses `noImplicitAny`, we have to make a decision for everyone.
 
The question is *do we allow all our users to do something potentially dangerous by accident, or do we require them to **explicitly** do that dangerous thing if they really want to?*

* Either we allow people to allow `createAction` without type argument, which then defaults to `createAction<any>` and eliminates essentially all type checks along the way. (This might cause a lot of bugs for ppl who normally use it but just forget it once in a while)
* Or we allow no-one to use `createAction` without a type argument (by making it useless, defaulting to something like void). They could still use `createAction<any>` if they really wanted, but they's have to specify that. (This might be a little inconvenient for ppl who use TS but don't care about type safety?)

As for the tests: currently, it seems to be a "business requirement" that it works as it currently does. I don't know who initially formulated it like this, and what they had in minde 🤷. I guess a few test would need to be replaced with other tests. But I also guess that would be fine.

markerikson:
Yeah, I think I can get behind making the typing more explicit here, especially since _most_ folks would be using `createSlice` anyway vs calling `createAction` themselves.

Would one of you be able to tackle this?  I'd want to get it into 0.8 as well, but I want to start working on updating the tutorial contents.

markerikson:
Okay, I re-created the branch off of `v0.8-integration`, and fixed up the type tests so they pass.  I'm not 100% sure the type test changes are _correct_, though.  Could both of you look them over?

type-tests/files/createAction.typetest.ts
```diff
@@ -22,16 +22,19 @@ function expectType<T>(p: T): T {
  * Test: PayloadAction type parameter is optional (defaults to `any`).
```
phryneas: `defaults to any` -> `should not default to any`

type-tests/files/createAction.typetest.ts
```diff
@@ -58,10 +61,12 @@ function expectType<T>(p: T): T {
       payload
     }),
     { type: 'action' }
-  ) as PayloadActionCreator
+  ) as PayloadActionCreator<number>
```
phryneas: I think this should be `number|undefined`

type-tests/files/createAction.typetest.ts
```diff
@@ -58,10 +61,12 @@ function expectType<T>(p: T): T {
...
 
   expectType<PayloadAction<number>>(actionCreator(1))
+  // typings:expect-error
```
phryneas: and these two should actually not fail then

type-tests/files/createAction.typetest.ts
```diff
@@ -110,22 +115,21 @@ function expectType<T>(p: T): T {
 }
 
 /*
- * Test: createAction() type parameter is optional (defaults to `any`).
+ * Test: createAction() type parameter is required, not inferred (defaults to `void`).
```
phryneas: this test doesn't really make any more sense (description does not match what it tests right now), so I would delete it

markerikson: yeah, lemme rewrite this one

phryneas:
Looks good from my side.

markerikson:
great, thanks!

PR #175, opened 2019-08-17
add ignoredActions to createSerializableStateInvariantMiddleware

### LINKED ISSUE #121: Question: How to use this with redux-persist?
I've been experimenting with getting `redux-persist` to integrate with my current project that's using `react-starter-kit` but I'm already hitting some walls and obscure errors.

*First issue* is a middleware that doesn't seem to know how to handle the register call for the persist middleware (which is caused by `serializable-state-invariant-middleware`).

*Second issue* is that `PersistGate` never renders, I can only assume this is because rehydrate is never called? I'm not really sure what's going on there, it's probably by own fault.

If this isn't the place to ask I'm sorry, but using `redux-persist` seems like a very common use case.

### Update 1:

I solved the `serializable-state-invariant-middleware` issue by simply adding thunk myself. **(Note it needs to be manually installed!)**

Here's my store.ts:
```typescript
import logger from "redux-logger";
import { configureStore } from "redux-starter-kit";
import thunk from "redux-thunk";
import rootReducer from "./reducers";

    reducer: rootReducer,
    middleware: [thunk, logger],
});

export default store;
```

### LINKED ISSUE #150: Persisting with redux-persist
I created the store using configureStore
I have an error when using redux-persist  tries to rehydrate my store.

Error:

> A non-serializable value was detected in an action, in the path: `%s`. Value: %o
> Take a look at the logic that dispatched this action:  %o.
> (See https://redux.js.org/faq/actions#why-should-type-be-a-string-or-at-least-serializable-why-should-my-action-types-be-constants), register, [Function register], Object {
>   "register": [Function register],
>   "rehydrate": [Function rehydrate],
>   "type": "persist/PERSIST",
> }

Is redux-persist even supported? I haven't seen any samples with redux persist integration.

ramyareye:
as a fix to error below while using redux-persist :

`A non-serializable value was detected in an action, in the path: register...`

I tried to add an `ignoredActions` array for cases like `persist/PERSIST`

a working todo example can be found [here](https://github.com/ramyareye/rsk-convert-todos-example)

fixes #121
fixes #150
fixes rt2zz/redux-persist#988

### DIFF
--- a/docs/api/otherExports.md
+++ b/docs/api/otherExports.md
@@ -15,7 +15,7 @@ Redux Starter Kit exports some of its internal utilities, and re-exports additio
 
 Creates an instance of the `serializable-state-invariant` middleware described in [`getDefaultMiddleware`](./getDefaultMiddleware.md).
 
-Accepts an options object with `isSerializable` and `getEntries` parameters.  The former, `isSerializable`, will be used to determine if a value is considered serializable or not. If not provided, this defaults to `isPlain`.  The latter, `getEntries`, will be used to retrieve nested values.  If not provided, `Object.entries` will be used by default.
+Accepts an options object with `isSerializable` and `getEntries` parameters.  The former, `isSerializable`, will be used to determine if a value is considered serializable or not. If not provided, this defaults to `isPlain`.  The latter, `getEntries`, will be used to retrieve nested values.  If not provided, `Object.entries` will be used by default. `ignoredActions` accepts an array of action types to ignore from serialization check
 
 Example:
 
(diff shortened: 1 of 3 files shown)


ramyareye:
@markerikson sry for mention, would you please take a look at this 
thanks

majid-vaghari:
@markerikson Same issue here. Can you take a look?

markerikson:
I'm fine with the idea of adding another option to the serializability middleware.  However, I don't like the way the options are being added to `getDefaultMiddleware`.

Let me see if I can rework this.

markerikson:
Resolved by #192 .

Issue #82, opened 2019-01-21
Discussion: Roadmap to 1.0


markerikson:
I'd been thinking about putting up a discussion issue like this already, and someone asked about it today, so figure I might as well.

I deliberately started out at 0.1.0 to give us room to fiddle with the API before reaching stability.  The "convert to TS" release put us at 0.4.0.  So, what do we feel would be needed to reach a stable 1.0 status?

Here's my rough thoughts on further changes I'm considering:

- Selectors
    - [x] I'm considering dropping Selectorator, re-exporting `createSelector` from Reselect, ~~adding a dependency on Re-reselect, and re-exporting its `createCachedSelector`~~
    - ~~We might want to have a tiny wrapper around those that _only_ accepts the array form of input selectors, not the arbitrary function arguments form~~ Looking like Reselect v5 may do this as a breaking change
- `createSlice`
    - [x] I want to add the ability to handle other action types besides the ones generated for the provided reducers.  Working on that now.
    - [x] Would like to consider the "customize payload callback" idea
    - [x] Change names of generated selectors (`selectFoo` instead of `getFoo`)?
    - [x] Returning a default selector that's just `state => state` is pointless.  Why are we doing that?  What should we do instead?
- `createAction`
    - [x] add `.type` to the action creators in addition to `.toString()`.  This helps with JS switch statements, as well as object lookup tables in TS.
- Async requests
    - [ ] I'm open to some options for simplifying async request handling.  Possibilities: 
        - Generating `"STARTED/SUCCESS/FAILURE"` action types
        - Generating thunks that call a provided API function and dispatch those action types based on promise lifecycle
        - Pulling in `redux-promise`
- More safety checks
    - ~~I'd like to see if I can make a middleware that checks for side effects that happen in a reducer (at least async requests and uses of randomness)~~ Skipping this idea
- Other
    - [ ] Might as well add `reduce-reducers` to the toolkit
    - [ ] Should fix tree shaking
    - [x] Allow `devTools` to be more options beyond a boolean
    - [x] Review `autodux` to see if there's any other functionality we want to swipe
    - [ ] Consider adding a simple middleware that lets users define callbacks to run in response to specific actions (which thunks can't do)

Not saying we _have_ to have all of these, but those are the ideas floating around my head.

Thoughts?  Suggestions?

denisw:
[...]

PR #197, opened 2019-09-22
Create slice changes

phryneas:
These are the changes you requested over in #82

### DIFF
--- a/docs/api/createSlice.md
+++ b/docs/api/createSlice.md
@@ -7,7 +7,7 @@ hide_title: true
 
 # `createSlice`
 
-A function that accepts an initial state, an object full of reducer functions, and optionally a "slice name",
+A function that accepts an initial state, an object full of reducer functions, and a "slice name",
 and automatically generates action creators and action types that correspond to the reducers and state.
 
 ## Parameters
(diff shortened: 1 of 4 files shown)


markerikson:
Great, thanks!

Ironically, I'm actually having second thoughts about the "return the reducer" thing:

https://twitter.com/acemarke/status/1175881949514219521

It's an interesting idea, and yet at the same time I'm not sure it buys much actual improvement.

Thoughts?

phryneas:
It's definitely nicer to use (took my a while to figure out that I could not use the slice directly in combineReducers), but the console.log in chrome is really bad :/
(Firefox works though: https://imgur.com/mIipct5 )

What about turning the idea around? Export a modified combineReducers from RSK that takes either functions or slices?

markerikson:
No, one of the main points of RSK is that it's "just Redux".  I don't want to have a special version of `combineReducers`.

Y'know, I think I've talked myself out of returning the reducer directly.  Returning the object gives us a bit more flexibility in case we ever do add something else to `createSlice`, and I don't think there's all that much benefit in returning the reducer by itself.  Let's drop that and just do the `slice? -> name` change.  

Sorry for the whiplash :)

markerikson:
@phryneas : just re-read your comment in the 1.0 thread.  

Yeah, sure, let's go ahead and include the individual reducers in the slice object (which is also another argument for keeping it as an object).

phryneas:
@markerikson okay. I'll revert the "reducer function" commit so that this can get merged and open a separate PR for the `caseReducers` (I think that name is fine) in the coming days.

markerikson:
One additional request: can we throw an error if `name` is falsy?  That would force the user to add a non-empty name.

markerikson:
Awright, looks good to me!

PR #200, opened 2019-09-30
Advanced tutorial: limit try blocks to API calls

mykter:
Thanks for the starter kit, it really helped me get going with redux. This is a PR reflecting our discussion here: https://twitter.com/acemarke/status/1171438698337947651

IMO the slight extra verbosity is worth avoiding the pain of trying to debug what's going on when the dispatch throws (for newcomers especially, speaking from personal experience!).


markerikson:
I just finished updating the tutorial examples and pages to use RSK 1.0.  I opted not to rewrite the actual data fetching code with your changes, but I did add a note on ways to do error handling in async thunks.  See https://github.com/reduxjs/redux-starter-kit/pull/245 .

Thanks!

PR #201, opened 2019-09-30
Remove remaining references to selectorator

### LINKED ISSUE #196: Outdated selectorator references
In #191, the selectorator library was removed. However, the [README](https://github.com/reduxjs/redux-starter-kit/blame/master/README.md#L35) and the [quickstart](https://github.com/reduxjs/redux-starter-kit/blame/04cdb02a20c227f76dece62f2785bdac47bc6768/docs/introduction/quick-start.md#L34) still mention this library.

mykter:
Closes #196


markerikson:
Can you rewrite those sentences to refer to `createSelector` from Reselect instead?

mykter:
> 
> 
> Can you rewrite those sentences to refer to `createSelector` from Reselect instead?

I'm probably not the best person for that job - I'm learning how to use reselect today.

mykter:
If anyone else does add some guidance on createSelector, it would be great to explain when you need to specify shallowEqual as a comparison function to useSelector, and when the selector will return the same object and that isn't needed. The example in the react-redux hooks docs doesn't use shallowEqual, but my experience (with arrays as return values) suggests it is sometimes necessary.

markerikson:
@mykter : that's exactly the issue.  If you are returning an object or array that is a new reference, but that points to the same contents as before, that new reference will cause a re-render unnecessarily.

markerikson:
Wound up going with #199 instead.  Thanks anyway!

PR #202, opened 2019-10-01
docs/api/createSelector.md: link to hooks guide

mykter:
When I first came to use createSelector it wasn't obvious to me how to use
it with hooks.

The reselect docs don't cover it, and I went down a few rabbit holes
before discovering that the react-redux Hooks docs cover it well.


markerikson:
Thanks!

PR #204, opened 2019-10-03
Add explanation of the second argument of createAction to the API docs

andreasgruenh:
I noticed that this argument was mentioned in the Intermedia Tutorial but not listed in the API docs. I'm not sure if this is intentional. If not, I started this PR to discuss possible explanations for this section of the docs.


markerikson:
Thanks!

PR #209, opened 2019-10-07
Include case reducers in createSlice result

markerikson:
Per discussion at https://github.com/reduxjs/redux-starter-kit/issues/82#issuecomment-533877434

### DIFF
--- a/src/createSlice.ts
+++ b/src/createSlice.ts
@@ -120,7 +120,11 @@ type ActionForReducer<R, ActionType = Action<string>> = R extends (
   action: PayloadAction<infer P>
 ) => any
   ? PayloadAction<P>
-  : ActionType
+  : R extends {
+      reducer(state: any, action: PayloadAction<infer P>): any
+    }
+  ? PayloadAction<P>
+  : unknown
 
 type CaseReducerActions<
   CaseReducers extends SliceCaseReducerDefinitions<any, any>
--- a/src/createSlice.ts
+++ b/src/createSlice.ts
@@ -1,4 +1,4 @@
-import { Reducer, Action } from 'redux'
+import { Reducer } from 'redux'
 import {
   createAction,
   PayloadAction,
@@ -36,7 +36,7 @@ export interface Slice<
    */
   actions: CaseReducerActions<CaseReducers>
 
-  caseReducers: OnlyCaseReducers<CaseReducers, State>
+  caseReducers: SliceDefinedCaseReducers<CaseReducers, State>
 }
 
 /**
@@ -90,10 +90,6 @@ type SliceCaseReducerDefinitions<State, PA extends PayloadActions> = {
     | CaseReducerWithPrepare<State, PA[ActionType]>
 }
 
-type SliceCaseReducers<State, PA extends PayloadActions> = {
-  [ActionType in keyof PA]: CaseReducer<State, PA[ActionType]>
-}
-
 type IfIsReducerFunctionWithoutAction<R, True, False = never> = R extends (
   state: any
 ) => any
@@ -115,7 +111,7 @@ type PrepareActionForReducer<R> = R extends { prepare: infer Prepare }
   ? Prepare
   : never
 
-type ActionForReducer<R, ActionType = Action<string>> = R extends (
+type ActionForReducer<R> = R extends (
   state: any,
   action: PayloadAction<infer P>
 ) => any
@@ -144,7 +140,7 @@ type CaseReducerActions<
   >
 }
 
-type OnlyCaseReducers<
+type SliceDefinedCaseReducers<
   CaseReducers extends SliceCaseReducerDefinitions<any, any>,
   State = any
 > = {
@@ -206,9 +202,9 @@ export function createSlice<
   const extraReducers = options.extraReducers || {}
   const reducerNames = Object.keys(reducers)
 
-  const sliceCaseReducersByName: SliceCaseReducers<State, any> = {}
-  const sliceCaseReducersByType: SliceCaseReducers<State, any> = {}
-  const actionCreators: any = {}
+  const sliceCaseReducersByName: Record<string, CaseReducer> = {}
+  const sliceCaseReducersByType: Record<string, CaseReducer> = {}
+  const actionCreators: Record<string, PayloadActionCreator> = {}
 
   reducerNames.forEach(reducerName => {
     const maybeReducerWithPrepare = reducers[reducerName]
@@ -237,7 +233,7 @@ export function createSlice<
   return {
     name,
     reducer,
-    actions: actionCreators,
+    actions: actionCreators as any,
     caseReducers: sliceCaseReducersByName as any
   }
 }
--- a/src/createSlice.ts
+++ b/src/createSlice.ts
@@ -1,4 +1,4 @@
-import { Reducer } from 'redux'
+import { Reducer, Action } from 'redux'
 import {
   createAction,
   PayloadAction,
@@ -17,8 +17,8 @@ import { createReducer, CaseReducers, CaseReducer } from './createReducer'
 export type SliceActionCreator<P> = PayloadActionCreator<P>
 
 export interface Slice<
-  State = any,
-  ActionCreators extends { [key: string]: any } = { [key: string]: any }
+  CaseReducers extends SliceCaseReducerDefinitions<State, PayloadActions>,
+  State = any
 > {
   /**
    * The slice name.
@@ -34,9 +34,9 @@ export interface Slice<
    * Action creators for the types of actions that are handled by the slice
    * reducer.
    */
-  actions: ActionCreators
+  actions: CaseReducerActions<CaseReducers>
 
-  caseReducers: SliceCaseReducers<State, any>
+  caseReducers: OnlyCaseReducers<CaseReducers, State>
 }
 
 /**
@@ -115,6 +115,13 @@ type PrepareActionForReducer<R> = R extends { prepare: infer Prepare }
   ? Prepare
   : never
 
+type ActionForReducer<R, ActionType = Action<string>> = R extends (
+  state: any,
+  action: PayloadAction<infer P>
+) => any
+  ? PayloadAction<P>
+  : ActionType
+
 type CaseReducerActions<
   CaseReducers extends SliceCaseReducerDefinitions<any, any>
 > = {
@@ -130,6 +137,16 @@ type CaseReducerActions<
       // else
       PayloadActionCreator<PayloadForReducer<CaseReducers[Type]>>
     >
+  >
+}
+
+type OnlyCaseReducers<
+  CaseReducers extends SliceCaseReducerDefinitions<any, any>,
+  State = any
+> = {
+  [Type in keyof CaseReducers]: CaseReducer<
+    State,
+    ActionForReducer<CaseReducers[Type]>
   >
 }
 
@@ -168,7 +185,7 @@ export function createSlice<
 >(
   options: CreateSliceOptions<State, CaseReducers> &
     RestrictCaseReducerDefinitionsToMatchReducerAndPrepare<State, CaseReducers>
-): Slice<State, CaseReducerActions<CaseReducers>>
+): Slice<CaseReducers, State>
 
 // internal definition is a little less restrictive
 export function createSlice<
@@ -176,7 +193,7 @@ export function createSlice<
   CaseReducers extends SliceCaseReducerDefinitions<State, any>
 >(
   options: CreateSliceOptions<State, CaseReducers>
-): Slice<State, CaseReducerActions<CaseReducers>> {
+): Slice<CaseReducers, State> {
   const { name, initialState } = options
   if (!name) {
     throw new Error('`name` is a required option for createSlice')
@@ -217,6 +234,6 @@ export function createSlice<
     name,
     reducer,
     actions: actionCreators,
-    caseReducers: sliceCaseReducersByName
+    caseReducers: sliceCaseReducersByName as any
   }
 }
--- a/src/createSlice.ts
+++ b/src/createSlice.ts
@@ -35,6 +35,8 @@ export interface Slice<
    * reducer.
    */
   actions: ActionCreators
+
+  caseReducers: SliceCaseReducers<State, any>
 }
 
 /**
@@ -42,7 +44,10 @@ export interface Slice<
  */
 export interface CreateSliceOptions<
   State = any,
-  CR extends SliceCaseReducers<State, any> = SliceCaseReducers<State, any>
+  CR extends SliceCaseReducerDefinitions<
+    State,
+    any
+  > = SliceCaseReducerDefinitions<State, any>
 > {
   /**
    * The slice's name. Used to namespace the generated action types.
@@ -74,15 +79,19 @@ type PayloadActions<Types extends keyof any = string> = Record<
   PayloadAction
 >
 
-type EnhancedCaseReducer<State, Action extends PayloadAction> = {
+type CaseReducerWithPrepare<State, Action extends PayloadAction> = {
   reducer: CaseReducer<State, Action>
   prepare: PrepareAction<Action['payload']>
 }
 
-type SliceCaseReducers<State, PA extends PayloadActions> = {
+type SliceCaseReducerDefinitions<State, PA extends PayloadActions> = {
   [ActionType in keyof PA]:
     | CaseReducer<State, PA[ActionType]>
-    | EnhancedCaseReducer<State, PA[ActionType]>
+    | CaseReducerWithPrepare<State, PA[ActionType]>
+}
+
+type SliceCaseReducers<State, PA extends PayloadActions> = {
+  [ActionType in keyof PA]: CaseReducer<State, PA[ActionType]>
 }
 
 type IfIsReducerFunctionWithoutAction<R, True, False = never> = R extends (
@@ -90,7 +99,7 @@ type IfIsReducerFunctionWithoutAction<R, True, False = never> = R extends (
 ) => any
   ? True
   : False
-type IfIsEnhancedReducer<R, True, False = never> = R extends {
+type IfIsCaseReducerWithPrepare<R, True, False = never> = R extends {
   prepare: Function
 }
   ? True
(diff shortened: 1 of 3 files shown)


markerikson:
@phryneas , @denisw , @Dudeonyx , @jessidhia : can I get some eyes on this to make sure I didn't do anything really stupid?  

The unit tests and type tests still pass, and it _looks_ like everything is being inferred correctly in the output.

src/createSlice.ts
```diff
@@ -35,14 +35,19 @@ export interface Slice<
    * reducer.
    */
   actions: ActionCreators
+
+  caseReducers: SliceCaseReducers<State, any>
```
phryneas: This is too generic, it needs some relation to the original reducers or it will broaden to accept any type of reducer with just any key. I'll add two example tests below.

type-tests/files/createSlice.typetest.ts
```diff
@@ -147,6 +147,25 @@ function expectType<T>(t: T) {
...
+    reducers: {
+      increment(state, action: PayloadAction<number>) {
+        return state + action.payload
+      }
+    }
+  })
+
+  expectType<(state: number, action: PayloadAction<number>) => number | void>(
+    counter.caseReducers.increment
+  )
```
phryneas: these two additional tests should be failing, but currently don't

```typescript
  // typings:expect-error
  expectType<(state: number, action: PayloadAction<string>) => number | void>(
    counter.caseReducers.increment
  )

  // typings:expect-error
  expectType<(state: number, action: PayloadAction<string>) => number | void>(
    counter.caseReducers.someThingNonExistant
  )
```

src/createSlice.ts
```diff
@@ -106,8 +115,10 @@ type PrepareActionForReducer<R> = R extends { prepare: infer Prepare }
   ? Prepare
   : never
 
-type CaseReducerActions<CaseReducers extends SliceCaseReducers<any, any>> = {
-  [Type in keyof CaseReducers]: IfIsEnhancedReducer<
+type CaseReducerActions<
+  CaseReducers extends SliceCaseReducerDefinitions<any, any>
```
nickmccurdy: Use unknown instead of any

src/createSlice.ts
```diff
@@ -153,16 +164,16 @@ function getType(slice: string, actionKey: string): string {
  */
 export function createSlice<
   State,
-  CaseReducers extends SliceCaseReducers<State, any>
+  CaseReducers extends SliceCaseReducerDefinitions<State, any>
```
nickmccurdy: Use unknown instead of any

src/createSlice.ts
```diff
@@ -172,35 +183,40 @@ export function createSlice<
...
+  const sliceCaseReducersByName: SliceCaseReducers<State, any> = {}
+  const sliceCaseReducersByType: SliceCaseReducers<State, any> = {}
+  const actionCreators: any = {}
+
+  reducerNames.forEach(reducerName => {
+    const maybeReducerWithPrepare = reducers[reducerName]
+    const type = getType(name, reducerName)
+
+    let caseReducer: CaseReducer<State, any>
+    let prepareCallback: PrepareAction<any> | undefined
```
nickmccurdy: See previous comment

markerikson:
@phryneas : 

> It needs the round-trip to something with the initial reducer types, or it just accepts anything.

Not sure I quite understand what you're saying here.  Can you clarify?

src/createSlice.ts
```diff
@@ -35,14 +35,19 @@ export interface Slice<
...
 
 /**
  * Options for `createSlice()`.
  */
 export interface CreateSliceOptions<
   State = any,
-  CR extends SliceCaseReducers<State, any> = SliceCaseReducers<State, any>
+  CR extends SliceCaseReducerDefinitions<
+    State,
+    any
```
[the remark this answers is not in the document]
markerikson: This one was there already, but yeah, we can try that.

src/createSlice.ts
```diff
@@ -172,35 +183,40 @@ export function createSlice<
...
-          : createAction(type, maybeEnhancedReducer.prepare)
-      return map
-    },
-    {} as any
-  )
+  const reducerNames = Object.keys(reducers)
+
+  const sliceCaseReducersByName: SliceCaseReducers<State, any> = {}
+  const sliceCaseReducersByType: SliceCaseReducers<State, any> = {}
+  const actionCreators: any = {}
```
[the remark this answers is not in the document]
markerikson: The `actionCreators : any` here was actually how the existing code worked as far as I can see, and the types of the exported action creators are still correct based on VS Code's inspection and the type tests.

src/createSlice.ts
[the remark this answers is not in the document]
markerikson: Nope.  Adding `unknown` in the places you're suggesting just broke all the existing type tests.

src/createSlice.ts
markerikson: Trying to poke at this now.  Got any suggestions for approaches?

Jessidhia: A possible solution (that is potentially breaking) is to change the type of the second generic argument on most things in this file; a lot of things here are overly broad, not just the changes done in this PR.

Make them all use the `PA extends PayloadActions` instead of types that are derived from it; put the derivation in the parameter list / helper types / interface declarations themselves. The fundamental units here are the `State` object and the `PayloadActions` that are handled.

Speaking of which, `PayloadActions` itself is erasing types; it does not preserve the "individuality" of each `PayloadAction` it contains; fixing that is even more involved than this,  though, and I think the current problems we're pointing out can be fixed before addressing that (higher kinded types yay 😓).

For example, this `Slice` interface should receive `State` and `PA extends PayloadActions`, the type of `actions` should be (what currently is) `CaseReducerActions<SliceCaseReducers<State, PA>>` (they're all already `createAction`ed, you can no longer have `CaseReducerWithPrepare` objects in the result), and type of of `caseReducers` should be `SliceCaseReducers<State, PA>`.

markerikson: Thanks for the suggestions.

I will freely admit that part of me wants to throw my hands up and say "someone else fix this!", and it may come to that.  But, I'd also like to start becoming at least not completely incompetent in this area.

For the moment, is there a simpler approach that would help improve the current PR without trying to redo the rest of the types right now?  Should we just get this PR in and tackle everything else in a follow-up?

src/createSlice.ts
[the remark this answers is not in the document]
nickmccurdy: What's the error? It's possible the type tests are running on an old version of TS, or that they need type guards.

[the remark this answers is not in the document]
markerikson: `redux-starter-kit/src/createSlice.ts (49, 5) Type 'unknown' does not satisfy the constraint 'Record<string,
WithPayload<any, Action<string>>>'.`

And no, we're on TS 3.4.3 right now.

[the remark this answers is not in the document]
nickmccurdy: I’m not sure why that doesn’t match, but if the return type uses any anyway, maybe this can be left alone for now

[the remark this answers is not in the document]
phryneas: You can pretty easily change all those parts in the `extends` statement from `any` to `unknown` without consequence. 
But doing it for the default value will in some places be a breaking change for people who didn't specify a type manually and where the type could not be inferred.
Arguably, this is how they shouldn't be using TS in the first place if they want it for type-safety, but it definitely _is_ used that way (see #165), so maybe that should be done with caution in a different PR.

src/createSlice.ts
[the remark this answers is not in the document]
phryneas: As the method has a defined return type, as long as the value you are returning somehow satisfies that type, you can do pretty much inside - the outside interface is defined. 
Even casting to any would be valid here, and in some places I think it is to make stuff less difficult. 

As an alternative, for example you could type the `actionCreators` to `CaseReducerActions<CaseReducers>`. As a result, the assignment of `{}` would fail, and you would have to assign it `{} as any` instead. 
But you've got to die one death here, and I think that would be acceptable, as it would at least guard all assignments. (Even though it could never make sure all the values were actually assigned, just that those being assigned are correct. But still, more typesafe than `any` alone)

But, as I said, even typing this as `any` as it is is probably okay - as long as the method is tested thoroughly, and that it is.

src/createSlice.ts
phryneas: For a simple approach:

* the createSlice method signature passes `CaseReducers` on to Slice instead of currently `CaseReducerActions<CaseReducers>`
* Slice changes it's second generic argument from `ActionCreators extends { [key: string]: any } = { [key: string]: any }`to `CaseReducers extends SliceCaseReducerDefinitions<State, unknown>` (I believe the default value would even satisfy that, as broad as it is)
* `actions: ActionCreators` becomes type `CaseReducerActions<CaseReducers>`

that way, you preserve given behaviour, but within Slice, you now have access to the `CaseReducers` instead of the derived `CaseReducerActions`.

Then you have to create a new type that maps  the values of CaseReducers to always the reducer, and use that for the caseReducers.

phryneas: Another option with maybe even less changes would be to change your definition of 
```typescript
// from
type SliceCaseReducers<State, PA extends PayloadActions> = {
  [ActionType in keyof PA]: CaseReducer<State, PA[ActionType]>
}
// to something like
type SliceCaseReducers<State, ActionCreators extends  { [key: string]: unknown }> = {
  [ActionType in keyof ActionCreators]: CaseReducer<State, ReturnValue<ActionCreators[ActionType]>>
}
```

And actually pass ActionCreators as a second parameter to this interface, instead of any.

But honestly, I'm not sure if that will work out in the long run (also, I've written this without actually testing it, so it might not work)

markerikson: Trying your `CaseReducers` suggestion now, and I'm getting two errors:

- `Type 'unknown' does not satisfy the constraint 'Record<string, WithPayload<any, Action<string>>>'`
- `Required type parameters may not follow optional type parameters.`

I can resolve the first one by replacing `unknown` with `PayloadAction` (the record type).  Trying to figure out what an appropriate default type might be here.

markerikson: Resolved the second error by simply swapping the order of the generics in `Slice` :)

So. Coming up with a type for the reducers. I feel like the `CaseReducerActions` type is roughly what I want to mimic. I "just" need to figure out how to reference the reducer functions instead.

markerikson: And I _think_ I have something!

type-tests/files/createSlice.typetest.ts
markerikson: Thanks!  These were a _huge_ help!

markerikson:
Awright, I _think_ I actually have a working solution here!

I added the two type tests suggested by @phryneas , and attempted to implement the suggested solution ( https://github.com/reduxjs/redux-starter-kit/pull/209#discussion_r332356920 ).

I fixed the "can't have a required type after a default type" error by swapping the order of the two generics in `Slice`.  

From there, I tried to set up some kind of mapping, loosely based on the `CaseReducerActions` type.  My first couple attempts got the `caseReducers.nonExistent` test failing as desired, but the mismatched payload test was incorrectly passing without error due to the payload being inferred as `any`:

```ts
type ExtractPayloadActions<T> = T extends SliceCaseReducerDefinitions<
  any,
  infer PA
>
  ? PA
  : never

type OnlyCaseReducers<
  CaseReducers extends SliceCaseReducerDefinitions<any, any>,
  State = any
> = {
  [Type in keyof CaseReducers]: CaseReducer<
    State,
    ExtractPayloadActions<CaseReducers>[Type]
  >
}
```

I eventually figured out that it was basically falling back to a default `PayloadAction<any>` somehow.

That clued me in that I need to extract the _specific_ action type for this situation.  Inspired by the `PayloadForReducer` type, I came up with this:

```ts
type ActionForReducer<R, ActionType = Action<string>> = R extends (
  state: any,
  action: PayloadAction<infer P>
) => any
  ? PayloadAction<P>
  : ActionType

type OnlyCaseReducers<
  CaseReducers extends SliceCaseReducerDefinitions<any, any>,
  State = any
> = {
  [Type in keyof CaseReducers]: CaseReducer<
    State,
    ActionForReducer<CaseReducers[Type]>
  >
}
```

That got both the "mismatched payload" and "nonexisting field" type tests throwing errors as expected.

I'm not sold on the name `OnlyCaseReducers`, but I'm open to suggestions.

phryneas:
You're on a good way here, but not quite there yet (and those types are complicated, so every little bit of frustration you might be feeling is perfectly understandable!).

The important thing here is to identify why your ExtractPayloadActions type or ActionForReducer might fail.

First your `ExtractPayloadActions` type: in theory that's perfectly fine! But the `SliceCaseReducerDefinitions` type itself is unfortunately so complex, that TS doesn't manage fitting "the right thing" into the Generic any more (at least that's how I explain it to myself, others to the rescue).

In your `ActionForReducer`, you're doing exactly the right next step: when TS can't figure it out, you need to be more specific. And for the first part, that does look good to me. The "else" part needs a bit works still, though.

Take a look at potential shapes that could be passed to the `R` parameter.
The default shape is the one you already identified: `(  state: any,  action: PayloadAction<P>) => any`
But then you're just falling back to `ActionType` - and as `ActionType` has no further mentioning in the type, it can't be inferred by TS and will always be it's default value `Action<string>`.

But there *is* another shape you can test for that we definitely know of:

```
{
  reducer(state: any,  action: PayloadAction<P>): any
}
```
so you can do a second nested ternary there, testing for that type and inferring P from there.

As for the "does not match any of the above" branch, I would not go for your ActionType or any Action, but just for `never`, `void`, `unknown` or the likes. If it doesn't match those, we're in a case that *should* never happen, so there's no need for a sensible fallback, in contrary: that fallback might prevent us from catching bugs if it isn't abviously unfitting.

I would suggest you add another test for the case with an enhanced reducer (including expect-error tests like I wrote above) - and then you're really super close :)

markerikson:
@phryneas : thanks for coaching me through this!  As I said, part of me just wants someone to fix all this and make it work so we can push RSK up to 1.0 ASAP, but at the same time I really need to start becoming not-totally-incompetent with this complex type stuff.

I'll try to tackle your suggested directions myself and see what happens.

markerikson:
@phryneas : had to read through your instructions a couple times, but turns out you were _really_ walking me through this one, weren't you :)

(I recognize the "tell the clueless n00b the exact steps they need to do, politely" phrasing.  Goodness knows I've done that enough times myself.)

Added the additional type test, implemented the ternary, and saw that test start failing as expected.  

From there, I cleaned up a few of the types in `createSlice` a bit, and decided to rename `OnlyCaseReducers` to `SliceDefinedCaseReducers`.  (There are entirely too many uses of the words "Slice", "Case", and "Reducers" in this file already, but oh well!)

How's this look now?

src/createSlice.ts
```diff
@@ -106,8 +111,21 @@ type PrepareActionForReducer<R> = R extends { prepare: infer Prepare }
   ? Prepare
   : never
 
-type CaseReducerActions<CaseReducers extends SliceCaseReducers<any, any>> = {
-  [Type in keyof CaseReducers]: IfIsEnhancedReducer<
+type ActionForReducer<R, S> = R extends (
```
phryneas: As you don't really care for the State type here, I believe you can just leave it out. 
```typescript
type ActionForReducer<R> =
  R extends (  state: any,  action: PayloadAction<infer P> ) => any
  ? PayloadAction<P>
  : R extends {      reducer(state: any, action: PayloadAction<infer P>): any    }
  ? PayloadAction<P>
  : unknown
```

src/createSlice.ts
```diff
@@ -122,6 +140,16 @@ type CaseReducerActions<CaseReducers extends SliceCaseReducers<any, any>> = {
   >
 }
 
+type SliceDefinedCaseReducers<
```
phryneas: as a consequence, no need for the State parameter here as well

type-tests/files/createSlice.typetest.ts
phryneas: I'd add one non-error-check for `decrement` here, as currently you're just testing for expect-errors on that one.

phryneas:
A little bit - but in the end, you did it, with all the failures that come along on the way. 

Just a little secret: for the reducer with payload stuff, I wrote that whole thing at least four times from scratch, completely resetting because something was not behaving as I wanted and I was going routine-blinded with my existing code. Wanted to spare you of *that* frustration, because it is not a learning one ;)

The code looks good now, I've added two more small comments, but I believe you did it :)

It's early in the morning here and I haven't had my coffee yet, so I'll do a second round of review in the evening, just in case I missed something.

src/createSlice.ts
markerikson: Yeah, that's basically what I originally had.  @nickmccurdy suggested I might as well use `State` instead of `any` for consistency, and it doesn't seem to actually hurt anything.

phryneas: It doesn't hurt - it's just one more thing to carry around :)

Explanation why I think it's not necessary here:
What you are doing here is shape-matching. You're either looking for an object with a `reducer` method or for a method itself, both methods with at least two arguments. From that moment on you just care about the type of the second argument.
Sure, you can choose to only match if the first argument is `State`, but that means a second parameter to the generic, also caring about that second parameter when calling. So, essentially: more code, without any direct benefit. 
So I'd just let it out. But if you choose to let it in there, don't forget to do the same check in line 120, too.

src/createSlice.ts
```diff
@@ -17,8 +17,8 @@ import { createReducer, CaseReducers, CaseReducer } from './createReducer'
 export type SliceActionCreator<P> = PayloadActionCreator<P>
 
 export interface Slice<
-  State = any,
-  ActionCreators extends { [key: string]: any } = { [key: string]: any }
+  CaseReducers extends SliceCaseReducerDefinitions<State, PayloadActions>,
+  State = any
```
phryneas: You could bring this back into the original order by giving the CaseReducers the same default value as ActionCreators had before (that should be working). That way, things might still break for ppl who used this type directly, but maybe not as significantly.

type-tests/files/createSlice.typetest.ts
phryneas: something like this one.
 ```typescript
  // Should match positively for reducers with prepare callback
  expectType<(state: number, action: PayloadAction<number>) => number | void>(
    counter.caseReducers.decrement
  )
 ```

src/createSlice.ts
markerikson: Sure, that works.

markerikson:
Awright, let's do this!

PR #210, opened 2019-10-08
Update usage-guide.md to reflect updates

stern-shawn:
I think the `createAction` function has since been updated to support `prepare` for adding attributes like `meta` to an action. Updating documentation so users don't get discouraged and think they can't use `meta` along with `redux-starter-kit`. (This was the first page I landed on when searching how to add `meta` in conjunction with this library and almost thought I'd need to abandon my work on migrating :yikes:)

Edit: and of course, thank you so much for creating this repo! Currently migrating an internal application that I inherited with a bunch of hand-made types and inconsistent reducer structuring which I'm hoping this will help address 🎉


markerikson:
Thanks!

Issue #198, opened 2019-09-22
Definition for rule '@typescript-eslint/consistent-type-assertions' was not found


nidble:
Hi all,

As usual, I cloned the main repo and executed `npm install`/`npm start`. But unfortunately app crashed very soon reporting the error in the subject.

From this thread https://github.com/typescript-eslint/typescript-eslint/issues/863 I fixed my local setup replacing into package.json react scripts lib: `"react-scripts": "^3.1.2",`.

I hope it helps. Thanks for your work.

markerikson:
Ah... _what_ repo did you clone, and for what purpose?

This is a library, not an app.  

I'm kind of confused, and this doesn't seem like an actual issue with this library.

nidble:
Whops! I'm sorry @markerikson. 

My fault, the correct repo is https://github.com/markerikson/rsk-github-issues-example. What do you prefer? I can repost this issue on the correct repo, or considering the owner :), do I leave its to you?

markerikson:
Hmm.  I would be very surprised if there was a real problem with that repo - it's just a standard CRA project.  

If you can consistently reproduce the problem, then maybe file an issue there, but I'm not sure there's really anything I can do about it.  Also, I suggest trying `yarn` instead of `npm`.

nidble:
Hi Mark, I tried with `yarn` and everythings works fine. Currently the "issue" is only with npm (6.9.0). The steps to get involved are quite simple: `npm install && npm run start`.

markerikson:
Yeah, I really don't see anything actionable to do here.  It's a plain standard CRA app, and installing the packages works with Yarn.  If NPM is having problems, that's outside the scope, and I can't think of anything I could change to affect that.

nidble:
Yes, sure. My comment was only to give you an insight about. And a reference to other people that meeting this issue. Thanks.

markerikson:
Yep, appreciate the heads-up!

markerikson:
So, this is bizarre.  I just started seeing this myself, in both my local repo and in CI.

PR #212, opened 2019-10-12
chore(build): swap in tsdx

RichiCoder1:
There's still work that needs to happen as when I last worked on this there were still some important differences in the output around externals, but I figured I'd issue this PR so one can see the differences in a rough state.

/cc @markerikson


markerikson:
Great, thanks! What are the differences in output atm?

markerikson:
I'm poking around at this, and I'm seeing some failing type tests that... shouldn't be.

In particular, these for `createSlice`:

```ts
  // typings:expect-error
  const stringReducer: Reducer<string, PayloadAction<number>> = slice.reducer
  // typings:expect-error
  const anyActionReducer: Reducer<string, AnyAction> = slice.reducer
```

@phryneas , any idea why these are failing?  Not sure if it's due to a TS update, or the fact that we're using TS directly now.

Also, I see that `ts-jest` is flagging unused variables in tests as errors, which seems overly strict (especially when a few of the tests are doing things like grabbing `extraArg` from a thunk and asserting on it, and not using `dispatch` or `getState`).  That's... annoying.

RichiCoder1:
> Great, thanks! What are the differences in output atm?

I think the biggest difference is you had externals for a few of the packages setup differently than how it's currently configured. I forget the specifics. I think there's also at least one rollup plugin I hadn't brought over yet.

> I'm poking around at this, and I'm seeing some failing type tests that... shouldn't be.

> In particular, these for createSlice:

Weird. Maybe a typescript bump error? 

> Also, I see that ts-jest is flagging unused variables in tests as errors, which seems overly strict (especially when a few of the tests are doing things like grabbing extraArg from a thunk and asserting on it, and not using dispatch or getState). That's... annoying.

That should be pretty easy to fix, just need to crank that option down.

markerikson:
Yeah, looking at the build output, the biggest immediate differences I see are:

- The names of the UMD files changed (`r-s-k.umd[.min].js` for current, `r-s-k.umd[.development].js` for  this branch)
- The UMD files are no longer embedding the other libs directly, but rather expecting them as globals

I'm not as overly concerned about the UMD naming, especially since this is pre-1.0, but I do want our UMD files to be entirely self-contained (ie, I should be able to link the UMD file and write working Redux code right there).

RichiCoder1:
I think both are those are pretty fixable. Just need to drop the externals plugin from the UMD build and tweak the output name. I can try both those later.

phryneas:
> ```ts
> stringReducer
> ```

Seems to be a bug introduced with TypeScript 3.6. State is inferred to `any` in that version, no matter what is being passed in. No idea why though. That'll need some investigation.

markerikson:
Bah, closed by the TS fix PR.  Reopening.

markerikson:
Welllll, something is definitely still not right here:

```
    src/createSlice.test.ts:104:42 - error TS2769: No overload matches this call.
      Overload 1 of 2, '(payload?: undefined): WithPayload<undefined, Action<"ADD_MORE">>', gave the following error.
        Argument of type '{ amount: number; }' is not assignable to parameter of type 'undefined'.
      Overload 2 of 2, '(payload?: void | undefined): WithPayload<void, Action<"ADD_MORE">>', gave the following error.
        Argument of type '{ amount: number; }' is not assignable to parameter of type 'void'.
    104       const result = reducer(10, addMore({ amount: 5 }))
                                                 ~~~~~~~~~~~~~
```

markerikson:
Well, it builds now!

I fixed the now-broken uses of `createAction`, and patched up the types in `createReducer.test.ts` enough to let it run.  (The point of that file is to test _behavior_, not the types, so I'm not worried if it's a little iffy.)

So, now the question is what the output is like.

I gotta head to bed, but will hopefully be able to look at it again tomorrow.

markerikson:
Okay, fixed up the UMD externals and the dev filename.  Output now looks basically equivalent to what we have already, minus a few Babel helpers and such.  

Also, interestingly, the CJS file appears to already be minified, which is surprising.

I'm going to publish an alpha build and see if we can get some folks to test it out.

markerikson:
Published `v0.9.0-alpha.0` as `redux-starter-kit@build-test`.

Can folks try it out and see how it looks?

phryneas:
You can look at the files in the package over at https://unpkg.com/browse/redux-starter-kit@0.9.0-alpha.0/ :

I'm not really happy that we're shipping the src folder (I already know I'll be importing from there on accident), and in addition to that all those .d.ts files seperately instead of a rolled-up .d.ts (I already know that those will be problems with my `babel`+`babel-preset-typescript` webpack setup :cry: ).

@nickmccurdy can we do a rollup from that? This seems to be an open issue with https://github.com/jaredpalmer/tsdx/issues/80 - that's why I personally prefer `pika` with `@pika/plugin-bundle-types` over `tsdx`. But maybe we could take some inspiration from jaredpalmer/jsdx#80 or how @pika/plugin-bundle-types does it?

phryneas:
As this is a quite exotic case, I've created a reproduction repo here: https://github.com/phryneas/reproduction-non-rollup-types/blob/master/src/index.ts

Given that by now, almost everything is re-exported from index.d.ts, I don't actually think this is too much of a problem any more (it was in the early days of RSK for me and I still have some workarounds for that in place!), but I'd love to see the types as one single file for 1.0.

phryneas:
> @phryneas I personally haven't been able to get rollup to generate a bundled `.d.ts` file. The issue is that the declarations are generated by the TS compiler, which isn't a bundler and cannot concatenate the files. Still, TypeScript will work fine with multiple `.d.ts` files even if there's only a single `.js` entry file.

TS itself works perfectly - but webpack can crash like in the reproduction above when using babel-loader or ts-loader with isolatedModules when provoking some edge cases.
 
There's a trail of issues in those loaders' repos, but there isn't much the loaders can do about it themselves. But as I said, at the current state it's probably fine, even though I suspect we will have some stray error reports over the next moths.

markerikson:
Why are multiple types files problematic for certain build setups?

FWIW, unless this is a huge problem with this PR, my intent is to merge it, publish an RC or two, and get 1.0 out the door in the next few days.

phryneas:
@markerikson tanspilation is often done without file-spanning typechecking to save time (and doing the typechecking in a second thread). While that is much faster, the transpiler has no information if a certain import is a type, or a value. Most of the time, that can be seen from the context the value/type is used within the file, but there are scenarios (directly re-exporting an imported type, for example) where that is not possible.
That, in combination with a bundler can lead to problems, as can be seen when trying to build my reproduction (in the editor, it looks fine).

While not critical (and nothing that should stop you from publishing 1.0), as I said, I would really like RSK to have a single-file output to prevent such problems.
I guess I'll give it a timeboxed spin, maybe I can make this work in a short amount of time.

phryneas:
Okay, doing a dts-rollup with api-extractor works without any modifications to the RSK sources.

That would mean these changes to the build: https://github.com/RichiCoder1/redux-starter-kit/compare/switch_to_tsdx...phryneas:dtsRollup?expand=1

build output would look like this: https://unpkg.com/browse/@phryneas/experimental-redux-starter-kit@0.8.1-donotuse-dtsRollup-2/dist/

phryneas:
Oh, and one thing that is definitely a bug and needs to be adressed:
This file https://unpkg.com/browse/redux-starter-kit@0.9.0-alpha.0/dist/index.js links to `./redux-starter-kit.cjs.production.min.js`, which does not exist.

markerikson:
@phryneas : y'know, I just realized that's why we only have a minified CJS file.  It's because the custom Rollup config is overriding the output in both cases, by reading from one of the `package.json` fields.  We'd want to change the `package.json` field to point to `index.js` instead.

markerikson:
Been working on this this evening.  Fixed the CJS filename, and was working on copying the changes @phryneas suggested.

I just updated TSDX to 0.10.x on general principle... and it looks like it's now bundling all the typedefs into `index.d.ts` automatically due to a change in Rollup TS plugins, no `api-extractor` package needed:

https://github.com/jaredpalmer/tsdx/issues/200

https://github.com/jaredpalmer/tsdx/pull/208

Based on that, I'm going to drop it from the setup for now.  If someone wants to see what other options it offers down the road, I'm fine with that.

markerikson:
Oooo-kay, and that's _completely_ dying in the tslint check with:

`Definition for rule '@typescript-eslint/no-angle-bracket-type-assertion' was not found  @typescript-eslint/no-angle-bracket-type-assertion`

For _every_ TS source file.  What in the world?

markerikson:
Whoa, this is bizarre.  That actually showed up previously over in #198 .

markerikson:
Still don't know what's _causing_ that weird lint error, but I was able to shut it up by explicitly disabling that lint rule.

markerikson:
yayyyyyy and we're green!

okay, I'm gonna go publish another alpha now.

markerikson:
Okay, `v0.9.0-alpha.1` is out.  Unpkg looks good:

https://unpkg.com/browse/redux-starter-kit@0.9.0-alpha.1/dist/

markerikson:
Tried forking the last "Advanced Tutorial" sandbox and updating it to `v0.9.0-alpha.1`.  Fixed the `slice` -> `name` changes, and it runs.

Any further comments or objections before I merge this in and publish 0.9.0?

phryneas:
I'd publish it as 0.9 - that way we get some real-world feedback until you publish it as 1.0

markerikson:
Agreed.  All right, I'll merge this in and put out 0.9.

markerikson:
Published https://github.com/reduxjs/redux-starter-kit/releases/tag/v0.9.0

RichiCoder1:
@markerikson Glad I could be of help!

markerikson:
Yep, thanks!

PR #213, opened 2019-10-13
Update tutorials for v0.8

markerikson:
Initial tutorial changes to match 0.8.  

So far this only includes the `slice?` -> `name` changes.

The intermediate and advanced tutorial repos still need to be rewritten, and at that point we'll have to fix up the commit links here to point to the new commits, and redo the CodeSandbox sandboxes as well.

I also based this PR off of #174 , so once that gets merged we'll need to rebase this off `v0.8-integration` and force-push.

### DIFF
--- a/docs/tutorials/advanced-tutorial.md
+++ b/docs/tutorials/advanced-tutorial.md
@@ -206,7 +206,7 @@ let initialState: CurrentDisplayState = {
 }
 
 const issuesDisplaySlice = createSlice({
-  slice: 'issuesDisplay',
+  name: 'issuesDisplay',
   initialState,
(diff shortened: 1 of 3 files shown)


markerikson:
Given that I've got to actually push 0.8 out the door before I can rewrite the commits, I think the plan will be to merge this to `master` and release 0.8 first.  The _content_ will be correct, it'll only be the links to the specific commits that will be outdated (and the commit contents).  

Once the 0.8 release is out, I can go rewrite the tutorial commits using 0.8, push those, and fix up the commit links.

Also, I want to at least integrate a note about async `try/catch` per #200 .

Issue #214, opened 2019-10-15
Payload type unknown when defining optional property?


mcchrish:
This is for version 0.8.0.

```js
const action = createAction<{ input?: string }>('ACTION')
type ActionType = ReturnType<typeof action> // ActionType.payload === unknown


const action = createAction<{ input?: string, input2: string }>('ACTION')
type ActionType = ReturnType<typeof action> // ActionType.payload === { input?: string | undefined; input2: string }
```

Somehow, if there is only one property and it's optional, it converts it type to `unknown`

markerikson:
Oh dear.

@phryneas , @jessidhia : any ideas?

phryneas:
Phew. Yup.

The feature detection for TS<3.5 was leaking both cases (TS<3.5 and TS>=3.5), exposing the pre-3.5 behaviour. 
Back then, we could not distinguish between the empty object default case and `unknown`, so it would fall back to `unknown` behaviour if the object looked like an empty object (which is the case here).

We can fix this for TS>=3.5, but in old versions this will be the behaviour if something possibly looks like an empty object.

( fixed TS version feature detection would look like this:
```typescript
export type AtLeastTS35<True, False> = [True, False][IsUnknown<ReturnType<<T>() => T>, 0, 1>]
```
)

Question is: is it worth it supporting old TS versions, which should be our minimum requirement? I guess we should still support 3.4, which is from march 2019, even if it hurts a little :/

markerikson:
Yeah, 3.4 seems reasonable to support, even if it's a bit of a pain.

phryneas:
I might be even able to fix it for pre-3.5... PR coming in.

markerikson:
Fix published as 0.8.1.  Hopefully that should resolve this - let us know if there's still an issue!

PR #215, opened 2019-10-15
Fix #212 #214

phryneas:
This should fix a bug with TS3.6 found in #212 and the bug reported in #214.

This should be working fine, but I'm currently preparing a conference talk, so I don't have the time to test this super-excessively and the bug fix for #214 is something I'm not feeling 100% sure of if I did everything right. 
@dragomirtitian, could I ask you to take a look that I didn't do anything incredibly stupid there?

### DIFF
--- a/src/tsHelpers.ts
+++ b/src/tsHelpers.ts
@@ -24,7 +24,11 @@ export type IsEmptyObj<T, True, False = never> = T extends any
  * * versions below 3.5 will return `{}` for unresolvable interference
  * * versions above will return `unknown`
  * */
-export type AtLeastTS35<True, False> = [True, False][IsUnknown<ReturnType<<T>() => T>, 0, 1>]
+export type AtLeastTS35<True, False> = [True, False][IsUnknown<
+  ReturnType<<T>() => T>,
+  0,
+  1
+>]
 
 export type IsUnknownOrNonInferrable<T, True, False> = AtLeastTS35<
   IsUnknown<T, True, False>,
--- a/src/tsHelpers.ts
+++ b/src/tsHelpers.ts
@@ -13,8 +13,8 @@ export type IsUnknown<T, True, False = never> = unknown extends T
   : False
 
 export type IsEmptyObj<T, True, False = never> = T extends any
-  ? {} extends T
-    ? IsUnknown<T, False, IsAny<T, False, True>>
+  ? keyof T extends never
+    ? IsUnknown<T, False, True>
     : False
   : never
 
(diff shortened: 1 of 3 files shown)


src/tsHelpers.ts
```diff
@@ -13,8 +13,8 @@ export type IsUnknown<T, True, False = never> = unknown extends T
   : False
 
 export type IsEmptyObj<T, True, False = never> = T extends any
```
phryneas: This is what I'm most concerned with. I believe this should be good to find really empty objects (the one before also matched objects with only optional properties), but I'm not 100% sure I'm not missing something.

markerikson:
Great, thanks!

Out of curiosity for the still mostly-TS-clueless: any chance you could walk through how you went about debugging this problem?

phryneas:
Hm, that failing build was to be expected. @markerikson, would you be okay if I added an any-cast in that failing line?

phryneas:
> Great, thanks!
> 
> Out of curiosity for the still mostly-TS-clueless: any chance you could walk through how you went about debugging this problem?

I went to the `PayloadActionCreator` type and replaced each possible result with a string, to see which path it goes. It went to `ActionCreatorWithPayload`, so I did the same thing there, only to see that it took a wrong branch in `IsUnknownOrNonInferrable` - and so on.

markerikson:
Ah, ye olde "brute force logging" approach :)

Yeah, I'm okay with whatever works to actually resolve the problem.

phryneas:
Argh, failing due to prettier issues. I really have to stop procastinating, I'll fix that up tomorrow. Alternatively, @markerikson could you please just push a prettier fix?

markerikson:
At work atm, but yeah, will do so later.

src/tsHelpers.ts
karol-majewski: Your intuition is right! As far as I know, that's the most reliable way to match an empty object. `{}` is very tricky, since [it doesn't really mean _empty object_](https://gist.github.com/OliverJAsh/381cd397008309c4a95d8f9bd31adcd7). 😉

phryneas: Thanks a lot, that gives me some assurance :)

phryneas:
> Also, Y U NO PRETTIER AUTOFORMAT ON SAVE?!?!?

Not my work laptop, and vscode defaults to some other autoformat behaviour than prettier. But it does something. So I always forget to install the plugin, because it looks right. Fixed that on two machines and today, wrote the fix on my third machine :D

src/tsHelpers.ts
dragomirtitian: The empty object detection looks good to me too. `keyof T extends never` is the most reliable way to ensure that the type has no keys.

markerikson:
Thanks for the fast solution!

PR #217, opened 2019-10-15
Test with multiple TypeScript minor versions in CI

nickmccurdy:
Fixes #216

- [x] Add build matrix to Travis for multiple TypeScript versions
- [ ] Update development version to 3.6

PR #218, opened 2019-10-16
Use TypeScript 3.6 in development

nickmccurdy:
I originally meant to include this in #217.

ESLint now gives a warning but I tried to update `typescript-eslint` and it doesn't seem to be supported yet.


markerikson:
Yeah, it's that stupid "this version of TS is newer, blah blah" thing.

PR #222, opened 2019-10-19
feat(action): support optional error field

tvanier:
Allow `prepareAction` to return an optional `error` field as per the Flux specification
https://github.com/redux-utilities/flux-standard-action#actions

### DIFF
--- a/type-tests/files/createAction.typetest.ts
+++ b/type-tests/files/createAction.typetest.ts
@@ -156,6 +156,8 @@ function expectType<T>(p: T): T {
 
   // typings:expect-error
   expectType<string>(strLenAction('test').payload)
+  // typings:expect-error
+  expectType<boolean>(strLenAction('test').error)
 }
 
 /*
@@ -171,6 +173,38 @@ function expectType<T>(p: T): T {
 
   // typings:expect-error
   expectType<string>(strLenMetaAction('test').meta)
+  // typings:expect-error
+  expectType<boolean>(strLenMetaAction('test').error)
+}
+
+/*
+ * Test: adding boolean error with prepareAction
+ */
+{
+  const boolErrorAction = createAction('boolError', (payload: string) => ({
+    payload,
+    error: true
+  }))
+
+  expectType<boolean>(boolErrorAction('test').error)
+
+  // typings:expect-error
+  expectType<string>(boolErrorAction('test').error)
+}
+
+/*
+ * Test: adding string error with prepareAction
+ */
+{
+  const strErrorAction = createAction('strError', (payload: string) => ({
+    payload,
+    error: 'this is an error'
+  }))
+
+  expectType<string>(strErrorAction('test').error)
+
+  // typings:expect-error
+  expectType<boolean>(strErrorAction('test').error)
 }
 
 /*
--- a/type-tests/files/createAction.typetest.ts
+++ b/type-tests/files/createAction.typetest.ts
@@ -157,7 +157,7 @@ function expectType<T>(p: T): T {
   // typings:expect-error
   expectType<string>(strLenAction('test').payload)
   // typings:expect-error
-  expectType<boolean>(strLenAction('test').error)
+  const error: any = strLenAction('test').error
 }
 
 /*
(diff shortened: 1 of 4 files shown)


markerikson:
Y'know, that's a good point.

It's frankly the least-used part of the FSA spec (most folks just dispatch `SOME_ACTION_ERROR` instead), but this would be a reasonable thing to include.

Can you add some types tests for this functionality as well?

@phryneas , any thoughts here?

phryneas:
Looks good to me, but this still need type-tests.

@tvanier : take a look over here : https://github.com/reduxjs/redux-starter-kit/blob/master/type-tests/files/createAction.typetest.ts

Some additional tests that if prepareAction returns an error, it is also defined on the created Action with the same type. And additionally, just add some `// expect-error` cases to existing tests, meaning that if no error is returned from prepareAction, no `error` property is present on the returned action.

markerikson:
If we can get this in in the next day or two, we can get this into 1.0.  Otherwise, we'll just add it afterwards (1.0.1 or something).

tvanier:
Sorry for having missed some tests, I am not too familiar with type-testing, just added some as suggested by @phryneas , let me know what you think. And please do not block the `1.0` release if this PR is not ready! (although supporting `error` would help our project)

type-tests/files/createAction.typetest.ts
```diff
@@ -156,6 +156,8 @@ function expectType<T>(p: T): T {
 
   // typings:expect-error
   expectType<string>(strLenAction('test').payload)
+  // typings:expect-error
+  expectType<boolean>(strLenAction('test').error)
```
phryneas: Please replace this with something like  
```typescript
// typings:expect-error
const x = strLenMetaAction('test').error
```
you want to see that the property does not exist at all, while the 
```typescript
// typings:expect-error 
expectType<boolean>(...)
```
 just ensures that it isn't boolean.

type-tests/files/createAction.typetest.ts
```diff
@@ -171,6 +173,38 @@ function expectType<T>(p: T): T {
 
   // typings:expect-error
   expectType<string>(strLenMetaAction('test').meta)
+  // typings:expect-error
+  expectType<boolean>(strLenMetaAction('test').error)
```
phryneas: same here

type-tests/files/createAction.typetest.ts
```diff
@@ -171,6 +173,38 @@ function expectType<T>(p: T): T {
...
+{
+  const boolErrorAction = createAction('boolError', (payload: string) => ({
+    payload,
+    error: true
+  }))
+
+  expectType<boolean>(boolErrorAction('test').error)
+
+  // typings:expect-error
+  expectType<string>(boolErrorAction('test').error)
```
phryneas: from here on it's fine - that's excatly what we want to test here :)

type-tests/files/createAction.typetest.ts
tvanier: I updated the test as suggested.
Out of curiosity, would it be possible to add Jest expectations here?
```ts
  // typings:expect-error
  const error: any = strLenAction('test').error
  expect(error).toBeUndefined()
```

type-tests/files/createAction.typetest.ts
tvanier: done

type-tests/files/createAction.typetest.ts
tvanier: sounds good!

type-tests/files/createAction.typetest.ts
phryneas: I don't think that code is actually ever run, it should only be compiled to look for compiler warnings. Doing something like this would work better in the normal tests.

We just want to make sure that there's no regression in the types & autocompletion.

tvanier:
Thanks @phryneas .
@markerikson It would be great if this could be merged before releasing `1.0`, I am adding the starter kit to a project where `action.error` is used already.

markerikson:
Cool.  Awright, in it goes!

tvanier:
Great, thanks @markerikson and @phryneas !
Do we have an ETA for `1.0`, or `0.10` maybe?

markerikson:
Coming Soon (TM) :)

PR #225, opened 2019-10-24
Correct log output when introing createSlice

gjastrab:
Initial example block demonstrating `createSlice` didn't include
type including the slice's name as the prefix.

Also fix missing tilde before block.


markerikson:
Thanks!

PR #226, opened 2019-10-24
Migrate Docs to Docusaurus 2 Alpha

aswathkk:
Live Preview:
[https://redux-starter-kit.netlify.com/](https://redux-starter-kit.netlify.com/)
Here are some screenshots:
Home
Quick Start

[Issue](https://github.com/facebook/docusaurus/issues/1834)
cc: @yangshun, @endiliey


website/docusaurus.config.js
```diff
@@ -0,0 +1,105 @@
+/**
+ * Copyright (c) 2017-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// See https://docusaurus.io/docs/site-config for all the possible
```
endiliey: delete this. That's v1 remnant

website/docusaurus.config.js
```diff
@@ -0,0 +1,105 @@
+/**
```
endiliey: delete copyright

website/package.json
```diff
@@ -1,13 +1,29 @@
...
   "scripts": {
-    "start": "docusaurus-start",
-    "build": "docusaurus-build",
-    "publish-gh-pages": "docusaurus-publish",
-    "write-translations": "docusaurus-write-translations",
-    "version": "docusaurus-version",
-    "rename-version": "docusaurus-rename-version"
+    "start": "docusaurus start",
+    "build": "docusaurus build",
+    "publish-gh-pages": "docusaurus publish",
```
endiliey:
```suggestion
    "publish-gh-pages": "docusaurus deploy",
```

website/package.json
```diff
@@ -1,13 +1,29 @@
...
+    "write-translations": "docusaurus write-translations",
```
endiliey:
```suggestion
```

website/package.json
```diff
@@ -1,13 +1,29 @@
...
+    "version": "docusaurus version",
```
endiliey:
```suggestion
```

website/package.json
```diff
@@ -1,13 +1,29 @@
...
+    "rename-version": "docusaurus rename-version"
```
endiliey:
```suggestion
```

website/src/pages/index.js
```diff
@@ -0,0 +1,153 @@
...
+
+import React from 'react';
+import classnames from 'classnames';
+import Layout from '@theme/Layout';
+import Link from '@docusaurus/Link';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+import useBaseUrl from '@docusaurus/useBaseUrl';
+import styles from './styles.module.css';
+
+const siteConfig = require(`../../docusaurus.config.js`)
```
endiliey: consume it from React context. Its because using require like this is not a good idea and all the defaulted value of siteConfig won't be loaded

website/src/pages/index.js
```diff
@@ -0,0 +1,153 @@
...
+    </Layout>
+  );
+}
+
+function docUrl(doc, language) {
+  return `${siteConfig.baseUrl}${language ? `${language}/` : ''}${doc}`
+}
+
+function imgUrl(img) {
+  return `${siteConfig.baseUrl}img/${img}`
```
endiliey: use `useBaseUrl` react hooks instead of this

website/src/pages/index.js
```diff
@@ -0,0 +1,153 @@
...
+            </div>
+          </section>
+        )}
+      </main>
+    </Layout>
+  );
+}
+
+function docUrl(doc, language) {
+  return `${siteConfig.baseUrl}${language ? `${language}/` : ''}${doc}`
```
endiliey: is this even used ? there's no language here. It would be nice to understand whats happening on the source code

website/src/pages/index.js
aswathkk: a commit has been made to resolve these

aswathkk:
@endiliey Live preview is available at https://redux-starter-kit.netlify.com/

markerikson:
As a first quick glance: looking at the Advanced Tutorial, the code block font seems smaller, and the `diff`-language blocks are not getting highlighted.

website/src/pages/404.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+
+function ErrorPage() {
+  const context = useDocusaurusContext();
+  const { siteConfig = {} } = context;
+
+  function getTrackingScript() {
+    if (!siteConfig.gaTrackingId) {
+      return null
```
endiliey: there is no such field

https://v2.docusaurus.io/docs/docusaurus.config.js

endiliey: btw why do you need 404.js ? docusaurus handle it for you

website/docusaurus.config.js
```diff
@@ -0,0 +1,98 @@
...
+              label: 'GitHub',
+              href: 'https://www.github.com/reduxjs/redux/',
+            },
+          ],
+        },
+      ],
+      copyright: `Copyright (c) 2015-present Dan Abramov and the Redux documentation authors.`,
+    },
+    image: 'img/redux-logo-landscape.png',
+    twitterImage: 'img/redux-logo-twitter.png',
```
endiliey: does this exist ? i dont think so

website/src/pages/index.js
```diff
@@ -0,0 +1,145 @@
...
+              to={useBaseUrl('introduction/quick-start')}
+            >
+              Get Started
+            </Link>
+          </div>
+        </div>
+      </header>
+      <main>
+        {features && features.length && (
+          <section className={[styles.features, styles.lightBackground].join(' ')}>
```
wgao19: the `lightBackground` style here is breaking dark mode now

website/src/pages/index.js
```diff
@@ -0,0 +1,145 @@
...
+            <div className={`container ${styles.featureBlock}`}>
+              <div className="row">
+                {features.map(({ image, title, content }, idx) => (
+                  <div
+                    key={idx}
+                    className={classnames('col col--3', styles.feature)}
+                  >
+                    {image && (
+                      <div className={`text--center ${styles.blockImage}`}>
+                        <img
```
wgao19: if we use ``, they won't be able to pick up colors and therefore will not adapt to dark theme here neither. Probably will need to change the implementation here.

endiliey:
Regarding the font size and the diff not showing any color. It's mostly due to theme. The default theme we chose: Palenight has that problem.

Here's some comparison
palenight

dracula

vsDark

website/src/css/custom.css
```diff
@@ -0,0 +1,23 @@
...
+/* You can override the default Infima variables here. */
+:root {
+  --ifm-color-primary: #764ABC;
+  --ifm-color-primary-dark: #613D9A;
+  --ifm-color-primary-darker: #4C3078;
+  --ifm-color-primary-darkest: #362256;
+  --ifm-color-primary-light: #825AC2;
+  --ifm-color-primary-lighter: #9B7BCE;
+  --ifm-color-primary-lightest: #B49CDA;
+}
```
wgao19: I believe you can add some more css here to address the styling issue addressed by @markerikson 

> As a first quick glance: looking at the Advanced Tutorial, the code block font seems smaller, and the diff-language blocks are not getting highlighted.

The `--ifm-font-size-base` defaults to `15px` from Infima, the underlying CSS framework that Docusaurus 2 uses. Where the base font size in the current RSK site is `16px`.

Although, directly changing it to `16px` causes some of the navbar item to wrap text. So you may need to tweak there a little bit.

image

Also the default `--ifm-code-font-size` is `85%` which causes the code block to become smaller.

image

Meanwhile, the classes in the `diff` language blocks are properly picked up, but there's no styling. This also can be fixed by adding some custom styling -- just some quick thoughts, there can be plugins / themes that already pick this up.

image

markerikson: I believe we're currently using Monokai as the syntax highlighting theme.

website/src/pages/404.js
aswathkk: Docusaurus is returning a blank content for 404. Don't know why.

endiliey: Try https://v2.docusaurus.io/sdasjasad

aswathkk: But why https://redux-starter-kit.netlify.com/fdsafdsf is showing a blank content?
I tried removing src/pages/404.js also. But no use.

endiliey: oh its because redux route basepath is empty so it routes to upper level docpage. Its equivalent to https://v2.docusaurus.io/docs

Seems like a bug

endiliey: Did adding 404.js solve the issue though ?

endiliey: This is fixed in next alpha release. 

You can safely remove the 404.js here

aswathkk: Adding 404.js did not resolve the issue.
> oh its because redux route basepath is empty so it routes to upper level docpage. Its equivalent to https://v2.docusaurus.io/docs
Sorry. I didn't get this part.

aswathkk:
@markerikson Can we use vsDark as the syntax highlighting theme?

website/src/pages/404.js
endiliey: It's a react router concept. You can ignore it for now.

But just to share knowledge
```js
// routes.js
export default [
  {
    path: "/docs",
    component: ComponentCreator("/docs"),
    routes: [
      {
        path: "/docs/advanced-themes",
        component: ComponentCreator("/docs/advanced-themes"),
        exact: true
      },
      {
        path: "/docs/advanced-plugins",
        component: ComponentCreator("/docs/advanced-plugins"),
        exact: true
      },
    ]
  },
  {
    path: "*",
    component: ComponentCreator("*")
  }
];
```
imagine above scenario. It will match ;/docs/;

For redux case, it;s not `/docs` but `/` that's why you dont have 404 but an empty page, its rendering the parent nested route

markerikson:
@aswathkk : I'd really prefer to stick with Monokai.

So where does this PR stand overall at this point?

wgao19:
hey @markerikson seems there hasn't been much updates for a week, is it possible that we merge this PR to an intermediate branch at RSK then we (as Docusaurus maintainers) can work on top of that?

Perhaps it is because monokai is not an available theme by [prism-react-renderer](https://github.com/FormidableLabs/prism-react-renderer) that Docusaurus 2 uses, but i guess we can convert [this](https://github.com/Swaagie/prismjs-monokai/blob/master/prism-monokai.css).

Looks to me we're mainly missing customized styling at this point.

markerikson:
Given that there's a docs preview here, do we actually need a separate branch?

yangshun:
Agree with @wgao19's approach. I think we need a separate branch so `master` can still keep the existing docs website and changes can be made to it.

endiliey:
> Given that there's a docs preview here, do we actually need a separate branch?

I think the main idea here is to let the maintainer of Docusaurus to continue @aswathkk (not a maintainer) work on the migration. 

Given that this PR isn't really ready yet, its not a good idea to push this to master. 
If we were to send another PR to RSK that supersedes this PR, that won't be nice to aswathkk since he has put lot of effort into this PR.
The concern here is that the migration has take a while and rounds of reviews is gonna put things more delayed

On another note deploy preview isn't working because the build directory has changed in v2. I recommend using a netlify config like CRA so that not only those who have access to RSK netlify.com setting can modify it.
https://github.com/facebook/create-react-app/blob/master/netlify.toml

markerikson:
Okay, just created a `docusaurus-v2-integration` branch.

Looks like there's conflicts from my recent changes to the tutorials.  Fix those and we can merge this into the integration branch.

aswathkk:
@markerikson I've resolved the conflicts.

markerikson:
It's failing to build atm, but yeah, let's go ahead and get this into the integration branch.

PR #231, opened 2019-10-26
Add prepare callback usage to createSlice.md

ronaldborman:
Added a description of the usage of a prepare callback to the createSlice API reference page. Also adapted the multiply case reducer in the example to demonstrate the usage.

### DIFF
--- a/docs/api/createSlice.md
+++ b/docs/api/createSlice.md
@@ -82,6 +82,8 @@ to force the TS compiler to accept the computed property.)
 Each function defined in the `reducers` argument will have a corresponding action creator generated using [`createAction`](./createAction.md)
 and included in the result's `actions` field using the same function name.
 
+If you need to customize the creation of the payload value of an action creator by means of a [`prepare callback`](https://redux-starter-kit.js.org/api/createaction#using-prepare-callbacks-to-customize-action-contents), the value of the appropriate field of the `reducers` argument object should be an object instead of a function. This object must contain two properties: reducer and prepare. The value of the reducer field should be the case reducer function while the value of the prepare field should be the prepare callback function.
+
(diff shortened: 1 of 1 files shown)


docs/api/createSlice.md
```diff
@@ -82,6 +82,8 @@ to force the TS compiler to accept the computed property.)
 Each function defined in the `reducers` argument will have a corresponding action creator generated using [`createAction`](./createAction.md)
 and included in the result's `actions` field using the same function name.
 
+If you need to customize the creation of the payload value of an action creator by means of a [`prepare callback`](https://redux-starter-kit.js.org/api/createaction#using-prepare-callbacks-to-customize-action-contents), the value of the appropriate field of the `reducers` argument object should be an object instead of a function. This object must contain two properties: reducer and prepare. The value of the reducer field should be the case reducer function while the value of the prepare field should be the prepare callback function.
```
markerikson:
```suggestion
If you need to customize the creation of the payload value of an action creator by means of a [`prepare callback`](./createAction.md#using-prepare-callbacks-to-customize-action-contents), the value of the appropriate field of the `reducers` argument object should be an object instead of a function. This object must contain two properties: reducer and prepare. The value of the reducer field should be the case reducer function while the value of the prepare field should be the prepare callback function.
```

docs/api/createSlice.md
```diff
@@ -82,6 +82,8 @@ to force the TS compiler to accept the computed property.)
 Each function defined in the `reducers` argument will have a corresponding action creator generated using [`createAction`](./createAction.md)
```
markerikson: Looks like the formatting check failed.  Please run the changes through Prettier.

markerikson:
Thanks!

Issue #224, opened 2019-10-23
Possibly incorrect type for `createAction`


ianp:
I don't think that the recently added `error` field is getting included in the type returned by `createAction`, see

```ts
import { createAction, createSlice } from 'redux-starter-kit'

type MySliceState = { counter: number; error?: any }

const oops = createAction('oops', (error: any) => ({ payload: {}, error }))

const slice = createSlice({
  name: 'mySlice',
  initialState: { counter: 1 } as MySliceState,
  reducers: {},
  extraReducers: {
    [oops.type]: (state, action: ReturnType<typeof oops>) => {
      state.error = action.error // !!! DOES NOT COMPILE !!!
    }
  }
})
```

I get an error like property `error` does not exist on type `WithPayload<{}, Action<"oops">>`, but I think that the inferred type should include `WithOptional`?

Of course it's also possible that I'm specifying the action type incorrectly!

markerikson:
I _think_ this is due to declaring `error` as `any`.

Using your example, the return type of `oops` is:

```ts
const oops: WithTypeProperty<"oops", (error: any) => WithPayload<{}, Action<"oops">>>
```

However, if I just change `error: any` to `error: boolean`, it suddenly becomes:

```ts
const oops: WithTypeProperty<"oops", (error: boolean) => WithOptional<void, boolean, WithPayload<{}, Action<"oops">>>>
```

Ditto if I use `string`.

So, something about `any` vs a meaningful type kicks it into a different type definition path.  Seems like the immediate solution is just to give `error` a non-`any` type.

@phryneas, @jessidhia , @tvanier : just out of curiosity, any idea why we're hitting that path, and if it can be tightened up?

ianp:
Yep, that's looks right! Interestingly, it works find with `unknown` which is probably a better option than `any` here anyway.

phryneas:
Yes, `any` is a little bit of a problem child, as `any` always takes both paths in ternaries:
`type X = any extends {} ? "yes" : "no"` will result in X being `"yes" | "no"`.

Will have to catch that as early as possible - maybe I find the time to look into this on the weekend.

ianp:
Maybe this for `PrepareAction`?

```ts
export type PrepareAction<P, M = void, E = void> = (
  ...args: any[]
) => Omit<WithOptional<M, E, WithPayload<P, Action>>, 'type'>
```

(or something similar using `Exclude` to support older TS versions).

[...]

PR #233, opened 2019-10-26
use `never` instead of `void` for PayloadAction Meta & Error

phryneas:
see #224
This prevents `any` from taking the `void` branch in the PayloadAction type - really distinguishing between the two was quite difficult, so I changed to `never`, which might have been the more logical choice in the first place.


markerikson:
Thanks!

Issue #228, opened 2019-10-24
How to integrate redux-observable with typescript?


on3iro:
I am currently trying to find a way to integrate redux-observable with typescript.
It looks like I need to specify a root epic type  for the observable-middleware like this `epicMiddleWare<RootAction, RootAction, RootState>`.

Unfortunately I am not quite sure how I can build a union type of all my actions created with `createSlice` so that it is usable by the observable-middleware. Can you point me into any direction of how to integrate redux-observable?

markerikson:
Hmm. I haven't worked with redux-observable myself.  You might get a better answer asking over on that repo.

The generated action types are available as `someSlice.actions.someActionCreator.type`.  You should be able to somehow define a union of all of those.

on3iro:
Alright - thanks for the quick response :)

I'll head over to the observable repo and ask my question there. If I find a good way to solve this I'll leave another comment here :+1:

on3iro:
@markerikson am I correct in assuming, that the action types produced by `createSlice` are just `string` and no type specific constants? That would mean that it is impossible to narrow down action types inside other functions if I am not mistaken? (I am by no means a typescript expert)

markerikson:
As far as I know, yes.  That's because we're concatenating together two strings to generate the type, not defining a specific literal up front.

@phryneas could confirm.

phryneas:
As typescript can not concatenate literal string types, yes, we cannot generate those. There are multiple issues over in the TS repo ion that.

phryneas:
Would https://github.com/reduxjs/redux-starter-kit/pull/239 solve your problem?

on3iro:
@phryneas that looks like it might work :)

phryneas:
Then let's see what @markerikson thinks about it :)

erodactyl:
I found a neat solution for getting the RootAction type from actions - you can take the union type of all your action in the form ReturnType<typeof someAction>

ChristianUlbrich:
@erodactyl `Code` or it did not _happen_. :D

erodactyl:
Well, here you go :D 

`export type Action =
  | ReturnType<typeof addMessageAction>
  | ReturnType<typeof setTypingAction>
  | ReturnType<typeof getConversationAction>;
`
and then

`const epic: Epic<Action, Action, void, Dependencies> = ....`

efossvold:
[...]

PR #239, opened 2019-10-29
RFC: add "match" method to actionCreator

phryneas:
This was triggered by #228 and inspired by https://github.com/aikoven/typescript-fsa - their actionCreators have a `match` method that works as a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) and can apparently [be used with redux-observable](https://github.com/aikoven/typescript-fsa#redux-observable).

This would add the same to our actionCreators. What do you think?

I've added a special typeTest for that useCase: https://github.com/phryneas/redux-starter-kit/blob/actionCreator-match/type-tests/files/createAction.typetest.ts#L279-L290


markerikson:
Can you rebase this against master now that #233 got merged in?

phryneas:
Rebased :+1:

markerikson:
Looks neat!

PR #240, opened 2019-10-30
update immer to v5

phryneas:
*it might be better to let this rest for a week or so*

Immer v5 was released today, with support for Map and Set. 
While I believe it is usually adviced against using those in Redux, I also know that people do so anyways - those might benefit from this update.

On the other hand, this will bump the size of the immer dependency from gzipped 4.5kb to 6.3kb. ( https://bundlephobia.com/result?p=immer@5.0.0 ).

There is https://github.com/immerjs/immer/pull/449 which aims to reduce bundle size again, and this is very new, so it might be a good idea to monitor https://github.com/immerjs/immer/pull/354 if any other issues arise and decide on merging this in a week or two, so this PR is more of a reminder to check back later ;)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3868,9 +3868,9 @@
       "dev": true
     },
     "immer": {
-      "version": "4.0.1",
-      "resolved": "https://registry.npmjs.org/immer/-/immer-4.0.1.tgz",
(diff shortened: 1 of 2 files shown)


markerikson:
The biggest issue here is that our serialization check middleware will explicitly throw errors if it sees a `Map` or `Set` in the state or action.  So, this doesn't really buy us anything useful.

I'm not opposed to updating eventually, but that selling point doesn't help here.

phryneas:
Good point, hadn't thought of that!

PR #241, opened 2019-10-31
Remove the unused `ActionCreator` import in the tutorial doc

christineoo:
The `ActionCreator` import in the **app/store.ts** diff is not needed and can be removed. I have double-checked it by looking at the [commit](https://github.com/reduxjs/rsk-github-issues-example/commit/2ac93bb089705847a8ce349864d885a5039eff4b) of adding AppThunk type.

Please review and merge. Thanks!


markerikson:
Thanks!

PR #251, opened 2019-11-13
use modern syntax in redux-observable example

evertbouw:
RxJS 6 deprecated method operators in favor of pipeline operators. These can be imported individually which improves developer experience and bundle size.


markerikson:
Thanks!

PR #257, opened 2019-11-16
Inline values in error messages

### LINKED ISSUE #253: Error messages appear to use placeholders
Hey there, I'm trying to migrate to this from a new-ish standalone React/Redux set up. I'm using React Native through Expo. Obviously there's something wrong with my code and I'm trying to debug where I've gone wrong, but a number of the error messages I'm receiving simply have placeholders without complete context.

```
A non-serializable value was detected in the state, in the path: `%s`. Value: %o
Take a look at the reducer(s) handling this action type: %s.
(See https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state), messages.0.createdAt, 2019-11-14T05:30:38.729Z, messages/addMessages
- node_modules/expo/build/environment/muteWarnings.fx.js:27:24 in error
- node_modules/@reduxjs/toolkit/dist/redux-toolkit.cjs.development.js:165:24 in <unknown>
- node_modules/redux-immutable-state-invariant/dist/index.js:55:36 in <unknown>
* screens/messages.js:23:38 in onSend
- node_modules/react-native-gifted-chat/lib/GiftedChat.js:115:34 in <unknown>
- node_modules/react-native-gifted-chat/lib/Send.js:26:27 in onPress
- ... 23 more stack frames from framework internals

A non-serializable value was detected in an action, in the path: `%s`. Value: %o
Take a look at the logic that dispatched this action:  %o.
(See https://redux.js.org/faq/actions#why-should-type-be-a-string-or-at-least-serializable-why-should-my-action-types-be-constants), payload.0.createdAt, 2019-11-14T05:30:43.585Z, Object {
  "payload": Array [
    Object {
      "_id": "3f72cd2c-fb9e-4806-a14b-3efcd6c77b75",
      "createdAt": 2019-11-14T05:30:43.585Z,
      "text": "Testing",
      "user": Object {
        "_id": 1,
      },
    },
  ],
  "type": "messages/addMessages",
}
- node_modules/expo/build/environment/muteWarnings.fx.js:27:24 in error
- node_modules/@reduxjs/toolkit/dist/redux-toolkit.cjs.development.js:155:24 in <unknown>
- node_modules/redux-immutable-state-invariant/dist/index.js:55:36 in <unknown>
* screens/messages.js:23:38 in onSend
- node_modules/react-native-gifted-chat/lib/GiftedChat.js:115:34 in <unknown>
- node_modules/react-native-gifted-chat/lib/Send.js:26:27 in onPress
- ... 23 more stack frames from framework internals

```

kevin940726:
Fix #253.

Inline the values in the error messages to fix that some environment won't properly display [string substitutions](https://developer.mozilla.org/en-US/docs/Web/API/console#Using_string_substitutions). By using `,` to separate each value, we can remain native formatter in each environment (node.js or all major browsers devtool). Since the value we want to log here might be non-serializable, converting them to strings often loose context and make the developer confused.

I also rewrite the tests to better represent how it's actually seen in the logs, so that we can guard it when something like this issue ever occurred in the future. A caveat here is that by doing so, our tests might only be able to run in node v10+.

### DIFF
--- a/src/serializableStateInvariantMiddleware.test.ts
+++ b/src/serializableStateInvariantMiddleware.test.ts
@@ -1,4 +1,6 @@
 import { Reducer } from 'redux'
+import { Console } from 'console'
+import { Writable } from 'stream'
 import { configureStore } from './configureStore'
 
 import {
@@ -79,8 +81,33 @@ describe('findNonSerializableValue', () => {
 })
 
 describe('serializableStateInvariantMiddleware', () => {
+  let log = ''
+  const originalConsole = window.console
+
   beforeEach(() => {
-    console.error = jest.fn()
+    log = ''
+
+    const writable = new Writable({
+      write(chunk, encoding, callback) {
+        log += chunk
+        callback()
+      }
+    })
+
+    const mockConsole = new Console({
+      stdout: writable,
+      stderr: writable
+    })
+
+    Object.defineProperty(window, 'console', {
+      value: mockConsole
+    })
+  })
+
+  afterEach(() => {
+    Object.defineProperty(window, 'console', {
+      value: originalConsole
+    })
   })
 
   it('Should log an error when a non-serializable action is dispatched', () => {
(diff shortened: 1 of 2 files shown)


src/serializableStateInvariantMiddleware.test.ts
```diff
@@ -98,19 +125,12 @@ describe('serializableStateInvariantMiddleware', () => {
...
-      value,
-      action
-    ] = (console.error as jest.Mock).mock.calls[0]
-
-    expect(message).toContain('detected in an action, in the path: `%s`')
-    expect(keyPath).toBe('type')
-    expect(value).toBe(type)
-    expect(action).toBe(dispatchedAction)
+    expect(log).toMatchInlineSnapshot(`
+      "A non-serializable value was detected in an action, in the path: \`type\`. Value: Symbol(SOME_CONSTANT)
```
markerikson: Are the backslashes actually needed here?  They should be escape characters and thus not appears in the actual output string, I would thinkg.

kevin940726: It’s auto generated by Jest via `toMatchInlineSnapshot`. I could manually change that if you agree though.

Issue #237, opened 2019-10-29
Add an action listener callback middleware


markerikson:
Thunks are easy to use and a good default, but the biggest weakness is that they don't let you respond to dispatched actions.  Sagas and observables are very powerful (_too_ powerful for most apps), but they do let you kick off additional logic in response to actions.

I've been considering adding some kind of middleware that would be in between - something that lets you run callback functions in response to specific actions, but without the complex overhead of sagas and observables.

I have links to a bunch of existing middleware under the [Middleware](https://github.com/markerikson/redux-ecosystem-links/blob/master/middleware.md) and [Middleware - Async](https://github.com/markerikson/redux-ecosystem-links/blob/master/middleware-async.md) sections of my addons list.  One of those might do what we need, or we can take inspiration.

Glitch appears to have one at [`handler-middleware.js`](https://gist.github.com/modernserf/e890a43887b1f7a80e27300481bffb4b) , which maybe @modernserf could say something about.

modernserf:
This is how I described this approach to my coworkers:
---
There are essentially two schools of thought for working with side effects with redux.

The first approach is to put the side effects in the actions -- e.g. instead of dispatching a simple value, dispatch a promise, or a function that calls side effects. Middleware like redux-promise and redux-thunk intercept these non-value actions and run them.

The second approach is to put the side effects directly into the middleware -- the actions are always simple values, but the middleware listens to these actions, and can run side effects in response to them. This is the approach used by redux-saga and redux-loop; it's also how logging and metrics-reporting middleware work.

I prefer the second approach because it encourages you to keep the "how" with the state, and the "what" with the UI. However, the tools associated with this approach, particularly redux-saga, tend to have a pretty steep learning curve. The approach I'm using here is less expressive, but much closer to the complexity of redux-thunk.

markerikson:
Yep, that's the idea I'm going for here.

One possible improvement I'd like to consider is adding and removing subscriptions at runtime.  The likely approach would be `dispatch(addActionListener(type, callback))`.

dpyzo0o:
What's the status now? It seems to me that the `extraReducers` can be used to respond to dispatched actions. Or are you working on something else?

phryneas:
> What's the status now? It seems to me that the `extraReducers` can be used to respond to dispatched actions. Or are you working on something else?

Keep in mind, these are still reducers and should not have any side-effects.

[...]

PR #262, opened 2019-11-23
alternative callback-builder-style notation for actionsMap

phryneas:
This is what I had suggested on twitter. Right now, it's just implemented on createReducer, but if you agree that we go forward with this, I'll add it to extraReducers, too.


type-tests/files/createReducer.typetest.ts
```diff
@@ -63,3 +63,55 @@ function expectType<T>(p: T) {}
...
+      expectType<{ type: 'increment' }>(action)
+      // typings:expect-error
+      expectType<{ type: 'decrement' }>(action)
+      // typings:expect-error - this cannot be inferred and has to be manually specified
+      expectType<{ type: 'increment'; payload: number }>(action)
+    })
+
+    builder.addCase(
+      increment,
+      (state, action: ReturnType<typeof increment>) => state
```
markerikson: Somehow I'd actually kind of figure this action type could get inferred automatically, but maybe that's not the case.

markerikson:
Looks pretty neat.  I assume we wouldn't want to allow this for `createSlice`'s `reducers` arg, because the point of that is we're using that to _generate_ the action types, right?  This is only good for when we've already got action types defined and are now generating reducers to match them.

type-tests/files/createReducer.typetest.ts
phryneas: It does (see line 86), this is just there to test if it can actually also be specified without error.

phryneas:
> Looks pretty neat. I assume we wouldn't want to allow this for `createSlice`'s `reducers` arg, because the point of that is we're using that to _generate_ the action types, right? This is only good for when we've already got action types defined and are now generating reducers to match them.

Exactly. I'm seeing a use case for `createReducer`'s `actionMap`, and `createSlice`s `extraReducers`.

Maybe also for some kind of async middleware in the future.

markerikson:
Yeah, go ahead and rig it up for `extraReducers`, then.

> Maybe also for some kind of async middleware in the future.

How so?

phryneas:
Intersting sidenote: there's no codesandbox being generated for this. Maybe because it's a Draft PR? 🤔

phryneas:
> Yeah, go ahead and rig it up for `extraReducers`, then.
> 
> > Maybe also for some kind of async middleware in the future.
> 
> How so?

The most simple async middleware I could think of would be something like an actionEffectsMap, either as an argument on `configureStore`, or even on a per-slice-level. Something like

```typescript
configureStore({
// ...
asyncEffects: {
  [todosRequested.type]:  async (action: ReturnType<typeof todosRequested>, { getState, dispatch }) => {
    // async stuff here
  } 
}
});
```

If we did something like that, that could benefit from a similar alternative notation.

markerikson:
Hey, good point.

And yes, a simple action -> callback map is basically what I'm picturing.  Already opened up a threat to discuss it over in #237 .

markerikson:
btw, after this, it'd be nice if we could add an actual "Usage with TypeScript" docs page that rephrases the info in the "Advanced Tutorial" and talks about using this builder.

phryneas:
Yeah. Also, the docblocks will need some updating. Do you know how to correctly write docblocks for overloaded methods?

markerikson:
Not specifically, no, and tbh the types I've listed in the API reference pages are sorta hand-wavy anyway (like references to a `ReducerFunction` type).

I'd say write the page for the callback builder first, then update the other couple pages to point to that and say that the args can either be a hand-written object or the result of the builder.

phryneas:
Code-wise this should be it.

I'm gonna take a look at the docs tomorrow - I should get some sleep now :)

markerikson:
@phryneas : maybe we lost the CodeSandbox integration due to the repo name change?

phryneas:
> @phryneas : maybe we lost the CodeSandbox integration due to the repo name change?

I guess so. Could you remove & re-add the github integration for a test?

I've now added the missing docblock comments.

As for the documentation page, writing just doesn't come to me today.
But I've created a codeSandbox with basically the content I would add to that page, without all the fancy words inbetween. If you're in a writing mood, could you take over from there?
https://codesandbox.io/s/rtk-ts-documentation-z8phk

Otherwise, I'll try again in a few days. That writing mood is unpredictable.

markerikson:
I didn't even set up the actual integration - I filled out the form and got an email from someone, I think.

Lemme see if I can find that email and reply asking for an update.

I'm trying to convince myself to work on a blogpost atm, so I'm gonna hold off on trying to write docs for this.

markerikson:
I just created a new issue for the TS usage page and copied these comments over there.

phryneas:
> I didn't even set up the actual integration - I filled out the form and got an email from someone, I think.

I think you have to go to https://github.com/apps/codesandbox and re-activate it for the renamed sandbox.

Aside from that - is this ready in your eyes/can I merge this or do you want to wait for the documentation to catch up?

garethx:
You'll need to enable the new repo in the CodeSandbox CI app - we only see reduxjs/redux-starting-kit enabled atm.

markerikson:
Yeah, we really oughta merge this in :)

Can you add info on this to the API docs as well?  We can do that as a separate PR.

phryneas:
@garethx by now, codesandbox builds are running again, but I just noticed that codeSandbox cannot find types on the CI builds. See this sandbox: https://codesandbox.io/s/friendly-hoover-g0sni
The same code in a non-CI-build seems to work: https://codesandbox.io/s/relaxed-bhabha-iqp0u
Is this a known issue? If not, where would be the best place to open an issue? The codesandbox repo structure is a bit confusing to me.

PR #268, opened 2019-12-01
document alternative callback-builder-style notation for actions…

phryneas:
the missing documentation für #262

### DIFF
--- a/docs/api/createSlice.md
+++ b/docs/api/createSlice.md
@@ -111,21 +111,21 @@ for references in a larger codebase.
 
 ## Examples
 
-```js
-import { createSlice } from '@reduxjs/toolkit'
+```ts
+import { createSlice, createAction, PayloadAction } from '@reduxjs/toolkit'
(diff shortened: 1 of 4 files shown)


docs/api/createSlice.md
```diff
@@ -107,6 +115,8 @@ for references in a larger codebase.
 import { createSlice } from '@reduxjs/toolkit'
 import { createStore, combineReducers } from 'redux'
 
+const incrementBy = createAction < number > 'incrementBy'
```
phryneas: ummmh... wtf is prettier doing here...?

phryneas: @markerikson I can't convince prettier of `const incrementBy = createAction<number>('incrementBy')` here...
Do you face the same problem?

markerikson: The code block is marked as `js`, not `ts`, so it's being mis-formatted.

And given that we aren't using any other TS in there, the example is kinda inconsistent anyway.

PR #269, opened 2019-12-01
Fix docs: builder.add -> builder.addCase

le0nik:


markerikson:
Thanks!

phryneas:
Thanks, good catch.

PR #271, opened 2019-12-05
(chore) update test snapshot

phryneas:
Since a few days, the tests are failing - it seems that updating this inline snapshot has been forgotten in #257 or some serializer was added afterwards. No matter why, this will make the tests green again.

### DIFF
--- a/src/serializableStateInvariantMiddleware.test.ts
+++ b/src/serializableStateInvariantMiddleware.test.ts
@@ -226,12 +226,9 @@ describe('serializableStateInvariantMiddleware', () => {
       store.dispatch({ type: ACTION_TYPE })
 
       // since default options are used, the `entries` function in `serializableObject` will cause the error
-      expect(log).toMatchInlineSnapshot(`
-        "A non-serializable value was detected in the state, in the path: \`testSlice.a.entries\`. Value: [Function: entries] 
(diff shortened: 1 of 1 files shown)


phryneas:
Okay, that makes codesandbox green and travis red, before that it was the other way round. :thinking: 

Maybe this is something with different node versions? I'll investigate this later and leave this PR up as a reminder.

phryneas:
Ah well, since the value is serialized differently in different systems I guess the most practical approach is just to match until the beginning of the serialized string. It's all about that path anyways.

phryneas:
@markerikson are you fine with me merging this?

markerikson:
Sure, go ahead.

kevin940726:
Oops! Actually I've encountered similar issue not long after sending the PR. Turns out that `util.inspect` outputs different format in different node versions. I created [`console-testing-library`](https://github.com/kevin940726/console-testing-library) to solve this issue. It uses `pretty-format` from jest to normalize the output across different environments.

This solution with regex works fine, but if we want the tests to more represent the actual output, I can try integrate it in another PR. (Though I think it might be an overkill for this library.)

phryneas:
@kevin940726 I'd love this to be less hacky - and since your `console-testing-library` doesn't add any additional dependencies we aren't already using (besides itself), I'd say go for it :)

PR #277, opened 2019-12-12
Use `console-testing-library` to get consistent snapshots

kevin940726:
As mentioned in https://github.com/reduxjs/redux-toolkit/pull/271#issuecomment-564184500, `console-testing-library` is a simple package I built to help test console output more align to what users see.

It's new, so it's probably not stable for every cases, but I think it's quite stable for our simple use case.

Under the hood, it uses `new Console()` as in #257 did, and also uses `pretty-format` from jest to generate consistent output across different node versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2463,6 +2463,16 @@
       "integrity": "sha512-KbS1Y0jMtyPgIxjO7ZzMAuUpAKMt1SzCL9fsrKsX6b0zJPTaT0SiSPmewwVZg9UAO83HVIlEhZF84LIjZ0lmAw==",
       "dev": true
     },
+    "console-testing-library": {
+      "version": "0.2.2",
+      "resolved": "https://registry.npmjs.org/console-testing-library/-/console-testing-library-0.2.2.tgz",
+      "integrity": "sha512-GhVHh/kQHxBhhNNPJCBheg8fECTbVLohgjjoylQhRAxVohaWSLzdyT+Vv6EYRtwuOq9ziQ2XTEC5Mvvd7XxuWA==",
+      "dev": true,
+      "requires": {
+        "jest-snapshot": "^24.9.0",
+        "pretty-format": "^24.9.0"
+      }
+    },
     "contains-path": {
       "version": "0.1.0",
       "resolved": "https://registry.npmjs.org/contains-path/-/contains-path-0.1.0.tgz",
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "@types/jest": "^24.0.11",
     "@types/node": "^10.14.4",
     "@types/redux-immutable-state-invariant": "^2.1.1",
+    "console-testing-library": "^0.2.2",
     "eslint-config-react-app": "^5.0.1",
     "prettier": "^1.18.2",
     "react": "^16.8.6",
--- a/src/serializableStateInvariantMiddleware.test.ts
+++ b/src/serializableStateInvariantMiddleware.test.ts
@@ -1,6 +1,5 @@
 import { Reducer } from 'redux'
-import { Console } from 'console'
-import { Writable } from 'stream'
+import { getLog } from 'console-testing-library'
 import { configureStore } from './configureStore'
 
 import {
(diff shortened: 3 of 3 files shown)


phryneas:
Now both CIs work nicely - thank you very much!

markerikson:
Hmm. I just pulled master, and the serializable test file is failing with:

```
 FAIL  src/serializableStateInvariantMiddleware.test.ts
  ● Test suite failed to run

    Call retries were exceeded

      at ChildProcessWorker.initialize (node_modules/jest-worker/build/workers/ChildProcessWorker.js:193:21)
```

Any idea what's going on here?  Does this work on Windows?

markerikson:
It specifically seems to be the test on line 83, `"'Should log an error when a non-serializable action is dispatched'"`.

If I skip that one, everything else passes:

```
 PASS  src/serializableStateInvariantMiddleware.test.ts
  findNonSerializableValue
    √ Should return false if no matching values are found (5ms)
    √ Should return a keypath and the value if it finds a non-serializable value (3ms)
    √ Should return the first non-serializable value it finds (1ms)
    √ Should return a specific value if the root object is non-serializable (1ms)
    √ Should accept null as a valid value (2ms)
  serializableStateInvariantMiddleware
    √ Should log an error when a non-serializable value is in state (13ms)
    √ Should use the supplied isSerializable function to determine serializability (1ms)
    √ should not check serializability for ignored action types (1ms)
    ○ skipped Should log an error when a non-serializable action is dispatched
    consumer tolerated structures
      √ Should log an error when a non-serializable value is nested in state (1ms)
      √ Should use consumer supplied isSerializable and getEntries options to tolerate certain structures (2ms)
```

If I include it, the test file fails with no actual error output:

```
 RUNS  src/serializableStateInvariantMiddleware.test.ts
npm ERR! Test failed.  See above for more details.
```

I want to get 1.2.0 out, but this needs to be resolved first.  Can one of you take a look at this?

kevin940726:
@markerikson I’ve seen similar result on older version of node 13, I’ve resolved it by just upgrading node to the latest version. What version are you on?

markerikson:
13.2, looks like.

What about the lib makes it dependent on a specific Node version that way, and why aren't all the other tests failing like that too?

kevin940726:
TBH, I just assumed it to be a node bug, I haven’t looked into the root cause. Related issue https://github.com/nodejs/node/issues/30730, https://github.com/facebook/jest/issues/8769

markerikson:
Uh.... I just commented out the entire body of that test, and the test run still dies the same way.  But, if I `.skip()` the test, all tests in that file pass.

wat.

kevin940726:
Looks like node 13.2 specifically is the only node version which has this problem, and it’s somehow unpredictable, at least in user-land

markerikson:
I'll try upgrading Node when I'm free later.

phryneas:
Currently I assume that this is some kind of memory leak in https://github.com/kevin940726/console-testing-library/blob/master/index.js#L177-L194 that only appears on node 13.2 with a minimum amount of tests being run - but I don't have a system where I could reproduce it. @kevin940726 , could you keep investigating please?

kevin940726:
yeah sure! but we’re around 3 am here, let me do it tmr 😅

phryneas:
Sure :D get some sleep!

kevin940726:
Did some digging, unfortunately nothing new. Found out that it's a V8 bug, which has been [fixed](https://github.com/nodejs/node/pull/30681) in Node. The bug has been happening in many tests, but what exactly triggers it is still unknown to me.

I don't think it'd be a blocker since it's just a dev dep, and we could just upgrade our node version to the latest. However, I will keep an eye on this and hope that we don't introduce any potential bugs.

phryneas:
Well, given it's a known (and fixed) node bug that leads to a real segfault, that would be fine with me personally. Depends on what @markerikson thinks about it.

One thing though: After reviewing this a second time, I'm not too fond that this automatically does a `beforeEach` and `afterEach` on import. 
That way, when writing a test all my console.logs I'm doing just to validate a value or something are lost and it's not clear why.
I would prefer it if these `beforeEach` and `afterEach` were in our test, for transparency. Or a call to `mockConsoleInAllTests` or something similar.

kevin940726:
Yeah the manual cleaning up makes sense to me. Thanks for the suggestion!

The problem is that I’m not totally sure what causes the problem, from my understanding so far is it’s a bug related to global inline cache. A warning would be nice but at the same time it might seems overly aggressive. FWIW it never happens in the test of the lib itself :thinking:

markerikson:
I tried pulling `master` and was still seeing the same error with the serializable middleware, but it did indeed go away after I upgraded from Node 13.2.0 to 13.5.0.  *shrug*

PR #280, opened 2019-12-13
Fix module bundling & ESM runtime error

### LINKED ISSUE #278: ESM module usage causes Uncaught ReferenceError: require is not defined
Hello there and thanks for this library!

I recently encountered an issue when trying to consume the ESM version of this library using Rollup.
The issue happens at runtime due to the following usage of `require`:

https://github.com/reduxjs/redux-toolkit/blob/afc11d9aaf559f6143274fd0ab530ad1ad7fd602/src/getDefaultMiddleware.ts#L57-L62

As mentioned [here]( (https://github.com/rollup/rollup/issues/1058)) and [here](https://github.com/apollographql/apollo-link/issues/558), the relevant Rollup plugin is unable to handle mixed use cases.
> Mixing import and require is definitely discouraged. The only way for Rollup to handle require statements is with rollup-plugin-commonjs, but that plugin will skip any files with import or export statements.
> — https://github.com/rollup/rollup/issues/1058#issuecomment-254187433

---

I've been working successfully with a local fix by changing `getDefaultMiddleware` to an async function, and altering the import to:
```diff
- var createImmutableStateInvariantMiddleware = require('redux-immutable-state-invariant').default 
+ var createImmutableStateInvariantMiddleware = await import('redux-immutable-state-invariant');
```

Let me know if you'd like me to make a PR for these changes.

Thanks again 👋

alex-ketch:
Closes #278 

I've tested the ESM build locally and confirmed that I no longer see the original issue.
When building for production, the `if (process.env.NODE_ENV !== 'production') {` is evaluated to `false` by Webpack and the chunk of code is omitted from output.

The contents of the generated UMD file (`redux-toolkit.umd.min.js`), hasn't changed, so there shouldn't be a problem there.
However the development build has changed (screenshot attached below).
I believe this is because `redux-immutable-state-invariant` was not being instantiated before. I might be missing context here, but let me know if the old behaviour was correct and can fix the logic.

Screen Shot 2019-12-13 at 17 11 10

### DIFF
--- a/tsdx.config.js
+++ b/tsdx.config.js
@@ -15,19 +15,26 @@ module.exports = {
       case 'umd':
         delete config.external
         config.output.indent = false
-        config.plugins.push(
+        config.plugins.unshift(
           replace({
-            '// UMD-DEV-ONLY: ': ''
+            '// UMD-ONLY: ': '',
+            delimiters: ['', '']
           })
         )
-        config.plugins.push(
+        config.plugins.unshift(
           stripCode({
-            // Remove the `require()` import of RISI so we use the import statement
+            // Remove the `import` of RISI so we use the dynamic `require()` statement
             start_comment: 'START_REMOVE_UMD',
             end_comment: 'STOP_REMOVE_UMD'
           })
         )
         if (env === 'production') {
+          config.plugins.unshift(
+            stripCode({
+              start_comment: 'PROD_START_REMOVE_UMD',
+              end_comment: 'PROD_STOP_REMOVE_UMD'
+            })
+          )
           config.output.file = join(__dirname, pkg.unpkg)
         } else {
           config.output.file = config.output.file.replace(
--- a/src/getDefaultMiddleware.ts
+++ b/src/getDefaultMiddleware.ts
@@ -1,6 +1,8 @@
 import { Middleware } from 'redux'
 import thunkMiddleware from 'redux-thunk'
-// UMD-DEV-ONLY: import createImmutableStateInvariantMiddleware from 'redux-immutable-state-invariant'
+/* START_REMOVE_UMD */
+import createImmutableStateInvariantMiddleware from 'redux-immutable-state-invariant'
+/* STOP_REMOVE_UMD */
 
 import {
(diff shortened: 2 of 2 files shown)


tsdx.config.js
```diff
@@ -15,19 +15,26 @@ module.exports = {
       case 'umd':
         delete config.external
         config.output.indent = false
-        config.plugins.push(
+        config.plugins.unshift(
```
alex-ketch: The plugins were being called after TS compilation, and weren't behaving as desired. I've moved them to the front of the plugin pipeline so that the compiled code takes the modifications into account.

tsdx.config.js
```diff
@@ -15,19 +15,26 @@ module.exports = {
...
           replace({
-            '// UMD-DEV-ONLY: ': ''
+            '// UMD-ONLY: ': '',
+            delimiters: ['', '']
```
alex-ketch: This replacement was never matching due to it being constrained to word boundaries as [explained here](https://github.com/rollup/plugins/tree/master/packages/replace#word-boundaries).

tsdx.config.js
```diff
@@ -15,19 +15,26 @@ module.exports = {
...
+            // Remove the `import` of RISI so we use the dynamic `require()` statement
             start_comment: 'START_REMOVE_UMD',
             end_comment: 'STOP_REMOVE_UMD'
           })
         )
         if (env === 'production') {
+          config.plugins.unshift(
+            stripCode({
+              start_comment: 'PROD_START_REMOVE_UMD',
+              end_comment: 'PROD_STOP_REMOVE_UMD'
```
alex-ketch: This `replace` instance allows for instantiation of `redux-immutable-state-invariant` in `development` builds only

markerikson:
Huh. Yeah, that actually looks like the UMD Dev build is currently broken.

Thanks for fixing this!

phryneas:
As reported in #297, the `PROD_START_REMOVE_UMD` annotation does not seem to have the intended effect. @alex-ketch, would you mind to take a second look? :)

PR #284, opened 2019-12-18
Use defined INCREMENT const instead 'INCREMENT' string

yuri-sakharov:


markerikson:
Hi.  Appreciate the PR, but this isn't actually what I'm trying to show there.  That snippet is intended to show that usually you _wouldn't_ have a separate `const INCREMENT = 'INCREMENT'` variable when using `createAction()`.  Thanks anyway!

PR #285, opened 2019-12-18
Fix createReducer Link in Docs

MarceloAlves:
Noticed that the link at the bottom of the Usage with Typescript page wouldn't go anywhere. The anchor should be all lowercase


markerikson:
Thanks!

Issue #286, opened 2019-12-20
Creating Generic Slices


MarceloAlves:
Stemming from this [thread](https://twitter.com/celo_fett/status/1207470181225598977)

For better or worse a lot of our state has the same structure and so I'm trying to create helper where we can pass a `name`, the `initialState` and end up with a slice that contains most of everything we need.

I'm running into some issues with how to type it correctly. I'm pretty green to more advanced types so this could all be me not understanding how to type it. An example of what I'm trying to achieve:

```typescript
export interface StateSlice<T = any> {
  data: T
  isFetching: boolean
  receivedAt: number | null
  error: boolean
  errorMessage: string | null
  errorCode: number | null
}

const requestStart: CaseReducer<StateSlice, PayloadAction> = (state: StateSlice) => {
  state.isFetching = true
}

const requestSuccess: CaseReducer<StateSlice, PayloadAction> = (state, { payload }) => {
  state.isFetching = false
  state.receivedAt = Date.now()
  state.error = false
  state.errorMessage = null
  state.errorCode = null
  state.data = payload
}

export const createSlice = ({
  name = '',
  initialState,
  reducers,
  extraReducers = {}
}: {
  name: string
  initialState: StateSlice
  reducers?: Record<string, CaseReducer<StateSlice, AnyAction>> // I've tried many different types here. Not final
}): Slice<StateSlice, AnyAction> => {
  return createSliceToolkit({
    name,
    initialState,
    reducers: {
      requestSuccess,
      requestStart,
      ...reducers
    }
  })
}
```

The issues I end up with:

1. When using this new `createSlice` and trying to pass it more reducers:

    ```typescript
    const slice = createSlice({
      name: 'generic/name',
      initialState,
      reducers: {
        updateSuccess: (state: StateSlice<Item[]>, { payload }: PayloadAction<Item>) => {}
      }
    })
    ```

    This error happens:

    >Type '{ updateSuccess: (state: StateSlice<Item[]>, { payload }: WithPayload<Item, Action<string>>) => void; }' is not assignable to type 'Record<string, CaseReducer<StateSlice<any>, AnyAction>>'.
    >Property 'updateSuccess' is incompatible with index signature.
    >Type '(state: StateSlice<Item[]>, { payload }: WithPayload<Item, Action<string>>) => void' is not assignable to type 'CaseReducer<StateSlice<any>, AnyAction>'.
    >  Types of parameters '__1' and 'action' are incompatible.
[...]

Issue #276, opened 2019-12-10
Export Types


wuifdesign:
Would it be possible to export the types used in this project. i want to extend some functionallity and need to copy&paste many of them to my code, so i can access them.

i.e: `CaseReducerActions`, `RestrictCaseReducerDefinitionsToMatchReducerAndPrepare`, ...

phryneas:
Hi @wuifdesign, some of these types are deliberately not exported, as they are really just useful internally and we might replace them with simpler types as we find better solutions.
As soon as we export an internal type, essentially we have to support it forever and can never change it to not break things for people. 
Especially `RestrictCaseReducerDefinitionsToMatchReducerAndPrepare` is a type that I am expecting to remove soon using the [technique displayed in this playground](https://www.typescriptlang.org/play/#code/C4TwDgpgBAwg9gOwM7AE4EMCWDgB4AqUEAHsBAgCZJQDeUA2gNYQgBcUKq2A5gLrsAKAHQj0qbknboEIerwCUUALwA+KNJBQAvmqW0AsACgoDANJRsUZiDgAzKPn5QB6AMbBMidgHkAtpjwABTF0XwgyVCQCelNeFXoABl4AGigAIlBINLUAMlooTIh2cy1FVSgANzhMCiMtAG4jI1dEFCgAC19fZSgCIlJyKlhWtCwcAhUVAWB2fDK1YEbDZpGoYh7O3wEaIxMyFBd3TwR2OkL2AHJ94AvUsHQQABs4dAp2AFcERgQ4AHcEbSKOgAegAVCIhKDgdpknV5E1DC1kMAoJo9JttrsChADm4PF58ucoBcfsB8DibncHs9Xh8vj9-oD8mCIVCYXCEUA).

Exporting `CaseReducerActions` would also require us to export `SliceCaseReducerDefinitions`, which relies on `PayloadActions` and some of these might also become obsolete using the same technique.

Could you please explain with a few code snippets what you are trying to do?
Maybe I can help you around your problems with only our exported types - and maybe that'll even lead to something we can add to our TypeScript documentation afterwards :)

wuifdesign:
i wont to extend the createSlice a bit. something like that (only as example, i want to add thunks or sagas or something else):

```
export function createAdvancedSlice<State, CaseReducers extends SliceCaseReducerDefinitions<State, any>>(
[...]

PR #290, opened 2019-12-21
exportable case reducers type, bugfix, documentation

phryneas:
This is what I'm currently working on and it has already gotten a bit bigger than expected :smile: 

INCLUDED:
- [x] change signatue of `createSlice` and add new types (exported) to be used in situations like #276 and #286 :
  - `SliceCaseReducerDefinitions`
  - `CaseReducerActions`
  - `CaseReducerWithPrepare` (not really necessary to export this, but `CaseReducer` was already exported so it might not hurt to have this)
- [x] fix a bug where `PayloadActionCreator<any>` becomes `ActionCreatorWithoutPayload` which was a regression introduced in #273 (also opened https://github.com/joonhocho/tsdef/pull/4 )
- [x] type `_ActionCreatorWithPreparedPayload` is now marked `@internal` (was `ActionCreatorWithPreparedPayload` before #273, since we renamed it anyway might as well hide it completely)
- [x] `@internal` types will be stripped from the `dist/typings.d.ts` file
- [x] document ALL types
- [x] add release tags (@public or @internal) to all types
- [x] add a type test that successfully wraps `createSlice`
- [x] ~~api-extractor trims too agressively (complete removal instead of "not export") - I'll have to see if that works out. (see https://github.com/microsoft/rushstack/issues/1664 )~~ (solved by manually exporting)
- [x] fix TS 3.3
- [x] add "Wrapping createSlice" chapter to the TS docs 
- [x] add "Defining the type of your initialState"  chapter to the TS docs

### DIFF
--- a/src/createReducer.ts
+++ b/src/createReducer.ts
@@ -7,6 +7,11 @@ import {
 
 /**
  * Defines a mapping from action types to corresponding action object shapes.
+ *
+ * @deprecated This should not be used manually - it is only used for internal
+ *             interference purposes and should not have any further value.
+ *             It might be removed in the future.
+ * @public
  */
 export type Actions<T extends keyof any = string> = Record<T, Action>
 
@@ -23,6 +28,8 @@ export type Actions<T extends keyof any = string> = Record<T, Action>
  * cause the store state to be mutated directly; instead, thanks to
  * [immer](https://github.com/mweststrate/immer), the mutations are
  * translated to copy operations that result in a new state.
+ *
+ * @public
  */
 export type CaseReducer<S = any, A extends Action = AnyAction> = (
   state: Draft<S>,
@@ -31,6 +38,12 @@ export type CaseReducer<S = any, A extends Action = AnyAction> = (
 
 /**
  * A mapping from action types to case reducers for `createReducer()`.
+ *
+ * @deprecated This should not be used manually - it is only used
+ *             for internal interference purposes and using it manually
+ *             would lead to type erasure.
+ *             It might be removed in the future.
+ * @public
  */
 export type CaseReducers<S, AS extends Actions> = {
   [T in keyof AS]: AS[T] extends Action ? CaseReducer<S, AS[T]> : void
@@ -51,6 +64,8 @@ export type CaseReducers<S, AS extends Actions> = {
  * @param initialState The initial state to be returned by the reducer.
  * @param actionsMap A mapping from action types to action-type-specific
  *   case reducers.
+ *
+ * @public
  */
 export function createReducer<
   S,
@@ -70,6 +85,8 @@ export function createReducer<
  * @param initialState The initial state to be returned by the reducer.
  * @param builderCallback A callback that receives a *builder* object to define
  *   case reducers via calls to `builder.addCase(actionCreatorOrType, reducer)`.
+ *
+ * @public
  */
 export function createReducer<S>(
   initialState: S,
--- a/src/createReducer.ts
+++ b/src/createReducer.ts
@@ -9,7 +9,7 @@ import {
  * Defines a mapping from action types to corresponding action object shapes.
  *
  * @deprecated This should not be used manually - it is only used for internal
- *             interference purposes and should not have any further value.
+ *             inference purposes and should not have any further value.
  *             It might be removed in the future.
  * @public
  */
@@ -40,7 +40,7 @@ export type CaseReducer<S = any, A extends Action = AnyAction> = (
  * A mapping from action types to case reducers for `createReducer()`.
  *
  * @deprecated This should not be used manually - it is only used
- *             for internal interference purposes and using it manually
+ *             for internal inference purposes and using it manually
  *             would lead to type erasure.
  *             It might be removed in the future.
  * @public
--- a/src/createSlice.ts
+++ b/src/createSlice.ts
@@ -103,8 +103,6 @@ export type CaseReducerWithPrepare<State, Action extends PayloadAction> = {
 
 /**
  * The type describing a slice's `reducers` option.
- * Also checks itself, so it has to be passed "itself" as it's second option.
- * See the method signature of `createSlice`.
  *
  * @public
  */
@@ -172,9 +170,11 @@ type SliceDefinedCaseReducers<CaseReducers extends SliceCaseReducers<any>> = {
 type NoInfer<T> = [T][T extends any ? 0 : never]
 
 /**
- * Used on a `reducers` object.
+ * Used on a SliceCaseReducers object.
  * Ensures that if a CaseReducer is a `CaseReducerWithPrepare`, that
  * the `reducer` and the `prepare` function use the same type of `payload`.
+ *
+ * Might do additional such checks in the future.
  *
  * This type is only ever useful if you want to write your own wrapper around
  * `createSlice`. Please don't use it otherwise!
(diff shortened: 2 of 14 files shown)


src/createSlice.ts
```diff
@@ -84,84 +94,97 @@ export interface CreateSliceOptions<
...
 }
 
-type SliceCaseReducerDefinitions<State, PA extends PayloadActions> = {
-  [ActionType in keyof PA]:
-    | CaseReducer<State, PA[ActionType]>
-    | CaseReducerWithPrepare<State, PA[ActionType]>
-}
+/**
+ * The type describing a slice's `reducers` option.
+ * Also checks itself, so it has to be passed "itself" as it's second option.
```
markerikson: wat :)

markerikson:
Your ability to crank out TS improvements continues to amaze me :)

No specific feedback from me on the changes, but as usual I still don't know quite enough to provide much feedback here anyway.  

Do you want to try to get the other type doc changes in here too, or save those for later?

src/createSlice.ts
phryneas: Yeah, I wouldn't have found out about that one myself as it is completely counter-intuitive. But a generic can `extend` (which essentially just means "match this shape") something that *depends on the generic being matched*.

So you can write
```ts
function something<X extends Restrict<X>>(x: X) {}
```

and then `Restrict` can be a type that does some crazy restrictions on the original input object while referencing it. 

This is something I learned from @jcalz over on StackOverflow: https://stackoverflow.com/questions/59200095/is-there-a-simpler-way-of-associating-a-key-to-a-value-in-a-mapped-obect?noredirect=1#comment104621839_59200095
Thank you again Joe!

phryneas:
> Your ability to crank out TS improvements continues to amaze me :)

This project is a wonderful learning experience, but it kills time like nothing else. I think i've started this Refactor over four or five times, each time putting in like 3-5 hours and then throwing everything away because it just wasn't working out.
So in the end, it just comes down to trial, error and my determination to bash my head against it until I find out how it works ;)

> Do you want to try to get the other type doc changes in here too, or save those for later?

I'm planning on doing everything in this PR, now that I already mixed different things in here. 
We'll probably have to wait for https://github.com/microsoft/rushstack/issues/1664 anyways and I need to find out why this doesn't work with TS3.3 again :woman_shrugging:

phryneas:
So it seems that relying on `@internal` won't be a possibility for us to exclude certain types from the dts-rollup (although it has a nice documenting effect when reading the source :smile: ). Which means I'll have to replace the `export * from ...` in `index.ts` with explicit export statements.
@markerikson would that be okay with you? I believe that might even help bundlers to do tree-shaking, although that's only a gut feeling.

The alternative would be to move the type to a different file where nothing is exported from, but right now I'd say that the type is in the file it belongs to.

markerikson:
Yeah, if that's necessary to keep internal types from leaking, we can do that.

phryneas:
Okay, that's it for today. All that's left is fixing TS3.3 - I hope that's gonna work out, but I'll leave that problem for "future me" :smile:

phryneas:
So, to make this work with TS <3.4, unfortunately I had to move `ValidateSliceCaseReducers` out of `SliceCaseReducers` (I renamed that one from `SliceCaseReducerDefintions`) again, because the `generic that extends itself to restrict itself` tactic seems not to be working in older TS versions.
This means, we have one more exported type: `ValidateSliceCaseReducers`.

But all in all, given that `SliceCaseReducers` is now a normal type and we aren't going over that "anoymous `PayloadActions` type" that was in there that was absolutely confusing, I think this whole refactor still has some good value and we can actually expose those types in good consciousness.

type-tests/files/createSlice.typetest.ts
```diff
@@ -312,3 +314,57 @@ function expectType<T>(t: T) {
...
+          state.status = 'loading'
+        },
+        success(state: GenericState<T>, action: PayloadAction<T>) {
+          state.data = action.payload
+          state.status = 'finished'
+        },
+        ...reducers
+      }
+    })
+  }
```
phryneas: This is what wrapping `createSlice` would look like now.

phryneas:
Oh, seriously? Now TS is acting up, only in 3.4 :roll_eyes: 

`Draft<GenericState<T>>` is not compatible with `GenericState<T>`. Suure.

phryneas:
Yay, we're all green, without dirty hacks. 

I was almost rage-quitting there o_O

src/createSlice.ts
markerikson: Still waiting for that blog post / talk on "all the bizarre stuff I learned working on RTK's TS types" :)

src/createReducer.ts
```diff
@@ -7,6 +7,11 @@ import {
 
 /**
  * Defines a mapping from action types to corresponding action object shapes.
+ *
+ * @deprecated This should not be used manually - it is only used for internal
+ *             interference purposes and should not have any further value.
```
markerikson: Should this be `inference` instead of `interference`?

src/createReducer.ts
```diff
@@ -31,6 +38,12 @@ export type CaseReducer<S = any, A extends Action = AnyAction> = (
 
 /**
  * A mapping from action types to case reducers for `createReducer()`.
+ *
+ * @deprecated This should not be used manually - it is only used
+ *             for internal interference purposes and using it manually
```
markerikson: Ditto.

src/createSlice.ts
```diff
@@ -42,18 +47,20 @@ export interface Slice<
    */
   actions: CaseReducerActions<CaseReducers>
 
-  caseReducers: SliceDefinedCaseReducers<CaseReducers, State>
+  /**
+   * The reducers defined by `reducers` for easy access if they were defined inline when calling createSlice.
```
markerikson:
```suggestion
   * The individual case reducer functions that were passed in the `reducers` parameter.  
   * This enables reuse and testing if they were defined inline when calling `createSlice`.
```

type-tests/files/createSlice.typetest.ts
```diff
@@ -312,3 +314,57 @@ function expectType<T>(t: T) {
     }
   })
 }
+
+/** Test: wrapping createSlice should be possible */
```
markerikson: Nice to see that A) this is possible, and B) we have a test for it.

markerikson: Also, can we add this example to the "Usage with TS" docs page?

markerikson:
No immediate complaints about the TS stuff, but my eyes also mostly glazed over, so take it as you will :)

Saw a couple possible comment typos, suggested a rewording, and it'd be nice to see the "generic slice" example in the docs, but other than that looks okay.

src/createSlice.ts
phryneas: Yup, got something in the making :) Meetup somewhere mid-february, maybe I'll also submit it to a few conferences. Also, some ideas to experiment with twitter-content :)

src/createReducer.ts
phryneas: Yup. Thanks for catching!

type-tests/files/createSlice.typetest.ts
phryneas: Honestly, without these tests I wouldn't have the confidence to change on single letter in the typings. Those are so invaluable!

phryneas:
Added a "Wrapping `createSlice`" chapter to the TS docs. While I was at it, I also added a "Defining the type of your `initialState`" chapter, as that was something of importance we had not mentioned before.

markerikson:
Awright, let's get this in.

PR #302, opened 2020-01-02
Fix another typo in the docs

jdreesen:


markerikson:
Good grief, apparently I can't type.

Thanks!

Issue #289, opened 2019-12-21
`configureStore()` middleware action typings are wrong


slikts:
Currently the `middleware` option for `configureStore()` does not affect what action types `store.dispatch()` accepts; thunks are always accepted even if the thunk middleware is missing, and adding other middleware ignores its dispatch signature.

This is a regression from Redux, where the dispatchable action types are picked up correctly by `createStore()`.

Example:
```ts
const store = configureStore({
  reducer: rootReducer,
  // default thunk middleware disabled
  middleware: []
});
// expected compiler error missing
store.dispatch(() => {});
```

I've also made a [longer working example][example] that defines a number middleware that adds a number action type and shows that the types work with plain Redux but not RTK. I've tried to work around the issue by passing explicit type parameters like so: `configureStore<RootState, AnyAction | number>`, but it doesn't work due to always expecting a type property on the action.

[example]: https://codesandbox.io/s/redux-store-dispatch-overload-u8kcp

phryneas:
Fascinating, I always assumed that `redux` did not do that (as it is impossible in TS to type this for an arbitrary number of middlewared`, but apparently the current types actually support the dispatch type of up to 5 middlewares.

Thank you for bringing that to my attention. @markerikson I guess we should add support for this as well then, shouldn't we?

In the meantime, the way to currently work around this is documented in the "Usage with TypeScript" documentation: [Extending the Dispatch type](https://redux-toolkit.js.org/usage/usage-with-typescript#extending-the-dispatch-type)

slikts:
Thanks for the workaround, although it'd be slightly nicer if it were possible to use the type parameters instead.

markerikson:
Yeah, definitely be nice if we can fix this.

phryneas:
@slikts I just finished my work on the PR, so this will most likely make it into the next version of RTK.
If you want to already use it, you can use the [codesandbox build for the PR](https://ci.codesandbox.io/status/reduxjs/redux-toolkit/pr/304/builds/7607). Early feedback is highly welcome :)

markerikson:
Should be fixed in https://github.com/reduxjs/redux-toolkit/releases/tag/v1.2.2 .

PR #304, opened 2020-01-03
correctly infer dispatch from provided middlewares

phryneas:
So this is my first attempt at solving #289.

Still a bit rough around the edges, but that's it for today.

- [x] feature itself is working
- [x] type documentation
- [x] skip tests using `as const` in TS 3.3
- [x] update typescript usage documentation


src/getDefaultMiddleware.ts
```diff
@@ -28,6 +29,24 @@ interface GetDefaultMiddlewareOptions {
   serializableCheck?: boolean | SerializableStateInvariantMiddlewareOptions
 }
 
+export type DefaultMiddlewareResult<
+  S,
+  O extends GetDefaultMiddlewareOptions
+> = Unshift<
```
markerikson: *whimper* :)

phryneas: Yeah, wanna REALLY hurt? Take a look at the implementation of Unshift ;)

Here, I had the choice between
* make `getDefaultMiddleware` always return `[ThunkMiddleware]`, which is untrue
* create a type with one case for every possible combination of `thunk`, `immutableCheck` and `serializableCheck` (I had this in for a while)
* this one: create the tuple of length 0 to 3, and one check per possible configuration option. The only way to make a tuple longer in TS is currently to **prepend** values, to I'm starting with the empty tuple `[]` and prepending the three middlewares depending on their configuration value from the innermost to the outermost `Unshift`. At least it doesn't repeat itself too much :man_shrugging:

markerikson: So now what happens if we add another entry to `getDefaultMiddleware`?

phryneas: This way: another level of `Unshift`, essentially two more lines. I might still extract the types in there out, that will make it more readable.

The way I had it before? Go from `3!` cases to `4!` cases. 

Problem is: now that the `Dispatch` type relies on the value of `middleware`, `getDefaultMiddleware` needs to return it's value accurately.

Although I can think of a fourth way right now: instead of the tuple, it could also just return an array. That might be simpler to type - it was just not possible with an earlier implementation of `DispatchForMiddlewares`. I'll investigate that tomorrow.

phryneas: Okay, it worked out - that's much simpler now.

phryneas:
Okay, documentation is updated, too. Could you please give this a look @markerikson ? :)

markerikson:
Awright, putting this in.

PR #310, opened 2020-01-09
Match code to tutorial description / tests

vthommeret:
For `toggleTodo` reducer, the tutorial says:

> For `toggleTodo`, the only value we need is the `id` of the todo being changed. We could have made that the `payload`, but I prefer always having `payload` be an object, so I made it `action.payload.id` instead.

But the code appears to just check the payload directly:

`const todo = state.find(todo => todo.id === action.payload)`

The tests also pass in the payload as an object, so they don't pass with the code in the tutorial:

```
{
  type: toggleTodo.type,
  payload: {
    id: 1
  }
}
```

### DIFF
--- a/docs/tutorials/intermediate-tutorial.md
+++ b/docs/tutorials/intermediate-tutorial.md
@@ -173,7 +173,7 @@ const todosSlice = createSlice({
       state.push({ id, text, completed: false })
     },
     toggleTodo(state, action) {
-      const todo = state.find(todo => todo.id === action.payload)
+      const todo = state.find(todo => todo.id === action.payload.id)
       if (todo) {
         todo.completed = !todo.completed
       }


markerikson:
Hmm.  The current code in the actual example repo just checks `action.payload`.  It's possible I mis-matched things somewhere along the way:

https://github.com/reduxjs/rtk-convert-todos-example/blob/138cc162b1cc9c64ab67fae0a1171d07940414e6/src/features/todos/todosSlice.js#L19

I'll have to look at this in more detail.

vthommeret:
Ah yeah, and it does look like the tests pass just the ID as the payload (I was pattern matching based on the example for `addTodo`) but looks like the payload guidance is out-of-sync with the actual code.

markerikson:
Obsolete due to #905 .

PR #317, opened 2020-01-15
fix: UMD dev build by removing runtime usage of require statement

### LINKED ISSUE #297: Counter Vanilla example broken
The counter vanilla example does not work when viewed on codesandbox or when downloaded locally (macos, tried Chrome/Firefox/Safari). Page loads, but an error message about 'require' being undefined is thrown, buttons don't update or change the state in any way.

Locally, I managed to fix the issue by changing:

```
<script src="https://unpkg.com/@reduxjs/toolkit@latest/dist/redux-toolkit.umd.js"></script>
```

to 

```
<script src="https://unpkg.com/@reduxjs/toolkit@latest/dist/redux-toolkit.umd.min.js"></script>
```

Not sure if the codesandbox demo is supposed to be interactive, but I feel like it should be. The same fix didn't work there though.

Hopefully this is all clear, let me know if any further info is needed.

alex-ketch:
closes #297 

Thanks for your patience on this @phryneas.
I've simplified the build a little. Now it just uses the `PROD_START_REMOVE_UMD` comments throughout, and avoids a dynamic `require` statement.
The dev (unminified) UMD bundle is a few KB bigger, but the production one hasn't changed.

### DIFF
--- a/src/getDefaultMiddleware.ts
+++ b/src/getDefaultMiddleware.ts
@@ -1,8 +1,8 @@
 import { Middleware } from 'redux'
 import thunkMiddleware from 'redux-thunk'
-/* START_REMOVE_UMD */
+/* PROD_START_REMOVE_UMD */
 import createImmutableStateInvariantMiddleware from 'redux-immutable-state-invariant'
-/* STOP_REMOVE_UMD */
+/* PROD_STOP_REMOVE_UMD */
 
 import {
   createSerializableStateInvariantMiddleware,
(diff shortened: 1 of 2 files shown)


markerikson:
Yeah, diffing 1.2.1's UMD build vs the CSB build of this PR shows that we've removed the `require()` line, and inlined all of `invariant()` and `redux-immutable-state-invariant`, as desired.

Thanks!  I'll try to put out a new build this evening.

PR #320, opened 2020-01-17
Add ignoredPaths option to ignore serializability check

### LINKED ISSUE #319: Add option to skip serializability check for state slices
The `redux-immutable-state-invariant` middleware has an option for skipping checks for selected slices of state:  https://github.com/leoasis/redux-immutable-state-invariant#api .  However, our homegrown `serializable-state-invariant` middleware only allows determining if a given value is serializable, and skipping specific actions, not skipping slices of state.

Per https://stackoverflow.com/questions/59773345/react-toolkit-and-redux-first-router , it looks like Redux-First-Router in particular tries to store thunks directly in the store. It's a bad idea, but seems like we should at least update the serializable middleware to accept a list of state slices to skip.

Since our serializable middleware was directly inspired by the immutable middleware to begin with, we should be able to look at how the immutable middleware implements this, and reuse the same approach.

kevin940726:
Fix #319

Add a new option to `createSerializableStateInvariantMiddleware` called `ignoredPaths`. It accepts an array of dot-separated path strings to be ignored when checking for serializability in state.

```js
const serializableStateInvariantMiddleware = createSerializableStateInvariantMiddleware(
  {
    ignoredPaths: ["testSlice.a", "testSlice.b.c"]
  }
);
```

### DIFF
--- a/src/serializableStateInvariantMiddleware.ts
+++ b/src/serializableStateInvariantMiddleware.ts
@@ -35,13 +35,9 @@ export function findNonSerializableValue(
   path: ReadonlyArray<string> = [],
   isSerializable: (value: unknown) => boolean = isPlain,
   getEntries?: (value: unknown) => [string, any][],
-  ignoredSlices: string[] = []
+  ignoredPaths: string[] = []
 ): NonSerializableValue | false {
   let foundNestedSerializable: NonSerializableValue | false
-
-  if (ignoredSlices.includes(path[0])) {
-    return false
-  }
 
   if (!isSerializable(value)) {
     return {
@@ -59,6 +55,10 @@ export function findNonSerializableValue(
   for (const [property, nestedValue] of entries) {
     const nestedPath = path.concat(property)
 
+    if (ignoredPaths.indexOf(nestedPath.join('.')) >= 0) {
+      continue
+    }
+
     if (!isSerializable(nestedValue)) {
       return {
         keyPath: nestedPath.join('.'),
@@ -72,7 +72,7 @@ export function findNonSerializableValue(
         nestedPath,
         isSerializable,
         getEntries,
-        ignoredSlices
+        ignoredPaths
       )
 
       if (foundNestedSerializable) {
@@ -109,9 +109,9 @@ export interface SerializableStateInvariantMiddlewareOptions {
   ignoredActions?: string[]
 
   /**
-   * An array of slice names to ignore when  checking for serializability, Defaults to []
+   * An array of dot-separated path strings to ignore when checking for serializability, Defaults to []
    */
-  ignoredSlices?: string[]
+  ignoredPaths?: string[]
 }
 
 /**
@@ -130,7 +130,7 @@ export function createSerializableStateInvariantMiddleware(
     isSerializable = isPlain,
     getEntries,
     ignoredActions = [],
-    ignoredSlices = []
+    ignoredPaths = []
   } = options
 
   return storeAPI => next => action => {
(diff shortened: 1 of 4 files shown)


markerikson:
Can we try doing `ignoredPaths` instead?

kevin940726:
@markerikson Sure! But I'm wondering what's the use case for making just part of the slices ignored for checking serializability? From my understanding, it should mainly comes from an uncontrolled third-party library which is also rare. I'm open to both solutions though, I'll make the change later :)

phryneas:
`createSlice` slices do not necessarily have to correspond to a key with their name in the rootReducer, which this seems to be assuming if I read it right - so it introduces a second interpretation of `slice` that is distinct from `createSlice`.

markerikson:
I'm mostly just looking for a granularity thing here.

kevin940726:
Updated! Thx @markerikson @phryneas ❤️

Do we want to also support passing array of array here? even though dot-separated path is quite common.

```js
const serializableStateInvariantMiddleware = createSerializableStateInvariantMiddleware(
  {
    ignoredPaths: [["testSlice", "a"], ["testSlice", "b", "c"]]
  }
);
```

src/serializableStateInvariantMiddleware.ts
```diff
@@ -59,6 +55,10 @@ export function findNonSerializableValue(
   for (const [property, nestedValue] of entries) {
     const nestedPath = path.concat(property)
 
+    if (ignoredPaths.indexOf(nestedPath.join('.')) >= 0) {
```
markerikson: As a tiny perf improvement, can we first check to see if `ignoredPaths` has a length > 0, outside the loop, and then check this as `if (hasIgnoredPaths && ignoredPaths.indexOf()` ?

phryneas:
> Updated! Thx @markerikson @phryneas heart
> 
> Do we want to also support passing array of array here? even though dot-separated path is quite common.
> 
> ```js
> const serializableStateInvariantMiddleware = createSerializableStateInvariantMiddleware(
>   {
>     ignoredPaths: [["testSlice", "a"], ["testSlice", "b", "c"]]
>   }
> );
> ```

Hm. I believe once we allow for array notation there, people are gonna assume that we have TS support for that (which I wouldn't want to add here as it's too much complication for a niche use case), so I'd stick with dot notation only.

markerikson:
Great, thanks!

jakeboone02:
Is anyone working on updating the [documentation](https://redux-toolkit.js.org/api/getdefaultmiddleware/#api-reference) for this?  I may have just volunteered...

markerikson:
Hmm.  Yeah, I missed that the doc update was only for the "other exports" page, and we should mention something about this on the `getDefaultMiddleware` page too.

kevin940726:
Oops, I didn't know that there is documentation for this in that page. Maybe a link point to the `other exports" page would be fine?

markerikson:
Mmm... let's at least update the type signature in the `getDefaultMiddleware` page, and then update the "API Reference" section to also point to the relevant API options links for `redux-immutable-state-invariant` and `redux-thunk` in their repos, and the serializable middleware (the "Other Exports" page).

kevin940726:
Sounds good to me. Currently `redux-immutable-state-invariant` and `redux-thunk` already have links pointing to their repos, but are not clear at all. Should we fix that too?

image

Live shown above, `redux-immutable-state-invariant` is actually clickable but `serializable-state-invariant-middleware` is not

markerikson:
Yeah, I know I made some CSS tweaks to the Redux core repo to make sure links in code snippets are actually colored.  Probably ought to pull those over.

kevin940726:
PR opened: https://github.com/reduxjs/redux-toolkit/pull/327

Issue #324, opened 2020-01-20
Describe inferring RootState type from store.getState


markerikson:
The docs show inferring `type RootState = ReturnType<typeof rootReducer>`.  But, if you pass the slice reducers directly to `configureStore()`, you never created the root reducer yourself, so you need to infer it as `type RootState = ReturnType<typeof store.getState>` instead.

We should update the Advanced Tutorial and Usage with TypeScript pages to mention that.

firstred:
That works indeed! I just arrived at that section of the documentation this morning. Here are my results with TypeScript 3.7.5:

It fails with `combineReducers()` or passing them directly to `configureStore()`. The type of `RootState` would always be `unknown`.
The only thing that works is, indeed, using `type RootState = ReturnType<typeof store.getState>` and passing the reducers directly to `configureStore()`. `combineReducers()` would still turn the type into `unknown`.

Based on my experience, I would recommend to leave out `combineReducers()` in the TypeScript documentation until a solution for `combineReducers()` has been found. Until then passing the reducers directly to `configureStore()` and using `ReturnType<typeof store.getState>` would be a nice alternative.

markerikson:
@firstred : ah.... I'm very confused.  Inferring the return type of `combineReducers _does_ work, as I showed in the Advanced Tutorial.  See this sandbox:

https://codesandbox.io/s/rtk-github-issues-example-03-final-ihttc?fontsize=14&hidenavigation=1&module=%2Fsrc%2Fapp%2FrootReducer.ts&theme=dark

If I hover over `export type RootState = ReturnType<typeof rootReducer>`, I see:

```ts
const rootReducer: Reducer<CombinedState<{
    issuesDisplay: CurrentDisplayState;
    repoDetails: RepoDetailsState;
    issues: IssuesState;
    comments: CommentsState;
}>, AnyAction>
```

Where and how are you seeing problems with this?

firstred:
PhpStorm 2019.3.1 + TypeScript 3.7.5. This is what I am seeing:

`typeof rootReducer` + `combineReducers()`:
![rootreducer](https://user-images.githubusercontent.com/6775736/72829182-93b6d480-3c7e-11ea-925b-77d9a57d705f.png)

`typeof store.getState` + `combineReducers()`:
![getstate](https://user-images.githubusercontent.com/6775736/72829191-974a5b80-3c7e-11ea-8cf8-f40f72c59288.png)

`typeof rootReducer`:
[...]

PR #326, opened 2020-01-21
docs: root state type with TypeScript and configureStore

firstred:
When passing the root reducers directly to `configureStore()`, it is not possible to get the root state type directly from the reducers. For the type, you will then have to access `store.getState()` instead.

So, instead of `type RootState = ReturnType<typeof rootReducer>` you will have to use `type RootState = ReturnType<typeof store.getState>`. This PR adds this little piece of information.


markerikson:
The info is just a bit off.  The problem comes when you don't define the root reducer explicitly, and instead pass the _slice reducers_ to `configureStore()`, like:

```ts
    reducer: {
        first: firstReducer,
        second: secondReducer,
    }

});
```

There's no `rootReducer` to reference, so we have to refer to `store.getState` instead.

Can you tweak the wording and example to match that?

firstred:
I am going to familiarize myself a bit more with the material before I continue with this PR. I thought I knew a lot about the Redux Starter Kit / Toolkit already, but today showed that it still had a few surprises for me.
Thank you for being so patient and helpful. It's much appreciated!!

markerikson:
Sure, and thanks for putting the PR together!

firstred:
I just updated the PR, taking this feedback into account:
- No `rootReducer` to refer: https://github.com/reduxjs/redux-toolkit/issues/324#issuecomment-583297569
- Possible TS circular reference when adding middleware: https://github.com/reduxjs/redux-toolkit/pull/326#issuecomment-576936660

docs/usage/usage-with-typescript.md
```diff
@@ -28,6 +28,35 @@ const rootReducer = combineReducers({})
...
+  reducer: rootReducer,
+  middleware: [
+    ...customMiddleware,
+    ...getDefaultMiddleware<RootState>(),
+  ] as const,
+})
+// TS2456: Type alias 'RootState' circularly references itself.
+export type RootState = ReturnType<typeof store.getState>
+```
+
```
phryneas: Could you mention here that it is still possible to get `RootState` in that case by using `ReturnType<typeof rootReducer>` (in some cases rootReducer would have to be created by calling `combineReducers(reducerObject)`, but in that case you can't help it).

firstred: That makes sense, thanks! I thought it was clear from the explanation that you could use both, but now that I read it again I agree it's indeed a bit vague.

markerikson:
This PR needs to be updated to take into account the changes from #549 and #559  .

docs/usage/usage-with-typescript.md
```diff
@@ -28,6 +28,35 @@ const rootReducer = combineReducers({})
...
+const store = configureStore({
+  reducer: rootReducer,
+  middleware: [
+    ...customMiddleware,
+    ...getDefaultMiddleware<RootState>(),
+  ] as const,
+})
+// TS2456: Type alias 'RootState' circularly references itself.
+export type RootState = ReturnType<typeof store.getState>
+```
```
phryneas, line 58: As of #549, I believe this section could just be removed altogether and this could get merged.

phryneas:
```suggestion
```

PR #329, opened 2020-01-24
Better generic argument for default Thunk Middleware

### LINKED ISSUE #321: `store.dispatch` does not work in 1.2.2
After migrating to 1.2.2 `store.dispatch` stopped working with the following error message:

```
Argument of type 'ThunkAction<void, CombinedState<{ auth: AuthState; }>, null, Action<string>>' is not assignable to parameter of type 'AnyAction'.
  Property 'type' is missing in type 'ThunkAction<void, CombinedState<{ auth: AuthState; }>, null, Action<string>>' but required in type 'AnyAction'.
```

Dispathing through dispatcher from hook works fine though:

```
  ...
  dispatch(logOut());  // OK
```

I can't use React Hooks in the axios interceptor, so I was using this:

```
import store from '../../store';

store.dispatch(logOut());
```

phryneas:
This should fix #321

### DIFF
--- a/docs/tutorials/advanced-tutorial.md
+++ b/docs/tutorials/advanced-tutorial.md
@@ -541,7 +541,7 @@ Before we go any further, let's add a type declaration we can reuse instead.
 
 export type AppDispatch = typeof store.dispatch
 
-+export type AppThunk = ThunkAction<void, RootState, null, Action<string>>
++export type AppThunk = ThunkAction<void, RootState, unknown, Action<string>>
 ```
 
 The `AppThunk` type declares that the "action" that we're using is specifically a thunk function. The thunk is customized with some additional type parameters:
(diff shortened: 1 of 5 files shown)


markerikson:
As usual, I don't know enough to say if this is actually an improvement or not, so I'll go with your expertise here :)

Two additional things that would potentially need to be updated accordingly:

- The Redux core "Usage with TS" page
- The RTK "Advanced Tutorial" example repo.  This one's more complicated, unfortunately, because the repo and the sandboxes based on a specific commit history.  We'd really need to go rewrite the commit history with this one change.  I've done that a few times already, but it's a pain.

phryneas:
Okay, should now work with `null`, `undefined` and `unknown` in the AppThunk parameter. As for documentation updates, I guess I have covered everything in various PRs.

See https://github.com/reduxjs/redux-toolkit/pull/348 and https://github.com/reduxjs/redux/pull/3693

PR #334, opened 2020-01-27
fix(int-tutorial): adds a link for jsconfig.json

kutyepov:
I saw a reference to `jsconfig.json` for the first time (though I am familiar with tsconfig.json) and it was a bit confusing what reads that config. Adding link to documentation will definitely help all subsequent readers with that same "gap" in knowledge :)


markerikson:
Thanks!

PR #335, opened 2020-01-27
chore(deps): removes 1 moderate and 12735 high vulnerabilities

kutyepov:
I think this could be automated though :)


markerikson:
Not terribly concerned with warnings in build tools, but thanks, appreciate it!

PR #348, opened 2020-02-09
advanced tutorial update links from commit ids to tags

phryneas:
I updated the AppThunk example from `undefined` to `unknown` in the advanced tutorial. This updates all links in the tutorial accordingly, using the tags we talked about.

This should work as soon as you force-push the current commit history from https://github.com/phryneas/rsk-github-issues-example/tree/change-appthunk-from-null-to-unknown over to the upstream repo and apply all the tags there.


markerikson:
Sweet.  Also need to update the sandbox URLs, but I'll do that momentarily

Issue #333, opened 2020-01-25
Consider adding types and logic for handling normalized state


markerikson:
Given that we [recommend normalizing state](https://redux.js.org/style-guide/style-guide#normalize-complex-nestedrelational-state), it would be nice to have at least some amount of built-in tooling to help with that.  This was also on [Dan's wishlist of things to include in a Redux addon layer](https://github.com/reduxjs/redux/issues/2295#issuecomment-287960516):

> - Containing built-in helpers for indexing, normalization, collections, and optimistic updates.

I previously asked on Twitter about the idea of adding something like this:

https://twitter.com/acemarke/status/1216766299688316928

The NgRX platform has a library called `@ngrx/entity` ( https://ngrx.io/guide/entity ), which contains some type definitions and helpers for working with normalized state.  Per discussion with an NgRX dev, [that library is already almost completely framework-agnostic](https://twitter.com/MikeRyanDev/status/1216820265658789888) - it just has a couple NgRX/Angular-specific imports.

I just opened up https://github.com/ngrx/platform/issues/2333 to ask if they might be able to tweak it and make it truly agnostic, in which case we might be able to depend on `@ngrx/entity` ourselves.

Alternately, we could copy-paste the code with attribution.

There's also a bunch of other libraries for dealing with entity management, which I have listed at https://github.com/markerikson/redux-ecosystem-links/blob/master/entity-collection-management.md .

Key considerations:

- Don't want to go down the rabbit hole of making this incredibly complex
- Needs to be tree-shakeable, and I think we still aren't doing that yet ( #78 )
- How does this overlap with ["model actions as events, not setters"](https://redux.js.org/style-guide/style-guide#model-actions-as-events-not-setters)?
- I can picture having a wrapper around `createSlice` to generate entity actions and reducers or something, but we've already seen that can get tricky in TypeScript
- Probably also some overlap with #76 as well for defining fetch logic

markerikson:
https://resthooks.io/docs/guides/redux looks interesting.

markerikson:
Ultimately, what I'm thinking here is:

- Fetching data from an API and putting it in Redux is a common thing to do
- The docs describe some "standard" but complex patterns to work with:
    - async action triplets (start/success/failure)
    - dispatching those actions in a thunk
    - normalizing the state as you load it in
- I'd like to come up with something of a 50-60% solution that minimizes the amount of code you have to write for the most common operations

markerikson:
[...]

Issue #76, opened 2019-01-10
Create Async Action


jamescmartinez:
`redux-starter-kit` includes `redux-thunk` but doesn't include an async version of `createAction`. I want to float the idea of a `createAsyncAction` function that uses thunks; I can create a pull request if all looks good.

Here's a rough initial API proposal:
```js
// To be added:
// (Inspired by the current implementation of `createAction`)
function createAsyncAction(type, thunk) {
  const action = payload => thunk(payload);
  action.toString = () => type;
  return action;
}

// Usage example:
const someAsyncAction = createAsyncAction("SOME_ASYNC_ACTION", payload => {
  return dispatch => {
    return setTimeout(() => {
      // One second later...
      dispatch(someAsyncActionSuccess());
    }, 1000);
  };
});

const someAsyncActionSuccess = createAction("SOME_ASYNC_ACTION_SUCCESS");
```

The only new proposal is at the top of that snippet, the function `createAsyncAction`. It takes two parameters: type (same as `createAction`) and `thunk`.

What are everyone's thoughts?

denisw:
A helper function for thunk actions sounds like a great idea to me! 

I have one thought about the design. In the case of async actions without payload, I could see myself forgetting about specifying the outer function (the one returning the thunk) and directly writing the thunk instead.

```js
// FAIL: I pass a thunk instead of a function returning a thunk
const someAsyncAction = createAsyncAction('SOME_ASYNC_ACTION', dispatch => {
  // ...
})

// I need to remember to do this instead
const someAsyncAction = createAsyncAction('SOME_ASYNC_ACTION', () => dispatch => {
  // ...
})
```

Not sure how much of a problem it is, especially if you are using TypeScript. But maybe there is a tweak that could be done to mitigate this issue; perhaps throwing a helpful error if no thunk is returned from the passed function. Something along the lines of:

```
function createAsyncAction(type, thunkCreator) {
  const action = payload => {
    const thunk = thunkCreator(payload);

    if (!(thunk instanceof Function)) {
      throw new TypeError(
        `The function you passed to createAsyncAction('${type}') did not `
        'return a thunk. Did you maybe pass a thunk directly instead of a '
        'function returning a thunk?')
    }

    return thunk;
  };

  action.toString = () => type;
  return action;
}

const someAsyncAction = createAsyncAction('SOME_ASYNC_ACTION', dispatch => {
  // ...
})

someAsyncAction() // TypeError
```

[...]

PR #352, opened 2020-02-12
Port ngrx/entity and add createAsyncThunk

markerikson:
Per #76 and #333 , I've decided that there is a lot of potential benefit in having an official minimal solution for dealing with the most common data fetching and normalized data storage scenarios.

To that end, I've ported https://ngrx.io/guide/entity to Redux Toolkit, but with a bunch of changes:

- Rewrote the internals to use Immer and simplify the logic
- Reordered the arguments to the generated `EntityAdapter` methods to take `state` first and also feature-detect `PayloadAction`s, so that they can be used directly as Redux case reducers

(I'd hoped that [`@ngrx/entity` might be made more generic](https://github.com/ngrx/platform/issues/2333) so it could be used by both NgRx and RTK, but the changes I ended up making are too drastic.)

I've also added a `createAsyncThunk` function as originally written in https://github.com/reduxjs/redux-toolkit/issues/76#issuecomment-575955644 .  

This should allow usage patterns like:

```ts
const usersAdapter = createEntityAdapter();

const fetchUsers = createRequestAction(
    "users/fetch",
    () => usersAPI.fetchAll()
);

// `fetchUsers` is now a typical thunk action creator, but also has
// four more action creators attached:
// pending, fulfilled, finished, rejected
// it will automatically dispatch those based on the promise lifecycle

const usersSlice = createSlice({
    name: "users",
    initialState: usersAdapter.getInitialState(),
    reducers: {
        userAdded: usersAdapter.addOne,
        userRemoved: usersAdapter.removeOne,
        // etc
    },
    extraReducers: {
        // would also want to handle the loading state cases, probably with a state machine,
        // using the lifecycle actions attached to fetchusers
        [fetchUsers.fulfilled](state, action) {
            return usersAdapter.upsertMany(state, action.payload.result)
        }
    }
});
```

Published it as https://github.com/reduxjs/redux-toolkit/releases/tag/v1.3.0-alpha.0


markerikson:
I'm already not happy with how `createAsyncThunk` is passing in values to the generated action creators. I'd rather use prepare callbacks so that they're called like `fulfilled(result, args)`, and we get `{type: "someData/fulfilled", payload: result, meta: {args})`, rather than `payload: {result, args}`.  That way we can reuse the entity adapter methods and directly handle the results because they're in the payload.

markerikson:
I removed the `removeMany(predicate)` overload since we discourage passing functions in through actions, and I don't want to have folks trying to do that when it's used as a reducer.  Straightforward enough to calculate the list of IDs to remove yourself.

I also reworked the thunk actions to put the thunk args in as `action.meta.args`.

Unfortunately, I realized that before releasing `alpha.0`, my attempt to put all the fetch callback parameters into an options object probably broke the ability to correctly infer the type of the `args` param correctly (ie, what the user is supposed to pass into the thunk as they're dispatching it).  Also, the inferred types are incredibly long and hard to read.

@phryneas , can you take a look at those issues?  I'm getting over my head trying to make changes there.

lacolaco:
As far as I understood, `users/fetch` doesn't seem following the redux style guide.
https://redux.js.org/style-guide/style-guide/#model-actions-as-events-not-setters

Is `fetch` an event? Should I choose `users/fetched` rather?
I'd like to know the thought about the recommendation of async actions.

markerikson:
@lacolaco : I'm not worried about style guide stuff atm.  I'm just trying to get the API working the way I want it to, and throwing together small snippets to try to exercise the code.  Don't read anything into the action type names in the tests.

lacolaco:
@markerikson I see, thank you. Instead, I leave the same comment on the original issue.

src/entities/unsorted_state_adapter.ts
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  function updateOneMutably(update: Update<T>, state: R): void
+  function updateOneMutably(update: any, state: any): void {
+    return updateManyMutably([update], state)
+  }
+
+  function updateManyMutably(updates: Update<T>[], state: R): void
+  function updateManyMutably(updates: any[], state: any): void {
+    const newKeys: { [id: string]: string } = {}
```
phryneas: what if two `newKeys` overlap?

src/entities/unsorted_state_adapter.ts
```diff
@@ -0,0 +1,175 @@
...
+    updates = updates.filter(update => update.id in state.entities)
+
+    const didMutateEntities = updates.length > 0
+
+    if (didMutateEntities) {
+      const didMutateIds =
+        updates.filter(update => takeNewKey(newKeys, update, state)).length > 0
+
+      if (didMutateIds) {
+        state.ids = state.ids.map((id: any) => newKeys[id] || id)
```
phryneas: if two newKeys overlap (and we allow that), then one of the ids would be duplicated here, so `state.ids.length > Object.values(state.entities).length`

src/entities/sorted_state_adapter.ts
```diff
@@ -0,0 +1,167 @@
...
+import { createStateOperator } from './state_adapter'
+import { createUnsortedStateAdapter } from './unsorted_state_adapter'
+import { selectIdValue } from './utils'
+
+export function createSortedStateAdapter<T>(
+  selectId: IdSelector<T>,
+  sort: Comparer<T>
+): EntityStateAdapter<T>
+export function createSortedStateAdapter<T>(selectId: any, sort: any): any {
+  type R = EntityState<T>
```
phryneas: When you're done I'd try if we can type this somehow that we can get a Key type somewhere here and there's not so much confusion about number and string ids everywhere in the typings.

src/entities/sorted_state_adapter.ts
```diff
@@ -0,0 +1,167 @@
...
+      if (a[i] === b[i]) {
+        continue
+      }
+      return false
+    }
+    return true
+  }
+
+  function merge(models: T[], state: R): void
+  function merge(models: any[], state: any): void {
```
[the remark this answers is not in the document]
markerikson: Correct, standard normalized state shape as described in the Redux docs:

https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape

src/entities/sorted_state_adapter.ts
markerikson: Totally agreed.

src/entities/sorted_state_adapter.ts
```diff
@@ -0,0 +1,167 @@
...
+    const newKey = selectIdValue(updated, selectId)
+
+    delete state.entities[update.id]
+
+    models.push(updated)
+
+    return newKey !== update.id
+  }
+
+  function updateManyMutably(updates: Update<T>[], state: R): void
```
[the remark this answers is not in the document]
markerikson: Interesting question.  Yeah, I haven't dug too far into the minutia of how these actually behave.

src/entities/sorted_state_adapter.ts
```diff
@@ -0,0 +1,167 @@
...
+      merge(models, state)
+    }
+  }
+
+  function mapMutably(map: EntityMap<T>, state: R): void
+  function mapMutably(updatesOrMap: any, state: any): void {
+    const updates: Update<T>[] = state.ids.reduce(
+      (changes: any[], id: string | number) => {
+        const change = updatesOrMap(state.entities[id])
+        if (change !== state.entities[id]) {
```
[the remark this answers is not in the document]
markerikson: Also something that should be looked at, yeah.

docs/api/createAsyncThunk.md
```diff
@@ -0,0 +1,278 @@
...
+
+### `payloadCreator`
+
+A callback function that should return a promise containing the result of some asynchronous logic. It may also return a value synchronously. If there is an error, it should return a rejected promise containing either an `Error` instance or a plain value such as a descriptive error message.
+
+The `payloadCreator` function can contain whatever logic you need to calculate an appropriate result. This could include a standard AJAX data fetch request, multiple AJAX calls with the results combined into a final value, interactions with React Native `AsyncStorage`, and so on.
+
+The `payloadCreator` function will be called with two arguments:
+
+- `arg`: a single value, containing the first parameter that was passed to the thunk action creator when it was dispatched. This is useful for passing in values like item IDs that may be needed as part of the request. If you need to pass in multiple values, pass them together in an object when you dispatch the thunk, like `dispatch(fetchUsers({status: 'active', sortBy: 'name'}))`.
```
alexkrolick: Suggestion: treat the object as the default approach. Otherwise it is tempting to add positional args after the first one, which will not work, but you might not even notice for a while depending on how defaults are handled.

```suggestion
`arg`: a single value, containing the first parameter that was passed to the thunk action creator when it was dispatched. This is useful for passing in values like item IDs that may be needed as part of the request. Passing an object is recommended for most cases,
```

docs/api/createAsyncThunk.md
```diff
@@ -0,0 +1,278 @@
...
+`createAsyncThunk` returns a standard Redux thunk action creator. The thunk action creator function will have plain action creators for the `pending`, `fulfilled`, and `rejected` cases attached as nested fields.
+
+When dispatched, the thunk will:
+
+- dispatch the `pending` action
+- call the `payloadCreator` callback and wait for the returned promise to settle
+- when the promise settles:
+  - if the promise resolved successfully, dispatch the `fulfilled` action with the promise value as `action.payload`
+  - if the promise failed, dispatch the `rejected` action with a serialized version of the error value as `action.error`
+- Return a fulfilled promise containing the final dispatched action (either the `fulfilled` or `rejected` action object)
```
alexkrolick: API comment: if awaiting the thunk returns a usable payload, it's very tempting to use the value directly (for example, by passing it to a local setState) instead of reading from redux state. I've seen this is a lot in an app that has a similar utility.

docs/api/createAsyncThunk.md
markerikson: Are you talking about at strictly a documentation level, or something at the code level too?

docs/api/createAsyncThunk.md
markerikson: You can already do that with thunks today, so I don't see any difference here conceptually.

Most of the time, folks are just interested in kicking off some kind of additional behavior after the thunk promise resolves:

```js
dispatch(someThunk())
  .then(() => {})
```

The issue here is that since we're catching errors internally, we always return a fulfilled promise, and it's hypothetically possible the user _might_ A) be interested in the original API call result, or B) want to deal with the API call having rejected.

So, we're trying to provide a way to enable that.

markerikson:
Hmm.  So it looks like `AbortController` definitely isn't supported in IE:

https://developer.mozilla.org/en-US/docs/Web/API/AbortController

How do we want to handle that?  Handle it ourselves and no-op?  Tell users to polyfill it themselves?

markerikson:
Next point of concern:

It seems like trying to actually _use_ `getState` is going to require hand-specifying a whole lot of generics.

I was just able to reorder things and add some defaults, and get it down to a more passable:

```ts
const thunk = createAsyncThunk<Return, Arg, State>("a", (a: Arg) => {})
```

However, this sounds like a neat potential approach:

https://twitter.com/ArrowoodTech/status/1228911887997448192
https://github.com/fastify/fastify/blob/next/types/request.d.ts
https://github.com/Ethan-Arrowood/fastify/blob/typescript-docs/docs/TypeScript.md#using-generics

src/createAsyncThunk.ts
```diff
@@ -0,0 +1,183 @@
...
+ *
+ * @param type
+ * @param payloadCreator
+ *
+ * @alpha
+ */
+export function createAsyncThunk<
+  Returned,
+  ThunkArg = void,
+  State = unknown,
```
Ethan-Arrowood: Good use of defaulting with `unknown`. That provides a really good developer experience IMO

src/createAsyncThunk.ts
```diff
@@ -0,0 +1,183 @@
...
+}
+
+/**
+ *
+ * @param type
+ * @param payloadCreator
+ *
+ * @alpha
+ */
+export function createAsyncThunk<
```
Ethan-Arrowood: As far as I can tell you could make this whole parameter list into an interface.

Even though i haven't tried it, I believe you can even make certain properties on the _named property_ generic required. 

i.e.

```ts
interface createAsyncThunkI {
  Returned: unknown, // this is required
  ThunkArg?: void
  ...
}

export function createAsyncThunk<G extends createAsyncThunkI = createAsyncThunkI>() {}
```

markerikson: Right, that's exactly what I'm asking for help on figuring out :)

src/createAsyncThunk.ts
```diff
@@ -0,0 +1,183 @@
...
+  Returned,
+  ThunkArg = void,
+  State = unknown,
+  Extra = unknown,
+  DispatchType extends Dispatch = Dispatch,
```
Ethan-Arrowood: These expressions, and their use in the `ThunkAPI` parameter may cause a problem. Not 100% certain on this, but I think I hit this problem when I tried implementing this pattern with a more complex generic type inside fastify

src/createAsyncThunk.ts
markerikson: Playing around with this locally now atm.  Lemme see if I can use this tiny snippet as inspiration...

src/createAsyncThunk.ts
```diff
@@ -0,0 +1,183 @@
...
+  ActionType extends string = string,
+  ThunkAPI extends BaseThunkAPI<any, any, any> = BaseThunkAPI<
+    State,
+    Extra,
+    DispatchType
+  >
+>(
+  type: ActionType,
+  payloadCreator: (
```
Ethan-Arrowood: Could these generics be passed in through this `payloadCreator` method? 

Maybe something this could work:

```ts
type payloadCreator<Returned, ThunkArg = void, State = unknown, Extra = unknown, DispatchType = unknown>(
  arg: ThunkArg,
  thunkAPI: BaseThunkAPI<State, Extra, DispatchType>
) => Promise<Returned> | Returned

export function createAsyncThunk<
  Returned,
  ThunkArg = void,
  State = unknown,
  Extra = unknown,
  DispatchType = unknown
>(
  type: ActionType,
  payloadCreator: payloadCreator<Returned, ThunkArg, State, Extra, DispatchType>
) { ... }

This can help reduce how many generics you have in the createAsyncThunk definition

Comment (markerikson) [reply src/createAsyncThunk.ts]: Just tried playing with this, but it's 1 AM and my head hurts.  Gonna go to bed now :)

@phryneas , if you've got time tomorrow, please give this a shot and see what happens.

**Comment by markerikson** [comment]
Published the current rearranged types as `alpha.4` as a stopgap.  Would be great if we can figure out how to get this "specific fields" approach to work.

**Comment by phryneas** [comment]
> Hmm. So it looks like `AbortController` definitely isn't supported in IE:
> 
> https://developer.mozilla.org/en-US/docs/Web/API/AbortController
> 
> How do we want to handle that? Handle it ourselves and no-op? Tell users to polyfill it themselves?

People already have to transpile and (at least) polyfill `regeneratorRuntime` if they want to target IE, probably they already have a lot additional polyfills in place.
I'd just tell users to use one more polyfill. Silently no-oping might cause unexpected behaviour rather than help.

File: src/createAsyncThunk.ts (hunk 12 above)
Comment (phryneas) [reply src/createAsyncThunk.ts]: I'll take a look. I had hit similar stopgaps (inference around parameters of functions that are represented by an inferred generic seems to be pretty weak), but I was hurrying because I wanted to give you that worked at all.
Now I've got some extra time & can take a thorough look.

But there's a more pressing matter before that: the generated types are plain wrong. This is a known bug in api-extractor, I know a workaround & will apply it.

**Comment by phryneas** [comment]
So there was a bug with the .d.ts rollup.

image
(screenshot of https://unpkg.com/browse/@reduxjs/toolkit@1.3.0-alpha.4/dist/typings.d.ts) 

These `import` statements should not be there after the rollup and cast all associated stuff to `any`.

This is [a bug of api-extractor](https://github.com/microsoft/rushstack/issues/1050) in combination with a very specific pattern generated by `tsc` and solved by importing the types that lead to this pattern in the file where they are used.
I just pushed https://github.com/reduxjs/redux-toolkit/pull/352/commits/39c5b3176bdea0ebeea12e15af703813dcf3d616  to this branch to solve that issue.

File: docs/api/createAsyncThunk.md (hunk 7 above)
Comment (alexkrolick) [reply docs/api/createAsyncThunk.md]: Just docs

File: src/entities/state_adapter.ts [hunk 13]
```diff
@@ -0,0 +1,26 @@
...
+  mutator: (arg: any, state: any) => void
+): any {
+  return function operation<S extends EntityState<V>>(
+    state: any,
+    arg: R | PayloadAction<R>
+  ): S {
+    // @ts-ignore createNextState() produces an Immutable<Draft<S>> rather
+    // than an Immutable<S>, and TypeScript cannot find out how to reconcile
+    // these two types.
+    return createNextState(state, (draft: Draft<EntityState<V>>) => {
```
msutkowski: If state is already an immer instance, would it make sense to keep the existing draft so it could be further modified in the reducer after any entity updates happen instead of returning a transformed object?

markerikson: Hmm.  Y'know, that may be a potential solution for the problem I'm seeing over in https://github.com/immerjs/immer/issues/377 .

It looks like there is an `isDraft` function exported by Immer:

https://immerjs.github.io/immer/docs/original

So yeah, I could very likely check that, and if it's a draft, just immediately call the `mutator` function instead of wrapping in `createNextState`.  Otherwise, assume this is being called standalone.

msutkowski: Makes sense to me. That behavior would make the entities usage really straightforward.

markerikson: Yeah, that worked!  Published as [`alpha.6`](https://github.com/reduxjs/redux-toolkit/releases/tag/v1.3.0-alpha.6)

src/entities/models.ts
```diff
@@ -0,0 +1,140 @@
+import { PayloadAction } from '../createAction'
+
+export type ComparerStr<T> = (a: T, b: T) => string
```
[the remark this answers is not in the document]
markerikson: Yeah, the `Array.sort()` signature clearly says the result is a number (which of course should be 1/0/-1).  Good catch!

src/entities/sorted_state_adapter.ts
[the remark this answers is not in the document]
markerikson: Yes, just confirmed that if I have items `["a", "b", "c"]`, and I do `updateOne({id: "b", changes: {id: "c"}})`, it will replace the previous `"c"` entry entirely.

src/entities/unsorted_state_adapter.ts
markerikson: Yeah, there's definitely a bug here.  Take this case:

```ts
    const withA = adapter.addOne(state, { id: 'a', title: 'First' })

    const withUpdates = adapter.updateMany(withA, [
      { id: 'a', changes: { id: 'b' } },
      { id: 'a', changes: { id: 'c' } }
    ])

    const { ids, entities } = withUpdates
```

The resulting state looks like:

```js
    {
      ids: [ 'c' ],
      entities: { b: { id: 'b', title: 'First' }, c: { id: 'c' } }
    }
```

The rename for B went through fully, but then the `'c'` rename didn't have an existing object to merge with.  The contents of just the update object were inserted, and we saved the `a -> c` ID mapping, so that's the final ID that ends up in `state.ids`

markerikson: Fixed it by changing the logic to merge together all updates targeting the same ID:

```ts
    const updatesPerEntity: { [id: string]: Update<T> } = {}

    updates.forEach(update => {
      // Only apply updates to entities that currently exist
      if (update.id in state.entities) {
        // If there are multiple updates to one entity, merge them together
        updatesPerEntity[update.id] = {
          ...updatesPerEntity[update.id],
          ...update
        }
      }
    })

    updates = Object.values(updatesPerEntity)
```

markerikson: and of course the spread operator _conveniently_ doesn't do anything with falsy values, allowing that line to work whether it's the tenth update for that item or the first.

src/entities/sorted_state_adapter.ts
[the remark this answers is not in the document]
markerikson: If there was some mutation inside, it should have been caught by Immer already, because we're either in a `produce()` _outside_ the entire entity logic (`createSlice`), or in a `produce()` up one level from here.  So, in the end Immer should have tracked that update regardless.

src/entities/sorted_state_adapter.ts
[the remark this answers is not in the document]
markerikson: Documented.

markerikson:
At this point I think I feel pretty good about the overall shape of the APIs.

I'm going to go ahead and merge this PR into the `v.1.3.0-integration` branch.  If we do have any more tweaks, we can make them there.

Next step will be to start working on porting `redux-immutable-state-invariant` over.

src/entities/models.ts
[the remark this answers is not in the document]
LMFinney: I'm late to the party, but I noticed that bug in NgRx a few weeks ago, and it has since been removed.

Issue #341, opened 2020-02-03
Use generic to type the name of a slice


tgouala:
Specifying a generic type for the property `name` of a slice would help typescript to deduce the type of objects built from slice.

As a practical example, I would like to put all my action creators inside an object I can then type and pass to Thunk extra arguments, or a react-redux mapDispatchToProps. This looks like this :
```ts
const clientSlice = createSlice({name:'clients', reducers: {get: (state) => {}}});
const categorySlice = createSlice({name:'category', reducers: {save: (state) => {}}});
const uiSlice = createSlice({name:'ui', reducers: {goToHome: (state) => {}}});

export const actionCreators = {
   [clientSlice.name]: {...clientSlice.actions},
   [categorySlice.name]: {...categorySlice.actions},
   [uiSlice.name]: {...uiSlice.actions}
}

export type ActionCreators = typeof actionCreators
```
If name was a generic extending string, I could have autocompletion in code like 
```ts
const connectMyComponent = (dispatch: Dispatch) : MapDispatchToProp => (actionCreators : ActionCreators) => ({
   getClients: dispatch(actionCreators.clients.get()),
   saveCategory: dispatch(actionCreators.category.save()),
   changePage: dispatch(actionCreators.ui.goToHome())
})
```

Unfortunately, this won't work since `clientSlice.name` is typed `string`.

There is a workaround you can see below, but integrating a generic type would be quite easy and straightforward (I can do a PR if wanted). 

**Workaround :**
It's possible to wrap the function createSlice like describe [here in rtk doc](https://redux-toolkit.js.org/usage/usage-with-typescript#wrapping-createslice) and return an object with a type for the property `name` :

```js
const createGenericSlice = <
  Name extends string
  T,
  Reducers extends SliceCaseReducers<T>
>({
  name = "",
  initialState,
  reducers
}: {
  name: Name
  initialState: T
  reducers: ValidateSliceCaseReducers<T, Reducers>
}) => {
  const slice = createSlice({
    name,
    initialState,
    reducers: {
      ...reducers
    }
  });

  return slice as typeof slice & {name: Name};
}

const clientSlice = createGenericSlice({
  name: 'clients',
  initialState: { status: 'loading' } as GenericState<string>,
  reducers: {
    {get: (state) => {}}
  }
})
```

then `typeof clientSlice ['name']` will be the literal 'clients'.

tgouala:
[...]

PR #354, opened 2020-02-12
strongly type slice name

tgouala:
This is a proposal for #341


phryneas:
Please put the new generic argument at the end of the list, and not at the beginning. Also, please let it default to `string`. Otherwise this would be a breaking change for everyone who actually specified those parameters. Usually those are inferred and it wouldn't matter, but this would be a BC for everyone who [wraps createSlice](https://redux-toolkit.js.org/usage/usage-with-typescript#wrapping-createslice).

tgouala:
Thanks for the feedback @phryneas . Let me know if anything else is still missing.

phryneas:
Sorry for the delay. So far it looks good to me - @markerikson if it works for you too we could get this merged.

PR #361, opened 2020-02-15
createAsyncThunk return fulfilled/rejected action instead of re-…

phryneas:
This is an alternative suggestion to #359.

I've created [a codesandbox](https://codesandbox.io/s/still-pond-7208i) that references the current alpha and implements a small data-fetching workflow.
The problem that becomes apparent is that, if there is an error within fetch and that error is not caught from dispatch, CRA in Dev Mode displays the red overlay of "something went terribly wrong":
image

This happens during an event handler as well as during `useEffect`.

In production, this would probably not happen, but realisticly, devs would react to this by wrapping every `dispatch` call in an async function with try/catch or append a `.catch()` to it.
Doing the first thing is quite complicated, because you can't just change an effect to an async function (because that would return a promise), so you come up with something like 
```
  React.useEffect(() => {
    async function doFetch() {
      try {
        dispatch(fetchDataThunk({}));
      } catch {
        // do nothing
      }
    }
    if (state === "initial") {
      doFetch();
    }
  }, [state, dispatch]);
```
for something that would otherwise just have been
```
  React.useEffect(() => {
    if (state === "initial") {
        dispatch(fetchDataThunk({}));
    }
  }, [state, dispatch]);
```

So this makes stuff *much*  more complicated - also, event handlers that would just have been `() => dispatch(fetchDataThunk({}))` now become either `() => dispatch(fetchDataThunk({})).catch()` or `async () => { try {dispatch(fetchDataThunk({}))} catch {} }`.

What I want to say: avoiding this gets really messy.

So this is my suggestion of another solution:
The thunk would not return the promise return value in this case, but the "last dispatched action", which means either the `fulfilled` or the `rejected` action. That way, the error *can* be accessed from React, but not everyone is forced to handle it correctly.

### DIFF
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -101,34 +101,33 @@ export function createAsyncThunk<
     ) => {
       const requestId = nanoid()
 
-      let result: Returned
+      let finalAction: ReturnType<typeof fulfilled | typeof rejected>
       try {
         dispatch(pending(requestId, args))
 
-        result = (await payloadCreator(args, {
-          dispatch,
-          getState,
-          extra,
-          requestId
-        } as TA)) as Returned
+        finalAction = fulfilled(
+          await payloadCreator(args, {
+            dispatch,
+            getState,
+            extra,
+            requestId
+          } as TA),
+          requestId,
+          args
+        )
       } catch (err) {
         const serializedError = miniSerializeError(err)
-        dispatch(rejected(serializedError, requestId, args))
-        // Rethrow this so the user can handle if desired
-        throw err
+        finalAction = rejected(serializedError, requestId, args)
       }
 
       // We dispatch "success" _after_ the catch, to avoid having any errors
       // here get swallowed by the try/catch block,
       // per https://twitter.com/dan_abramov/status/770914221638942720
       // and https://redux-toolkit.js.org/tutorials/advanced-tutorial#async-error-handling-logic-in-thunks
-      return dispatch(fulfilled(result!, requestId, args))
+      dispatch(finalAction)
+      return finalAction
     }
   }
 
-  actionCreator.pending = pending
-  actionCreator.rejected = rejected
-  actionCreator.fulfilled = fulfilled
-
-  return actionCreator
+  return Object.assign(actionCreator, { pending, rejected, fulfilled })
 }
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -2,6 +2,8 @@ import { Dispatch } from 'redux'
 import nanoid from 'nanoid'
 import { createAction } from './createAction'
 
+type Await<P> = P extends PromiseLike<infer T> ? T : P
+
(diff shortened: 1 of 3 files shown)


src/createAsyncThunk.ts
```diff
@@ -101,34 +101,33 @@ export function createAsyncThunk<
...
+      return finalAction
     }
   }
 
-  actionCreator.pending = pending
-  actionCreator.rejected = rejected
-  actionCreator.fulfilled = fulfilled
-
-  return actionCreator
+  return Object.assign(actionCreator, { pending, rejected, fulfilled })
```
phryneas: Oh, just to bring your attention to this: before, `pending`, `rejected` and `fulfilled` where `any`-typed, the approach with `Object.assign` is the easiest way to preserve the types here.

markerikson:
Yeah, a user was alpha-testing `createAsyncThunk` last night and pointed this out, and it's exactly what I was also wondering might be a point of concern.

I've semi-frequently seen folks try to chain off thunks in components, but they've generally only cared about the success case.  

Having to check whether the promise return value is a failure action is awkward, but probably the best alternative here.

markerikson:
hah, you forgot to update the public API file :)  I've only done that FOR EVERY ATTEMPT TO PUBLISH AN ALPHA SO FAR!

phryneas:
Yeah, it's not perfect, but I think it can still be written pretty compact using the `.match` method:
`dispatch(async(3)).then(result => async.fulfilled.match(result) && /* do something */console.log(result))`
We'll just have to document that.

We could even add a helper that would re-introduce a `throw` for people that want it, so
`dispatch(async(3)).then(async.throwErrorsFromPromise).then(/*...*).catch(/*...*/)`
could be a possibility.

phryneas:
> hah, you forgot to update the public API file :) I've only done that FOR EVERY ATTEMPT TO PUBLISH AN ALPHA SO FAR!

Yap :D
I even created it and force-pushed an amended commit where I forgot to add the change before amending -.-

markerikson:
Sure, wanna put that throw helper in real fast?

phryneas:
Yap, why not :)

phryneas:
Okay, it's in there. If you have a better idea, feel free to change that method name :D

markerikson:
How about `rejectIfError`?

````js
dispatch(fetchStuff())
    .then(rejectIfError)
    .catch(handleError)
```

etc

phryneas:
> How about `rejectIfError`?
> 
> ```js
> dispatch(fetchStuff())
>     .then(rejectIfError)
>     .catch(handleError)
> ```
> 
> etc

Hmm. Both names don't really reflect the other side of it, that the return value would also be unwrapped. :thinking: 

`unwrapAndRejectIfError` would reflect what it really does, but is a bit long.. `splitPromise` makes sense if you know what it does... `normalizePromise` likewise, but from the alternatives I can come up I think it's the best-sounding.... not really sure there. :man_shrugging:

PR #362, opened 2020-02-15
add .abort() to the createAsyncThunk thunkAction

phryneas:
It's a bit... async, sorry - I don't think this is much simpler to solve.

The tests should give you a pretty good idea of what's actually happening.

### DIFF
--- a/etc/redux-toolkit.api.md
+++ b/etc/redux-toolkit.api.md
@@ -102,18 +102,20 @@ export function createAction<P = void, T extends string = string>(type: T): Payl
 export function createAction<PA extends PrepareAction<any>, T extends string = string>(type: T, prepareAction: PA): PayloadActionCreator<ReturnType<PA>['payload'], T, PA>;
 
 // @alpha (undocumented)
-export function createAsyncThunk<ActionType extends string, Returned, ActionParams = void, TA extends AsyncThunksArgs<any, any, any> = AsyncThunksArgs<unknown, unknown, Dispatch>>(type: ActionType, payloadCreator: (args: ActionParams, thunkArgs: TA) => Promise<Returned> | Returned): ((args: ActionParams) => (dispatch: TA["dispatch"], getState: TA["getState"], extra: TA["extra"]) => Promise<import("./createAction").PayloadAction<Returned, string, {
+export function createAsyncThunk<ActionType extends string, Returned, ActionParams = void, TA extends AsyncThunksArgs<any, any, any> = AsyncThunksArgs<unknown, unknown, Dispatch>>(type: ActionType, payloadCreator: (args: ActionParams, thunkArgs: TA) => Promise<Returned> | Returned): ((args: ActionParams) => ((dispatch: TA["dispatch"], getState: TA["getState"], extra: TA["extra"]) => Promise<import("./createAction").PayloadAction<Returned, string, {
     args: ActionParams;
     requestId: string;
 }, never> | import("./createAction").PayloadAction<undefined, string, {
     args: ActionParams;
     requestId: string;
-}, Error>>) & {
+}, any>>) & {
+    abort: (reason?: string) => void;
+}) & {
     pending: import("./createAction").ActionCreatorWithPreparedPayload<[string, ActionParams], undefined, string, never, {
         args: ActionParams;
         requestId: string;
     }>;
-    rejected: import("./createAction").ActionCreatorWithPreparedPayload<[Error, string, ActionParams], undefined, string, Error, {
+    rejected: import("./createAction").ActionCreatorWithPreparedPayload<[Error, string, ActionParams], undefined, string, any, {
         args: ActionParams;
         requestId: string;
     }>;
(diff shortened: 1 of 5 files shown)


markerikson:
Yipes. Yeah, getting complicated.

It's almost like there are other asynchronous logic abstractions that are perhaps better meant for cancellation or something!

src/createAsyncThunk.ts
```diff
@@ -129,21 +183,25 @@ export function createAsyncThunk<
...
-    return returned.payload
+    return Object.assign(thunkAction, { abort })
   }
 
   return Object.assign(actionCreator, {
     pending,
     rejected,
-    fulfilled,
-    unwrapResult
+    fulfilled
```
markerikson: Pretty sure we need to include `aborted` in here so it can be handled?

src/createAsyncThunk.ts
```diff
@@ -84,24 +83,67 @@ export function createAsyncThunk<
     }
   )
 
+  const aborted = createAction(
```
markerikson: Do we actually need a separate `aborted` action type?

Could we use `rejected`, and have `aborted: true` as part of `meta`?

phryneas: That would be an option, too. I'm fine either way - your call :)

phryneas:
> It's almost like there are other asynchronous logic abstractions that are perhaps better meant for cancellation or something!

Well, definitely not promises :D

This is what the fetch API uses, so it would fit most use cases I guess - and actually using it is pretty straightforward, too (see the tests).

src/createAsyncThunk.test.ts
```diff
@@ -104,3 +109,111 @@ describe('createAsyncThunk', () => {
...
+      expect.any(Object),
+      expect.objectContaining({ type: 'test/pending' }),
+      expect.objectContaining({ type: 'test/fulfilled' })
+    ])
+  })
+
+  test('abort after dispatch', async () => {
+    const thunkAction = asyncThunk({})
+    const promise = store.dispatch(thunkAction)
+    thunkAction.abort('AbortReason')
```
markerikson: The usage here seems unlikely to occur in a real app.  Most code isn't going to reference the `thunkAction` value at all, as it's going to be `dispatch(asyncThunk())`.  

I realize that the abort handling needs to be on a per-call basis, so I'm not sure what the right answer is here.

etc/redux-toolkit.api.md
```diff
@@ -102,32 +102,27 @@ export function createAction<P = void, T extends string = string>(type: T): Payl
 export function createAction<PA extends PrepareAction<any>, T extends string = string>(type: T, prepareAction: PA): PayloadActionCreator<ReturnType<PA>['payload'], T, PA>;
 
 // @alpha (undocumented)
-export function createAsyncThunk<ActionType extends string, Returned, ActionParams = void, TA extends AsyncThunksArgs<any, any, any> = AsyncThunksArgs<unknown, unknown, Dispatch>>(type: ActionType, payloadCreator: (args: ActionParams, thunkArgs: TA) => Promise<Returned> | Returned): ((args: ActionParams) => (dispatch: TA["dispatch"], getState: TA["getState"], extra: TA["extra"]) => Promise<import("./createAction").PayloadAction<Returned, string, {
+export function createAsyncThunk<ActionType extends string, Returned, ActionParams = void, TA extends AsyncThunksArgs<any, any, any> = AsyncThunksArgs<unknown, unknown, Dispatch>>(type: ActionType, payloadCreator: (args: ActionParams, thunkArgs: TA) => Promise<Returned> | Returned): ((args: ActionParams) => ((dispatch: TA["dispatch"], getState: TA["getState"], extra: TA["extra"]) => Promise<import("./createAction").PayloadAction<Returned, string, {
```
markerikson: Missed the API file update again :)

src/createAsyncThunk.test.ts
phryneas: Yeah, I think cancellation is not the most common thing per se, as for many people ignoring earlier running requests might be enough - but if needed this api provides all tools to create a simple reusable abstraction:

```js
function useAbortingApi(asyncApiAction, args) {
  const abortRef = useRef();
  return () => {
    const thunkAction = asyncApiAction(args);
    if (abortRef.current) abortRef.current();
    abortRef.current = thunkAction.abort;
    dispatch(thunkAction);
  }
}

function MyComponent(){
  const clickHandler = useAbortingApi(api.fetch, {});
  return <button onClick={clickHandler}>...</button>;
}
```

phryneas: That said, I *could* try to put an `abort` method on the promise returned by the thunk... Might even be pretty simple.

phryneas: That eliminated the "aborted before dispatch" case and allowed for a lot of simplifications.
I hadn't even considered that :man_facepalming:

phryneas: But the diff is ruined now *sigh*

src/createAsyncThunk.ts
```diff
@@ -87,63 +86,94 @@ export function createAsyncThunk<
   const rejected = createAction(
     type + '/rejected',
     (error: Error, requestId: string, args: ActionParams) => {
+      if (error.name === 'AbortError') {
```
markerikson: Any chance we can collapse these two cases to shorten it a bit?

phryneas: Better?

markerikson:
Awright, good enough for now.  Great work!

PR #364, opened 2020-02-16
use ThunkApiConfig for optional type arguments

phryneas:
I think this is what @Ethan-Arrowood suggested in [this tweet](https://twitter.com/ArrowoodTech/status/1228908646551183360).

It actually is quite a different pattern from what we've been using, but also quite interesting.

@markerikson, @Ethan-Arrowood could you take a look?

### DIFF
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -1,10 +1,12 @@
-import { Dispatch } from 'redux'
+import { Dispatch, AnyAction } from 'redux'
 import nanoid from 'nanoid'
 import {
   createAction,
   PayloadAction,
   ActionCreatorWithPreparedPayload
 } from './createAction'
+import { ThunkDispatch } from 'redux-thunk'
+import { FallbackIfUnknown } from './tsHelpers'
 
 // @ts-ignore we need the import of these types due to a bundling issue.
 type _Keep = PayloadAction | ActionCreatorWithPreparedPayload<any, unknown>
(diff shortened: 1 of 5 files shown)


src/createAsyncThunk.ts
```diff
@@ -60,20 +89,12 @@ export const miniSerializeError = (value: any): any => {
...
-  DispatchType extends Dispatch = Dispatch,
-  ActionType extends string = string,
-  ThunkAPI extends BaseThunkAPI<any, any, any> = BaseThunkAPI<
-    State,
-    Extra,
-    DispatchType
-  >
+  ThunkApiConfig extends AsyncThunkConfig = {}
 >(
-  type: ActionType,
```
phryneas: I've removed this type completely since it does not seem to have any value - all action types are concatenation to that string, so it'll never be used again.

src/createAsyncThunk.ts
```diff
@@ -50,6 +52,33 @@ export const miniSerializeError = (value: any): any => {
   return value
 }
 
+type AsyncThunkConfig = {
+  state?: unknown
+  dispatch?: Dispatch
+  extra?: unknown
+}
+
+type GetS<ThunkApiConfig> = ThunkApiConfig extends {
```
phryneas: The pattern here is essentially

```
GetX<ConfigType> = ConfigType extends { x: infer X } ? X : DefaultValue
```

phryneas:
https://github.com/reduxjs/redux-toolkit/blob/f5e32045278564402963081ea4e354577543b233/type-tests/files/createAsyncThunk.typetest.ts#L78-L85

This is essentially how it would be used. I've left the `ReturnValue` and `ThunkArgs` parameters as-is, as these should be automatically inferred in simple use cases. The third parameter takes optional properties and other properties fall back to sane defaults. The `dispatch` fallback in this case even references the user-provided `state` and `extra` types.

PR #366, opened 2020-02-16
RFC: Add predicate to `createSlice` reducers.

phryneas:
This would make it more natural to write slice reducers as state machines.

It's just an idea on how to handle this and we'll definitely release 1.3 before we'll add this, but why not get some early feedback? :)

@davidkpiano, I think especially your feedback would be very valuable. Could you take a look?

Usage would look like this:
https://github.com/reduxjs/redux-toolkit/blob/26271da274a673f402ba024776c0ca9a53a53e4b/src/createSlice.test.ts#L207-L228

### DIFF
--- a/etc/redux-toolkit.api.md
+++ b/etc/redux-toolkit.api.md
@@ -62,8 +62,10 @@ export type CaseReducer<S = any, A extends Action = AnyAction> = (state: Draft<S
 // @public
 export type CaseReducerActions<CaseReducers extends SliceCaseReducers<any>> = {
     [Type in keyof CaseReducers]: CaseReducers[Type] extends {
-        prepare: any;
-    } ? ActionCreatorForCaseReducerWithPrepare<CaseReducers[Type]> : ActionCreatorForCaseReducer<CaseReducers[Type]>;
+        prepare: AnyFunction;
+    } ? ActionCreatorForCaseReducerWithPrepare<CaseReducers[Type]> : CaseReducers[Type] extends {
+        reducer: AnyFunction;
+    } ? ActionCreatorForCaseReducer<CaseReducers[Type]['reducer']> : ActionCreatorForCaseReducer<CaseReducers[Type]>;
 };
 
 // @public @deprecated
@@ -74,7 +76,8 @@ export type CaseReducers<S, AS extends Actions> = {
 // @public
 export type CaseReducerWithPrepare<State, Action extends PayloadAction> = {
     reducer: CaseReducer<State, Action>;
-    prepare: PrepareAction<Action['payload']>;
+    prepare?: PrepareAction<Action['payload']>;
+    predicate?: Predicate<State, Action>;
 };
 
 // @public
@@ -202,9 +205,17 @@ export { ThunkAction }
 // @public
 export type ValidateSliceCaseReducers<S, ACR extends SliceCaseReducers<S>> = ACR & {
     [T in keyof ACR]: ACR[T] extends {
+        prepare: AnyFunction;
         reducer(s: S, action?: infer A): any;
     } ? {
         prepare(...a: never[]): Omit<A, 'type'>;
+    } : {};
+} & {
+    [T in keyof ACR]: ACR[T] extends {
+        predicate: AnyFunction;
+        reducer(s: S, action?: infer A): any;
+    } ? {
+        predicate: Predicate<S, A>;
     } : {};
 };
 
--- a/src/createSlice.test.ts
+++ b/src/createSlice.test.ts
@@ -202,3 +202,65 @@ describe('createSlice', () => {
     })
   })
 })
+
+test('createSlice with predicate for reducer', () => {
+  const slice = createSlice({
+    name: 'test',
+    initialState: {
+      status: 'idle'
+    } as { status: 'working' | 'idle'; workItem?: string },
+    reducers: {
+      startWork: {
+        predicate: state => state.status === 'idle',
+        reducer(state, action: PayloadAction<string>) {
+          state.status = 'working'
+          state.workItem = action.payload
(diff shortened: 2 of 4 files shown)


city41:
I'm just a random redux user who saw your tweet. I have to admit I don't think I'm a fan of this. I like that reducers currently are just functions. Making them objects with specific keys increases the cognitive load and required learning a tad, but with not much benefit because the predicate can just be an if statement at the top of the reducer function. I also think I would prefer to keep everything about the reducer contained in the function.

On the other hand, it does make the idea that reducers should be state machines a bit more explicit.

davidkpiano:
While this does help, predicates don't necessarily make it explicit that the behavior is for a specific state (multiple predicates may return `true`, as they are not mutually exclusive).

A helper for state machines would have to deviate from `createSlice`, which is action-first by its nature. Here's an idea:

```js
const userStatesReducer = createStates({
  key: 'status', // the key where the finite state lives
  initial: {
    status: 'idle',
    user: null,
    error: null
  },
  reducers: {
    // determined based on the key location
    idle: (state, action) => {/* ... */},
    loading: (state, action) => {/* ... */},
    failure: (state, action) => {/* ... */},
    success: (state, action) => {/* ... */}
  }
});
```

IMO, that API feels pretty nice! What do you think?

davidkpiano:
Actually, you can combine the above idea with `createSlice` and have a really nice, state-machine-based interface:

```js
const userMachine = createStates({
  key: 'status', // the key where the finite state lives
  initialState: {
    status: 'idle',
    user: null,
    error: null
  },
  reducers: {
    idle: {
      // same as `.reducers` in createSlice()
      fetch: (state, action) => {/* ... */}
    },
    loading: {
      cancel: (state, action) => {/* ... */},
      resolve: (state, action) => {/* ... */},
      reject: (state, action) => {/* ... */}
    },
    failure: {
      fetch: (state, action) => {/* ... */}
    },
    success: {
      fetch: (state, action) => {/* ... */}
    },
  }
});

dispatch(userMachine.actions.fetch(42)); // same as createSlice()
```

phryneas:
> I'm just a random redux user who saw your tweet. I have to admit I don't think I'm a fan of this. I like that reducers currently are just functions. Making them objects with specific keys increases the cognitive load and required learning a tad, but with not much benefit because the predicate can just be an if statement at the top of the reducer function. I also think I would prefer to keep everything about the reducer contained in the function.
> 
> On the other hand, it does make the idea that reducers should be state machines a bit more explicit.

This is not a new notation - we are already using this notation [for an optional `prepare` function](https://redux-toolkit.js.org/api/createSlice#examples), so this would just extend that existing notation. Of course, the "simple" syntax of passing just a normal reducer function will definitely be possible in the future, no matter how we decide on this.

phryneas:
@davidkpiano 
> (multiple predicates may return `true`, as they are not mutually exclusive).

I'm not sure if I really brought across what these would do (that alone is valuable feedback).

Multiple predicates being true would not be a problem, as these reducers are still bound to their respective action - the `startWork` predicate and reducer would only be evaluated if a `test/startWork` action would be dispatched.

> Actually, you can combine the above idea with createSlice and have a really nice, state-machine-based interface

As the notation from above just extends an [existing notation that takes an object with the keys `reducer` and `prepare`](https://redux-toolkit.js.org/api/createSlice#examples), we can't create a completely different functionality with overlapping semantics.


So, I'll take a more detailed step of explaining what this should do.

Before this RFC, you would currently be writing this:

```ts
const slice = createSlice({
    name: 'test',
    initialState: {
        status: 'idle'
    } as { status: 'working' | 'idle'; workItem?: string },
    reducers: {
        startWork(state, action: PayloadAction<string>) {
            if (state.status !== 'idle') {
                return;
            }
            state.status = 'working'
            state.workItem = action.payload
        },
        finishWork(state) {
            if (state.status !== 'working') {
                return
            }
            state.status = 'idle'
            state.workItem = undefined
        }
    }
}) 
```

This new `predicate` parameter would just move that early exit condition out of the reducer.

In the end, it would be possible to define something like
```ts
function fromStatus(status) {
  return (state) => state.status === status;
}
```

That could either (old notation) be used as 
```ts
        finishWork(state) {
            if (!fromStatus(state, 'working')) { // I tweaked it here for this example to keep it legible
                return
            }
            state.status = 'idle'
            state.workItem = undefined
        }
```
or with the new notation like
```ts
        finishWork: {
            predicate: fromStatus('working'),
            reducer(state){
                state.status = 'idle'
                state.workItem = undefined
            }
        }
```


The point here is essentially not to create a completely new functionality (for that, we could go WAY better than what I suggested), but to enhance the existing functionality of `createSlice` to make it easier to work with state machines.
That said, of course we might still be adding something like `createStates` in the future, but that doesn't feel like a short-term goal.

markerikson:
I have mixed thoughts on this.

- In one sense, this doesn't provide any new capabilities.  As the examples show, you can already manually check states and make logic exclusive in a case reducer.
- But, having a bit of API syntax for this specific case may make it more apparent and cut down the repetition a bit.
- As @phryneas said, we've already got `createSlice` set up to take just a case reducer function as the default syntax, or `{prepare, reducer}` if you need to customize the generated action creator.  So, the object field syntax exists, we'd just be adding a new possible key in there.
- Switching from a function to an object of functions doesn't really gain anything in terms of saving the couple lines of code needed for `if(someCondition) { ... }`
- I'm on the fence as to whether it actually really adds any real benefit, or if it would be better to leave `createSlice` alone and either add a separate function that's more focused on state machines, or just recommend that folks grab `@xstate/fsm` and be done with it.

phryneas:
It was a nice idea, but this is probably really not necessary. Closing :)

PR #367, opened 2020-02-17
createAsyncThunk improvements

phryneas:
The asyncThunk now exits early when the `abort()` method is called and doesn't keep waiting for the `payloadCreator` to finish.
This should be an improvement especially in cases where the `payloadCreator` did not support the signal, but something outside was awaiting the promise from `dispatch`.
Also, before this, a non-signal-aware `payloadCreator` could still finish which would have caused a dispatch of a "fulfilled" action after an "rejected" (abort) action, which was confusing.

Also I've removed some logic that made sense to me at the moment but really was just confusing and useless.
Also removed the `meta.abortReason` property as it is not possible any more to override the error in the "rejected"(aborted) return value, so that doesn't make sense any more.

### DIFF
--- a/docs/api/createAsyncThunk.md
+++ b/docs/api/createAsyncThunk.md
@@ -133,7 +133,6 @@ interface RejectedAction<ThunkArg> {
     requestId: string
     arg: ThunkArg
     aborted: boolean
-    abortReason?: string
   }
 }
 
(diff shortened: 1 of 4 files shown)


markerikson:
Given that you are much more familiar with how the abort stuff is supposed to work, could you add some kind of a writeup to the `createAsyncThunk` API docs page?  I know you've said docs writing isn't your strength, but if you can at least get in a basic description, I can tweak it if needed.

markerikson:
aaaaand take a wild guess why CI is failing :)

phryneas:
Okay, I've added some docs.

Honestly, I can't think of any real good examples of implementing those myself, but also it's midnight, so not the best time to be creative. Better than nothing for now I guess.

That example code *should*  work, but I haven't run it.

markerikson:
Cool.  I'll merge it in to the feature branch, clean up some of the wording, and put out yet another alpha.

PR #369, opened 2020-02-18
Update Docusaurus and add lockfile to 43 version

lex111:
This should fix the sidebar showing (see https://redux-toolkit.js.org/api/getdefaultmiddleware/)

cc @markerikson


phryneas:
This project is already shipping with a `package-lock.json`, so please do that update with `npm` instead of `yarn` and submit an updated `package-lock.json` instead of `yarn.lock` :)

lex111:
@phryneas oh yes, thanks for the tip, fixed!

markerikson:
Okay, after fixing up a couple issues with our `netlify.toml` file, this _does_ look like it fixes the "no sidebar on direct link" issue:

https://deploy-preview-369--redux-starter-kit-docs.netlify.com/api/getdefaultmiddleware/

Thanks!

markerikson:
PR 352's preview was still behaving wrong after I merged down `master`, but clearing the Netlify cache and redeploying seems to have finally gotten it working:

https://deploy-preview-352--redux-starter-kit-docs.netlify.com/api/getdefaultmiddleware/

- Sidebar on left side even though we went straight to the page URL
- Page TOC on the right side
- Content is the correct width

So, I'm going to claim this is indeed good to go now.

Thanks for poking at this enough to fix things!

PR #374, opened 2020-02-19
v1.3.0 Integration

markerikson:
WIP draft PR to collect all our v1.3.0 changes together, giving us docs preview, bundle size checks, etc.


markerikson:
@phryneas I just saw someone in Discord report that they were getting non-serializable values in an action dispatched by `createAsyncThunk`, in a field like `meta.args.someTimestamp`.  Looks like they're passing a `Date` instance to the payload creator.

This is an interesting issue, because strictly speaking passing non-serializable values to the payload creator _should_ be entirely fine.  It's only a problem because we're attaching those to the dispatched action.

Any thoughts on what the right approach is for dealing with that?  Some possibilities:

- Just document it and say "either don't pass non-serializable args to the payload creator, or explicitly ignore that action type in the serializability check"
- Edit the serializability middleware to ignore `action.meta.args` by default

phryneas:
> @phryneas I just saw someone in Discord report that they were getting non-serializable values in an action dispatched by `createAsyncThunk`, in a field like `meta.args.someTimestamp`. Looks like they're passing a `Date` instance to the payload creator.
> 
> This is an interesting issue, because strictly speaking passing non-serializable values to the payload creator _should_ be entirely fine. It's only a problem because we're attaching those to the dispatched action.
> 
> Any thoughts on what the right approach is for dealing with that? Some possibilities:
> 
>     * Just document it and say "either don't pass non-serializable args to the payload creator, or explicitly ignore that action type in the serializability check"
> 
>     * Edit the serializability middleware to ignore `action.meta.args` by default

One step further: have the serializability middleware ignore all `action.meta` by default? Would be a nice escape hatch to pass non-serializable data to middlewares etc without having to stop action propagation.

markerikson:
Mmm... the serializable middleware should generally be after all others anyway unless you've reordered things, which is why it's okay to dispatch thunk functions in the first place.

But yeah, let's at least specifically exempt `meta.args`, since that's not even something the user is explicitly adding to the action themselves.

markerikson:
3, 2, 1, let's jam

PR #377, opened 2020-02-20
remove `any` where applicable

phryneas:
This should make the types much more resilient. I'm on a train with very flaky internet, so I just added my "review comments" as comments in the code - I can remove those before we merge this ;)

### DIFF
--- a/src/createAction.ts
+++ b/src/createAction.ts
@@ -296,10 +296,6 @@ export function createAction(type: string, prepareAction?: Function): any {
   return actionCreator
 }
 
-/**
- * The previous typings implied some assertion for the types of Payload/Error/Meta, which was just not the case.
- * I'd suggest we just do something like this - assert the shape of a FSA, but make no assumptions about the contents.
- */
 export function isFSA(
   action: unknown
 ): action is {
--- a/src/createAction.ts
+++ b/src/createAction.ts
@@ -296,11 +296,18 @@ export function createAction(type: string, prepareAction?: Function): any {
   return actionCreator
 }
 
-export function isFSA<
-  Payload = undefined,
-  Type extends string = string,
-  Meta = undefined
->(action: any): action is PayloadAction<Payload, Type, Meta> {
+/**
+ * The previous typings implied some assertion for the types of Payload/Error/Meta, which was just not the case.
+ * I'd suggest we just do something like this - assert the shape of a FSA, but make no assumptions about the contents.
+ */
+export function isFSA(
+  action: unknown
+): action is {
+  type: string
+  payload?: unknown
+  error?: unknown
+  meta?: unknown
+} {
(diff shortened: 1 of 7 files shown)


src/entities/models.ts
```diff
@@ -24,9 +24,13 @@ export interface DictionaryNum<T> {
 
 /**
  * @alpha
+ * why was this a class, not an interface?
  */
 export abstract class Dictionary<T> implements DictionaryNum<T> {
-  [id: string]: T | undefined
```
markerikson: FWIW, this definition came straight from `@ngrx/entity`.

I think the intent here is more that the user-level code understands that `booksState.entities["doesNotExist"]` returns `undefined`, vs a `Record<string, Book>` where it doesn't.

So, yeah, the internal casts are a bit weird, but it seems like an okay tradeoff to get the right behavior at the user level?

src/createAction.ts
```diff
@@ -296,11 +296,18 @@ export function createAction(type: string, prepareAction?: Function): any {
...
-export function isFSA<
-  Payload = undefined,
-  Type extends string = string,
-  Meta = undefined
->(action: any): action is PayloadAction<Payload, Type, Meta> {
+/**
+ * The previous typings implied some assertion for the types of Payload/Error/Meta, which was just not the case.
+ * I'd suggest we just do something like this - assert the shape of a FSA, but make no assumptions about the contents.
+ */
+export function isFSA(
```
markerikson: Sure, makes sense to me.

src/entities/models.ts
```diff
@@ -58,71 +57,48 @@ export interface EntityDefinition<T> {
 }
 
 export interface EntityStateAdapter<T> {
-  addOne<S extends EntityState<T>>(state: S, entity: TypeOrPayloadAction<T>): S
```
markerikson: What's the net benefit of dropping `TypeOrPayloadAction` here?

phryneas: You had the function overload only taking the PayloadAction there anyways, so it's either remove that or cut this back to prevent duplication. 

Since I remembered you mentioned that you had some problem with this `TypeOrPayloadAction` and having two overloads with different arguments somehow looked "cleaner" to me anyways, I decided to go the "two overload" way.

src/entities/models.ts
phryneas: Yeah. That's the lifelong discussion in the TS issues if an index signature should actually be undefined per default and it comes down to "if you really want to signal this, add the `| undefined`.
I guess in this place it makes sense, so I'll re-add it.

markerikson: And yeah, I never understood why this was an `abstract class` to begin with either

markerikson:
Looks sweet!

Issue #380, opened 2020-02-21
DevTools Extension package is not being tree-shaken in production


markerikson:
Per #78 , and similar to #363 , the `redux-devtools-extension` package is not written as ES Modules, and is breaking Webpack's ability to optimize it out of bundles.

Per that discussion, the package is itself tiny, and the fact that it results in _all_ of Redux being pulled in is not likely to cause issues in practice because you're almost definitely using Redux.  But, as a general principle, we'd like it to be shakeable.

I did go ahead and file https://github.com/zalmoxisus/redux-devtools-extension/issues/718 to see if we can get it converted to ESM or TS.  However, the package has been irregularly maintained for a while, so I don't expect a fast response.

Given that, I've gone ahead and updated the repo fork I'd previously made at https://github.com/reduxjs/redux-devtools-extension , and am currently attempting to convert the package to TypeScript.  I've successfully published a first attempt on NPM as `@reduxjs/redux-devtools-extension-fork@0.0.1`.  

Following the instructions at https://github.com/reduxjs/redux-toolkit/issues/78#issue-398469913 , I created a minimal project that only imports `createAction` and runs it through Webpack.  With the existing RTK@1.2.5, the output was:

```
      Asset     Size  Chunks                   Chunk Names
    main.js   28 KiB       0  [emitted]        main
main.js.map  154 KiB       0  [emitted] [dev]  main
Entrypoint main = main.js main.js.map
 [7] (webpack)/buildin/global.js 472 bytes {0} [built]
     ModuleConcatenation bailout: Module is not an ECMAScript module
 [8] (webpack)/buildin/harmony-module.js 573 bytes {0} [built]
     ModuleConcatenation bailout: Module is not an ECMAScript module
[14] ./src/index.js + 3 modules 35.1 KiB {0} [built]
     ModuleConcatenation bailout: Cannot concat with ./node_modules/immer/dist/immer.module.js (<- Module uses injected variables (process))     ModuleConcatenation bailout: Cannot concat with ./node_modules/nanoid/index.browser.js (<- Module is not an ECMAScript module)
     ModuleConcatenation bailout: Cannot concat with ./node_modules/redux-devtools-extension/index.js (<- Module is not an ECMAScript module)
     ModuleConcatenation bailout: Cannot concat with ./node_modules/redux/es/redux.js (<- Module is referenced from these modules with unsupported syntax: ./node_modules/redux-devtools-extension/index.js (referenced with cjs require))
[...]

PR #384, opened 2020-02-21
Fork redux-devtools-extension

markerikson:
Per #380 , I've gone ahead and converted the `redux-devtools-extension` package to TS and dropped in the one entry point file that we actually use.  This appears to improve the tree shaking behavior overall.


markerikson:
Huh. Surprised the CJS prod size went up, but we'll see how it works out.

PR #387, opened 2020-02-21
refactor devtoolsExtension

phryneas:
This enhances `window` with the properties `__REDUX_DEVTOOLS_EXTENSION_COMPOSE__` and `__REDUX_DEVTOOLS_EXTENSION__`.

Also, it uses optional chaining & nullish coalescense to make the code itself a little more compact. Will only work once you've updated TS to at least 3.7.


phryneas:
We won't be able to use optional chaining & nullish coalescense for quite some while because that would kill our tests. Closing this :)

Issue #390, opened 2020-02-24
Accessing the original error in createAsyncThunk


msutkowski:
With the current implementation of `createAsyncThunk`, there is no way to access any errors returned from a server. Here is a specific scenario which I imagine is extremely common:

You attempt a form submission with a set of data like `POST /users` with a body of `{ first_name: 'valid', last_name: 'valid', email: 'in@valid' }` and the API responds with a 400:

```
{
    "details": [
        {
            "message": "\"email\" must be a valid email",
            "path": [
                "email"
            ],
            "type": "string.email",
            "context": {
                "value": "in@valid",
                "key": "email",
                "label": "email"
            }
        }
    ],
    "source": "payload",
    "meta": {
        "id": "1582509238404:1c50114e-ef8e-4d97-b3e4-dbc995e8fcff:20:k6zpbe70:10030",
        "api": "1.0.0",
        "env": "production"
    }
}
```

In this case, I need some way to access that error so that I could update my UI - for example, by unwrapping the `details` field and using Formik's `setErrors` to provide feedback. I imagine it'd also be valuable for detailed toasts as well, where people `useEffect` and pop an alert if the error message changes.

After looking through `createAsyncThunk`, the only real solution I can think of is
 - Passing a new parameter for `rethrow`  and rethrowing the error instead of returning `finalAction`.

If we were able to rethrow, we could do this sort of thing:

```js
<Form 
   onSubmit={async (values, helpers) => {
          try {
            await dispatch(myAsyncThunk(values));
          } catch (err) {
            const joiErrors = err.response.details;
            if (joiErrors) {
              helpers.setErrors(unwrapErrors(joiErrors));
            }
          }
        }}
      />

```

markerikson:
Per DMs, my inclination at this point is that if you need to reformat _anything_ from the server, it's your job to do it in the payload creator.  

Axios requests where you need to grab something out of `response.data`, normalization of entity data, extracting important values from a server error... we don't know anything about that ourselves.  Do it yourself in the payload creator, and whatever you return or throw will end up in the corresponding action.

markerikson:
Hmm.  After further conversation, I'm still not convinced that "rethrowing" is the right idea.  However, if the payload creator extracts some kind of an object from the error and tries to rethrow that, `miniSerializeError()` will assume it's actually an `Error` due to `typeof error === 'object'` and tried to read out the standard fields, vs just passing through the thrown object to the rejected action.  That seems insufficient.

[...]

PR #391, opened 2020-02-24
Add rethrow flag to createAsyncThunk

msutkowski:
Addresses comments in #390. Usage would look like:

Sample action creator:
```js
export const genericAxiosRequestThunk = createAsyncThunk(
  "app/dataFetched",
  async ({ httpMethod, path, baseUrl, data }) => {
    try {
      const result = await axios.request({
        method: httpMethod,
        url: `${baseUrl}/${path}`,
        data
      });
    return result.data;
  }
);
```

Usage in component:
```js
    try {
      await dispatch(
        genericAxiosRequestThunk(
          {
            httpMethod: "POST",
            baseUrl: "https://mycoolapi.com/v1",
            path: "users",
            data: {
              email: "bademail",
              password: "short",
              first_name: "John",
              last_name: "Smith",
            }
          },
          true
        )
      );
    } catch (err) {
      console.log("Original error in component", err.response.data);
    }
```

Notes: I don't know if I really care for this being an extra argument. Would it make more sense to check in the args?

### DIFF
--- a/etc/redux-toolkit.api.md
+++ b/etc/redux-toolkit.api.md
@@ -103,7 +103,7 @@ export function createAction<P = void, T extends string = string>(type: T): Payl
 export function createAction<PA extends PrepareAction<any>, T extends string = string>(type: T, prepareAction: PA): PayloadActionCreator<ReturnType<PA>['payload'], T, PA>;
 
 // @alpha (undocumented)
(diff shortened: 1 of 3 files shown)


msutkowski:
Just some additional comments on this:

> if you need to do more formatting, catch the error in the payload creator and re-throw with just the data you care about. in the same way that if you want to extract the data for the payload, you do that in the payload creator too

This idea is good but _currently_ doesn't solve the given use case. The way that `miniSerializeError` currently works is limited in regards to handling more complex error types. While rethrow might not be the 'best option' per se, it at least gives the control of handling 100% back to the user. No matter what we do in the current situation, we're still going to have trouble handling errors and `unwrapResult`ing them right now assuming anyone ever wants anything more than a basic HTTP error messages.

Will try to come up with some better ideas tomorrow.

msutkowski:
Closing in favor of #393

PR #393, opened 2020-02-25
RFC createAsyncThunk: reject with typed value

phryneas:
Starting from #390, @msutkowski and I have been discussing this further over discord.

This would be one possible way to reject from a `payloadCreator`, but pass typed data (which would be impossible using `throw`, as that cannot be typed):
Somewhere within `payloadCreator`, an object passed through `rejectWithValue` is returned. That object then ends up as `payload` of a `rejected` action that is finally dispatched & returned by the async thunk.

See

https://github.com/phryneas/redux-toolkit/blob/44a550193646d309931405a107d66ea350f70f43/type-tests/files/createAsyncThunk.typetest.ts#L104-L123

This PR is essentially a starting point so @msutkowski can experiment with it in his example app.

### DIFF
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -11,12 +11,18 @@ import { FallbackIfUnknown } from './tsHelpers'
 // @ts-ignore we need the import of these types due to a bundling issue.
 type _Keep = PayloadAction | ActionCreatorWithPreparedPayload<any, unknown>
 
-export type BaseThunkAPI<S, E, D extends Dispatch = Dispatch> = {
+export type BaseThunkAPI<
+  S,
+  E,
+  D extends Dispatch = Dispatch,
+  RejectedValue = undefined
+> = {
   dispatch: D
   getState: () => S
   extra: E
   requestId: string
   signal: AbortSignal
+  rejectWithValue(value: RejectedValue): RejectWithValue<RejectedValue>
 }
 
 /**
@@ -29,15 +35,19 @@ export interface SerializedError {
   code?: string
 }
 
-const commonProperties: (keyof SerializedError)[] = [
+const commonProperties: Array<keyof SerializedError> = [
   'name',
   'message',
   'stack',
   'code'
 ]
 
+class RejectWithValue<RejectValue> {
+  constructor(public readonly value: RejectValue) {}
+}
+
 // Reworked from https://github.com/sindresorhus/serialize-error
-export const miniSerializeError = (value: any): any => {
+export const miniSerializeError = (value: any): SerializedError => {
   if (typeof value === 'object' && value !== null) {
     const simpleError: SerializedError = {}
     for (const property of commonProperties) {
@@ -49,13 +59,14 @@ export const miniSerializeError = (value: any): any => {
     return simpleError
   }
 
-  return value
+  return { message: String(value) }
 }
 
 type AsyncThunkConfig = {
   state?: unknown
   dispatch?: Dispatch
   extra?: unknown
+  rejectValue?: unknown
 }
 
 type GetState<ThunkApiConfig> = ThunkApiConfig extends {
@@ -82,8 +93,15 @@ type GetDispatch<ThunkApiConfig> = ThunkApiConfig extends {
 type GetThunkAPI<ThunkApiConfig> = BaseThunkAPI<
   GetState<ThunkApiConfig>,
   GetExtra<ThunkApiConfig>,
-  GetDispatch<ThunkApiConfig>
+  GetDispatch<ThunkApiConfig>,
+  GetRejectValue<ThunkApiConfig>
 >
+
+type GetRejectValue<ThunkApiConfig> = ThunkApiConfig extends {
+  rejectValue: infer RejectValue
+}
+  ? RejectValue
+  : unknown
 
 /**
  *
@@ -101,8 +119,13 @@ export function createAsyncThunk<
   payloadCreator: (
     arg: ThunkArg,
     thunkAPI: GetThunkAPI<ThunkApiConfig>
-  ) => Promise<Returned> | Returned
+  ) =>
+    | Promise<Returned | RejectWithValue<GetRejectValue<ThunkApiConfig>>>
+    | Returned
+    | RejectWithValue<GetRejectValue<ThunkApiConfig>>
 ) {
+  type RejectedValue = GetRejectValue<ThunkApiConfig>
+
   const fulfilled = createAction(
     type + '/fulfilled',
     (result: Returned, requestId: string, arg: ThunkArg) => {
@@ -125,11 +148,16 @@ export function createAsyncThunk<
 
   const rejected = createAction(
     type + '/rejected',
-    (error: Error, requestId: string, arg: ThunkArg) => {
-      const aborted = error && error.name === 'AbortError'
+    (
+      error: Error | null,
+      requestId: string,
+      arg: ThunkArg,
+      payload?: RejectedValue
+    ) => {
+      const aborted = !!error && error.name === 'AbortError'
       return {
-        payload: undefined,
-        error: miniSerializeError(error),
+        payload,
+        error: miniSerializeError(error) || { message: 'Rejected' },
         meta: {
           arg,
           requestId,
@@ -173,9 +201,17 @@ export function createAsyncThunk<
                 getState,
                 extra,
                 requestId,
-                signal: abortController.signal
+                signal: abortController.signal,
+                rejectWithValue(value: RejectedValue) {
+                  return new RejectWithValue(value)
+                }
               })
-            ).then(result => fulfilled(result, requestId, arg))
+            ).then(result => {
+              if (result instanceof RejectWithValue) {
+                return rejected(null, requestId, arg, result.value)
+              }
+              return fulfilled(result, requestId, arg)
+            })
           ])
         } catch (err) {
           finalAction = rejected(err, requestId, arg)
@@ -199,14 +235,23 @@ export function createAsyncThunk<
   })
 }
 
+type ActionTypesWithOptionalErrorAction =
+  | { error: any }
+  | { error?: never; payload: any }
+type PayloadForActionTypesExcludingErrorActions<T> = T extends { error: any }
+  ? never
+  : T extends { payload: infer P }
+  ? P
+  : never
+
 /**
  * @alpha
  */
-export function unwrapResult<T>(
-  returned: { error: any } | { payload: NonNullable<T> }
-): NonNullable<T> {
+export function unwrapResult<R extends ActionTypesWithOptionalErrorAction>(
+  returned: R
+): PayloadForActionTypesExcludingErrorActions<R> {
   if ('error' in returned) {
     throw returned.error
   }
-  return returned.payload
+  return (returned as any).payload
 }
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -167,6 +167,34 @@ export function createAsyncThunk<
     }
   )
 
+  let displayedWarning = false
+
+  const AC =
+    typeof AbortController !== 'undefined'
+      ? AbortController
+      : class implements AbortController {
+          signal: AbortSignal = {
+            aborted: false,
+            addEventListener() {},
+            dispatchEvent() {
+              return false
+            },
+            onabort() {},
+            removeEventListener() {}
+          }
+          abort() {
+            if (process.env.NODE_ENV === 'development') {
+              if (!displayedWarning) {
+                displayedWarning = true
+                console.info(
+                  `This platform does not implement AbortController. 
+If you want to use the AbortController to react to \`abort\` events, please consider importing a polyfill like 'abortcontroller-polyfill/dist/abortcontroller-polyfill-only'.`
+                )
+              }
+            }
+          }
+        }
+
   function actionCreator(arg: ThunkArg) {
     return (
       dispatch: GetDispatch<ThunkApiConfig>,
@@ -175,7 +203,7 @@ export function createAsyncThunk<
     ) => {
       const requestId = nanoid()
 
-      const abortController = new AbortController()
+      const abortController = new AC()
       let abortReason: string | undefined
 
       const abortedPromise = new Promise<never>((_, reject) =>
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -157,7 +157,7 @@ export function createAsyncThunk<
       const aborted = !!error && error.name === 'AbortError'
       return {
         payload,
-        error: miniSerializeError(error) || { message: 'Rejected' },
+        error: miniSerializeError(error || 'Rejected'),
         meta: {
           arg,
           requestId,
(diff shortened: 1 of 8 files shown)


markerikson:
Cool, glad you two have been busy while I've been at work :)

phryneas:
Apparently api-extractor doesn't like that I added a method on createAsyncThunk. To be honest, I'm not even sure if I like that method on there, so I'll leave it complaining for the moment. 
Feedback on that is very welcome.

phryneas:
Okay, moved it on the `thunkApi` parameter. I think that fits a lot better because it doesn't need an extra import and also comes pre-typed correctly.

markerikson:
Yep, makes sense to me.  (And this is why I pushed to make it an options object instead of positional args.  I didn't know we'd add this specific thing, but as soon as we started looking at more than just `(dispatch, getState)`, it made sense to bundle them.)

msutkowski:
Sorry about the delay in providing feedback here. 

### General feedback
These changes are great and very straightforward to work with. I don't see any problems with how the implementation is done, but I'm running into type issues. I'll do my best to outline the things I've found.

#### What I'm trying to achieve with `createAsyncThunk`
1. Handle the success and set the payload
2. On a known/expected error type I want to return it via `rejectWithValue`
3. On an unknown error, I'd like to throw it and just let `miniSerializeError` wrap it up


#### Issue when trying try/catching
When defining `get<CallsResponse>`, the type signature on `return data` is `Call[]`, matching the return type specificed `Call[]`. 
```ts
export const fetchLiveCallsError = createAsyncThunk<
  Call[],
  string,
  { state: RootState; rejectValue: RejectedErrorPayload<ValidationErrorsResponse> }
>("calls/fetchLiveCalls", async (organizationId, { rejectWithValue }) => {
  try {
    const {
      data: { data }
    } = await apiRequest.get<CallsResponse>(`organizations/${organizationId}/calls/live/iwill404`);
    return data;
  } catch (err) {
    let error: AxiosError = err; // cast for access to AxiosError properties
    return rejectWithValue({
      error: "just a test message",
      data: error.response?.data,
    });
  }
});
```
It throws this error (slightly cleaned up):
```  Type 'Promise<Call[] | RejectWithValue<RejectedErrorPayload<ValidationErrorsResponse>>>' is not assignable to type 'Call[] | Promise<Call[]> | RejectWithValue<RejectedErrorPayload<ValidationErrorsResponse>> | Promise<...>'.
    Type 'Promise<Call[] | RejectWithValue<RejectedErrorPayload<ValidationErrorsResponse>>>' is not assignable to type 'Promise<Call[]>'.
      Type 'Call[] | RejectWithValue<RejectedErrorPayload<ValidationErrorsResponse>>' is not assignable to type 'Call[]'.
        Type 'RejectWithValue<RejectedErrorPayload<ValidationErrorsResponse>>' is missing the following properties from type 'Call[]': length, pop, push, concat, and 28 more.
```

If I pull the Axios type of `CallsResponse` out, data is set to `any` and there are no errors but that doesn't seem like the right answer. 

#### Is it possible to add a type guard or something along the lines of this (`isRejectedValue`):

```ts
builder.addCase(fetchLiveCallsError.rejected, (state, action) => {
      // is something like this a good idea to export from the lib that would tell TS the rejected payload is correct and drop the | undefined-related errors
      const isRejectedValue = (check: typeof action) => check.error === 'Rejected'; 
     
      if (isRejectedValue(action)) {
        // this was rejected via rejectWithValue, handle it
        state.error = action.payload?.error;
      } else {
        /** 
         * was actually thrown and would have a mini serializable error format
         * action.error.name, action.error.stack, action.error.code, action.error.message should be available
         * but appear to be getting dropped
         * */
      }
    })
```

#### A full example with all handling methods
Per point 3, I imagine a 'complete' handling scenario would look like the below. Does this seem right? If so, how can we get the point after for component usage to be aware of the difference between the success result, rejected with value result, and the general rejected data?

```ts
export const fetchLiveCallsError = createAsyncThunk<
  Call[],
  string,
  { state: RootState; rejectValue: RejectedErrorPayload<ValidationErrorsResponse> }
>("calls/fetchLiveCalls", async (organizationId, { rejectWithValue }) => {
  try {
    const {
      data: { data }
    } = await apiRequest.get<CallsResponse>(`organizations/${organizationId}/calls/live/iwill404`);
    return data;
  } catch (err) {
    let error: AxiosError = err;
    // This is an expected error with validation errors - format + reject it
    if (error.response?.data?.details) {
      return rejectWithValue({
        error: "just a test message",
        data: error.response?.data,
      });
    }
    // just throw the unexpected error?
    throw err;
  }
});
```


#### Leveraging the resulting rejected/fulfilled values in a component
A common request is to be able to access the action like this:

```ts
<Button onClick={async() => {
              const resultAction = await dispatch(fetchLiveCallsError(organizationId)); // type is PayloadAction and resultAction.error never exists
              if  (resultAction.error === 'Rejected') {
                //  set form errors from the info that was passed into the rejected payload, etc etc.
              }
            }}>Perform call</Button>```

In this instance, someone could check for truthiness and cast it _(or if we can do the type guard idea from above, great)_, but is the ideal usage? Either way, it would appear that type union returned isn't accurate.

**Comment by phryneas** [comment]
> #### Issue when trying try/catching

...

> It throws this error (slightly cleaned up):
> 
> ```
>     Type 'Promise<Call[] | RejectWithValue<RejectedErrorPayload<ValidationErrorsResponse>>>' is not assignable to type 'Promise<Call[]>'.
>       Type 'Call[] | RejectWithValue<RejectedErrorPayload<ValidationErrorsResponse>>' is not assignable to type 'Call[]'.
>         Type 'RejectWithValue<RejectedErrorPayload<ValidationErrorsResponse>>' is missing the following properties from type 'Call[]': length, pop, push, concat, and 28 more.
> ```

This should be fixed, thanks for noticing!

> If I pull the Axios type of `CallsResponse` out, data is set to `any` and there are no errors but that doesn't seem like the right answer.
> #### Is it possible to add a type guard or something along the lines of this (`isRejectedValue`):

You are setting
```ts
let error: AxiosError = err // cast for access to AxiosError properties
```
which defaults to `AxiosError<any>`, so in the rejected value, data will result in `any`.

You can use `AxiosError<MyApiErrorType>` instead.

> 
> ```ts
> builder.addCase(fetchLiveCallsError.rejected, (state, action) => {
>       // is something like this a good idea to export from the lib that would tell TS the rejected payload is correct and drop the | undefined-related errors
>       const isRejectedValue = (check: typeof action) => check.error === 'Rejected'; 
>      
>       if (isRejectedValue(action)) {
>         // this was rejected via rejectWithValue, handle it
>         state.error = action.payload?.error;
>       } else {
>         /** 
>          * was actually thrown and would have a mini serializable error format
>          * action.error.name, action.error.stack, action.error.code, action.error.message should be available
>          * but appear to be getting dropped
>          * */
>       }
>     })
> ```

I'd do this the other way around: instead of `if (isRejectedValue(action))`, just to `if (action.payload)`, because the existence of a payload is exactly what separates a rejection with value from one without.


For the "mini serializable error format": yes, but now. Currently, `miniSerializeError` returns `any`, because if you put a non-object like a string into it, it will return exactly that and not a `SerializedError`. This might be unwanted behaviour though. @markerikson what do you think?

> 
> #### A full example with all handling methods
> 
> Per point 3, I imagine a 'complete' handling scenario would look like the below. Does this seem right? If so, how can we get the point after for component usage to be aware of the difference between the success result, rejected with value result, and the general rejected data?
> 
> ```ts
> export const fetchLiveCallsError = createAsyncThunk<
>   Call[],
>   string,
>   { state: RootState; rejectValue: RejectedErrorPayload<ValidationErrorsResponse> }
> >("calls/fetchLiveCalls", async (organizationId, { rejectWithValue }) => {
>   try {
>     const {
>       data: { data }
>     } = await apiRequest.get<CallsResponse>(`organizations/${organizationId}/calls/live/iwill404`);
>     return data;
>   } catch (err) {
>     let error: AxiosError = err;
>     // This is an expected error with validation errors - format + reject it
>     if (error.response?.data?.details) {
>       return rejectWithValue({
>         error: "just a test message",
>         data: error.response?.data,
>       });
>     }
>     // just throw the unexpected error?
>     throw err;
>   }
> });
> ```
> 
> #### Leveraging the resulting rejected/fulfilled values in a component
> 
> A common request is to be able to access the action like this:
> 
> ```ts
> <Button onClick={async() => {
>               const resultAction = await dispatch(fetchLiveCallsError(organizationId)); // type is PayloadAction and resultAction.error never exists
>               if  (resultAction.error === 'Rejected') {
>                 //  set form errors from the info that was passed into the rejected payload, etc etc.
>               }
>             }}>Perform call</Button>```
> 
> In this instance, someone could check for truthiness and cast it _(or if we can do the type guard idea from above, great)_, but is the ideal usage? Either way, it would appear that type union returned isn't accurate.
> ```

I'd probably do something along the lines of

```ts
 if (fetchLiveCallsError.fulfilled.match(result)) {
      //success
      expectType<ReturnType<typeof fetchLiveCallsError['fulfilled']>>(result)
      expectType<Call[]>(result.payload)
    } else {
      expectType<ReturnType<typeof fetchLiveCallsError['rejected']>>(result)
      if (result.payload) {
        // rejected with value
        expectType<RejectedErrorPayload<ValidationErrorsResponse>>(
          result.payload
        )
      } else {
        // rejected by throw
        expectType<undefined>(result.payload)
        // result.error is `any`, because `miniSerializeError` currently returns any (if you pass in a non-object, it returns it and not a SerializedError)
        expectType<IsAny<typeof result['error'], true, false>>(true)
      }
    }
```

msutkowski:
Thanks, this all works as expected now. The component usage is also straightforward as you explained and usage is really simple.

markerikson:
Sorry, what's the question / suggestion around serialized errors?

phryneas:
> Sorry, what's the question / suggestion around serialized errors?

Sorry, my bad.

Currently, `reject` returns the result of `miniSerializeError` on it's `error` property. That is not a `SerializedError` though, but `any`, as `miniSerializeError` in the current form does not always return a `SerializedError`. It does so, if the passed argument is an `object`, but otherwise it just returns it's input.
So my question is: is this intentional and do we want to leave it that way? Or do we want to make a change alike this, to make the shape of `error` more predictable

```diff
-export const miniSerializeError = (value: any): any => {
+export const miniSerializeError = (value: any): SerializedError => {
  if (typeof value === 'object' && value !== null) {
    // ...
    return simpleError
  }

-  return value
+  return { message: String(value) }
}
```

markerikson:
If you feel that would be more consistent, sure, we can go with that.

phryneas:
So I'm not really sure, are the CI fails just from immer not working with 3.5/3.6 right now?
Other than that I *think* this should be good go go.
Note: I did add `axios` as a `devDependency` because I used their types in a type-test. Could do another test, but it gives it kind of a nice "real-life" spin. Could we keep that?

markerikson:
Yeah, I expect TS 3.5 and 3.6 to fail right now due to the Immer alpha.

Can you update the docs as well?

src/createAsyncThunk.ts
```diff
@@ -173,9 +201,17 @@ export function createAsyncThunk<
                 getState,
                 extra,
                 requestId,
-                signal: abortController.signal
```
markerikson: Can you add that notional `AbortController` pseudo-polyfill in this PR as well?

phryneas: Gonna do that in an extra PR in the next few days, I think this has been blocked long enough :)

phryneas:
I just rewrote the history of this, so we can merge this without squashing to keep @msutkowski's contributions under his name.

phryneas:
And we have a mocked AbortController. It looks good from the tests, but I have no IE at hand, so I can only test this in theory. Anyways, I guess we'll get that feedback from some alpha testers ^^.

phryneas:
The AbortController seems to cost us quite a bit though.

Stats before last commit:



| Filename | Size | Change | |
|:--- |:---:|:---:|:---:|
| `dist/redux-toolkit.cjs.development.js` | 9.69 kB | +92 B (0%) |  |
| `dist/redux-toolkit.cjs.production.min.js` | 3.99 kB | +69 B (1%) |  |
| `dist/redux-toolkit.esm.js` | 9.6 kB | +99 B (1%) |  |
| `dist/redux-toolkit.umd.js` | 20.9 kB | +94 B (0%) |  |
| `dist/redux-toolkit.umd.min.js` | 9.49 kB | +65 B (0%) |  |

Stats after:


| Filename | Size | Change | |
|:--- |:---:|:---:|:---:|
| `dist/redux-toolkit.cjs.development.js` | 9.97 kB | +367 B (3%) |  |
| `dist/redux-toolkit.cjs.production.min.js` | 4.18 kB | +256 B (6%) | 🔍 |
| `dist/redux-toolkit.esm.js` | 9.88 kB | +381 B (3%) |  |
| `dist/redux-toolkit.umd.js` | 21.1 kB | +369 B (1%) |  |
| `dist/redux-toolkit.umd.min.js` | 9.67 kB | +246 B (2%) |  |

markerikson:
Hmm. Yeah. I wonder if it's partly because of the class being transpiled. Any way to mimic that as a plain object / function that might end up being smaller?

phryneas:
I think I just forgot to remove the comment around my NODE_ENV check... checking it right now.

phryneas:
Yeah, now i'ts looking better:



| Filename | Size | Change | |
|:--- |:---:|:---:|:---:|
| `dist/redux-toolkit.cjs.development.js` | 9.95 kB | +351 B (3%) |  |
| `dist/redux-toolkit.cjs.production.min.js` | 4.05 kB | +126 B (3%) |  |
| `dist/redux-toolkit.esm.js` | 9.87 kB | +371 B (3%) |  |
| `dist/redux-toolkit.umd.js` | 21.1 kB | +351 B (1%) |  |
| `dist/redux-toolkit.umd.min.js` | 9.55 kB | +127 B (1%) |  |



Oh, by the way: the error message from the nonSerializableMiddleware seems to take up ~0.5k - maybe we should shorten that in the prod build?

markerikson:
Both the immutable and serializable middleware are stripped out in prod entirely, so they'll never show up in the actual prod bundle.

phryneas:
Nope, this is from `redux-toolkit.cjs.production.min.js`:

```
a&&console.error("A non-serializable value was detected in an action, in the path: `"+a.keyPath+"`. Value:",a.value,"\nTake a look at the logic that dispatched this action: ",i,"\n(See https://redux.js.org/faq/actions#why-should-type-be-a-string-or-at-least-serializable-why-should-my-action-types-be-constants)");var c=t(i),s=l(e.getState(),[],r,n,u);return s&&console.error("A non-serializable value was detected in the state, in the path: `"+s.keyPath+"`. Value:",s.value,"\nTake a look at the reducer(s) handling this action type: "+i.type+".\n(See https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state)"
```

markerikson:
Oh dear.  That shouldn't be in there.

Tell you what.  Let's get this PR merged in now.  I'm at work and theoretically supposed to be doing something productive atm.  Can you take a look at why that error is even ending up in the prod build?

phryneas:
Can do :)

PR #396, opened 2020-02-27
Use Immer 6 alpha

markerikson:
Update to Immer 6 alpha for better bundle sizes, per https://github.com/immerjs/immer/pull/536


markerikson:
Travis is failing because Immer 6 alpha currently only supports TS 3.7+.  Immer PR https://github.com/immerjs/immer/pull/541 should fix that once it's merged in and let us keep supporting TS 3.5.  But, I want to go ahead and put out an alpha with this update anyway.

PR #407, opened 2020-03-04
bump immer to v6

phryneas:


markerikson:
Weird. Travis failed for TS 3.5/3.6, because of a whole bunch of failures to find "immer".  

Y'know what, it's not even worth trying to debug that for now.  Let's just merge this.

PR #411, opened 2020-03-05
fix(docs): typo in createEntityAdapter example

jsonberry:
Not sure if this is 100% a typo but it looks like it.

Thanks for the hard work on this!


jsonberry:
Did some further reading, like like `reducer` might actually be available :( sorry was just reading through and thought I caught something.

Gonna close for now since it does look like `createSlice` does have `reducer` as an option.

Issue #416, opened 2020-03-10
createAsyncThunk docs imply payloadCreator executes before pending action


dlwalsh:
I understand `createAsyncThunk` executes in the following order:
1. Pending action
2. Payload creator
3. Fulfilled/rejected action

However this [code snippet](https://deploy-preview-374--redux-starter-kit-docs.netlify.com/api/createAsyncThunk#examples) implies something different:

```
const fetchUserById = createAsyncThunk(
  'users/fetchByIdStatus',
  async (userId, { getState }) => {
    const { loading } = getState().users
    if (loading !== 'idle') {
      return
    }
    const response = await userAPI.fetchById(userId)
    return response.data
  }
)
```

The `loading` flag is expected to be in the `idle` state. But by the time the payload creator is executed, the `loading` flag will have been changed to `pending`.

```
    [fetchUserById.pending]: (state, action) => {
      if (state.loading === 'idle') {
        state.loading = 'pending'
      }
    },
```

Is this a bug in the example code?

markerikson:
Hmm.  Yeah, that does seem like a bug in the snippet.  The hazards of trying to write snippets that aren't necessarily executable :)

Can you file a PR against the `v1.3.0-integration` branch to change it to something more appropriate?

dlwalsh:
Resolved by https://github.com/reduxjs/redux-toolkit/pull/417

PR #417, opened 2020-03-10
Fix bug in docs for createAsyncThunk example

dlwalsh:
Resolves https://github.com/reduxjs/redux-toolkit/issues/416

### DIFF
--- a/docs/api/createAsyncThunk.md
+++ b/docs/api/createAsyncThunk.md
@@ -325,9 +325,9 @@ import { userAPI } from './userAPI'
 
 const fetchUserById = createAsyncThunk(
   'users/fetchByIdStatus',
-  async (userId, { getState }) => {
-    const { loading } = getState().users
-    if (loading !== 'idle') {
+  async (userId, { getState, requestId }) => {
+    const { currentRequestId, loading } = getState().users
+    if (loading !== 'pending' || requestId !== currentRequestId) {
       return
     }
     const response = await userAPI.fetchById(userId)
@@ -340,6 +340,7 @@ const usersSlice = createSlice({
   initialState: {
     entities: [],
     loading: 'idle',
+    currentRequestId: undefined,
     error: null
   },
   reducers: {},
@@ -347,18 +348,23 @@ const usersSlice = createSlice({
     [fetchUserById.pending]: (state, action) => {
       if (state.loading === 'idle') {
         state.loading = 'pending'
+        state.currentRequestId = action.meta.requestId
       }
     },
     [fetchUserById.fulfilled]: (state, action) => {
-      if (state.loading === 'pending') {
+      const { requestId } = action.meta
+      if (state.loading === 'pending' && state.currentRequestId === requestId) {
         state.loading = 'idle'
-        state.push(action.payload)
+        state.entities.push(action.payload)
+        state.currentRequestId = undefined
       }
     },
     [fetchUserById.rejected]: (state, action) => {
-      if (state.loading === 'pending') {
+      const { requestId } = action.meta
+      if (state.loading === 'pending' && state.currentRequestId === requestId) {
         state.loading = 'idle'
         state.error = action.error
+        state.currentRequestId = undefined
       }
     }
   }


markerikson:
Hmm.  Now that I look at this, I feel like this needs a bit more thought put into it.

How _would_ you distinguish between "this specific thunk has dispatched the `pending` action and is now doing the payload work", and "some other request is already in flight"?

@msutkowski , @phryneas - any thoughts on this?

dlwalsh:
I guess you'd need to store the `requestId` and check against that.

msutkowski:
@markerikson With the way `createAsyncThunk` currently works, the user would have to make a simple wrapper that tracked the promise resolution state and use it like: 
`const myTrackedPromise = promiseState(dispatch(myLongRunningAsyncThunk()))`

If this seems like something we'd want to handle internally, a quick and low-effort solution would be to expose a `getStatus` or `isProcessing` function on the promise and have it return the internal state.

```js
const everlongPromise = dispatch(myLongRunningThunk());
// ... do a lot of processing and such?
if (promise.getStatus() === 'processing') {
  // we know we're processing still...
}
```
or

```js
const everlongPromise = dispatch(myLongRunningThunk());
// ... do a lot of processing and such?
if (promise.isProcessing()) {
  // we know we're processing still...
}
```

Would someone actually want more of the internals like `requestId`, or just to know if we're still in the middle of processing?

msutkowski:
Here is a basic sample of that idea: https://github.com/reduxjs/redux-toolkit/pull/418. Switched the naming from my example above to match the standardized naming.

phryneas:
I also think we should just go with the `requestId`. 
Also, I don't know if the `payloadCreator` itself should generally check that it should be running. It might make sense in this case, but that is a very edgy case and I wouldn't really use that as an example then.
Generally: It has been started and not been cancelled, so it runs. I would assume that it's only the job of the reducer to ignore (or not) actions sent from the thunk.

So this would be my attempt:

```ts
import { createSlice, createAsyncThunk, SerializedError } from "@reduxjs/toolkit";

function fetchById(id: string) {
  return new Promise<{ data: string }>(resolve => setTimeout(resolve, 1000, { data: "test" }));
}

const fetchUserById = createAsyncThunk("users/fetchByIdStatus", async (userId: string) => {
  if (getState().slice.requestId !== requestId){
    // if you really do want to do something like this, I'd do it this way
    return
  }
  const response = await fetchById(userId);
  return response.data;
});

const usersSlice = createSlice({
  name: "users",
  initialState: {
    currentRequestId: undefined as undefined|string,
    entities: [] as string[],
    loading: "idle",
    error: null as null | SerializedError
  },
  reducers: {},
  extraReducers: builder =>
    builder
      .addCase(fetchUserById.pending, (state, action) => {
        if (state.loading === "idle") {
          state.loading = "pending";
          state.currentRequestId = action.meta.requestId
        }
      })
      .addCase(fetchUserById.fulfilled, (state, action) => {
        if (state.currentRequestId === action.meta.requestId) {
          state.loading = "idle";
          state.entities.push(action.payload);
        }
      })
      .addCase(fetchUserById.rejected, (state, action) => {
        if (state.currentRequestId === action.meta.requestId) {
          state.loading = "idle";
          state.error = action.error;
        }
      })
});
```

dlwalsh:
Updated PR to implement the pattern above.

docs/api/createAsyncThunk.md
```diff
@@ -340,25 +340,29 @@ const usersSlice = createSlice({
...
   reducers: {},
   extraReducers: {
     [fetchUserById.pending]: (state, action) => {
       if (state.loading === 'idle') {
         state.loading = 'pending'
+        state.currentRequestId = action.meta.requestId
       }
     },
     [fetchUserById.fulfilled]: (state, action) => {
       if (state.loading === 'pending') {
```
phryneas: Doing a `(state.currentRequestId === action.meta.requestId)` test here is much more important than looking for the correct loading state.

Even if the asyncThunk exits early, it will dispatch a pending action and a rejected or resolved action. So you'd run into this:


* state idle, currentRequestId undefined, entities []
* users/fetchByIdStatus/pending (requestId 1)
* state pending, currentRequestId 1
* users/fetchByIdStatus/pending (requestId 2)
* action is ignored
* payloadCreator for request 2 exits early
* users/fetchByIdStatus/resolved (requestId 2)
* state idle, currentRequestId undefined, entities [undefined]
* users/fetchByIdStatus/resolved (requestId 1)
* action is ignored

docs/api/createAsyncThunk.md
```diff
@@ -340,25 +340,29 @@ const usersSlice = createSlice({
...
         state.loading = 'idle'
         state.push(action.payload)
```
phryneas:
```suggestion
        state.entities.push(action.payload)
```

dlwalsh:
Thanks for the feedback @phryneas. I've updated the code snippet accordingly.

You've also clarified my own confusion around the `createAsyncThunk` API. I suspect the fact that the fulfilled action still fires on early exit will be surprising to some. It was to me.

phryneas:
> Thanks for the feedback @phryneas. I've updated the code snippet accordingly.
> 
> You've also clarified my own confusion around the `createAsyncThunk` API. I suspect the fact that the fulfilled action still fires on early exit will be surprising to some. It was to me.

A return statement is a return statement and will trigger a `fulfilled` action. 
Maybe `throw`ing something in that example might be better - then it'd be a `rejected` action. 
But something will definitely happen - it also does when cancelling the thunk by calling `.abort()` on it - it's always guaranteed to have a `pending` action followed by either an `fulfilled` or `rejected` action.

PR #419, opened 2020-03-10
Add "Create a React Redux App" to quick start

nickmccurdy:
See reduxjs/redux#3726

### DIFF
--- a/docs/introduction/quick-start.md
+++ b/docs/introduction/quick-start.md
@@ -48,6 +48,16 @@ yarn add @reduxjs/toolkit
 It is also available as a precompiled UMD package that defines a `window.RTK` global variable.
 The UMD package can be used as a [`<script>` tag](https://unpkg.com/@reduxjs/toolkit/dist/redux-toolkit.umd.js) directly.
 
+## Create a React Redux App
+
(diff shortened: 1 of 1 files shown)


phryneas:
Could we add something like "a TS version is currently in the making"? 
I'd love to emphasize TS as being a first-class citizen as much as possible.

nickmccurdy:
This is already live on the Redux docs and the TS version is almost done, so I figured it would be best to mention the JS version and then link it to the TS version once that's done

phryneas:
That's fair :)

nickmccurdy:
@markerikson @timdorr The site hasn't updated, is this an issue with Docusaurus' versioned docs (see discussion in https://github.com/reduxjs/react-redux/pull/1543#issuecomment-597119087)?

markerikson:
Hmm.  No, we don't use versioning for RTK.  Also, I don't see a PR preview here.  Odd.

markerikson:
Netlify says:

```
2:56:28 PM: Failing build: Failed to build site
2:56:28 PM: failed during stage 'building site': Deploy directory 'website/website/build' does not exist
2:56:28 PM: Finished processing build request in 1m5.805041136s
```

May be a path mismatch somewhere.

markerikson:
Just fixed it with #422 , I think, although no idea why this was causing issues or seemed to work previously.

nickmccurdy:
Thanks!

Issue #412, opened 2020-03-06
Consider adding a warning if default middlewares run too slow


matthieudumas:
I've switched to redux toolkit for a new project built with create react app, and encountered a big performance issue during development. Long story short, I was storing large amounts of json in the store and redux toolkit's default dev middlewares (redux-immutable-state-invariant and serializable-state-invariant-middleware) were taking around 400ms to validate each dispatch...

I'm an experienced react/redux developer, yet I didn't know about those middlewares. Chrome profiler didn't prove very useful, and it took me about 2 hours to nail down the problem.

I suspect other people might have the same problem than me. So what would you think about those middlewares to implement a self timer and raise a warning in the console if they take more than a certain time to run (eg 50ms, ie 20 FPS) ?

markerikson:
Yeah, that seems like a reasonable thing to do.

Can you go ahead and file a PR against the `v1.3.0-integration` branch to add that?

markerikson:
Done in 1.3.

LuisGarcia9000:
> I've switched to redux toolkit for a new project built with create react app, and encountered a big performance issue during development. Long story short, I was storing large amounts of json in the store and redux toolkit's default dev middlewares (redux-immutable-state-invariant and serializable-state-invariant-middleware) were taking around 400ms to validate each dispatch...
> 
> I'm an experienced react/redux developer, yet I didn't know about those middlewares. Chrome profiler didn't prove very useful, and it took me about 2 hours to nail down the problem.
> 
> I suspect other people might have the same problem than me. So what would you think about those middlewares to implement a self timer and raise a warning in the console if they take more than a certain time to run (eg 50ms, ie 20 FPS) ?

@matthieudumas I have the same problem, How do you solved it?

markerikson:
@LuisGarcia9000 The dev check middlewares have done that for the last year or so.  Also, we recently updated the middlewares in 1.5.x to run _much_ faster in most cases. Please make sure you are using the latest version of RTK (1.6.2+)

LuisGarcia9000:
@markerikson I have updated to RTK 1.6.2, but the problem still remains, even though I have placed:
```
middleware: getDefaultMiddleware => [
    ...getDefaultMiddleware({
      immutableCheck: false,
      serializableCheck: false
    })
```
¿any siggestion?

phryneas:
You might also want to try `immutableCheck: false`

markerikson:
[...]

PR #427, opened 2020-03-11
display a warning if `immutableStateInvariantMiddleware` or `ser…

phryneas:
This should address #412. I had to split up the measurement so that `next(action)` is excluded and it only warns for the time taken by the middleware itself, not by something else.
We *could*  add a general "slow reducer" warning while we're at it though.

| Filename | Size | Change | |
|:--- |:---:|:---:|:---:|
| `dist/redux-toolkit.cjs.development.js` | 10.4 kB | +406 B (3%) |  |
| `dist/redux-toolkit.esm.js` | 10.3 kB | +410 B (3%) |  |
| `dist/redux-toolkit.umd.js` | 21.5 kB | +374 B (1%) |  |

<details><summary>ℹ️ <strong>View Unchanged</strong></summary>

| Filename | Size | Change | |
|:--- |:---:|:---:|:---:|
| `dist/index.js` | 149 B | 0 B |  |
| `dist/redux-toolkit.cjs.production.min.js` | 3.69 kB | 0 B |  |
| `dist/redux-toolkit.umd.min.js` | 9.22 kB | 0 B |  |

</details>

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,0 +1,21 @@
+export function getTimeMeasureUtils(maxDelay: number, fnName: string) {
+  let elapsed = 0
+  return {
+    measureTime<T>(fn: () => T): T {
+      const started = Date.now()
+      try {
+        return fn()
+      } finally {
+        const finished = Date.now()
+        elapsed += finished - started
+      }
+    },
+    warnIfExceeded() {
+      if (elapsed > maxDelay) {
+        console.warn(`${fnName} took ${elapsed}ms, which is more than the warning threshold of ${maxDelay}ms. 
+If you are passing very large objects into your state, you might to disable the middleware as it might cause too much of a slowdown in development mode. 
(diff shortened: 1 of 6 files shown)


src/utils.ts
```diff
@@ -0,0 +1,21 @@
...
+        return fn()
+      } finally {
+        const finished = Date.now()
+        elapsed += finished - started
+      }
+    },
+    warnIfExceeded() {
+      if (elapsed > maxDelay) {
+        console.warn(`${fnName} took ${elapsed}ms, which is more than the warning threshold of ${maxDelay}ms. 
+If you are passing very large objects into your state, you might to disable the middleware as it might cause too much of a slowdown in development mode.
```
markerikson:
```suggestion
If your state or actions are very large, you may want to disable the middleware as it might cause too much of a slowdown in development mode. See https://redux-toolkit.js.org/api/getDefaultMiddleware for instructions.
```

Issue #426, opened 2020-03-11
createSlice actions TypeScript typings for the payload does not have Intellisense support


RPDeshaies:
## Context
When using the `createSlice` api with Typescript, it generates actions that can be called using two overloads

1. `payload?: undefined`
2. `<PT extends MY_PAYLOAD>(payload?: PT)` where `MY_PAYLOAD` would be the typing provided inside the `PayloadAction` of the reducer

I have the impression that because of that, Typescript is not able to generate the autocomplete properly even though it is still able to validate the object.

## Example

```typescript
const mySlice = createSlice({
  name: "mySlice",
  initialState: {
    foo: 3
  },
  reducers: {
    setFoo(
      state,
      action: PayloadAction<{
        foo: number;
      }>
    ) {
      state.foo = action.payload.foo;
    }
  }
});

mySlice.actions.setFoo(); // is valid
mySlice.actions.setFoo({}); // is not valid 
```

The error message you get is

![image](https://user-images.githubusercontent.com/6224111/76424795-1571d700-637f-11ea-8036-a2afbf9034ff.png)

Which is perfect, but if you try to toggle the intellisense inside the anonymous object of the second call to `setFoo` you get this:

![image](https://user-images.githubusercontent.com/6224111/76424863-30444b80-637f-11ea-8610-72b4316a1817.png)

## Expected behavior

We should be able to see that the only property the payload can receive is `foo: number`

## Versions
```
"@reduxjs/toolkit": "1.2.5",
"typescript": "3.7.2",
```

phryneas:
I guess you have `strictNullChecks: false` in your `tsconfig.json`?

Please see [this answer](https://github.com/reduxjs/redux-toolkit/issues/399#issuecomment-592691295)

RPDeshaies:
It's turned off by default and turning it on does indeed fix the issue.
Just read the other **Issue**. Thanks for looking into it.

Closing the issue

phryneas:
Every `tsconfig.json` generated by `tsc --init` since TypeScript 2.3 should have `strict: true`, which turn on all the strict options.
It's only off by default to preserve backwards compatibility with pre-2017-code. You might want to check your other `strict` settings ;)

phryneas:
Hey @RPDeshaies, I put in a PR that might improve the autocompletion for people with `strictNullChecks: false` (#428) - could you give it a test and give me feedback if it improves the situation to before?

```
[...]

Issue #399, opened 2020-02-28
Action payload is still possibly undefined even when a value is given


maxijonson:
Considering this example:

```ts
type TestType = { counter: number; step: number; };
const test = createSlice({
    name: "test",
    initialState: { counter: 0, step: 1 } as TestType,
    reducers: {
        increment: (state) => {
            state.counter += state.step;
        },
        setCounterAndStep: (state, action: PayloadAction<TestType>) => {
            state.counter = action.payload.counter;
            state.step = action.payload.step;
        }
    }
});
```

When calling the `setCounterAndStep` action with an empty object as payload... (or an incomplete object)

```ts
test.actions.setCounterAndStep({  })
```

the function tries to apply the `payload` type to undefined AND TestType, which comes from two function overloads when it really should be applying it only to TestType, since an empty object is not undefined. 

```
No overload matches this call.
  Overload 1 of 2, '(payload?: undefined): { payload: undefined; type: string; }', gave the following error.
    Argument of type '{}' is not assignable to parameter of type 'undefined'.
  Overload 2 of 2, '(payload?: { counter: number; step: number; }): { payload: { counter: number; step: number; }; type: string; }', gave the following error.
    Argument of type '{}' is not assignable to parameter of type '{ counter: number; step: number; }'.
      Type '{}' is missing the following properties from type '{ counter: number; step: number; }': counter, stepts(2769)
```

Additionally, Intellisense fails to suggest `counter` and `step` because of the `undefined` overload. 

When trying to understand why this happens, I stumbled upon the `ActionCreatorWithOptionalPayload` interface of typings.d.ts. It seems the first overload is described by

```ts
(payload?: undefined): PayloadAction<undefined, T>;
```

and the second by

```ts
<PT extends Diff<P, undefined>>(payload?: PT): PayloadAction<PT, T>;
```

I found the first overload strange... Why would the `payload` be optional AND typed as `undefined`? I think there should only be the second overload, which already has `payload` optional, so it has the same effect as the first overload.

I removed the first overload from the typings and it fixed the issue. All of these calls work as expected:

```ts
test.actions.setCounterAndStep({}) // Error as expected. Type '{}' is missing the following properties from type '{ counter: number; step: number; }': counter, step
test.actions.setCounterAndStep({ counter: 0, step: 2 }) // Good. Type is satisfied
test.actions.increment() // Good. Payload is undefined
```

[...]

PR #428, opened 2020-03-11
simplify signatures of `ActionCreatorWithOptionalPayload` and `A…

phryneas:
This would probably reduce the pain of #426, #399 and a few other issues where people had `strictNullChecks: false`. Also, it might improve IntelliSense for `ActionCreatorWithPayload` with a little tradeoff.

Changes for `ActionCreatorWithOptionalPayload`:
Before, assuming actionCreator was a `ActionCreatorWithOptionalPayload<string|undefined>
* calling actionCreator() returned a `PayloadAction<undefined>`
* calling actionCreator(undefined) returned a `PayloadAction<undefined>`
* calling actionCreator("test") returned a `PayloadAction<"test">`
* calling actionCreator(someStringTypedVariable) returned a `PayloadAction<string>`

Now, all those calls will return `PayloadAction<string | undefined>` - but autocompletion will work quite better for object types payloads.

Simlarily, assuming actionCreator was a `ActionCreatorWithPayload<string>
* calling actionCreator("test") returned a `PayloadAction<"test">`
* calling actionCreator(someStringTypedVariable) returned a `PayloadAction<string>`

Now, both those cases will return `PayloadAction<string>`, but autocompletion will probably work better for object type payloads.

I've been thinking about this for a bit and I believe the extra granularity we had before might not be worth the frustration people are having with this. What do you think about this simplification?

### DIFF
--- a/etc/redux-toolkit.api.md
+++ b/etc/redux-toolkit.api.md
@@ -28,8 +28,7 @@ export interface ActionCreatorWithNonInferrablePayload<T extends string = string
 
 // @public
 export interface ActionCreatorWithOptionalPayload<P, T extends string = string> extends BaseActionCreator<P, T> {
-    (payload?: undefined): PayloadAction<undefined, T>;
-    <PT extends Diff<P, undefined>>(payload?: PT): PayloadAction<PT, T>;
+    (payload?: P): PayloadAction<P, T>;
 }
 
 // @public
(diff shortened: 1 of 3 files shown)


markerikson:
If you think it's an improvement, we can go with it.  Is this good as-is?

phryneas:
I think so. It might be breaking if someone actually relied on the old behaviour, but I doubt anyone really used that.

We might want to ask for feedback in the old issues though.

markerikson:
Let's go with it.

Issue #434, opened 2020-03-14
Entity adapter methods do not cause TS errors when used as case reducers illegaly


markerikson:
Per https://codesandbox.io/s/entityadapter-test-o5190 , `savedQueriesSliceScoped.ts` has a state slice with two different normalized sections inside of it: `{public: EntityState, private: EntityState}`.  

That means that passing `someAdapter.updateMany` as a case reducer should be illegal and cause a TS error, because the adapter methods should only be getting an `EntityState` object as the `state` argument.  Instead, TS doesn't seem to error.  

If I manually write a reducer and call that same update method inside, TS correctly flags it as being illegal:

```ts
const { actions, reducer } = createSlice({
  name: "savedQueriesScoped",
  initialState: initialSavedQueriesState,
  reducers: {
    // This line should cause errors, but doesn't
    savedQueriesUpdated: savedQueriesAdapter.updateMany,
    otherReducer(state, action: PayloadAction<KustoQueryDefinition[]>) {
      // the `state` arg here will be flagged as an error
      savedQueriesAdapter.updateMany(state, action);
    }
  },
```

@phryneas , any ideas?

phryneas:
I just took a quick look.

Used in this position, S in 
```ts
declare interface EntityStateAdapter<T> {
    addOne<S extends EntityState<T>>(state: S, entity: T): S;
```
seems to somehow resolve to `any` or somewhat similar, making it usable everywhere.

A possible solution would probably be to write it more complicated, eliminating `any` and enforcing `EntityState<T>` at the least, like

```ts
type IgnoreAny<T> = IsAny<T, unknown, T>;

declare interface EntityStateAdapter<T> {
    addOne<S extends EntityState<T>>(state: IgnoreAny<S> & EntityState<T>, entity: T): IgnoreAny<S> & EntityState<T>;
```

Something a *little* simpler like

```ts
type FallbackIfAny<T, Fallback = unknown> = IsAny<T, Fallback, T>;

declare interface EntityStateAdapter<T> {
    addOne<S extends EntityState<T>>(s: FallbackIfAny<S, EntityState<T>>, entity: T): FallbackIfAny<S, EntityState<T>>;
```

also seems to do the trick.

This just as a first feedback, might need a little more investigation later.

markerikson:
Should be fixed by #436 .

PR #436, opened 2020-03-15
prevent any-casting of S generic on entityAdapter reducer-like f…

phryneas:
This would prevent what we were seeing in #434. I'm preventing these methods from having their `state` parameter from falling back to `any`.
Note that I do *not* do this for their return type though, because that would make the "should be usable in a slice with extra properties" test I added fail. 
I think like this it's a good compromise of a fallback.

Additionally, I've added type-tests.

Also, I've removed the entityAdapter.map method, as that cannot be used as a reducer (action would be non-serializable), is easily re-implemented in userland and the name `map` implies a potentially completely different return type of a non-mutating copy, which is not really the case here.

### DIFF
--- a/docs/api/createEntityAdapter.md
+++ b/docs/api/createEntityAdapter.md
@@ -118,8 +118,6 @@ export interface Dictionary<T> extends DictionaryNum<T> {
 
 export type Update<T> = { id: EntityId; changes: Partial<T> }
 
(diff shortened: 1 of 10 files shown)


markerikson:
Can you update the docs page as well?

phryneas:
Sure.

Issue #441, opened 2020-03-24
Add usage guide docs for `createEntityAdapter`


markerikson:
v1.3.0 is now out and includes `createEntityAdapter`.  However, while we have API docs at https://redux-toolkit.js.org/api/createEntityAdapter , we don't have any usage guide info in https://redux-toolkit.js.org/usage/usage-guide .

I'd like a section added that covers:

- Using the adapter CRUD methods as both case reducers and "mutating" helper functions
- Patterns for handling normalization both by hand and with a tool like `normalizr`, such as running `normalizr` in your action creation logic and extracting `action.payload.entities` in your reducers and passing to the adapter APIs.
- It'd also be interesting to see if there's a good way to have a centralized `state.entities` structure as well

PR #444, opened 2020-03-26
Add dictionary support to createEntityAdapter many methods

msutkowski:
#### Overview
This PR allows a user to pass in an object with the shape of `{ 1: { id: 1, other: 'value' }` directly to the `upsertMany`, `addMany`, and `setAll` methods of an entity adapter. This ultimately makes integration with a normalization library like [normalizr](https://github.com/paularmstrong/normalizr) much simpler.

~~Todo~~
- [x] Review `updateMany`
- [x] Review tests
- [x] Expand type tests

### DIFF
--- a/src/entities/models.ts
+++ b/src/entities/models.ts
@@ -57,16 +57,22 @@ export interface EntityStateAdapter<T> {
     action: PayloadAction<T>
   ): S
 
-  addMany<S extends EntityState<T>>(state: PreventAny<S, T>, entities: T[]): S
   addMany<S extends EntityState<T>>(
     state: PreventAny<S, T>,
-    entities: PayloadAction<T[]>
+    entities: T[] | Record<string, T>
+  ): S
+  addMany<S extends EntityState<T>>(
+    state: PreventAny<S, T>,
+    entities: PayloadAction<T[] | Record<string, T>>
   ): S
 
-  setAll<S extends EntityState<T>>(state: PreventAny<S, T>, entities: T[]): S
   setAll<S extends EntityState<T>>(
     state: PreventAny<S, T>,
-    entities: PayloadAction<T[]>
+    entities: T[] | Record<string, T>
+  ): S
+  setAll<S extends EntityState<T>>(
+    state: PreventAny<S, T>,
+    entities: PayloadAction<T[] | Record<string, T>>
   ): S
 
   removeOne<S extends EntityState<T>>(state: PreventAny<S, T>, key: EntityId): S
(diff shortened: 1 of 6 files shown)


src/entities/models.ts
```diff
@@ -57,16 +57,22 @@ export interface EntityStateAdapter<T> {
     action: PayloadAction<T>
   ): S
 
-  addMany<S extends EntityState<T>>(state: PreventAny<S, T>, entities: T[]): S
   addMany<S extends EntityState<T>>(
     state: PreventAny<S, T>,
-    entities: PayloadAction<T[]>
+    entities: T[] | Record<string, T>
```
markerikson: Use `EntityId` instead of `string`, since IDs could be numbers.

Hmm.  It looks like `normalizr`'s types are:

`{ [key:string]: { [key:string]: T } | undefined}`

Do these mesh okay?  Should we be using the `Dictionary<T>` type instead?

msutkowski: That's a good question. `{ [key: string]: T}` types _should_ play nicely with `Record<EntityId, T>` for the most part without the index key signature features. I'll try experimenting some, but maybe @phryneas has an idea here?

msutkowski: Just wanted to leave a record of where I got to. I struggled a lot getting normalizr to give me the actual types of the processed data. If we define what the `ReturnType` _should_ be, there are no issues and the types pass through as expected without errors in the reducer.

```ts
export const fetchArticle = createAsyncThunk<
  {
    articles: { [key: string]: Article };
    users: { [key: string]: Author };
    comments: { [key: string]: Comment };
  },
  number,
  any
>("articles/fetchArticle", async id => {
  const data = await fakeAPI.articles.show(id);
  // normalize the data so reducers can responded to a predictable payload, in this case: `action.payload = { users: { 1: { ... } }, articles: {}, comments: {} }`
  const normalized = normalize(data, articleEntity);
  return normalized.entities as any; // cast as any to avoid the type issues
});
```

msutkowski: Quick update: the normalizr types don't infer anything at all, so this about the 'cleanest' way I can get this work:

```ts
export const fetchArticle = createAsyncThunk(
  "articles/fetchArticle",
  async (id: number) => {
    const data = await fakeAPI.articles.show(id);
    // normalize the data so reducers can responded to a predictable payload, in this case: `action.payload = { users: {}, articles: {}, comments: {} }`
    const normalized = normalize<
      any,
      {
        articles: { [key: string]: Article };
        users: { [key: string]: Author };
        comments: { [key: string]: Comment };
      }
    >(data, articleEntity);
    return normalized.entities;
  }
);
```

markerikson: For the "Usage Guide" page, we can keep things as plain JS, so we don't have to worry about that here.

Perhaps we could make a brief note of this in the "Usage with TS" page, and possibly point to a TS sandbox.

msutkowski: Good point. I've already converted the sandbox mentioned in #441 to TS specifically for this code sample above.  If this PR is approved, I'll update the usage docs and sandbox references there in a cleanup pass.

phryneas:
Looks good to me.

PR #449, opened 2020-03-27
Export createImmutableStateInvariantMiddleware, fix typo

### LINKED ISSUE #448: createImmutableStateInvariantMiddleware not actually exported
Hello! 👋 

The [Other Exports](https://redux-toolkit.js.org/api/other-exports) docs say that `createImmutableStateInvariantMiddleware` is exported, but it doesn't seem to actually be there: https://github.com/reduxjs/redux-toolkit/blob/master/src/index.ts

exogen:
This adds `createImmutableStateInvariantMiddleware` (and the default `isImmutable`, and options type) to the package exports.

Also noticed a typo in an argument name: `ingorePaths` -> `ignorePaths` (doesn't change the API as it's just a normal arg and not an options object or anything).

Fixes #448

### DIFF
--- a/etc/redux-toolkit.api.md
+++ b/etc/redux-toolkit.api.md
@@ -136,7 +136,7 @@ export function createEntityAdapter<T>(options?: {
     sortComparer?: false | Comparer<T>;
 }): EntityAdapter<T>;
 
-// @public (undocumented)
+// @public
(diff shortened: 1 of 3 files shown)


markerikson:
Thanks!

PR #455, opened 2020-03-29
createEntityAdapter Examples reducer changed to reducers

ouweiya:
https://redux-starter-kit.js.org/api/createEntityAdapter#examples
Code example word error, "reducer" changed to "reducers"

### DIFF
--- a/docs/api/createEntityAdapter.md
+++ b/docs/api/createEntityAdapter.md
@@ -312,7 +312,7 @@ const booksSlice = createSlice({
   initialState: booksAdapter.getInitialState({
     loading: 'idle'
   }),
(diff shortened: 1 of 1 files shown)


markerikson:
Thanks!

Also, any reason why you're still using the `redux-starter-kit.js.org` address?  We kept that alive during the project rename several months ago, but the correct address is `redux-toolkit.js.org`

ouweiya:
I bookmarked this URL a long time ago, and I often browse to learn.

markerikson:
Ah. Yeah, please update your bookmark to https://redux-toolkit.js.org .  At some point we'll turn off the old URL.

PR #457, opened 2020-03-29
add ignoredActionPaths to createSerializableStateInvariantMiddleware

### LINKED ISSUE #453: createAsyncThunk cannot pass non-serializable arg
I try to use `createAsyncThunk` with non-serializable object like:

```ts
export const roundAsync = createAsyncThunk(
  "someAsync",
  ({ api }: { api: ApiClient }) => {
    return api.apiAction()
  }
)
```

and I got error

```
index.js:1 A non-serializable value was detected in an action, in the path: `meta.arg.api`. Value: ApiClient {} 
Take a look at the logic that dispatched this action:
```

I found this issue ignored `meta.args` by default
https://github.com/reduxjs/redux-toolkit/pull/410

but maybe `createAsyncThunk` fire `meta.arg`

Additionaly, I try  customize `serializableCheck.ignoredPaths` option, I cannot resolve this too.
This maybe not passed `ignorePaths` params  to `findNonSerializableValue`
https://github.com/reduxjs/redux-toolkit/blob/a1e8becbb2e4653d721bdc1d2363a45562c526a2/src/serializableStateInvariantMiddleware.ts#L157-L162

phryneas:
allow configuration of createSerializableStateInvariantMiddleware ignoredActionPaths, update default value, add tests

fixes #453

I've opted for the additional `ignoredActionPaths` option now, but I'm open for discussion :)

### DIFF
--- a/src/serializableStateInvariantMiddleware.test.ts
+++ b/src/serializableStateInvariantMiddleware.test.ts
@@ -330,6 +330,61 @@ describe('serializableStateInvariantMiddleware', () => {
     expect(numTimesCalled).toBeGreaterThan(0)
   })
 
+  describe('ignored action paths', () => {
+    function reducer() {
+      return 0
+    }
+    const nonSerializableValue = new Map()
+
+    it('default value: meta.arg', () => {
+      configureStore({
+        reducer,
+        middleware: [createSerializableStateInvariantMiddleware()]
+      }).dispatch({ type: 'test', meta: { arg: nonSerializableValue } })
+
+      expect(getLog().log).toMatchInlineSnapshot(`""`)
+    })
+
+    it('default value can be overridden', () => {
+      configureStore({
+        reducer,
+        middleware: [
+          createSerializableStateInvariantMiddleware({
+            ignoredActionPaths: []
+          })
+        ]
+      }).dispatch({ type: 'test', meta: { arg: nonSerializableValue } })
+
+      expect(getLog().log).toMatchInlineSnapshot(`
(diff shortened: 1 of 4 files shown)


phryneas:
The codesandbox CI build fails because we got a test taking more that 32ms, adding the "check took too long" message to the output. Not really sure how we should handle that :/

Also, shouldn't travis be running?

src/serializableStateInvariantMiddleware.test.ts
```diff
@@ -330,6 +330,61 @@ describe('serializableStateInvariantMiddleware', () => {
     expect(numTimesCalled).toBeGreaterThan(0)
   })
 
+  describe('ignored action paths', () => {
+    function reducer() {
+      return 0
+    }
+    const nonSerializableValue = new Map()
+
+    it('default value: meta.arg', () => {
```
markerikson: Let's throw in a use of `createAsyncThunk` here and dispatch one of its actions too, just to confirm that we that's consistent.

markerikson:
Yeah, not sure why Travis isn't running.

src/serializableStateInvariantMiddleware.test.ts
phryneas: I already added that over here: 

https://github.com/reduxjs/redux-toolkit/pull/457/files#diff-7ccfeb60873b37c2d4c0a8e4d255f298R466-R478

markerikson:
Tests pass locally, so we're good.

Issue #447, opened 2020-03-27
`extraReducers` type problem


yhnavein:
Hi,

I wanted to check out the new `createAsyncThunk` and I stumbled with the following issue raised by TypeScript:
![image](https://user-images.githubusercontent.com/516709/77745141-0e9fc100-701b-11ea-88e2-1016cfbf57d3.png)

I can write `[loadDates.fulfilled as any]` to get rid of this error, but this feel very off.
Am I doing something wrong?

I am using `"typescript": "3.8.3"`

phryneas:
https://redux-toolkit.js.org/usage/usage-with-typescript/#type-safety-with-extrareducers

@markerikson This seems to confuse people all the time - what are your thoughts on making the builder notation the default in the documentation?

rrebase:
You could also use the type explicitly e.g `[loadDates.fulfilled.type]`

phryneas:
He could, but then he wouldn't have inferred action types. Using the map object notation with TypeScript is pretty much useless.

markerikson:
I don't think we need to make it the _default_ in the docs, but I do think we should update the `extraReducers` section in the API reference with examples of both syntaxes, and emphasize there that the builder notation is preferred for TS usage.

markerikson:
Covered by #460 .

PR #460, opened 2020-03-31
Updates documentation for createSlice and createReducer

msutkowski:
Addresses some of the concerns mentioned in #447 by clarifying some of the recommendations and examples for TypeScript users regarding `createSlice` and `createReducer`, and linking to relevant TS usage doc sections.

I also included an example of using the `builder callback` API with 2 different actions, as the syntax for more than one wasn't shown in either section.


markerikson:
Thanks!

PR #461, opened 2020-03-31
Fix reference to new fetchIssuesCount

lukevp:
Minor doc update that confused me a bit on first read.  the new thunk creator function is called fetchIssuesCount and it replaces fetchIssueCount.


markerikson:
Thanks!

PR #473, opened 2020-04-02
Export ThunkDispatch from redux-thunk.

### LINKED ISSUE #472: Inferred type from `createAsyncThunk` causing TypeScript error: "The inferred type of ... cannot be named without a reference to ..."
Trying to use `createAsyncThunk` in TypeScript, with a pretty basic use case:
```typescript
export const fetchStationSpecs = createAsyncThunk("stations/fetchStationSpecs", async () => {
  const response = await new TestingDaemonApiRequest("station_specs").doFetch(true);
  const stations = (await response.json()).data as IStationCommandMap;
  return stations;
});
```

**tsc** throws this error:
> The inferred type of 'fetchStationSpecs' cannot be named without a reference to '../../../../../../../build/js/rush_temp/node_modules/.pnpm/nexus.skyd.io/redux-thunk/2.3.0/node_modules/redux-thunk'. This is likely not portable. A type annotation is necessary.

This looks to be due to some trickiness with transitive dependencies as described [here](https://github.com/microsoft/TypeScript/issues/30858) - the inferred return type from `createAsyncThunk` relies on the `ThunkDispatch` type from **redux-thunk**, which is unavailable to my package since it's a transitive dependency (this may not be an issue for other build setups, but appears to be a common class of problem for monorepos using **pnpm** or similar systems). 

Fortunately, this problem completely goes away if `ThunkDispatch` is exported directly from Redux Toolkit (thus making it available to the package that depends on it), which I think would be useful to developers for making thunk-related utilities anyway. Going to make the shortest pull request ever to fix this.

sammyers:
Provide the `ThunkDispatch` type as an export for developer utility and to fix a nasty TypeScript problem.

Fixes #472  (see the issue for details on what this is solving)

### DIFF
--- a/etc/redux-toolkit.api.md
+++ b/etc/redux-toolkit.api.md
@@ -307,6 +307,8 @@ export type SliceCaseReducers<State> = {
 
 export { ThunkAction }
 
+export { ThunkDispatch }
+
 // @alpha (undocumented)
 export function unwrapResult<R extends ActionTypesWithOptionalErrorAction>(returned: R): PayloadForActionTypesExcludingErrorActions<R>;
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,7 +8,7 @@ export {
   OutputSelector,
   ParametricSelector
 } from 'reselect'
-export { ThunkAction } from 'redux-thunk'
+export { ThunkAction, ThunkDispatch } from 'redux-thunk'
 
 // We deliberately enable Immer's ES5 support, on the grounds that
 // we assume RTK will be used with React Native and other Proxy-less


phryneas:
I don't really understand the exact circumstances here, but I think this would definitely not hurt. What do you think @markerikson ?

markerikson:
Sure.

VinceOPS:
Good morning @markerikson @phryneas.

I had the exact same problem (https://github.com/reduxjs/redux-toolkit/issues/472) with type `ThunkMiddleware`. Exporting this type from `redux-thunk` fixes it, just the way @sammyers did it with `ThunkDispatch`.

> The inferred type of '**configureXxxxxStore**' cannot be named without a reference to '.pnpm/redux-thunk@2.4.1_redux@4.2.0/node_modules/redux-thunk'. This is likely not portable. A type annotation is necessary.ts(2742)

(context: Monorepo using pnpm, latest RTK and TS 4.8.4).

Don't you mind if I create a PR to add these types to the exported ones?


___

(FYI, I also have the same error with `createSlice` but related to `immer`. I couldn't find the "missing" type yet...)

> The inferred type of '**xxxxxSlice**' cannot be named without a reference to '.pnpm/immer@9.0.15/node_modules/immer/dist/internal'. This is likely not portable. A type annotation is necessary.

Issue #474, opened 2020-04-03
Should updateOne/updateMany support nested updates?


msutkowski:
As of right now, `updateOne` and `updateMany` are only useful if your data has no nesting at all. In the scenario that a user uses a JSON API schema, they'll normally have a structure like `{ id: 1, type: 'example', attributes: { other: key, another: one } }`, depending on how they normalize their data. Should we consider supporting at least one level of nesting with mutable updates? If not, we should update the documentation around these methods.

_Note: Technically it'll process nested updates now, but it'll just replace the nested objects and this may be confusing to users who think immer is doing magic updates here._

markerikson:
It's an interesting question, but I'd rather not go there for now.  Too much potential complexity to figure out how deep updates should work.  

That said, the shallow update behavior should probably be documented.  Can you add that to the API reference for `createEntityAdapter`?

otakustay:
We have previously encountered this issue in out own library, as a solution we provide a `customMerge` function to something like `createEntityAdapter`, this can give more extensibility and keep a low complexity at the same time, I'd like to recommend to add a `mergeEntity(base: T, incoming: T)` option to `createEntityAdapter`

markerikson:
I'd be open to seeing some possibilities, but I have concerns about trying to provide a generic merging API.  I think there's too many complexities in how that _could_ behave for us to provide a single solution.

tannera:
a `customMerge` function would be helpful.

markerikson:
As I said, I'm open to seeing some _specific_ API proposals.  But, my inclination atm is that if you need custom behavior, you write your own reducer.  It's not _that_ hard :)  Especially with Immer available.

markerikson:
Going to close this for now, since there haven't been any actionable API suggestions.  If folks do have specific suggestions on what such an API might look like, please comment with details and we can discuss further.

vputsenko:
yes it is not hard, but if you look inside adapter sources, upsertMany code splits payload into update and merge parts following by resorting and this should be copy pasted just because customMerge predicate is not acceptable solution.

Issue #464, opened 2020-03-31
createEntityAdapter "updateOne" API doesn't work


ouweiya:
createEntityAdapter "updateOne" API doesn't work, data won't update? Where am I doing wrong?
codesandbox: 
https://codesandbox.io/s/hopeful-kowalevski-yet3g

```jsx
import React from 'react';
import { Provider, useDispatch, useSelector } from 'react-redux';
import {
  configureStore,
  createSlice,
  createEntityAdapter
} from '@reduxjs/toolkit';

const todosAdapter = createEntityAdapter();

const todosSlice = createSlice({
  name: 'todos',
  initialState: todosAdapter.getInitialState(),
  reducers: {
    todoAdded: todosAdapter.addOne,
    todoUpdated: todosAdapter.updateOne
  }
});

const { todoAdded, todoUpdated } = todosSlice.actions;

const store = configureStore({
  reducer: { todos: todosSlice.reducer }
});

const Updated = () => {
  const dispatch = useDispatch();
  const todos = useSelector(store => store.todos);
  console.log('todos: ', todos);

  return (
    <div>
      <button onClick={() => dispatch(todoAdded({ id: 1, todo: 'one' }))}>
        todoAdded
      </button>
      <button onClick={() => dispatch(todoUpdated({ id: 1, todo: 'two' }))}>
        todoUpdated
      </button>

      <h5>Todos: {JSON.stringify(todos)}</h5>
    </div>
  );
};

export default () => (
  <Provider store={store}>
    <Updated />
  </Provider>
);
```

markerikson:
It looks like you're using the API the wrong way.

`updateOne` takes a parameter that looks like this:

```ts
export type Update<T> = { id: EntityId; changes: Partial<T> }
```

So, you _should_ be dispatching:

```
dispatch(todoUpdated({id: 1, changes: {todo: "two"}))
```

I'm disturbed that TS isn't catching that.

@phryneas , do we need to do something similar to the `EntityState` tweaks to ensure that the user is passing in the right type here?  Do any of the other CRUD methods need similar changes?

msutkowski:
As a side note, this is a good use case for using `upsertOne` as it's easier to deal with.

phryneas:
> @phryneas , do we need to do something similar to the `EntityState` tweaks to ensure that the user is passing in the right type here? Do any of the other CRUD methods need similar changes?

[...]

Issue #454, opened 2020-03-28
createEntityAdapter with dynamic or fulfilled-order sortComparer


waynebloss:
Can you give some guidance on how to do a dynamic `sortComparer`, e.g. ngrx/platform#898
or if you would even want to do that?

The `sortComparer` defines the sort order for storage, so I'm not sure that having a dynamic sort for that purpose is a good idea. The first answer on the linked ngrx issue recommends just using a selector to sort, which is probably what I'd do instead of constantly changing the storage sort order.

However, for the default order - I'd prefer it to be fulfilled-order (insert-order). Since you make [no guarantees](https://redux-toolkit.js.org/api/createEntityAdapter#sortcomparer) about the default ordering when no `sortComparer` is provided - how can I provide a `sortComparer` that always sorts the entities based on the order that they were fetched / fulfilled / inserted into the Entity Adapter?

waynebloss:
I can see in the source code for the [unsorted adapter](https://github.com/reduxjs/redux-toolkit/blob/master/src/entities/unsorted_state_adapter.ts#L23) that insert-order is the default sort order.

Would you consider documenting that and guaranteeing that it won't change without a breaking change version?

markerikson:
Can you give an example of what you mean by "dynamic `sortComparer`?

My main thought here is that the built-in sorting behavior is primarily a convenience factor.  If you want something more complex or detailed, you should probably be managing your own sorting logic, in either reducers or selectors.

There's a bit of a parallel with the techniques shown in the post [Advanced Redux Entity Normalization](https://medium.com/@dcousineau/advanced-redux-entity-normalization-f5f1fe2aefc5), which talks about having multiple "keywindows", which are really just additional ID arrays with some specific subset of IDs in them.

waynebloss:
I'm trying to figure out exactly in what capacity should `sortComparer` be used. 

So, I was wondering if it was intended as a way to sort things on behalf of the user. I imagined that a "dynamic `sortComparer`" would be one that changes what we're sorting by after some action occurs. When I looked at the _nrgx_ issues I saw people asking about it there as well.

I don't think this is the true intention. I can see now that `sortComparer` is there simply to enforce a single, unchanging default sort order.

[...]

PR #479, opened 2020-04-05
Update createEntityAdapter docs

msutkowski:
- Add notes about shallow update behavior (#474)
- Add `updateOne` example In the usage guide (#464) 
- Add a comment about the `unsorted adapter`s behavior per (#454).
- Adds periods at the end of sentences to make things uniform


markerikson:
Looks good, thanks!

PR #492, opened 2020-04-10
Update docs, add test, alias ignore->ignoredPaths

### LINKED ISSUE #491: Outdated documentation of immutable invariant options on the homepage
The [API reference on the homepage](https://redux-toolkit.js.org/api/getDefaultMiddleware#api-reference) currently states that the paths to ignore in the immutable state middleware can be configured with the `ignore` key. This is not true; when the middleware was merged in the most recent version, the key was renamed to `ignoredPaths`, relying on the old `ignore` key has no effect. (The new middleware also has a `warnAfter` key for what it's worth). The API reference on the homepage should be updated to avoid confusion. Also, it would probably be a good idea to treat `ignore` as an alias for `ignoredPaths` in order not to break backwards compatibility in a minor release.

Thanks a lot in advance!

msutkowski:
Fixes #491

### DIFF
--- a/docs/api/getDefaultMiddleware.md
+++ b/docs/api/getDefaultMiddleware.md
@@ -110,7 +110,8 @@ interface ThunkOptions<E = any> {
 
 interface ImmutableStateInvariantMiddlewareOptions {
   isImmutable?: (value: any) => boolean
-  ignore?: string[]
+  ignoredPaths?: string[]
+  warnAfter?: number
 }
 
 interface SerializableStateInvariantMiddlewareOptions {
--- a/etc/redux-toolkit.api.md
+++ b/etc/redux-toolkit.api.md
@@ -278,6 +278,8 @@ export type IdSelector<T> = (model: T) => EntityId;
 // @public
 export interface ImmutableStateInvariantMiddlewareOptions {
     // (undocumented)
+    ignore?: string[];
+    // (undocumented)
     ignoredPaths?: string[];
     // (undocumented)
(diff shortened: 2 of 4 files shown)


msutkowski:
@markerikson When we inlined this middleware, I just moved the tests over as they were and didn't update anything outside of the types. I propose changing this test:

```js
  it('Should not print a warning if "next" takes too long', () => {
    const next: Dispatch = action => {
      const started = Date.now()
      while (Date.now() - started < 8) {}
      return action
    }

    const dispatch = middleware({ warnAfter: 4 })(next)

    const restore = mockConsole(createConsole())
    try {
      dispatch({ type: 'SOME_ACTION' })
      expect(getLog().log).toEqual('')
    } finally {
      restore()
    }
  })
})
```

to this:

```js
  it('Should not print a warning if execution completes before the warnAfter threshold', () => {
    const next: Dispatch = action => action

    const dispatch = middleware({ warnAfter: 32 })(next) 

    const restore = mockConsole(createConsole())
    try {
      dispatch({ type: 'SOME_ACTION' })
      expect(getLog().log).toEqual('')
    } finally {
      restore()
    }
  })
})
```

markerikson:
The point of that test is to make sure we're only measuring the time the middleware itself takes to execute. If a reducer takes a long time, that's not something we want to warn about.  Also, that's a test we specifically added while working on 1.3.0, when we added the `warnAfter` option (which isn't part of the original immutable middleware package).

Issue #489, opened 2020-04-08
TypeScript createAsyncThunk optional or no args


stephen-dahl:
I am trying to make an api call with an optional parameter however typescript complains
```
export const usersLoad = createAsyncThunk(
	'users/load',
	(filters: UserFilters = {}) => Api.get<User[]>('/api/users/search', filters)
);
...
dispatch(usersLoad());
```
`Expected 1 arguments, but got 0`

I have a similar issue with this action
```
export const usersLoadMore = createAsyncThunk(
	'users/loadMore',
	(_: SyntheticEvent | void, thunkApi) => {
		const state = thunkApi.getState() as RootState;
		const { filters, data } = state.users;
		return Api.get<User[]>('/api/users/search', {
			...filters,
			offset: data.length,
		});
	}
);
```
here I don't use the arg but I had to give it the SyntheticEvent or it complained when passing it to onClick, and I had to give it void to make it work when I call it directly. I would think that I would not have to set it to anything since I am not using it.

phryneas:
Yeah, that's not possible at the moment.
I will look into this in the near future, but in the meantime you can do something like

```ts
const _usersLoad = createAsyncThunk(
	'users/load',
	(filters: UserFilters) => Api.get<User[]>('/api/users/search', filters)
);

export const usersLoad = (filters: UserFilters = {}) => _usersLoad(filters);
```

I admit it's a bit hacky, but it should do as a workaround for now.

stephen-dahl:
ya i'm working around it.
still new to typescript but it looks like the typing returns a function with the signature
`(arg: ThunkArg) => ....` making arg required

phryneas:
Yes. As I said:

> that's not possible at the moment.
> I will look into this in the near future

We simply didn't anticipate this use case and none of our beta testers had it either, so we'll have to add it.

markerikson:
Should be resolved by https://github.com/reduxjs/redux-toolkit/releases/tag/v1.3.5 .

drydenwilliams:
@markerikson please could you show an example of an optional argument? That like looks like it just cancels it no? Also it seem using `createAsyncThunk` seems to be more boilerplate as you need to have: 

```typescript
// someThunk.ts
export const fetchSomeThunk = createAsyncThunk(
  'app/fetchSomeThunk',
  async (arg: IMyArg) => {
    // Async stuff here
    return axios.get('')
  }
)
```

and then in a component:

```
// myComponent.tsx

[...]

Issue #486, opened 2020-04-07
Extending createAsyncThunk with Typescript


dagstuan:
I'm trying to write a generic extension of `createAsyncThunk` in Typescript, so that I dont have to repeat the thunkApiConfig every time. I'm trying to write it such as this:

```ts
type ThunkApiConfig = {
  dispatch: MyDispatch:
  rejectValue: MyErrorType;
};

export const createAppThunk = <Returned, ThunkArg = void>(
  type: string,
  payloadCreator: (
    arg: ThunkArg,
    thunkAPI: GetThunkAPI<ThunkApiConfig>,
  ) =>
    | Promise<Returned | RejectWithValue<GetRejectValue<ThunkApiConfig>>>
    | Returned
    | RejectWithValue<GetRejectValue<ThunkApiConfig>>,
) => createAsyncThunk<Returned, ThunkArg, ThunkApiConfig>(type, payloadCreator);
```

However, this doesnt work since the types `GetThunkAPI`, `RejectWithValue`, and `GetRejectValue` are not exported. Is there any other recommended way of doing this? Or can you export the types needed to extend `createAsyncThunk`?

dsyncerek:
I encountered the same problem.

In `react-redux` types there is exported interface `DefaultRootState` (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-redux/index.d.ts#L47) which can be augmented by users to add default types for the root state. So instead of reapeating `useSelector<AppState, ...>()` or creating generic wrapper, we can write once:
```typescript
declare module 'react-redux' {
  interface DefaultRootState extends AppState {}
}
```

It would be great if something similar will be possible in `@reduxjs/toolkit`. I mean:
```typescript
declare module '@reduxjs/toolkit' {
  type AsyncThunkConfig = {
    state?: AppState;
    dispatch?: AppDispatch;
    extra?: null;
    rejectValue?: HttpError;
  }
}
```

phryneas:
Hmm. While most everything else would usually stay the same, especially the `rejectValue` would be something that could change on a `asyncThunk` to `asyncThunk` basis, so adding a `DefaultAsyncThunkConfig` that could be overwritten by module augmentation would imho be the wrong approach.

Also note that the `react-redux`-types are community-driven on DefinitelyTypes and this is the first time I hear of that type - this has probably been added without any of the redux maintainers even noticing. I'm not really sure if I'm a fan of that, to be honest.

[...]

PR #502, opened 2020-04-13
payloadCreator arg argument => asyncThunk arg type

phryneas:
This would address #489 without changing any external signatures to this point. So @stephen-dahl and @msutkowski  please take a look.

The complete behavioral change should be covered by this test:

https://github.com/phryneas/redux-toolkit/blob/47eac023cd18e86cf888224381a77b9776a97677/type-tests/files/createAsyncThunk.typetest.ts#L185-L230

My only concern is the handling of optional parameters. For people with `strictNullChecks: false` in their `tsconfig.json`, this would make the parameter optional for all their asyncThunks - which is only consequent, but may be undesirable.
I've gotta take a second look if this can be done better - on the other hand, we have the same behavior with `createSlice` actions already, so I would not try too much there.

Also, this *might* make extending createAsyncThunk with Typescript (#486) a little more difficult (we'd have to do some things for that anyways, but it might be simpler before this change), so I've got to evaluate that before we merge it.

But please already take a look.

### DIFF
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -5,7 +5,7 @@ import {
   ActionCreatorWithPreparedPayload
 } from './createAction'
 import { ThunkDispatch } from 'redux-thunk'
-import { FallbackIfUnknown } from './tsHelpers'
+import { FallbackIfUnknown, IsAny } from './tsHelpers'
 import { nanoid } from './nanoid'
 
 // @ts-ignore we need the import of these types due to a bundling issue.
@@ -216,12 +216,28 @@ If you want to use the AbortController to react to \`abort\` events, please cons
           }
         }
 
-  function actionCreator(arg: ThunkArg) {
-    return (
-      dispatch: GetDispatch<ThunkApiConfig>,
-      getState: () => GetState<ThunkApiConfig>,
-      extra: GetExtra<ThunkApiConfig>
-    ) => {
+  type AsyncActionThunkAction = (
+    dispatch: GetDispatch<ThunkApiConfig>,
+    getState: () => GetState<ThunkApiConfig>,
+    extra: GetExtra<ThunkApiConfig>
+  ) => Promise<ReturnType<typeof fulfilled | typeof rejected>> & {
+    abort(reason?: string): void
+  }
+
+  type ActionCreator = IsAny<
+    ThunkArg,
+    (arg: ThunkArg) => AsyncActionThunkAction,
+    unknown extends ThunkArg
+      ? (arg: ThunkArg) => AsyncActionThunkAction
+      : [ThunkArg] extends [void] | [undefined]
+      ? () => AsyncActionThunkAction
+      : [undefined] extends [ThunkArg] | []
+      ? (arg?: ThunkArg) => AsyncActionThunkAction
+      : (arg: ThunkArg) => AsyncActionThunkAction
+  >
+
+  function actionCreator(arg: ThunkArg): AsyncActionThunkAction {
+    return (dispatch, getState, extra) => {
       const requestId = nanoid()
 
       const abortController = new AC()
@@ -277,7 +293,7 @@ If you want to use the AbortController to react to \`abort\` events, please cons
     }
   }
 
-  return Object.assign(actionCreator, {
+  return Object.assign(actionCreator as ActionCreator, {
     pending,
     rejected,
     fulfilled
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -130,6 +130,64 @@ export type AsyncThunkPayloadCreator<
   thunkAPI: GetThunkAPI<ThunkApiConfig>
 ) => AsyncThunkPayloadCreatorReturnValue<Returned, ThunkApiConfig>
 
+type AsyncThunkReturnValue<ThunkArg, FulfilledValue, RejectedValue> =
+  | PayloadAction<FulfilledValue, string, { arg: ThunkArg; requestId: string }>
+  | PayloadAction<
+      undefined | RejectedValue,
+      string,
+      { arg: ThunkArg; requestId: string; aborted: boolean },
+      SerializedError
+    >
+/**
+ * A ThunkAction created by `createAsyncThunk`.
+ * Dispatching it returns a Promise for either a
+ * fulfilled or rejected action.
+ * Also, the returned value contains a `abort()` method
+ * that allows the asyncAction to be cancelled from the outside.
+ *
+ * @public
+ */
+export type AsyncThunkAction<
+  Returned,
+  ThunkArg,
+  ThunkApiConfig extends AsyncThunkConfig
+> = (
+  dispatch: GetDispatch<ThunkApiConfig>,
+  getState: () => GetState<ThunkApiConfig>,
+  extra: GetExtra<ThunkApiConfig>
+) => Promise<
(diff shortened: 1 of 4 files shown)


msutkowski:
@phryneas This looks pretty good! Would it be a bad idea to change `ActionCreator` to the type below _(just swapping the last two lines)_? I believe this would address the issue you mentioned with parameters being optional - but I'm not 100% sure 😅 

```ts
type ActionCreator = IsAny<
    ThunkArg,
    (arg: ThunkArg) => AsyncActionThunkAction,
    unknown extends ThunkArg
      ? (arg: ThunkArg) => AsyncActionThunkAction
      : [ThunkArg] extends [void] | [undefined]
      ? () => AsyncActionThunkAction
      : [undefined] extends [ThunkArg] | []
      ? (arg: ThunkArg) => AsyncActionThunkAction
      : (arg?: ThunkArg) => AsyncActionThunkAction
  >
```

I'm also not sure at all about how this would impact the issue you referenced with extending the types for `createAsyncThunk`, but I can help test that tomorrow. 

My 🧠 has trouble processing those type tests this late at night, so I made [a gist of the common typing patterns](https://gist.github.com/msutkowski/81a69be15e96028c848bf3764334a323) to check against as well. (I just dropped this in a `recipes` folder locally against your branch - if you have a better way of doing something similar for these purposes, please let me know).

If you look at the last sample in that gist, I did `arg: string | void`, and I don't feel like that's too awful to type to mimic `arg?: string`.

Thanks again for putting this up, and sorry in advance if this isn't helpful!

phryneas:
> ```ts
> type ActionCreator = IsAny<
>     ThunkArg,
>     (arg: ThunkArg) => AsyncActionThunkAction,
>     unknown extends ThunkArg
>       ? (arg: ThunkArg) => AsyncActionThunkAction
>       : [ThunkArg] extends [void] | [undefined]
>       ? () => AsyncActionThunkAction
>       : [undefined] extends [ThunkArg] | []
>       ? (arg: ThunkArg) => AsyncActionThunkAction
>       : (arg?: ThunkArg) => AsyncActionThunkAction
>   >
> ```

That would not prevent the issue, but introduce two new ones:

* a asyncThunk resulting from a payloadGenerator like `(arg?: number) => 0` would look like `(arg: number) => ...`
* a asyncThunk resulting from a payloadGenerator like `(arg: number) => 0` would look like `(arg?: number) => ...`

Essentially, the question is if we handle optional arguments or just replace the last three lines with `(arg: ThunkArg) => AsyncActionThunkAction`, making the argument mandatory in all cases where it *could* be different than `undefined`.

Something like `number | void` leading to `(arg?: number) => AsyncActionThunkAction` would be a possibility, too, of course, but that would not be very intuitive so I'm not sure if it would even be worth it to have it there.

Gotta run to work now :)

stephen-dahl:
> Something like `number | void` leading to `(arg?: number) => AsyncActionThunkAction` would be a possibility, too, of course, but that would not be very intuitive so I'm not sure if it would even be worth it to have it there.

would there be another way to handle `(arg: number | void, thunkApi) => {}`?
`arg` cant be `arg?` in this case we cant have an optional param before a required one.

figuring out types for some of these edge cases looks like a pain 😂

msutkowski:
@phryneas Gotcha. In that case, the only issue I see with the current types is this:

```ts
const usingInline = createAsyncThunk(
  'test/usingInline',
  (arg: string, { getState }) => {
    const { test } = getState() as RootState

    return {
      banana: 'test'
    }
  }
)

usingInline() // i would expect an error being that we say `arg: string`, but this doesn't get caught as the resulting type is `arg?: string`
```

If there is a way to get `arg: string` and `arg?: string` to work as expected, that'd be the ideal thing. If not, I think I would _personally_ prefer the `string | void` to imply an optional argument. The reason being is the typing can be weird. Consider these:

```ts
const usingInline = createAsyncThunk(
  'test/usingInline',
  (arg?: string, { getState }) => { // results in - A required parameter cannot follow an optional parameter.
    const { test } = getState() as RootState

    return {
      banana: 'test'
    }
  }
)
```

```ts
// you can't do `string?` here as the `arg` param, so how would a user accomplish this?
const usingInline = createAsyncThunk<{ banana: string }, string, { state: RootState }>(
  'test/usingInline',
  (arg, { getState }) => {
    const { test } = getState()

    return {
      banana: 'test'
    }
  }
)
```

My only other input is that I don't think `arg` will be typed as optional by a user very often. It seems it is either well-defined or void. I could be wrong here, but that's what I've seen so far in Discord. Thanks!

phryneas:
I've added lots of tests for usage with a seconds argument.

The "optional" cases will now be optional with `strictNullChecks: true`, but require `| void` declaration with `strictNullChecks: false` so not all arguments become optional. I believe this is the best compromise we can get.

Have I missed any weird case?

type-tests/files/createAsyncThunk.typetest.ts
```diff
@@ -181,3 +181,164 @@ const defaultDispatch = (() => {}) as ThunkDispatch<{}, any, AnyAction>
...
+    expectType<(arg?: number) => any>(asyncThunk)
+    asyncThunk()
+    asyncThunk(5)
+    // typings:expect-error
+    asyncThunk('string')
+  }
+
+  // two arguments, first specified as any: asyncThunk has required any argument
+  {
+    const asyncThunk = createAsyncThunk('test', (arg: any) => 0)
```
stephen-dahl, line 320: shouldnt this be
```suggestion
    const asyncThunk = createAsyncThunk('test', (arg: any, thunkApi) => 0)
```

phryneas: You're absolutely right... :facepalm:

phryneas:
Thanks... guess I was tired. Added those, I hopen that were all.

stephen-dahl:
thanks @phryneas I think this is the best experience I have had with getting something updated in a project. lgtm

phryneas:
@stephen-dahl  I still have to extract some helper types to make #486 less painful, but now with the tests in place & thus the behaviour reviewed from you it's gonna be pretty straightforward - I'll try to finish this up today or tomorrow :)

msutkowski:
@phryneas Once #486 is addressed, I'll open a PR for the docs with [common typing recipes](https://gist.github.com/msutkowski/81a69be15e96028c848bf3764334a323), and include an example of typing a generic helper per that thread.

phryneas:
Okay, I've just extracted these types out and exported one of them, `AsyncThunkAction` - that should help make things more readable in tooltips and error messages.

When the tests are finished, I believe this is ready to merge from my side. @markerikson what do you think?

src/createAsyncThunk.ts
```diff
@@ -103,6 +103,64 @@ type GetRejectValue<ThunkApiConfig> = ThunkApiConfig extends {
...
+  dispatch: GetDispatch<ThunkApiConfig>,
+  getState: () => GetState<ThunkApiConfig>,
+  extra: GetExtra<ThunkApiConfig>
+) => Promise<
+  AsyncThunkReturnValue<ThunkArg, Returned, GetRejectValue<ThunkApiConfig>>
+> & {
+  abort(reason?: string): void
+}
+
+type AsyncThunkActionCreator<
```
markerikson, line 137: I sure hope you know what you're doing here, because my eyes are glazing over

markerikson:
Rebased on master after merging #512 .

src/createAsyncThunk.ts
phryneas: I added more tests than I added code, guess we'll be fine :)

Issue #508, opened 2020-04-16
Passing an immer draft to a createReducer reducer returns no changes


AndrewCraswell:
I'm building a library that wraps existing reducers and provides change rollback functionality via Immer patches. This has been working great with Redux, except I've just realized that it doens't work with reducers that were created with `createReducer()` from Redux Toolkit.

```typescript
      function injectImmer(state: IEndpointState & State, action: AnyAction) {
        const result = produce(state, (draft: any) => {
          reducer(draft, action);
        }, (patches, inverse) => {
          // Library is listening for these patches...
          //   but none are received when using Redux Toolkit reducers
        });

        return result;
      }

      rootReducer(injectImmer(state, action), action);
```

I'm taking the reducer, wrapping it in a `produce` statement, and passing the `draft` to the `createReducer` reducer. The result I get back from the reducer has no changes, even though I can debug and see that the reducer executed, and changes were made.

Here is the relevant section in the `createReducer.ts` file:

```typescript
  return function(state = initialState, action): S {
    return createNextState(state, (draft: Draft<S>) => {
      const caseReducer = actionsMap[action.type]
      return caseReducer ? caseReducer(draft, action) : undefined
    })
  }
```
[[createReducer.ts; line 111](https://github.com/reduxjs/redux-toolkit/blob/master/src/createReducer.ts#L111)]

It doesn't seem like this block is checking if the `state` passed to this function `isDraft(state)`, although it is being checked in other areas of the toolkit:

```typescript
    if (isDraft(state)) {
      // we must already be inside a `createNextState` call, likely because
      // this is being wrapped in `createReducer` or `createSlice`.
      // It's safe to just pass the draft to the mutator.
      runMutator(state)

      // since it's a draft, we'll just return it
      return state
    } else {
      return createNextState(state, runMutator)
    }
```
[[state_adapter.ts; line 26](https://github.com/reduxjs/redux-toolkit/blob/a1e8becbb2e4653d721bdc1d2363a45562c526a2/src/entities/state_adapter.ts#L26)]

Unfortunately this is a blocker for my release in the current state.

markerikson:
What kind of change in behavior are you looking for, here?

AndrewCraswell:
It should be possible to nest producers without Redux Toolkit swallowing the changes. I can propose a PR change, but there are some typing issues that I will need advice on.

markerikson:
[...]

PR #509, opened 2020-04-16
Allow nested Immer produce statements with createReducer

AndrewCraswell:
Fixes issue #508, where passing an Immer `draft` object to a reducer created with `createReducer` will swallow any changes made, and they won't affect the initial `draft` object. This means any top-level `produce` statement will not be able to detect any change to the state, even if changes occurred.

This bumps into some typing issues, since `CaseReducers` are configured to contain a `Draft<S>` and return `S | void`. Someone will need to review this and advise how best to fix the typing issues, as I've just used the `as` keyword to assume a type on lines 114-115.

### DIFF
--- a/src/createReducer.ts
+++ b/src/createReducer.ts
@@ -111,7 +111,8 @@ export function createReducer<S>(
         // we must already be inside a `createNextState` call, likely because
         // this is being wrapped in `createReducer` or `createSlice`.
         // It's safe to just pass the draft to the mutator.
-        return caseReducer(state, action)
+        const draft = state as Draft<S> // We can aassume this is already a draft
+        return caseReducer(draft, action) || state
       } else {
         // @ts-ignore createNextState() produces an Immutable<Draft<S>> rather
         // than an Immutable<S>, and TypeScript cannot find out how to reconcile
--- a/src/createReducer.ts
+++ b/src/createReducer.ts
@@ -1,4 +1,4 @@
-import createNextState, { Draft } from 'immer'
+import createNextState, { Draft, isDraft } from 'immer'
 import { AnyAction, Action, Reducer } from 'redux'
 import {
   executeReducerBuilderCallback,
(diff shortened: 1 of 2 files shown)


src/createReducer.ts
```diff
@@ -105,12 +105,24 @@ export function createReducer<S>(
...
-      const caseReducer = actionsMap[action.type]
-      return caseReducer ? caseReducer(draft, action) : undefined
-    })
+    const caseReducer = actionsMap[action.type]
+    if (caseReducer) {
+      if (isDraft(state)) {
+        // we must already be inside a `createNextState` call, likely because
+        // this is being wrapped in `createReducer` or `createSlice`.
+        // It's safe to just pass the draft to the mutator.
+        const draft = state as Draft<S> // We can aassume this is already a draft
```
Jomik, line 114: This cast seems fine. It seems like isDraft is not typed "correctly". It should be a type guard.
https://github.com/immerjs/immer/blob/master/src/utils/common.ts#L22

src/createReducer.ts
```diff
@@ -105,12 +105,24 @@ export function createReducer<S>(
...
+        return caseReducer(draft, action) || state
```
Jomik, line 115: Shouldn't this use nullish coalescing?
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing
```suggestion
        return caseReducer(draft, action) ?? state
```

AndrewCraswell: Actually come to think of it, this won't work because it's trying to prevent the scenario of the return value being void, but the `?? operator only protects against `null` or `undefined`, but apparently not `void`.

markerikson:
I'd like to see a couple tests added that cover this scenario.

src/createReducer.ts
Jomik: Hm. Are you sure about that?
When you have a function that is returning void, you will get `undefined`.

```ts
function foo() {
  return;
}
console.log(foo() ?? "bar");
```
Will log "bar"

AndrewCraswell: Here's an example, this should hit some typing errors.
https://www.typescriptlang.org/play/#code/C4TwDgpgBAysCGxoF4oG8BQUoDd4BsBXCALigDtCBbAIwgCcBuDAX2Y3wmCipDkVKwESKAB9cAewCWAEyio0uAsTIBGAAxQ2GDADNC5AMbApE8lCQBnYAAprAsvxHic0mQEpHw6Jmz0uhPTm9iIA-KHoSkSCAExazCw6hmaWEpwAdPgSAOY2AFIwAPIAcunW9FLk2VK6IDZWtrxOEO6tjEA

AndrewCraswell:
@markerikson I've added the same immer checks that were being used, but with a draft state being passed to the reducer. This should give the same coverage for the new scenarios as existed for previous immer scenarios.

markerikson:
Thanks!

PR #513, opened 2020-04-19
allow to skip AsyncThunks using a condition callback

phryneas:
This is my first attempt at adding a `condition` callback so that people can skip `asyncThunk` execution before it even enters `pending` state, depending on `arg`, `getState()` and `extra`.

(Please don't merge this before #502 and #512, I guess order of merging might be of importance for merge conflicts here :laughing:)

### DIFF
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -188,10 +188,35 @@ type AsyncThunkActionCreator<
     : (arg: ThunkArg) => AsyncThunkAction<Returned, ThunkArg, ThunkApiConfig>
 >
 
+interface AsyncThunkOptions<
+  ThunkArg = void,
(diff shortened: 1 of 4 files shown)


src/createAsyncThunk.ts
```diff
@@ -220,6 +248,16 @@ If you want to use the AbortController to react to \`abort\` events, please cons
       const promise = (async function() {
         let finalAction: ReturnType<typeof fulfilled | typeof rejected>
         try {
+          if (
+            options &&
+            options.condition &&
+            !options.condition(arg, { getState, extra })
```
markerikson, line 254: I think we should make this a strict `=== false` comparison.  The user might return undefined or falsy by accident.  The assumption should be that the thunk is going to run unless they really say no.

phryneas: Good point.

PR #544, opened 2020-05-09
fix #541: invoke removeAll mutably within produce

### LINKED ISSUE #541: removeAll method of entity is not working.
https://github.com/reduxjs/redux-toolkit/blob/c89d5e2c6d3a4c1cc14a96db94d823314c4624e4/src/entities/unsorted_state_adapter.ts#L67-L72

Hi, teams.
I tested removeAll method, but it didn't work.
It looks incorrect usage of Object.assign.

```javascript
function removeAll(state: R): any { 
   return Object.assign(state, { 
     ids: [], 
     entities: {} 
   }) 
 } 
```

`state` parameter should be placed at first of params

can you check it?

phryneas:
This should invoke removeAll as a mutable action when wrapped in createNextState/produce and close #541.
Also, lots of tests.

### DIFF
--- a/src/entities/sorted_state_adapter.test.ts
+++ b/src/entities/sorted_state_adapter.test.ts
@@ -7,6 +7,7 @@ import {
   AClockworkOrange,
   AnimalFarm
 } from './fixtures/book'
+import { createNextState } from '..'
 
 describe('Sorted State Adapter', () => {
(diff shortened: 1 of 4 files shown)


markerikson:
LGTM.

MOAR TESTS!

PR #432, opened 2020-03-13
add createMiddleware utility

will-stone:
This is to accompany https://github.com/reduxjs/redux-toolkit/issues/431

I haven't written any docs yet; thought I'd wait for feedback first.

Allows for middleware to be written using the `builder` pattern:

```js
const customMiddleware = createMiddleware(builder =>
  builder
    .addCase(actionName, (action, dispatch) => {
      setInterval(() => dispatch(asyncActionHappened(action.payload)), 5000);
    })
    .addCase(anotherActionName, async (action, dispatch) => {
      const data = await fetch(action.payload).then(resp => resp.json());
      dispatch(dataFulfilled(data));
    })
);
```

PR #547, opened 2020-05-09
yet another attempt at actionListenerMiddleware

phryneas:
Reference: #237, #432, #272  .

So this is just the middleware and a few tests for it, I have not yet thought about how to usefully combine that into createDefaultMiddleware or even configureStore.

Let's discuss the middleware itself for now and take a look at that later.

The tests should showcase all functionality:

* subscribing via method on the middleware
* unsubscribing via method on the middleware
* unsubscribing via returned callback from subscribing
* subscribing via action
* unsubscribing via action
* unsubscribing via callback returned from dispatch

Also, I've added the following options to each individual subscription (taken some extra inspiration from https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener):

```ts
interface ActionListenerOptions<
  A extends AnyAction,
  S,
  _ extends Dispatch<AnyAction>
> {
  /**
   * Indicates that the listener should be removed after if has run once.
   */
  once?: boolean
  /**
   * If set to true, the action will not be forwarded to
   * * listeners that were registered after this listener
   * * middlewares later in the middleware chain
   * * reducers
   * If this listener is skipped due to `options.condition`, this has no effect.
   */
  preventPropagation?: boolean
  /**
   * A function that determines if the listener should run, depending on the action and probably the state.
   */
  condition?(action: A, getState: () => S): boolean
}
```

So, let's discuss this :)

### DIFF
--- a/src/createActionListenerMiddleware.ts
+++ b/src/createActionListenerMiddleware.ts
@@ -1,6 +1,7 @@
 import { Middleware, Dispatch, AnyAction, MiddlewareAPI, Action } from 'redux'
 import { TypedActionCreator } from './mapBuilders'
 import { createAction, BaseActionCreator } from './createAction'
+import { WithMiddlewareType } from './tsHelpers'
 
 export type When = 'before' | 'after' | undefined
 type WhenFromOptions<
@@ -349,5 +350,9 @@ export function createActionListenerMiddleware<
     }
   }
 
-  return Object.assign(middleware, { addListener, removeListener })
+  return Object.assign(
+    middleware,
+    { addListener, removeListener },
+    {} as WithMiddlewareType<typeof middleware>
+  )
 }
--- a/src/createActionListenerMiddleware.ts
+++ b/src/createActionListenerMiddleware.ts
@@ -37,8 +37,6 @@ interface AddListenerAction<
   type: 'actionListenerMiddleware/add'
   payload: {
     type: string
-  }
-  meta: {
     listener: ActionListener<A, S, D>
     options?: ActionListenerOptions<A, S, D>
   }
@@ -61,21 +59,19 @@ export const addListenerAction = createAction(
 
     return {
       payload: {
-        type
-      },
-      meta: {
+        type,
         listener,
         options
       }
     }
   }
 ) as BaseActionCreator<
-  { type: string },
-  'actionListenerMiddleware/add',
   {
+    type: string
     listener: ActionListener<any, any, any>
     options: ActionListenerOptions<any, any, any>
-  }
+  },
+  'actionListenerMiddleware/add'
 > & {
   <C extends TypedActionCreator<any>, S, D extends Dispatch>(
     actionCreator: C,
@@ -98,8 +94,6 @@ interface RemoveListenerAction<
   type: 'actionListenerMiddleware/remove'
   payload: {
     type: string
-  }
-  meta: {
     listener: ActionListener<A, S, D>
   }
 }
@@ -120,19 +114,14 @@ export const removeListenerAction = createAction(
 
     return {
       payload: {
-        type
-      },
-      meta: {
+        type,
         listener
       }
     }
   }
 ) as BaseActionCreator<
-  { type: string },
-  'actionListenerMiddleware/remove',
-  {
-    listener: ActionListener<any, any, any>
-  }
+  { type: string; listener: ActionListener<any, any, any> },
+  'actionListenerMiddleware/remove'
 > & {
   <C extends TypedActionCreator<any>, S, D extends Dispatch>(
     actionCreator: C,
@@ -156,7 +145,7 @@ export function createActionListenerMiddleware<
     listener: ActionListener<any, S, D>
   }
 
-  const listenerMap: Record<string, Set<ListenerEntry>> = {}
+  const listenerMap: Record<string, Set<ListenerEntry> | undefined> = {}
   const middleware: Middleware<
     {
       (action: Action<'actionListenerMiddleware/add'>): Unsubscribe
@@ -167,19 +156,20 @@ export function createActionListenerMiddleware<
     if (addListenerAction.match(action)) {
       const unsubscribe = addListener(
         action.payload.type,
-        action.meta.listener,
-        action.meta.options
+        action.payload.listener,
+        action.payload.options
       )
-      delete action.meta
-      next(action)
+
       return unsubscribe
     }
     if (removeListenerAction.match(action)) {
-      removeListener(action.payload.type, action.meta.listener)
+      removeListener(action.payload.type, action.payload.listener)
+
+      return
     }
 
-    if (listenerMap[action.type]) {
-      const listeners = listenerMap[action.type]
+    const listeners = listenerMap[action.type]
+    if (listeners) {
       for (const entry of listeners) {
         if (!entry.condition || entry.condition(action, api.getState)) {
           entry.listener(action, api)
@@ -217,11 +207,9 @@ export function createActionListenerMiddleware<
         ? typeOrActionCreator
         : typeOrActionCreator.type
 
-    if (!listenerMap[type]) {
-      listenerMap[type] = new Set()
-    }
+    const listeners = getListenerMap(type)
 
-    let entry = findListenerEntry(listenerMap[type], listener)
+    let entry = findListenerEntry(listeners, listener)
 
     if (!entry) {
       entry = {
@@ -229,10 +217,17 @@ export function createActionListenerMiddleware<
         listener
       }
 
-      listenerMap[type].add(entry)
+      listeners.add(entry)
     }
 
-    return () => listenerMap[type].delete(entry!)
+    return () => listeners.delete(entry!)
+  }
+
+  function getListenerMap(type: string) {
+    if (!listenerMap[type]) {
+      listenerMap[type] = new Set()
+    }
+    return listenerMap[type]!
   }
 
   function removeListener<C extends TypedActionCreator<any>>(
@@ -252,13 +247,19 @@ export function createActionListenerMiddleware<
         ? typeOrActionCreator
         : typeOrActionCreator.type
 
-    let entry = findListenerEntry(listenerMap[type], listener)
+    const listeners = listenerMap[type]
+
+    if (!listeners) {
+      return false
+    }
+
+    let entry = findListenerEntry(listeners, listener)
 
     if (!entry) {
       return false
     }
 
-    listenerMap[type].delete(entry!)
+    listeners.delete(entry)
     return true
   }
 
--- a/src/createActionListenerMiddleware.ts
+++ b/src/createActionListenerMiddleware.ts
@@ -44,6 +44,9 @@ interface AddListenerAction<
   }
 }
 
+/**
+ * @alpha
+ */
 export const addListenerAction = createAction(
   'actionListenerMiddleware/add',
   function prepare(
@@ -101,6 +104,9 @@ interface RemoveListenerAction<
   }
 }
 
+/**
+ * @alpha
+ */
 export const removeListenerAction = createAction(
   'actionListenerMiddleware/remove',
   function prepare(
@@ -139,6 +145,9 @@ export const removeListenerAction = createAction(
   ): RemoveListenerAction<AnyAction, S, D>
 }
 
+/**
+ * @alpha
+ */
 export function createActionListenerMiddleware<
   S,
   D extends Dispatch<AnyAction> = Dispatch
--- a/src/createActionListenerMiddleware.ts
+++ b/src/createActionListenerMiddleware.ts
@@ -2,12 +2,20 @@ import { Middleware, Dispatch, AnyAction, MiddlewareAPI, Action } from 'redux'
 import { TypedActionCreator } from './mapBuilders'
 import { createAction, BaseActionCreator } from './createAction'
 
+export type When = 'before' | 'after' | undefined
+type WhenFromOptions<
+  O extends ActionListenerOptions<any, any, any, any>
+> = O extends ActionListenerOptions<any, any, any, infer W> ? W : never
+
 /**
  * @alpha
  */
-export interface ActionListenerMiddlewareAPI<S, D extends Dispatch<AnyAction>>
-  extends MiddlewareAPI<D, S> {
-  stopPropagation(): void
+export interface ActionListenerMiddlewareAPI<
+  S,
+  D extends Dispatch<AnyAction>,
+  O extends ActionListenerOptions<any, any, any, any>
+> extends MiddlewareAPI<D, S> {
+  stopPropagation: WhenFromOptions<O> extends 'after' ? undefined : () => void
 }
 
 /**
@@ -16,13 +24,15 @@ export interface ActionListenerMiddlewareAPI<S, D extends Dispatch<AnyAction>>
 export type ActionListener<
   A extends AnyAction,
   S,
-  D extends Dispatch<AnyAction>
-> = (action: A, api: ActionListenerMiddlewareAPI<S, D>) => void
+  D extends Dispatch<AnyAction>,
+  O extends ActionListenerOptions<any, any, any, any>
+> = (action: A, api: ActionListenerMiddlewareAPI<S, D, O>) => void
 
 export interface ActionListenerOptions<
   A extends AnyAction,
   S,
-  _ extends Dispatch<AnyAction>
+  _ extends Dispatch<AnyAction>,
+  W extends When = 'before'
 > {
   /**
    * Indicates that the listener should be removed after if has run once.
@@ -32,18 +42,25 @@ export interface ActionListenerOptions<
    * A function that determines if the listener should run, depending on the action and probably the state.
    */
   condition?(action: A, getState: () => S): boolean
+  /**
+   * Determines if the listener runs 'before' or 'after' the reducers have been called.
+   * If set to 'before', calling `api.stopPropagation()` from the listener becomes possible.
+   * Defaults to 'before'.
+   */
+  when?: W
 }
 
 export interface AddListenerAction<
   A extends AnyAction,
   S,
-  D extends Dispatch<AnyAction>
+  D extends Dispatch<AnyAction>,
+  O extends ActionListenerOptions<A, S, D, When>
 > {
   type: 'actionListenerMiddleware/add'
   payload: {
     type: string
-    listener: ActionListener<A, S, D>
-    options?: ActionListenerOptions<A, S, D>
+    listener: ActionListener<A, S, D, O>
+    options?: O
   }
 }
 
@@ -54,7 +71,7 @@ export const addListenerAction = createAction(
   'actionListenerMiddleware/add',
   function prepare(
     typeOrActionCreator: string | TypedActionCreator<string>,
-    listener: ActionListener<any, any, any>,
+    listener: ActionListener<any, any, any, any>,
     options?: ActionListenerOptions<AnyAction, any, any>
   ) {
     const type =
@@ -73,22 +90,31 @@ export const addListenerAction = createAction(
 ) as BaseActionCreator<
   {
     type: string
-    listener: ActionListener<any, any, any>
+    listener: ActionListener<any, any, any, any>
     options: ActionListenerOptions<any, any, any>
   },
   'actionListenerMiddleware/add'
 > & {
-  <C extends TypedActionCreator<any>, S, D extends Dispatch>(
+  <
+    C extends TypedActionCreator<any>,
+    S,
+    D extends Dispatch,
+    O extends ActionListenerOptions<ReturnType<C>, S, D, When>
+  >(
     actionCreator: C,
-    listener: ActionListener<ReturnType<C>, S, D>,
-    options?: ActionListenerOptions<ReturnType<C>, S, D>
-  ): AddListenerAction<ReturnType<C>, S, D>
+    listener: ActionListener<ReturnType<C>, S, D, O>,
+    options?: O
+  ): AddListenerAction<ReturnType<C>, S, D, O>
 
-  <S, D extends Dispatch>(
+  <
+    S,
+    D extends Dispatch,
+    O extends ActionListenerOptions<AnyAction, S, D, When>
+  >(
     type: string,
-    listener: ActionListener<AnyAction, S, D>,
-    options?: ActionListenerOptions<AnyAction, S, D>
-  ): AddListenerAction<AnyAction, S, D>
+    listener: ActionListener<AnyAction, S, D, O>,
+    options?: O
+  ): AddListenerAction<AnyAction, S, D, O>
 }
 
 interface RemoveListenerAction<
@@ -99,7 +125,7 @@ interface RemoveListenerAction<
   type: 'actionListenerMiddleware/remove'
   payload: {
     type: string
-    listener: ActionListener<A, S, D>
+    listener: ActionListener<A, S, D, any>
   }
 }
 
@@ -110,7 +136,7 @@ export const removeListenerAction = createAction(
   'actionListenerMiddleware/remove',
   function prepare(
     typeOrActionCreator: string | TypedActionCreator<string>,
-    listener: ActionListener<any, any, any>
+    listener: ActionListener<any, any, any, any>
   ) {
     const type =
       typeof typeOrActionCreator === 'string'
@@ -125,17 +151,17 @@ export const removeListenerAction = createAction(
     }
   }
 ) as BaseActionCreator<
-  { type: string; listener: ActionListener<any, any, any> },
+  { type: string; listener: ActionListener<any, any, any, any> },
   'actionListenerMiddleware/remove'
 > & {
   <C extends TypedActionCreator<any>, S, D extends Dispatch>(
     actionCreator: C,
-    listener: ActionListener<ReturnType<C>, S, D>
+    listener: ActionListener<ReturnType<C>, S, D, any>
   ): RemoveListenerAction<ReturnType<C>, S, D>
 
   <S, D extends Dispatch>(
     type: string,
-    listener: ActionListener<AnyAction, S, D>
+    listener: ActionListener<AnyAction, S, D, any>
   ): RemoveListenerAction<AnyAction, S, D>
 }
 
@@ -146,8 +172,8 @@ export function createActionListenerMiddleware<
   S,
   D extends Dispatch<AnyAction> = Dispatch
 >() {
-  type ListenerEntry = ActionListenerOptions<any, S, D> & {
-    listener: ActionListener<any, S, D>
+  type ListenerEntry = ActionListenerOptions<any, S, D, When> & {
+    listener: ActionListener<any, S, D, any>
   }
 
   const listenerMap: Record<string, Set<ListenerEntry> | undefined> = {}
@@ -177,18 +203,35 @@ export function createActionListenerMiddleware<
     if (listeners) {
       for (const entry of listeners) {
         if (!entry.condition || entry.condition(action, api.getState)) {
-          let preventPropagation = false
-          entry.listener(action, {
-            ...api,
-            stopPropagation() {
-              preventPropagation = true
-            }
-          })
           if (entry.once) {
             listeners.delete(entry)
           }
-          if (preventPropagation) {
-            return action
+
+          if (entry.when != 'after') {
+            /* before */
+            let stoppedPropagation = false
+            entry.listener(action, {
+              ...api,
+              stopPropagation() {
+                stoppedPropagation = true
+              }
+            })
+            if (stoppedPropagation) {
+              return action
+            }
+            return next(action)
+          } else {
+            /* after */
+            const result = next(action)
+            entry.listener(action, {
+              ...api,
+              stopPropagation: () => {
+                throw new Error(
+                  'stopPropagation can only be called by action listeners with the `when` option set to "before"'
+                )
+              }
+            })
+            return result
           }
         }
       }
@@ -198,20 +241,26 @@ export function createActionListenerMiddleware<
 
   type Unsubscribe = () => void
 
-  function addListener<C extends TypedActionCreator<any>>(
+  function addListener<
+    C extends TypedActionCreator<any>,
+    O extends ActionListenerOptions<ReturnType<C>, S, D, When>
+  >(
     actionCreator: C,
-    listener: ActionListener<ReturnType<C>, S, D>,
-    options?: ActionListenerOptions<ReturnType<C>, S, D>
+    listener: ActionListener<ReturnType<C>, S, D, O>,
+    options?: O
   ): Unsubscribe
-  function addListener(
-    type: string,
-    listener: ActionListener<AnyAction, S, D>,
-    options?: ActionListenerOptions<AnyAction, S, D>
+  function addListener<
+    T extends string,
+    O extends ActionListenerOptions<Action<T>, S, D, When>
+  >(
+    type: T,
+    listener: ActionListener<Action<T>, S, D, O>,
+    options?: O
   ): Unsubscribe
   function addListener(
     typeOrActionCreator: string | TypedActionCreator<any>,
-    listener: ActionListener<AnyAction, S, D>,
-    options?: ActionListenerOptions<AnyAction, S, D>
+    listener: ActionListener<AnyAction, S, D, any>,
+    options?: ActionListenerOptions<AnyAction, S, D, When>
   ): Unsubscribe {
     const type =
       typeof typeOrActionCreator === 'string'
@@ -243,15 +292,15 @@ export function createActionListenerMiddleware<
 
   function removeListener<C extends TypedActionCreator<any>>(
     actionCreator: C,
-    listener: ActionListener<ReturnType<C>, S, D>
+    listener: ActionListener<ReturnType<C>, S, D, any>
   ): boolean
   function removeListener(
     type: string,
-    listener: ActionListener<AnyAction, S, D>
+    listener: ActionListener<AnyAction, S, D, any>
   ): boolean
   function removeListener(
     typeOrActionCreator: string | TypedActionCreator<any>,
-    listener: ActionListener<AnyAction, S, D>
+    listener: ActionListener<AnyAction, S, D, any>
   ): boolean {
     const type =
       typeof typeOrActionCreator === 'string'
--- a/src/createActionListenerMiddleware.ts
+++ b/src/createActionListenerMiddleware.ts
@@ -5,8 +5,8 @@ import { WithMiddlewareType } from './tsHelpers'
 
 export type When = 'before' | 'after' | undefined
 type WhenFromOptions<
-  O extends ActionListenerOptions<any, any, any, any>
-> = O extends ActionListenerOptions<any, any, any, infer W> ? W : never
+  O extends ActionListenerOptions
+> = O extends ActionListenerOptions ? O['when'] : never
 
 /**
  * @alpha
@@ -14,9 +14,10 @@ type WhenFromOptions<
 export interface ActionListenerMiddlewareAPI<
   S,
   D extends Dispatch<AnyAction>,
-  O extends ActionListenerOptions<any, any, any, any>
+  O extends ActionListenerOptions
 > extends MiddlewareAPI<D, S> {
-  stopPropagation: WhenFromOptions<O> extends 'after' ? undefined : () => void
+  stopPropagation: WhenFromOptions<O> extends 'before' ? () => void : undefined
+  unsubscribe(): void
 }
 
 /**
@@ -26,36 +27,23 @@ export type ActionListener<
   A extends AnyAction,
   S,
   D extends Dispatch<AnyAction>,
-  O extends ActionListenerOptions<any, any, any, any>
+  O extends ActionListenerOptions
 > = (action: A, api: ActionListenerMiddlewareAPI<S, D, O>) => void
 
-export interface ActionListenerOptions<
-  A extends AnyAction,
-  S,
-  _ extends Dispatch<AnyAction>,
-  W extends When = 'before'
-> {
-  /**
-   * Indicates that the listener should be removed after if has run once.
-   */
-  once?: boolean
-  /**
-   * A function that determines if the listener should run, depending on the action and probably the state.
-   */
-  condition?(action: A, getState: () => S): boolean
+export interface ActionListenerOptions {
   /**
    * Determines if the listener runs 'before' or 'after' the reducers have been called.
    * If set to 'before', calling `api.stopPropagation()` from the listener becomes possible.
    * Defaults to 'before'.
    */
-  when?: W
+  when?: When
 }
 
 export interface AddListenerAction<
   A extends AnyAction,
   S,
   D extends Dispatch<AnyAction>,
-  O extends ActionListenerOptions<A, S, D, When>
+  O extends ActionListenerOptions
 > {
   type: 'actionListenerMiddleware/add'
   payload: {
@@ -73,7 +61,7 @@ export const addListenerAction = createAction(
   function prepare(
     typeOrActionCreator: string | TypedActionCreator<string>,
     listener: ActionListener<any, any, any, any>,
-    options?: ActionListenerOptions<AnyAction, any, any>
+    options?: ActionListenerOptions
   ) {
     const type =
       typeof typeOrActionCreator === 'string'
@@ -92,7 +80,7 @@ export const addListenerAction = createAction(
   {
     type: string
     listener: ActionListener<any, any, any, any>
-    options: ActionListenerOptions<any, any, any>
+    options: ActionListenerOptions
   },
   'actionListenerMiddleware/add'
 > & {
@@ -100,18 +88,14 @@ export const addListenerAction = createAction(
     C extends TypedActionCreator<any>,
     S,
     D extends Dispatch,
-    O extends ActionListenerOptions<ReturnType<C>, S, D, When>
+    O extends ActionListenerOptions
   >(
     actionCreator: C,
     listener: ActionListener<ReturnType<C>, S, D, O>,
     options?: O
   ): AddListenerAction<ReturnType<C>, S, D, O>
 
-  <
-    S,
-    D extends Dispatch,
-    O extends ActionListenerOptions<AnyAction, S, D, When>
-  >(
+  <S, D extends Dispatch, O extends ActionListenerOptions>(
     type: string,
     listener: ActionListener<AnyAction, S, D, O>,
     options?: O
@@ -173,7 +157,7 @@ export function createActionListenerMiddleware<
   S,
   D extends Dispatch<AnyAction> = Dispatch
 >() {
-  type ListenerEntry = ActionListenerOptions<any, S, D, When> & {
+  type ListenerEntry = ActionListenerOptions & {
     listener: ActionListener<any, S, D, any>
   }
 
@@ -202,55 +186,48 @@ export function createActionListenerMiddleware<
 
     const listeners = listenerMap[action.type]
     if (listeners) {
-      /* before */
-      for (const entry of listeners) {
-        if (entry.when == 'after') {
-          continue
-        }
-
-        if (!entry.condition || entry.condition(action, api.getState)) {
-          if (entry.once) {
-            listeners.delete(entry)
+      const defaultWhen = 'after'
+      let result: unknown
+      for (const phase of ['before', 'after'] as const) {
+        for (const entry of listeners) {
+          if (phase !== (entry.when || defaultWhen)) {
+            continue
           }
-
           let stoppedPropagation = false
+          let currentPhase = phase
+          let synchronousListenerFinished = false
           entry.listener(action, {
             ...api,
             stopPropagation() {
-              stoppedPropagation = true
+              if (currentPhase === 'before') {
+                if (!synchronousListenerFinished) {
+                  stoppedPropagation = true
+                } else {
+                  throw new Error(
+                    'stopPropagation can only be called synchronously'
+                  )
+                }
+              } else {
+                throw new Error(
+                  'stopPropagation can only be called by action listeners with the `when` option set to "before"'
+                )
+              }
+            },
+            unsubscribe() {
+              listeners.delete(entry)
             }
           })
+          synchronousListenerFinished = true
           if (stoppedPropagation) {
             return action
           }
         }
-      }
-
-      const result = next(action)
-
-      /* after */
-      for (const entry of listeners) {
-        if (entry.when != 'after') {
-          continue
-        }
-        if (!entry.condition || entry.condition(action, api.getState)) {
-          if (entry.once) {
-            listeners.delete(entry)
-          }
-
-          /* after */
-          entry.listener(action, {
-            ...api,
-            stopPropagation: () => {
-              throw new Error(
-                'stopPropagation can only be called by action listeners with the `when` option set to "before"'
-              )
-            }
-          })
+        if (phase === 'before') {
+          result = next(action)
+        } else {
+          return result
         }
       }
-
-      return result
     }
     return next(action)
   }
@@ -259,16 +236,13 @@ export function createActionListenerMiddleware<
 
   function addListener<
     C extends TypedActionCreator<any>,
-    O extends ActionListenerOptions<ReturnType<C>, S, D, When>
+    O extends ActionListenerOptions
   >(
     actionCreator: C,
     listener: ActionListener<ReturnType<C>, S, D, O>,
     options?: O
   ): Unsubscribe
-  function addListener<
-    T extends string,
-    O extends ActionListenerOptions<Action<T>, S, D, When>
-  >(
+  function addListener<T extends string, O extends ActionListenerOptions>(
     type: T,
     listener: ActionListener<Action<T>, S, D, O>,
     options?: O
@@ -276,7 +250,7 @@ export function createActionListenerMiddleware<
   function addListener(
     typeOrActionCreator: string | TypedActionCreator<any>,
     listener: ActionListener<AnyAction, S, D, any>,
-    options?: ActionListenerOptions<AnyAction, S, D, When>
+    options?: ActionListenerOptions
   ): Unsubscribe {
     const type =
       typeof typeOrActionCreator === 'string'
--- a/src/createActionListenerMiddleware.ts
+++ b/src/createActionListenerMiddleware.ts
@@ -201,40 +201,55 @@ export function createActionListenerMiddleware<
 
     const listeners = listenerMap[action.type]
     if (listeners) {
+      /* before */
       for (const entry of listeners) {
+        if (entry.when == 'after') {
+          continue
+        }
+
         if (!entry.condition || entry.condition(action, api.getState)) {
           if (entry.once) {
             listeners.delete(entry)
           }
 
-          if (entry.when != 'after') {
-            /* before */
-            let stoppedPropagation = false
-            entry.listener(action, {
-              ...api,
-              stopPropagation() {
-                stoppedPropagation = true
-              }
-            })
-            if (stoppedPropagation) {
-              return action
+          let stoppedPropagation = false
+          entry.listener(action, {
+            ...api,
+            stopPropagation() {
+              stoppedPropagation = true
             }
-            return next(action)
-          } else {
-            /* after */
-            const result = next(action)
-            entry.listener(action, {
-              ...api,
-              stopPropagation: () => {
-                throw new Error(
-                  'stopPropagation can only be called by action listeners with the `when` option set to "before"'
-                )
-              }
-            })
-            return result
+          })
+          if (stoppedPropagation) {
+            return action
           }
         }
       }
+
+      const result = next(action)
+
+      /* after */
+      for (const entry of listeners) {
+        if (entry.when != 'after') {
+          continue
+        }
+        if (!entry.condition || entry.condition(action, api.getState)) {
+          if (entry.once) {
+            listeners.delete(entry)
+          }
+
+          /* after */
+          entry.listener(action, {
+            ...api,
+            stopPropagation: () => {
+              throw new Error(
+                'stopPropagation can only be called by action listeners with the `when` option set to "before"'
+              )
+            }
+          })
+        }
+      }
+
+      return result
     }
     return next(action)
   }
--- a/src/createActionListenerMiddleware.ts
+++ b/src/createActionListenerMiddleware.ts
@@ -2,11 +2,24 @@ import { Middleware, Dispatch, AnyAction, MiddlewareAPI, Action } from 'redux'
 import { TypedActionCreator } from './mapBuilders'
 import { createAction, BaseActionCreator } from './createAction'
 
-type ActionListener<A extends AnyAction, S, D extends Dispatch<AnyAction>> = (
-  action: A,
-  api: MiddlewareAPI<D, S>
-) => void
-interface ActionListenerOptions<
+/**
+ * @alpha
+ */
+export interface ActionListenerMiddlewareAPI<S, D extends Dispatch<AnyAction>>
+  extends MiddlewareAPI<D, S> {
+  stopPropagation(): void
+}
+
+/**
+ * @alpha
+ */
+export type ActionListener<
+  A extends AnyAction,
+  S,
+  D extends Dispatch<AnyAction>
+> = (action: A, api: ActionListenerMiddlewareAPI<S, D>) => void
+
+export interface ActionListenerOptions<
   A extends AnyAction,
   S,
   _ extends Dispatch<AnyAction>
@@ -16,20 +29,12 @@ interface ActionListenerOptions<
    */
   once?: boolean
   /**
-   * If set to true, the action will not be forwarded to
-   * * listeners that were registered after this listener
-   * * middlewares later in the middleware chain
-   * * reducers
-   * If this listener is skipped due to `options.condition`, this has no effect.
-   */
-  preventPropagation?: boolean
-  /**
    * A function that determines if the listener should run, depending on the action and probably the state.
    */
   condition?(action: A, getState: () => S): boolean
 }
 
-interface AddListenerAction<
+export interface AddListenerAction<
   A extends AnyAction,
   S,
   D extends Dispatch<AnyAction>
@@ -172,11 +177,17 @@ export function createActionListenerMiddleware<
     if (listeners) {
       for (const entry of listeners) {
         if (!entry.condition || entry.condition(action, api.getState)) {
-          entry.listener(action, api)
+          let preventPropagation = false
+          entry.listener(action, {
+            ...api,
+            stopPropagation() {
+              preventPropagation = true
+            }
+          })
           if (entry.once) {
             listeners.delete(entry)
           }
-          if (entry.preventPropagation) {
+          if (preventPropagation) {
             return action
           }
         }
--- a/src/createActionListenerMiddleware.ts
+++ b/src/createActionListenerMiddleware.ts
@@ -1,6 +1,6 @@
 import { Middleware, Dispatch, AnyAction, MiddlewareAPI, Action } from 'redux'
-import { TypedActionCreator } from '../mapBuilders'
-import { createAction, BaseActionCreator } from '../createAction'
+import { TypedActionCreator } from './mapBuilders'
+import { createAction, BaseActionCreator } from './createAction'
 
 type ActionListener<A extends AnyAction, S, D extends Dispatch<AnyAction>> = (
   action: A,
--- a/src/entities/createActionListenerMiddleware.test.ts
+++ b/src/entities/createActionListenerMiddleware.test.ts
@@ -1,0 +1,236 @@
+import { configureStore } from '../configureStore'
+import {
+  createActionListenerMiddleware,
+  addListenerAction,
+  removeListenerAction
+} from './createActionListenerMiddleware'
+import { createAction } from '../createAction'
+
+const middlewareApi = {
+  getState: expect.any(Function),
+  dispatch: expect.any(Function)
+}
+
+describe('createActionListenerMiddleware', () => {
+  let store = configureStore({
+    reducer: () => ({}),
+    middleware: [createActionListenerMiddleware()] as const
+  })
+  let reducer: jest.Mock
+  let middleware: ReturnType<typeof createActionListenerMiddleware>
+
+  const testAction1 = createAction<string>('testAction1')
+  type TestAction1 = ReturnType<typeof testAction1>
+  const testAction2 = createAction<string>('testAction2')
+
+  beforeEach(() => {
+    middleware = createActionListenerMiddleware()
+    reducer = jest.fn(() => ({}))
+    store = configureStore({
+      reducer,
+      middleware: [middleware] as const
+    })
+  })
+
+  test('directly subscribing', () => {
+    const listener = jest.fn((_: TestAction1) => {})
+
+    middleware.addListener(testAction1, listener)
+
+    store.dispatch(testAction1('a'))
+    store.dispatch(testAction2('b'))
+    store.dispatch(testAction1('c'))
+
+    expect(listener.mock.calls).toEqual([
+      [testAction1('a'), middlewareApi],
+      [testAction1('c'), middlewareApi]
+    ])
+  })
+
+  test('subscribing with the same listener will not make it trigger twice (like EventTarget.addEventListener())', () => {
+    /**
+     * thoughts: allow to use this to override the options for a listener?
+     * right now it's just exiting if the listener is already registered
+     */
+
+    const listener = jest.fn((_: TestAction1) => {})
+
+    middleware.addListener(testAction1, listener)
+    middleware.addListener(testAction1, listener)
+
+    store.dispatch(testAction1('a'))
+    store.dispatch(testAction2('b'))
+    store.dispatch(testAction1('c'))
+
+    expect(listener.mock.calls).toEqual([
+      [testAction1('a'), middlewareApi],
+      [testAction1('c'), middlewareApi]
+    ])
+  })
+
+  test('unsubscribing via callback', () => {
+    const listener = jest.fn((_: TestAction1) => {})
+
+    const unsubscribe = middleware.addListener(testAction1, listener)
+
+    store.dispatch(testAction1('a'))
+    unsubscribe()
+    store.dispatch(testAction2('b'))
+    store.dispatch(testAction1('c'))
+
+    expect(listener.mock.calls).toEqual([[testAction1('a'), middlewareApi]])
+  })
+
+  test('directly unsubscribing', () => {
+    const listener = jest.fn((_: TestAction1) => {})
+
+    middleware.addListener(testAction1, listener)
+
+    store.dispatch(testAction1('a'))
+
+    middleware.removeListener(testAction1, listener)
+    store.dispatch(testAction2('b'))
+    store.dispatch(testAction1('c'))
+
+    expect(listener.mock.calls).toEqual([[testAction1('a'), middlewareApi]])
+  })
+
+  test('subscribing via action', () => {
+    const listener = jest.fn((_: TestAction1) => {})
+
+    store.dispatch(addListenerAction(testAction1, listener))
+
+    store.dispatch(testAction1('a'))
+    store.dispatch(testAction2('b'))
+    store.dispatch(testAction1('c'))
+
+    expect(listener.mock.calls).toEqual([
+      [testAction1('a'), middlewareApi],
+      [testAction1('c'), middlewareApi]
+    ])
+  })
+
+  test('unsubscribing via callback from dispatch', () => {
+    const listener = jest.fn((_: TestAction1) => {})
+
+    const unsubscribe = store.dispatch(addListenerAction(testAction1, listener))
+
+    store.dispatch(testAction1('a'))
+    // @ts-ignore TODO types
+    unsubscribe()
+    store.dispatch(testAction2('b'))
+    store.dispatch(testAction1('c'))
+
+    expect(listener.mock.calls).toEqual([[testAction1('a'), middlewareApi]])
+  })
+
+  test('unsubscribing via action', () => {
+    const listener = jest.fn((_: TestAction1) => {})
+
+    middleware.addListener(testAction1, listener)
+
+    store.dispatch(testAction1('a'))
+
+    store.dispatch(removeListenerAction(testAction1, listener))
+    store.dispatch(testAction2('b'))
+    store.dispatch(testAction1('c'))
+
+    expect(listener.mock.calls).toEqual([[testAction1('a'), middlewareApi]])
+  })
+
+  test('"condition" allows to skip the listener', () => {
+    const listener = jest.fn((_: TestAction1) => {})
+
+    middleware.addListener(testAction1, listener, {
+      condition(action) {
+        return action.payload !== 'b'
+      }
+    })
+
+    store.dispatch(testAction1('a'))
+    store.dispatch(testAction1('b'))
+    store.dispatch(testAction1('c'))
+
+    expect(listener.mock.calls).toEqual([
+      [testAction1('a'), middlewareApi],
+      [testAction1('c'), middlewareApi]
+    ])
+  })
+
+  test('"once" unsubscribes the listener automatically after one use', () => {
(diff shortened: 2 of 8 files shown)


src/entities/createActionListenerMiddleware.ts
```diff
@@ -0,0 +1,268 @@
...
+    D
+  > = api => next => action => {
+    if (addListenerAction.match(action)) {
+      const unsubscribe = addListener(
+        action.payload.type,
+        action.meta.listener,
+        action.meta.options
+      )
+      delete action.meta
+      next(action)
```
markerikson, line 165: Would we actually want to forward the `addListener` action onwards at all?

src/entities/createActionListenerMiddleware.ts
```diff
@@ -0,0 +1,268 @@
...
+    const type =
+      typeof typeOrActionCreator === 'string'
+        ? typeOrActionCreator
+        : (typeOrActionCreator as TypedActionCreator<string>).type
+
+    return {
+      payload: {
+        type
+      },
+      meta: {
```
markerikson, line 63: Feels like `listener` and `options` would be better suited for `payload`, conceptually. After all, those are key aspects of this particular action, and the middleware is explicitly using them.

src/entities/createActionListenerMiddleware.test.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    expect(reducer.mock.calls).toEqual([[{}, testAction1('a')]])
+  })
+
+  test('"preventPropagation" prevents actions from being forwarded to the store', () => {
+    reducer.mockClear()
+
+    const listener = jest.fn((_: TestAction1) => {})
+
+    middleware.addListener(testAction1, listener, { preventPropagation: true })
```
markerikson, line 208: Huh. This case is interesting.

Would we want `preventPropagation` to be a single boolean?  Would we want _this_ to be another `condition`-type callback as well?

phryneas: Might be an idea.
I was also thinking about doing something like `return preventPropagation` from the listener instead - or even adding it as an option on the action itself, so you could call `action.stopPropagation()`, just like `event.stopPropagation()`.

This seemed just the easiest for now.

src/entities/createActionListenerMiddleware.ts
phryneas: I do have something in the back of my mind that someone talked about maybe wanting to track all active subscriptions in a state or so - in that case it would definitely be useful. On the other hand, the `delete action.meta` would be counter-productive then.

Also, if it is passed thorugh, it can be observed in the devTools, wich might be nice.

src/entities/createActionListenerMiddleware.ts
phryneas: Yeah, no particular reason to put these in `meta` tbh, other than that I personally try to keep non-serializable stuff out of `payload`.

src/entities/createActionListenerMiddleware.test.ts
```diff
@@ -0,0 +1,236 @@
...
+      [testAction1('c'), middlewareApi]
+    ])
+  })
+
+  test('subscribing with the same listener will not make it trigger twice (like EventTarget.addEventListener())', () => {
+    /**
+     * thoughts: allow to use this to override the options for a listener?
+     * right now it's just exiting if the listener is already registered
+     */
+
```
phryneas, line 55: Any thoughts on that?

src/entities/createActionListenerMiddleware.ts
mpeyper: Yeah, [that was me](https://github.com/reduxjs/redux-toolkit/issues/237#issuecomment-590269935) (back when I still thought I'd have the time to look at this 😓), although I did concede that in practise, nothing ever actually reacted to those events.  The observability in the dev-tools has saved us on multiple occasions, so I'd advocate for keeping it.

The only thing I'd recommend is cleansing the action of any functions so that it serialisable by the time it hit the store. I thought RTK actually raised a warning if the action wasn't?

mpeyper: Just noticed the the `delete action.meta`.

Not sure if you want to modify the original action in case the dispatcher is reusing the same action (i.e. if they're adding and removing listeners on component mount/dismount).

markerikson: Is there really a meaningful benefit to having actions like "add/remove listener" show up in the DevTools? I've never been a fan of the idea of dispatching actions _just_ to get a logging effect.

src/entities/createActionListenerMiddleware.test.ts
markerikson: Maybe include `stopPropagation()` in the `api` object?

src/entities/createActionListenerMiddleware.test.ts
markerikson: Don't feel like there's a lot of reason to change the options for a listener once added.

markerikson:
Just updated the PR by moving the files from `src/entities` to `src`, and exporting the middleware.  That way we at least get a usable build out of CodeSandbox.

nogwater:
I started playing with this in codesandbox, starting from [priceless-gates-bu5x2](https://codesandbox.io/s/priceless-gates-bu5x2).  I made a [simple counter component](https://codesandbox.io/s/vigorous-brook-7z0je?file=/src/components/Counter.tsx) that listens for the displayRepo, increments the count, and displays the current count.  I updated store.ts with:

```TypeScript
const actionListenerMiddleware = createActionListenerMiddleware()
export const addActionListener = actionListenerMiddleware.addListener

const middleware = [...getDefaultMiddleware(), actionListenerMiddleware]

  middleware,
  reducer: rootReducer
})
```
Is this a good way to expose `actionListenerMiddleware.addListener`?

For the counter itself, I initially added this:
```TypeScript
  addActionListener(displayRepo, action => {
    console.log(action)
    setCount(count + 1)
  })
```
The count works as I'd want, but I noticed that it prints an additional log entry every time I click `Load Repo`.  So, the first time I click it, it prints one entry, the second time it prints two entries, and so on.  Clicking `Load Repo` is causing my Counter component to rerender, which causes a new listener to be registered.

To fix it, I used useEffect to unsubscribe the previous listener each time I add a new one:
```TypeScript
  useEffect(() => {
    const unsub = addActionListener(displayRepo, action => {
      console.log(action)
      setCount(count + 1)
    })
    return unsub
  }, [count, setCount])
```

Is this what I should be doing?  Should I make a custom `useActionListener` hook that handles the useEffect boilerplate?

markerikson:
@nogwater : the middleware has the ability to add listeners by dispatching an action, so you don't have to import that method from the middleware itself:

```js
const unsubscribe = store.dispatch(addListenerAction(testAction1, listener))
```

For this specific example, I'd suggest changing the setter call to use the functional updater form:

```js
setCount(count => count + 1)
```

That way you don't have to depend on `count` in the deps array, and you can just add the listener once on mount (ie, an empty deps array).  Also, you don't have to have `setCount` in the deps array, because `useState` setters are always a stable reference.

It's interesting you jumped straight to doing that in a component.  @phryneas had specific concerns about that as a possible anti-ptatern at https://github.com/reduxjs/redux-toolkit/issues/237#issuecomment-590081833 .  My thought a couple comments later was that [I'm not overly concerned about that myself](https://github.com/reduxjs/redux-toolkit/issues/237#issuecomment-590086649v).

nogwater:
@markerikson Thanks for the tips, and pointing to comments.  I'll probably take a crack at rebuilding this without putting the subscription in the component tomorrow.  Until then, I tried applying your suggestions and ended up with a couple warnings.

```TypeScript
  const dispatch = useDispatch()
  useEffect(() => {
    const unsubscribe = dispatch(
      addListenerAction(displayRepo, () => setCount(count => count + 1))
    )
    return unsubscribe
  }, [])
```

The first TS is that unsubscribe's type,  `AddListenerAction<...>`, isn't compatible with what `useEffect` expects for a callback, `EffectCallback`.  The second is that eslint wants `dispatch` added to the dependencies.

markerikson:
Yep, `[dispatch]` is a normal deps requirement, because the lint rule doesn't know that is going to stay stable.  (Technically it could switch if you swapped out the store instance, but the lint rule doesn't actually know that either.)

Looking at the React types, I see:

```ts
    // NOTE: callbacks are _only_ allowed to return either void, or a destructor.
    // The destructor is itself only allowed to return void.
    type EffectCallback = () => (void | (() => void | undefined));
```

So technically `EffectCallback` is the callback function you're passing to `useEffect` itself, not the cleanup function you return.

I think part of the issue here is that `dispatch()`, by default, returns the action object you've passed in.  Middleware can change that, but you have to modify the store type to make TS know that...

Oh.  Oh no.

@phryneas !!!!!  Guess what fun type issue we just ran into?

THE EXACT SAME THING I WAS DOING LAST YEAR WITH CONVINCING TS THAT DISPATCHING A SPECIFIC ACTION MAKES `dispatch` ACTUALLY RETURN A PROMISE! :)

image

It thinks that `unsubscribe` is the action we dispatched, not the cleanup function the middleware returns, and from TS's perspective an "action object" is certainly not the same type as a "function that returns void".

I know you'd come up with a nicer way to handle that, and I _think_ I'd used that technique in my work project, but I don't have it handy right this second.  Do you remember what it was?  If not, I'll have to look it up later.

phryneas:
Heh, you ran into my type-related TODO. Dispatch return type isn't correct automatically.

https://github.com/reduxjs/redux-toolkit/pull/547/files#diff-e0913c885878b627558b1f6b39f9a382R119

For now, just add a `// @ts-ignore` above the `return`, I'm gonna fix it before it's ready.

src/entities/createActionListenerMiddleware.ts
phryneas: Honestly, 90% of the time I look into the devtools to just look at the action names (sometimes payloads) to get a feeling for the flow what happened.
10% for state changes. Never use time travel.
So I'm very much in favor of forwarding it, but I agree that it has to have some use.

Right now I'm thinking about preventing the "listen" action from propagating and dispatching a new "started_listening" action, which might also contain a unique id that could be used to unsubscribe that listener.

src/entities/createActionListenerMiddleware.test.ts
phryneas: That's a good idea!

src/entities/createActionListenerMiddleware.test.ts
phryneas: Yeah, one could always unsubscribe & resubscribe with different options. No need to make it too complicated.

phryneas:
> The count works as I'd want, but I noticed that it prints an additional log entry every time I click `Load Repo`. So, the first time I click it, it prints one entry, the second time it prints two entries, and so on. Clicking `Load Repo` is causing my Counter component to rerender, which causes a new listener to be registered.
> 
> To fix it, I used useEffect to unsubscribe the previous listener each time I add a new one:
> 
> ```ts
>   useEffect(() => {
>     const unsub = addActionListener(displayRepo, action => {
>       console.log(action)
>       setCount(count + 1)
>     })
>     return unsub
>   }, [count, setCount])
> ```

Yup, you'd have to use a `useEffect` with an unsubscribe callback, because while the middleware prevents the *same* listener from being added twice, if you define the listener function within the render method, it will be a different function each time - so both functions would be subscribed since they're distinct from each other.

src/entities/createActionListenerMiddleware.ts
phryneas: Hmm. While trying to implement this "started_listening", it seems non-trivial:
* do we dispatch that action only when it is subscribed via `addListener` action?
* if we also dispatch it, what if the middleware isn't used by a store yet - or used by multiple stores?
* would we dispatch something different if a listener was not added because it was already in use? Same for removal.

So for now, I'm tending towards cutting that feature altogether. Let's get this out as simple as possible, we can always add it back later on, with some separate discussions. Right now it looks like it might not be worth the amount of work that would be required.

src/entities/createActionListenerMiddleware.ts
phryneas: :heavy_check_mark:

src/entities/createActionListenerMiddleware.ts
phryneas: Removed it for now. Let's re-evaluate that later, in a separate issue.

src/entities/createActionListenerMiddleware.test.ts
phryneas: Okay, changed that one.

src/createActionListenerMiddleware.ts
```diff
@@ -0,0 +1,278 @@
...
+          if (entry.once) {
+            listeners.delete(entry)
+          }
+          if (entry.preventPropagation) {
+            return action
+          }
+        }
+      }
+    }
+    return next(action)
```
nogwater, line 196: The `next(action)` call occurs after calling the actions with `entry.listener(action, api)`.  This means that you can't have a listener that uses the updated state, right?  Is this intentional?  I found https://github.com/reduxjs/redux-toolkit/issues/237#issuecomment-562287869 that says the reducers should run first.  The logic for this block might get tricky with checking conditions and preventPropagation, but might still be doable.

phryneas: Phew. That's a good point.
Ideally, we'd like to allow for both behaviours.

I could imagine adding an option `{ when: 'before' | 'after' }`, defaulting it to `before` and issuing an Error from the `stopPropagation` method when called in combination with `after`.
What do you think?

phryneas: Ouch. This one hurts quite a bit in the typings. Following through with it nevertheless ^^

phryneas: Okay, that's in. Comments? :)

phryneas:
Oh, btw. I wanna have the features on this finalized before I add type tests - those might add quite some type changes and I don't really want to do that multiple times.

src/createActionListenerMiddleware.ts
nogwater: If when is after, should the entry.condition be given the previous state or the next state?
You are returning from within the for loop which means not all listeners will be processed.

This is a bit verbose, and totally untested, but how about something like:
```TypeScript
if (listeners) {
  // process 'before' listeners
  for (const entry of listeners) {
    if (entry.when == 'after') {
      continue;
    }

    if (!entry.condition || entry.condition(action, api.getState)) {
      if (entry.once) {
        listeners.delete(entry)
      }

      let stoppedPropagation = false
      entry.listener(action, {
        ...api,
        stopPropagation() {
          stoppedPropagation = true
        }
      })
      if (stoppedPropagation) {
        return action
      }
    }
  }

  const result = next(action)

  // process 'after' listeners
  for (const entry of listeners) {
    if (entry.when != 'after') {
      continue;
    }

    if (!entry.condition || entry.condition(action, api.getState)) {
      if (entry.once) {
        listeners.delete(entry)
      }

      entry.listener(action, {
        ...api,
        stopPropagation: () => {
          throw new Error(
          'stopPropagation can only be called by action listeners with the `when` option set to "before"'
          )
        }
      })
    }
  }
  return result
}
return next(action)
```

I think this leaves the door open for having a `when: both` option, but I don't know why someone would want that.

phryneas: I think if someone wanted to do `when: both`, they could just add a `before` listener and an `after` listener. We really shouldn't add any extra logic for something like that.

But you are very right about the condition. That should definitely run after for the `after` case - good catch!

phryneas: Okay, I pushed a fix and added another two tests for that. Could you please take another look? :)

nogwater: That looks good.  I'm still unsure about which default (before or after) is better.  For people coming from Saga, having after be the default might be less confusing.  That said, if there's a good case for running before the reducers then it's probably better to default to before.  I'll be curious to hear more thoughts on this.

phryneas:
> @phryneas !!!!! Guess what fun type issue we just ran into?

Fixed it. Weirdest thing. TS can infer everything just fine from a 
```ts
Middleware<
  {(action: "foo") => "bar"},
  RootState
>
```

but once you add additional properties (in our case `addListener` and `removeListener`) on top, it falls back to `Middleware<unknown, unknown>`, so it loses the dispatch signature.

Let's just say I found a way to shove it somehow into TypeScript by force.

src/createActionListenerMiddleware.ts
phryneas: Personally, I'd like to have the ability to `stopPropagation` by default.  
I believe in sagas it's only the case because those are async AF and usually a million ticks have happened yielding 80 times before anything meaningful happens. (Doesn't sound like it, but I still really like saga ;) )

@markerikson do you have an opinion on this?

markerikson: Both sagas and observables explicitly let the action hit the reducers first:

https://github.com/redux-saga/redux-saga/blob/a8c968b77243f59fbe672ca56691761664e060bf/packages/core/src/internal/middleware.js#L27

https://github.com/redux-observable/redux-observable/blob/dfef883fdba9f669ea88139bdb7abd2d026ff7a8/src/createEpicMiddleware.ts#L88-L91

I'd say that's the preferable default.

src/createActionListenerMiddleware.ts
```diff
@@ -0,0 +1,358 @@
...
+    if (removeListenerAction.match(action)) {
+      removeListener(action.payload.type, action.payload.listener)
+
+      return
+    }
+
+    const listeners = listenerMap[action.type]
+    if (listeners) {
+      /* before */
+      for (const entry of listeners) {
```
markerikson, line 206: The logic for the "before" and "after" cases is _almost_ identical. Is it worth trying to deduplicate it?

src/createActionListenerMiddleware.ts
```diff
@@ -0,0 +1,358 @@
...
+  W extends When = 'before'
+> {
+  /**
+   * Indicates that the listener should be removed after if has run once.
+   */
+  once?: boolean
+  /**
+   * A function that determines if the listener should run, depending on the action and probably the state.
+   */
+  condition?(action: A, getState: () => S): boolean
```
mpeyper, line 45: Is there a particular feature this is satisfying that just having the condition around the body of the listener function wouldn't also provide?

```js
function myCondition(action, getState) {
  // ...
}

function myListener(action, { getState } ) {
  // ...
}

store.dispatch(addListener(myListener, { condition: myCondition }))
```
 vs. 

```js
function myCondition(action, getState) {
  // ...
}

function myListener(action, { getState } ) {
  if (myCondition(action, getState)) {
    // ...
  }
}

store.dispatch(addListener(myListener))
```

Similarly to reducers, a higher-order listener could be used it the same condition logic was needed or if the listener was being provided by a third party and I wanted to add a condition to it

```js
import someListener from 'some-dep'

function myListener(action, ...args) {
  if (action.type !== 'secret action type') {
    return someListener(action, ...args)
  }
}

store.dispatch(addListener(myListener))
```

I just feels like it's adding complexity to the internals without provide anything they can do (and relatively easily) in userland... unless I'm missing something?

markerikson: Hmm. Yeah, I could see it being straightforward to do conditional and `once` handling on your own, although we might need to document that as recipes.

src/createActionListenerMiddleware.ts
```diff
@@ -0,0 +1,358 @@
...
+          }
+
+          let stoppedPropagation = false
+          entry.listener(action, {
+            ...api,
+            stopPropagation() {
+              stoppedPropagation = true
+            }
+          })
+          if (stoppedPropagation) {
```
mpeyper, line 223: The current `stopPropagation` idea is troublesome for me.  I can sort of see the intent of wanting to stop other listeners from getting a shot at the action, but I don't think we should ever be preventing the action from going to the next middleware or to the store.  Even with the other listeners I'm struggling the think of a good reason why I'd want need to prevent them getting the action either.  In most cases, I think it will introduce complexity or subtle bugs that are difficult to debug ("Why isn't my action in the dev-tools? _3 weeks later..._ Dammit, the condition on this unrelated listener is too permissive and is smothering it!!!").

To me, it sort of goes beyond its charter as a "listener".  I think if someone wants to have that much control of what actions are able to propagate and which ones can't, then they have the option of using sagas or observables or even writing a custom middleware.

src/createActionListenerMiddleware.ts
mpeyper: Just a thought, could the listener behave a bit like `useEffect` and allow you to return a function to run after the state update? something like:

```js
function myListener(action, { getState() }) {
  let previousValue = getState().valueWeCareAbout
  return () => {
    let  newValue =  getState().valueWeCareAbout
    if (previousValue === 'Pending' && newValue === 'Failed') {
      logger.error('Failed to submit')
    }
  }
}
```

Then you could have before listener

```js
const beforeListener = () => {}
```

and after listener

```js
const afterListener = () => () => {}
```

or a before and after listener

```js
const bothListener = () => {
  return () => {}
}
```

markerikson: That also feels kinda complex, and the similarity in syntax with `useEffect` would actually seem like a reason to _not_ do it that way because they're two very different things.

src/createActionListenerMiddleware.ts
nogwater: The duplicated code is pretty readable, but here's a stab at deduping it.

```TypeScript

  const defaultStage = 'before'
  if (listeners) {
    const triggerListener = (stage, listener) => {
      let stoppedPropagation = false

      if (
        (listener.when || defaultStage) === stage &&
        (!listener.condition || listener.condition(action, api.getState))
      ) {
        if (listener.once) {
          listeners.delete(listener)
        }

        listener.listener(action, {
          ...api,
          stopPropagation() {
            if (stage === 'after') {
              throw new Error(
                'stopPropagation can only be called by action listeners with the `when` option set to "before"'
              )
            }
            stoppedPropagation = true
          },
        })
      }

      return stoppedPropagation
    }

    /* before */
    for (const entry of listeners) {
      const stoppedPropagation = triggerListener('before', entry)
      if (stoppedPropagation) {
        return action
      }
    }

    const result = next(action)

    /* after */
    for (const entry of listeners) {
      triggerListener('before', entry)
    }

    return result
  }
  return next(action)

```

src/createActionListenerMiddleware.ts
```diff
@@ -0,0 +1,358 @@
...
+ */
+export function createActionListenerMiddleware<
+  S,
+  D extends Dispatch<AnyAction> = Dispatch
+>() {
+  type ListenerEntry = ActionListenerOptions<any, S, D, When> & {
+    listener: ActionListener<any, S, D, any>
+  }
+
+  const listenerMap: Record<string, Set<ListenerEntry> | undefined> = {}
```
mpeyper, line 180: This might be wishful thinking, but the 1-M action type to listeners relationship almost feels backwards to me.  This becomes clumsy if I wanted to have a listener respond to a set or actions, or all actions especially, whereas the other way round (1-M listener to action types) does not.

Essentially I'm advocating for treating the listener more like a reducer where the action is used to conditionally run logic, and the middleware is just responsible for collecting the set of listeners and passing the actions to them.

There are likely repercussions to changing it around like this for things like the `once` option (how would the middleware know if it fired or not) and is probably more or a rethink of the whole approach than an iteration of this proposal.

src/createActionListenerMiddleware.ts
phryneas: Hm. I'd really like to keep the `once` option, as having a "react to only the next action of type X" seems really useful sometimes. 
In that context, the condition also has it's value. So I'm not really sure if I would want to let got of that.

src/createActionListenerMiddleware.ts
mpeyper: I guess I don't see them as _very_ different in terms of wanting to perform logic around state changing (dependency array for `useEffect`, redux state for this).  I could also argue that this primary use case for this middleware is to use effects in response to actions.

There are also some similarities in the api with exisiting redux concepts with thunks and middleware.  It could be considered a middleware-lite api so converting your listener to a custom middleware when the requirements grow is eased a bit.

src/createActionListenerMiddleware.ts
phryneas: The idea of this middleware is to give people an alternative to thunks, as dispatching functions and the usage of higher-order functions is counter-intuitive for many people - they's rather dispatch an action, and react to that. So it should feature-wise be at least en par with thunks while offering another approach.

Now, thunks can take non-serializable parameters, as the thunk itself never reaches the reducer. Likewise, I personally would have "trigger actions" with non-serializable payloads that would never have any effect in any reducer - I'd only use those to trigger my middleware. Did the same thing with sagas.
If those actions would land in the reducer, they would cause problems in the devtools and trigger the nonSerializableActionMiddleware. So there needs to be some way of stopping those actions. Otherwise, listeners would need to start modifying in-transit actions, which would arguably be much worse.

On the flipside, I don't really see the problem you are hinting at - using `stopPropagation` is a *very* deliberate act and nothing that would happen by accident.

On the point of the "listener" name and stopping propagation: Web developers have had "listeners" since the early  days of the DOM. Those "listeners" could always prevent further event propagation. So at least for me, it's a concept that is really entwined with listeners.

src/createActionListenerMiddleware.ts
nogwater: It sounds like this would be an internal change that on its own wouldn't be exposed to the user.  This map of action type to listeners approach should be more efficient if there are a lot of different action types with listeners attached, but I think you'd need a LOT to make it noticeable.

On the other hand, switching to a list of listeners that each have one or more action types that they are subscribed to would enable pattern-based subscriptions.  For example `addListenerAction('actionListenerMiddleware/*', listener)`.  Heck, if you do that, then you could listen for every action in the world and it would create another way to plug-in something like middleware `addListenerAction('*', listener)`.  I can't decide if I like this.

src/createActionListenerMiddleware.ts
phryneas: An alternative option to `once` *could* be to add an `unsubscribe` function to the api. Not 100% if I like the idea yet. Thoughts?

src/createActionListenerMiddleware.ts
phryneas: Honestly, at that point you really want to write a middleware by hand, because otherwise we'd provide a middleware that allowed you to write middlewares in the style of writing middlewares.

phryneas: @nogwater So far we've avoided this string-pattern matching everywhere else in RTK, and for good reason. Once you fall back to that, TypeScript really loses all type safety as every action just boils down to `any`.

src/createActionListenerMiddleware.ts
mpeyper: Can the conditional `once` functionality not be implemented as

```js
function myListener(action, { getState, dispatch } ) {
  if (myCondition(action, getState)) {
    dispatch(removeListener(myListener))
    // ...
  }
}
```

phryneas: Yes, but that's a lot less obvious to the user than 
```ts
function myListener(action, { getState, dispatch, unsubscribe } ) {
  if (myCondition(action, getState)) {
    unsubscribe()
  }
}
```
so I'd rather go that way, as it doesn't really cost much to add that.

src/createActionListenerMiddleware.ts
mpeyper: @phryneas, yeah as I said, that idea is probably a competing proposal more than an iteration of this one.

In my ideal implementation, the listener middleware would be a more reduxy implementation that leaned into the more functional concepts and loose action/reducer coupling while RTK would provide the opinionated, type-safe utilities to make writing listeners for the 95% use-cases straight forward and easier (in a similar way to how `redux-thunk` is supported).

src/createActionListenerMiddleware.ts
mpeyper: Ok, I see where you are coming from with this about the serialisable parameters in a trigger action.  I think my preferred usage would be to construct the listener with the non-serialisable aspects and keep the trigger action plain.  This adds an element of tracability, predictablilty and ultimately debuggability to the whole arrangement that, in my experience, is worth the dance around the "actions must be plain" restrictions (especially when working on large teams).

I also feel like this is the first side effect middleware I've seen that lets you swallow an action that isn't directly related to the middleware itself (e.g. the thunk itself, `addListener`/`removeListener`).  I'm not saying that it has to be like everyone else, but I have to wonder why the others didn't need this.

> On the flipside, I don't really see the problem you are hinting at

The problem is that it's a very deliberate action that can be performed on a listener that just happens to be first in the listening order that could prevent a completely unrelated listener (or reducer) from doing it's job.  Here's a scenario (it's a bit contrived, but not _that_ far off something we do where I work, so go with me on this):

We're building a really cool app and one of our devs sees this PR and gets an awesome idea.  What if we could use a listener to listen for a `ROUTE` action to move between routes using a `path` value on the action.  We agree that this could work, so we implement it.

Next someone implements a second listener on `ROUTE` that pings off an analytics event so we can track how our users are navigating through through the app.

A new feature request has come in to make one of the routes dynamic.  Luckily, all the details needed are stored in state, so the `ROUTE` action is updated to pass a function to derive the `path` from state and the middleware gets updated to support this.  By oh no, now we get a warning about the action not being serialisable so we google a bit, find the docs and discover `stopPropagation`, we add it at the end of the listener and away we go.

A few days later we notice a sudden drop in pages getting hit...  You can see where I'm heading with this.

There is very little to help the developer track this down as they can see the action getting dispatched, but their listener is never hit, the dev-tools are not recording it, but the route is definitely changing.

I guess the moral of the story is that is you ship a feature, someone will try to use it, and sometimes that usage will be implemented poorly (there's many things wrong with the above scenario, and many ways it could be worse) and the more someones in the mix, the more chance of something going wrong.  All we can hope for as library authors is when someone gets themselves in a pinch, they find some clue to pull themselves out of it.

> Web developers have had "listeners" since the early days of the DOM. Those "listeners" could always prevent further event propagation. So at least for me, it's a concept that is really entwined with listeners.

This might be where the friction comes from.  I joined the web dev world in the late Angular days and quickly moved to React/Redux, so for me a lot of "early days of the DOM" stuff has either never been used or never been needed, so the idea of preventing propagation on an event bus feels off to me.

At the end of the day, I don't have to like every part of the proposal, so feel free to dismiss this concern.  If I have to choose one method of preventing propagation, the `api. stopPropagation()` is preferred over the other suggestions I read previously.

nogwater:
Starting from the [pensive-allen-y3m72](https://codesandbox.io/s/pensive-allen-y3m72) CodeSandbox, I [replaced the fetchComments thunk with a listener](https://codesandbox.io/s/strange-mountain-9yyw3?file=/src/features/issueDetails/commentsSlice.ts).

commentsSlice.ts before:
```TypeScript
export const fetchComments = (issue: Issue): AppThunk => async dispatch => {
  try {
    dispatch(getCommentsStart())
    const comments = await getComments(issue.comments_url)
    dispatch(getCommentsSuccess({ issueId: issue.number, comments }))
  } catch (err) {
    dispatch(getCommentsFailure(err))
  }
}
```
commentsSlice.ts after:
```TypeScript
export async function getCommentsStartListener(action, api) {
  try {
    const issue: Issue = action.payload
    const comments = await getComments(issue.comments_url)
    api.dispatch(getCommentsSuccess({ issueId: issue.number, comments }))
  } catch (err) {
    api.dispatch(getCommentsFailure(err))
  }
}
```

IssueDetailsPage.tsx before:
```TypeScript
import { fetchComments } from './commentsSlice'
...
  useEffect(() => {
    if (issue) {
      dispatch(fetchComments(issue))
    }
  }, [issue, dispatch])
```
IssueDetailsPage.tsx after:
```TypeScript
import { getCommentsStart, getCommentsStartListener } from './commentsSlice'
...
  dispatch(addListenerAction(getCommentsStart, getCommentsStartListener))
...
  useEffect(() => {
    if (issue) {
      dispatch(getCommentsStart(issue))
    }
  }, [issue, dispatch])
```

I think overall, they're pretty similar.  The action listener approach just taps in to the existing `getCommentsStart` action instead of introducing the new `fetchComments` function.  I don't really like having to call `dispatch(addListenerAction(getCommentsStart, getCommentsStartListener))` from IssueDetailsPage.tsx.  Is there a way that could be done from commentsSlice.ts?  Is it confusing that the `getCommentsStart` action takes a payload that isn't used by the reducer?  Any other thoughts?

markerikson:
I wouldn't see fetching data thunk-style as a great use case for this, especially given the existence of `createAsyncThunk`.  (tbh it would probably be helpful if we _did_ have a list of some specific use cases we envision people trying to do, and seeing how those turn out with this API.)

msutkowski:
> I wouldn't see fetching data thunk-style as a great use case for this, especially given the existence of `createAsyncThunk`. (tbh it would probably be helpful if we _did_ have a list of some specific use cases we envision people trying to do, and seeing how those turn out with this API.)

Adding in a couple ideas for a _how_: these scenarios apply to a few our codebases and are currently implemented in a few different ways

1. `dispatch(event)` -> `listener: create ws conn, await websocket events` -> `do something once on specific event` -> `tear down handler` (we do this with 'admin waiting area's where an enrollment flow depends on a user on a different device completing an action, such as signing an agreement)
2. streamlining error reporting with `.rejected` from  `createAsyncThunk`s (ex: logging to bugsnag, having 'higher priority' rejected actions forwarding to a diff service, etc)

markerikson:
Y'know, I can see the "rejected" use case as a good argument for doing _some_ kind of pattern-matching or letting the listener decide whether it wants to handle a given action vs doing exact matching on a specific type, ie:

```js
const myListener = (action, getState) => {
  if(action.endsWith("/rejected") {
    // do something useful here
  }
}
```

Agreed with @phryneas 's comments that that sorta devolves into an almost middleware-on-its-own kind of thing, but A) I've seen a couple other "middleware that let you dynamically add and remove middleware from the pipeline" implementations, and B) there may be sufficient value that it's worth doing anyway.

mpeyper:
The other use case that is used by redux-saga and redux-observable and is not possible with redux think is a [cancellable vounter](https://github.com/redux-saga/redux-saga/tree/master/examples/cancellable-counter).

1. Click a increment
2. Countdown starts
3. Increment counter when it reaches 0
4. If you click the button again, the it cancels the countdown and no increment occurs

This usually involves listening for multiple actions and changing behaviour based on the order they were received.

src/createActionListenerMiddleware.ts
phryneas: those are now deduplicated & shrunk quite a bit

phryneas:
So I just went ahead and pushed some changes:

* default `when` to "after"
* remove `once` and `condition`
* add `unsubscribe` to api
* consolidate "before" and "after" handling

src/createActionListenerMiddleware.ts
```diff
@@ -202,55 +186,48 @@ export function createActionListenerMiddleware<
 
     const listeners = listenerMap[action.type]
```
markerikson, line 187: Would it maybe make sense to split it into _two_ maps, one for "before" and one for "after", so that we only iterate through listeners relevant to that phase?

phryneas: We could do that, but I wouldn't worry about it too much. It probably doesn't matter unless someone has hundreds of listeners for the same action type - and even then it's only a factor of 2.

markerikson: Fair point.  I trust that Node can iterate through a bunch of things fairly quickly :)

markerikson:
Gonna throw a bit of a monkey wrench into this discussion:

Several other libraries have similar APIs, and we ought to take some time to look those over and see what inspiration we can take from them:

- https://easy-peasy.now.sh/docs/api/listeners.html
- https://kea.js.org/docs/guide/concepts#listeners
- https://rematch.github.io/rematch/#/api-reference/api?id=effects

nogwater:
> * https://easy-peasy.now.sh/docs/api/listeners.html
> * https://kea.js.org/docs/guide/concepts#listeners
> * https://rematch.github.io/rematch/#/api-reference/api?id=effects

I've just taken a quick look at the docs for these.  They all have a very declarative feel.  I know it's too early for this, but if this approach were part of createSlice it might look like
```JavaScript
createSlice({
  name: 'counter',
  initialState: 0,
  reducers: {
    incrementBy: (state, action) => {
      state += action.payload
   }
  },
  listeners: {
    incrementBy: (action, api) => {
      console.log(`${action.type} called, new count is ${api.getState().counter}')
    }
  }
})
```

I think this makes the listeners much less dynamic.  The listeners are there from the beginning and don't change over the lifetime of the app.  Maybe these libraries also have advanced ways to dynamically add listeners?

Easy-peasy lets you listen for regular action types, thunks (at various stages: start, success, fail), and lists/combinations of these.  Listening for multiple targets probably cuts down on repeated code in some cases.

Rematch gives you access to dispatch() if you want it (I'm not sure if the others do), so you can have listeners trigger external actions.

It looks like the listeners all run after the reducers, but not all of the docs are clear about this.

Kia has [breakpoints](https://kea.js.org/docs/guide/additional/#breakpoints) to support debouncing and handling out of order network requests.

--
Here are some more, but less popular, libraries for action listeners.
https://github.com/si0ls/redux-subscribe-action
API for subscribing to actions before or after they've been processed by other middleware/reducers.  Unsubscribe looks like it removes all listeners of a type (eg all "before" listeners, or all listeners for a specific action type) instead of working with function references or with a callback.  Your listener code only gets access to the action itself, no store, state, or dispatch.

https://github.com/Gaya/redux-listeners
Very small api for adding a listener by action type.  Your listener callback receives the action, and the dispatch function, but not the store or state.  Looking at the code, the listeners run before the reducers.  The dispatch function you are given wraps the real dispatch with a setTimeout.  Is this necessary?  No unsubscribe/remove api.

https://github.com/rhythnic/redux-action-listeners
Has more of a declarative/OO feel.  You can get a reference to the store if you want it.  Runs after the other middleware/reducers.  Return what you want from the listener to change what dispatch returns.  It looks like you need to add the middleware multiple times to get multiple listeners.

https://github.com/rnsloan/redux-white-black#readme
Listen for a list of actions (whitelist), or all actions except some (blacklist).

https://github.com/heiskr/redux-listen
Subscribe by action type or regular expression pattern.  Your listener gets the action, dispatch(), and getState().  Add multiple listeners at once by passing a dictionary of action type to listener functions.  Unsubscribe/remove by action type, function reference, or both.  Some support for async, but I'm not sure how well baked it is.

https://github.com/Deseteral/redux-on-action
Targeted at async usecase.  Subscribe only (onAction) to a specific action type.  Your listener gets action, dispatch(), and getState().  Looking at the code, it looks like this completely intercepts the action so actions of that type can't be used for anything else, and it returns the state unchanged.

phryneas:
> I've just taken a quick look at the docs for these. They all have a very declarative feel. I know it's too early for this, but if this approach were part of createSlice it might look like
> 
> ```js
> createSlice({
>   name: 'counter',
>   initialState: 0,
>   reducers: {
>     incrementBy: (state, action) => {
>       state += action.payload
>    }
>   },
>   listeners: {
>     incrementBy: (action, api) => {
>       console.log(`${action.type} called, new count is ${api.getState().counter}')
>     }
>   }
> })
> ```
> 
> I think this makes the listeners much less dynamic. The listeners are there from the beginning and don't change over the lifetime of the app. Maybe these libraries also have advanced ways to dynamically add listeners?

That's what I've done over in https://github.com/reduxjs/redux-toolkit/pull/272 as an experiment - but that essentially creates a lot of "small middlewares" and a lot of follow-up-questions, like how to integrate that with `configureStore`. See the discussion in that PR :)

mpeyper:
I've been mulling over a few thoughts about this proposal.  It might be a bit of a brain dump, so apologies if it's a bit of a ramble.

The `when` option feels clunky to me.  Most, if not every, other major listener/effect middleware I've encountered exclusively fires _after_ the state update, and I believe there are very few situations where firing before has an appreciable difference for the listener.  It seems like in the majority of use cases, the listener is performing an async operation anyway, so even if the trigger was before the fold, the result is after.

Even in this PR, the primary reason (as far as I can tell) for offering to trigger the listener _before_ the action hits the reducers is so that the have the opportunity to use the `stopPropagation` API so that a non-pure trigger action can be used.  I've come around to this perhaps being useful to enabling some advanced use cases, even if iI feel like I'd never use it myself.

Unfortunately, by default, the `stopPropagation` feature is off limits because we're defaulting the `when` to `after`, which makes sense as this is likely to be the more common usage (reacting to an action in combination with the the resulting state).  This means that if a listener wants to to use the `stopPropagation` feature, it must be added with the `before` option

```js
const listener = (action, { getState, dispatch, stopPropagation }) => {
  stopPropagation()
  // ...
}

store.dispatch(addListener(action, listener, { when: 'before' }))
```

There is a disconnect here.  The listener is powerless to control, affect _when_ it is triggered, despite being completely reliant on it.  There's also a ripple effect, albeit fairly minor, when wanting to change the behaviour of a listener to start using or stop using the `stopPropagation` and/or post update state (if it was using the `before` option) that the dispatcher must also change.  This is more problematic, although less likely, for a listener being consumed from a library, as the change must now be considered breaking, rather than just an internal logic change.

Also, as a side note, if we do choose to continue with `when` as string value, can we get some constants exported for the valid values (e.g. `When.BEFORE`, `When.AFTER`)?  Can we use [TypeScript enums](https://www.typescriptlang.org/docs/handbook/enums.html) for this or is there some reason the string value is preferable here?

Now assuming we want to support non-plain trigger actions, we cannot just fall back to only triggering after the state update as this will trigger the warnings and all the bad things from having non-serialisable actions happen (has anyone here ever actually been bitten by this?), so I'd like to propose some alternative APIs.

Perhaps the simplest approach would be to change the interface of a listener to be an object with a `noPropagation` field  and `handle` function:

```js
const listener = {
  stopPropagation: true,
  handle(action, { getState, dispatch }) {
    action.someFunction()
  }
}
```

Obviously `stopPropagation` would default to `false` if not provided and I could envision that if just a function was provided, the it would use that as the the handle function of a default listener definition, so essentially the currently proposed API would be supported.

Alternatively we could change things up a bit and have a `setup` (or more appropriately named function) be provided in the listener interface that can use state and pre update state to provide an argument to the `handle` function.  If not provided, this would default to `() => action` and similar defaulting to before could be provided if just a function was passed.

```js
const listener = {
  setup(action, { getState, stopPropagation, }) {
    stopPropagation()
    return { someFunction, action.someFunction, preState: getState() }
  },
  handle({ someFunction, preState }, { getState, dispatch }) {
    const postState = getState()
    someFunction()
  }
}

const listener = {
  handle(action, { dispatch, getState }) {
    // ...
  }
}
```

I've suggested this before, which @markerikson shot down as looking complex, but having the listener return a function to run pre state update that _can_ return a function to run post update solves this problem as well as the allowing the pre/post update state to be used in the listener logic:

```js
function listener(action, { getState, dispatch, stopPropagation }) {
  stopPropagation()
  const preState = getState()

  return () => {
    const postState = getState()
    action.someFunction()
  }
}
```

If the _return a function_ approach is thought to be a bit scary or confusing, I would also like to propose an `async/await` API that works in a similar way:

```js
async function listener(action, { getState, stopPropagation, waitForPropagation }) {
  stopPropagation()
  const preState = getState()

  await waitForPropagation()

  const postState = getState()
  action.someFunction()
}
```

My biggest issue with this proposed API is that the "regular" case (allow propagation and use post update state) requires the extra `await`:

```js
async function listener(action, { getState, dispatch, waitForPropagation }) {
  await waitForPropagation()
  
  const state = getState()
  processState(state)
}
```

For what it's worth, I've got working versions of the middleware for all these variations, but they were quite lengthy to add to this post.  If you're keen to follow up on any of the ideas presented here, I'm happy to share them.  They're not extensively tested, but I've run the happy path through them.

Ultimately, I'm feeling like the simpler alternatives, options 1 and 2, are easier to understand and follow then the more complex alternative, options 3 and 4, despite them perhaps opening up more potential use cases.

Anyway, that's enough rambling for now.

mpeyper:
@markerikson I've also got a fairly complex, real world example of a situation where we used `redux-saga` specifically for it's listening capability.

I've tried writing out the scenario here and it got way too complicated and you'd need a deep understanding of our system, financial services, axios interceptors, promises and redux-saga, so I'm not going to try to go into too much detail.  Here's the short version:

1. a requests deep in the app fails with specific status code
2. axios response interceptor dispatches action to say a two-factor auth token is required and sets a saga listening for the 2FA token retrieved action
3. saga triggers to set a flag to show a global component
4. when 2FA token is retrieved, the saga resolves a promise set up by the axios interceptor
5. the axios interceptor replays the original request with the retrieved 2FA token attached and returns that back to the original requester

Basically the axios interceptor plus the saga allows any process in the app to hook into a global two-factor auth process before continuing where it left off by carefully orchestrating the resolution of a promise.  It was super complicated to get working initially, but the best part about it now is that the requester is completely unaware that the 2FA process has even occurred and we can reuse the process wherever 2FA is required in our app with very little effort.

nogwater:
I wonder if it makes sense to split the stopPropagation functionality into a different middleware, more of an action-filter middleware.  I'm sure there's a lot of similar logic between a filter and a listener (they both need a way to watch for actions and stop watching).  That said, I don't know how useful an action-filter would be if all it could do it throw actions in the trash.  Maybe it could also return replacement actions and act as more of an action translator?  Or even return an array of actions (which might include the original action) to fan-out?  Now that I think about it, it sounds more like flat-map than a filter.  I don't want to scope creep this.  All of these use cases could be done with the action listener today, right?  These would all require the when:before option.

markerikson:
Lotta good discussion here. I haven't assimilated all of it, tbh, but a few quick thoughts:

- "After" is definitely the typical approach. I don't immediately see a lot of reason to run stuff "before".
- Similarly, while I can see some point to the "stop propagation" stuff, it feels like there's a bunch of edge cases there that we ought to avoid for now
- On the flip sides, this means that even "signal actions" meant to trigger more logic really still have to be serializable or else the dev check middleware will yell at you.  They also trigger a reducer run and subscriber notifications.
- I am curious if there's maybe value in providing access to the _previous_ state besides the updated state

So, my inclination would be that we scale this back as an MVP.  Drop "before", run everything after, drop `stopPropagation`.  Keep it simpler for now, and if there's really an expressed desire for more flexibility later, we can add the options back in as people give us more concrete use cases.

Thoughts on adding `getPreviousState` to the options?

nogwater:
> Thoughts on adding `getPreviousState` to the options?

Do `getPreviousState` and `getState` need to be a passed as functions for listeners?  It looks like the reason for `getState` is a function is to check that it's not being accessed from a reducer, but if we are passing the state to a listener we already know we aren't in a reducer.  Could the api be `{ dispatch, currentState, previousState }` ?  (I don't know if i like `currentState`, `state`, or `nextState`.)  I assume to make `getPreviousState` work, we'd need to call `getState` before `next(action)`, save the results, and then wrap it in a function.

markerikson:
Hrm. Y'know, that brings up an interesting point.

By passing in `getState` to the listener, we enable them to read the "current" state _at any point in the future_, ie, in the middle of some `async/await` stuff.  So yeah, we definitely want `getState`.

What doesn't give you is the ability to know what the state was _before_ the listener even got triggered at all, which is why I was thinking about `getPreviousState`.

As a side note: these listeners are definitely way weaker than sagas and observables, in that I don't see how you'd listen for multiple distinct actions in sequence.  I'm okay with that, as this is deliberately intended to be a much more scoped API anyway.

mpeyper:
> As a side note: these listeners are definitely way weaker than sagas and observables, in that I don't see how you'd listen for multiple distinct actions in sequence. I'm okay with that, as this is deliberately intended to be a much more scoped API anyway.

Yeah, I definitely agree that that we don't want to offer a complete competitor to these solutions, but rather a lightweight alternative or stepping stone for the basic use cases.

I have previously stated my preference to allow listeners to react to multiple action types as a way of tracking more complex processes being undertaken, but I'm happy with concede the point as it's technically possible for a listener to subscribe new listeners or that that is a job for the more powerful alternatives.

On that point, the current proposal has included an `unsubscribe` function in the listener API to simplify "one shot" listeners, so I'm wondering if a similar `subscribe` function to start listening again should also be included.  This would allow something like this to be possible in a fairly easy to read an understand way:

```js
function createCompletionListener(onCompletion) {
  return (action, { unsubscribe }) => {
    unsubscribe()
    onCompletion()
  }
}

function triggerListener(action, { dispatch, subscribe, unsubscribe }) {
  unsubscribe()

  const completionListener = createCompletionListener(() => {
    subscribe()
    dispatch({ type: "PROCESS_COMPLETE" }))
  })
  dispatch(addListener('FINAL_ACTION', completionListener))
}
```

> What doesn't give you is the ability to know what the state was before the listener even got triggered at all, which is why I was thinking about getPreviousState.

Getting access to the previous state would be useful for a few things.  Responding to a change in state over time, enabling some kind of snapshot/restore functionality or generating a diff to report distinct changes in a log message come to mind.

My only concern is when there is async work happening, the `getPreviousState` will always return the previous state from when it triggered, not from before the last dispatch or anything other weird way users might interpret the word "previous" in the context they are calling it.  It might just be a naming thing (I don't have an alternative that isn't overly complex and verbose - `getStateBeforeTrigger`) or perhaps I'm just not giving people enough credit?

markerikson:
One other potential wrench in the gears:

Do we want to add `extraArgument` support ala thunks?  Is that something that would make sense?

mpeyper:
When I was looking at an implementation of this, I had included a `withExtraArgument` option to the API as I was trying to have a very _thunk-esque_ API.  I personally like the idea of keeping it familiar to `thunk` users, especially given that thunks are also supported by default in RTK.

Selfishly, we currently use the `withExtraArgument` feature of thunks where I work to inject preconfigured axios instances (for dealing with the previously mentioned 2FA feature, as well as some other global functionality) and when to do some finessing of the [sagas (barely documented) context feature](https://redux-saga.js.org/docs/api/), so having the option here as well would make my life easier.

With my own experimentation, the extra argument was added when creating the middleware:

```js
  let store = configureStore({
    reducer: () => ({}),
    middleware: [createActionListenerMiddleware(initialListeners).withExtraArgument({ api })]
  })
```

In hindsight, I'd probably not worry with changing functions (trying too hard to be _exactly_ like thunk) and just let it be passed into the `createActionListenerMiddleware` function directly, or as an options argument if we want to leave it open for more options in the future:

```js
  let store = configureStore({
    reducer: () => ({}),
    middleware: [createActionListenerMiddleware(initialListeners, { api })]
  })
```

```js
  let store = configureStore({
    reducer: () => ({}),
    middleware: [createActionListenerMiddleware(initialListeners, { extraArgument: { api } })]
  })
```

Although, I've just double checked the implementation here and it seems as though you cannot actually create the middleware with any listeners, so the thunk-like API could still be beneficial if the middleware was exported without needing to be created:

```js
  let store = configureStore({
    reducer: () => ({}),
    middleware: [
      thunk.withExtraArgument({ api }),
      listeners.withExtraArgument({ api })]
  })
```

I'd also be interested in exploring an option to include the extra argument (or additional extra arguments?) when adding a listener,  This would enable my above example of a short term listener to be reduced to:

```js
function completionListener(action, { unsubscribe }, onCompletion) {
    unsubscribe()
    onCompletion()
  }
}

function triggerListener(action, { getState, dispatch, subscribe, unsubscribe }) {
  unsubscribe()

  const onCompletion = () => {
    subscribe()
    const { processState } = getState()
    dispatch({ type: "PROCESS_COMPLETE", payload: processState })
  }

  dispatch(addListener('FINAL_PROCESS_ACTION', completionListener, {
    extraArgument: onCompletion
  ))
}

store.dispatch(addListener('FIRST_PROCESS_ACTION', triggerListener))
```

This may also alleviate some of the gymnastics required to avoid the plain action problem `stopPropogation` was intended to avoid:

```js
function listener(action, { dispatch }, effectMap) {
    effectMap[action.payload]().then(dispatch)
  }
}

const effectMap = {
  someEffect() {
    return getPromise()
  }
  someOtherEffect() {
    return getAnotherPromise()
  }
}

store.dispatch(addListener(action, listener, { extraArgument: effectMap) })

store.dispatch(action('someEffect'))
store.dispatch(action('someOtherEffect'))
```

phryneas:
Hmm. I've been thinking about this for a while.

We all seem to have very different things in mind when talking about "action listeners".

What about avoiding that term right now. Not sure if I like the name, given that it is loaded with meaning by react, but for now, what about "action effects" or even "action side effects"?

We could go with a minimal implementation, which means that "effects" are always triggered after an action has been handled by the store and reflect the state after. We *could* give access to the "before" state, but I guess if people were to implement something that benefitted from that, they were doing something more complex that might warrant a hand-written middleware, so I'd personally leave that out for now.

We could always later - when we see the use case coming up in the issues multiple times with valid use cases - add the concept of "action interceptors" or something like that for what "before" listeners where in my initial suggestion.

Splitting the naming up that way would make it very clear what these things do and avoid the confusion we were having here.

Also, I might have found a way of doing wildcard-like action matches. We could allow to pass a *type guard* as the first argument to what is now addListener.

```ts
  <A extends AnyAction, S, D extends Dispatch, O extends ActionListenerOptions>(
    typeGuard: (a: any) => a is A,
    listener: ActionListener<A, S, D, O>,
    options?: O
  ): AddListenerAction<A, S, D, O>
```

So that might be used in a simple way to just match multiple possible actions like this (we could provide something like `oneOf`)

```ts
const actionA = createAction<string>('a')
const actionB = createAction<number>('b')

function oneOf<T1, T2>(f1: (a: any) => a is T1, f2: (a: any) => a is T2) {
  return (a: unknown): a is T1 | T2 => f1(a) || f2(a)
}

// ends up with the signature `(a: any) => a is (ReturnType<typeof actionA> | ReturnType<typeof actionB>)`
const isAorB = oneOf(actionA.match, actionB.match)

addListenerAction(isAorB, action => {
  // action is inferred to actionA or actionB here
})
```

of for more "wildcardy" matches that would still be strongly typed (the logic of a valid runtime check would be up to the user)

```ts
function hasError(a: any): a is (AnyAction & { meta: { error: any }}) {
  return !!(a?.meta?.error)
}

addListenerAction(hasError, action => {
  // action is inferred to have `meta.error: any`
})
```

Teomik129:
Just thought I'd chime in with [another library](https://github.com/jameslnewell/redux-when) that might provide more inspiration for the api.
Slightly different approach where rather than listening for any specific action, the middleware waits for the provided condition to evaluate to true, providing access to the dispatching action and next state. Then, rather than the reacting function receiving the dispatch, the middleware just dispatches whatever is returned (action, thunk, promise, etc.).
A naming alternative to "action listener" that just popped into my head: "whence" which is both a cheeky combo of when + once and refers to the reacting function (`createAction` in this library) returning something to the dispatch whence the original triggering action came.

mpeyper:
I kinda like that redux-when API. It's basic and simple and doesn't have much crossover with the redux-thunk use cases (e.g. it's not trying to handle API calls). I can see this being used to orchestrate actions and thunks into complex sequences quite nicely.

The one thing I'm not sure it could handle is cancelling a promise, which is one of the main drawbacks of thunks vs something like saga or observable.

> A naming alternative to "action listener" that just popped into my head: "whence"

I see what where your going with that, but it doesn't work for me. I don't get to make the decisions here though, so you might still have a chance.

jakobo:
After chatting with @markerikson in #redux, he pointed me over to this ticket. I was able to get the middleware working, but ran into a few API rough spots. (examples in untyped JS)

**Listeners as a Set, not Map**
Unless we're exposing a `builder` similar to what's done in `createReducer` / `createSlice`, the listener collection will need to be a `Set()`. I ran into this rough edge because using rtkquery, we only receive a series of action matchers. Since these don't collapse neatly into strings for the `listenerMap` I changed things over to a Set. There's a small performance hit from needing to filter the listener collection, but it seemed negligible in a prototype with about a hundred side effect listeners.

```js
// expected to work. for example, calling more data in response to an API call
const actionListener = createActionListenerMiddleware();
actionListener.addListener(api.endpoints.getSomething.matchFulfilled, (action, api) => { /*...*/});
```

**Listener creator syntax**
The `createSlice` is where I'd ultimately expect listeners to go, similar to how we auto-generate the reducers, actions, etc. But since we're trying not to change `createSlice` with this, I just manually attached my listeners to my slice after generation. This resulted in needing some way to quickly define listeners. I extracted `prepare` since it's used for our standalone action creators, and for a custom `createListener` function with the signature `createListener(typeOrActionCreatorOrMatcher, listener)` which uses the same `prepare` internally for consistency.

```js
function prepare(typeOrActionCreator, listener, options) {
  const type =
    typeof typeOrActionCreator === "string" ||
    typeof typeOrActionCreator === "function"
      ? typeOrActionCreator
      : typeOrActionCreator.type;

  return {
    payload: {
      type,
      listener,
      options,
    },
  };
}

export const createListener = (typeOrActionCreatorOrMatcher, callback) =>
  prepare(typeOrActionCreatorOrMatcher, callback).payload;

export const addListenerAction = createAction(
  "actionListenerMiddleware/add",
  prepare
);

export const removeListenerAction = createAction(
  "actionListenerMiddleware/remove",
  prepare
);
```

**Bulk addListeners on initialization of middleware**
Low priority, but with a lot of listeners, the process of adding listeners results in a lot of `addListener` calls. I added a default `listeners` to an option bag as part of `createActionListenerMiddleware`

```js
function createActionListenerMiddleware({listeners: initialListeners = []} = {}) {
```

---

Generally, the middleware is easy to work with. There weren't too many runtime scenarios where I needed to add and remove listeners from the operation. Most commonly I needed to take an API result from RTKQ and do some follow-up async processing in a manner that would have made `queryFn` a very scary bit of code. This provided an obvious place for that lightweight side effect logic.

src/createActionListenerMiddleware.ts
```diff
@@ -0,0 +1,332 @@
...
+      let result: unknown
+      for (const phase of ['before', 'after'] as const) {
+        for (const entry of listeners) {
+          if (phase !== (entry.when || defaultWhen)) {
+            continue
+          }
+          let stoppedPropagation = false
+          let currentPhase = phase
+          let synchronousListenerFinished = false
+          entry.listener(action, {
```
FaberVitale, line 199: Sorry, I'm a bit late :)

### Issue

Have you considered what happens if a listener throws an exception?

In the current implementation an exception raised in a listener would act like a [stopImmediatePropagation](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopImmediatePropagation).

I do not think that is desirable to to have this behaviour especially when an exception is thrown unintentionally (e.g. `cannot read property X of undefined`).

In big apps it is extremely confusing and frustrating when an unrelated piece of code causes bugs to a feature.

### Example

```ts
listeners = [
  trackUserClick, // analytics e.g. call gtag or something similar...
  respondToUserClick // business logic
];
```

1. User  interaction triggers an event.
2. trackUserClick raises an exception for some reason:  
    integration error server down, `cannot read property X of undefined`, bug etc... 
3. respondToUserClick does not run, we have failed to respond to user interaction due to **implicit** `stopImmediatePropagation` behaviour.

---

### Other event listener APIs

#### DOM addEventlistener

Each listener runs in its own task, a subscriber exception does not stop the propagation.

#### Svelte stores

[It does not handle  subscriber exceptions](https://github.com/sveltejs/svelte/blob/master/src/runtime/store/index.ts#L73), unfortunately.

#### Observer proposal

[A subscriber cannot stop the propagation](https://github.com/tc39/proposal-observable/blob/master/src/Observable.js#L49)

#### rxjs

[A subscriber cannot stop the propagation](https://github.com/ReactiveX/rxjs/blob/6fafcf53dc9e557439b25debaeadfd224b245a66/src/internal/Subscriber.ts#L177) errors are reported in [another task](https://github.com/ReactiveX/rxjs/blob/6fafcf53dc9e557439b25debaeadfd224b245a66/src/internal/util/reportUnhandledError.ts#L13)


#### Suggested changes

I believe this event loop should call every listener in a try catch and should then notify the errors.

A global report handler should suffice, consider [redux-saga onError as an example](https://redux-saga.js.org/docs/api#createsagamiddlewareoptions),
but every listener could also  have the option to provide an error handler. 

The latter is implemented in this [event listener middleware I wrote few day ago](https://gist.github.com/FaberVitale/476e2977f59c38a703aa63cca56beaf7).

markerikson:
FYI, I just had a reason to want to try out this middleware in my own app at work.  Since it's not actually published yet, and this branch is stale relative to `master`, I decided to try copy-pasting the logic into my app directly.

I ended up making a few changes as well:

- Since this imports directly from several RTK-internal types, I had to copy-paste those
- As written here, it only supports listening for specific action type strings, either by passing the type string directly, or by passing an RTK-generated action creator with a `.type` field. I immediately had a need to use RTK's "matchers" instead, so I went ahead and updated the middleware to support that.  (This also required that I google up how to infer the generic type result from a type guard, in order to infer the correct action type from a matcher :)  )

I'll paste the entire implementation in here for visibility. Maybe we can go ahead and publish this standalone version as the separate package for now?

```ts
import { createAction, PayloadAction, Middleware, Dispatch, AnyAction, MiddlewareAPI, Action } from '@reduxjs/toolkit';


interface BaseActionCreator<P, T extends string, M = never, E = never> {
  type: T;
  match(action: Action<unknown>): action is PayloadAction<P, T, M, E>;
}

interface TypedActionCreator<Type extends string> {
  (...args: any[]): Action<Type>;
  type: Type;
}

type MatchFunction<T> = (v: any) => v is T;

export interface HasMatchFunction<T> {
  match: MatchFunction<T>;
}

export const hasMatchFunction = <T>(v: Matcher<T>): v is HasMatchFunction<T> => {
  return v && typeof (v as HasMatchFunction<T>).match === 'function';
};

/** @public */
export type Matcher<T> = HasMatchFunction<T> | MatchFunction<T>;

const declaredMiddlewareType: unique symbol = undefined as any;
export type WithMiddlewareType<T extends Middleware<any, any, any>> = {
  [declaredMiddlewareType]: T;
};

export type When = 'before' | 'after' | undefined;
type WhenFromOptions<O extends ActionListenerOptions> = O extends ActionListenerOptions ? O['when'] : never;

/**
 * @alpha
 */
export interface ActionListenerMiddlewareAPI<S, D extends Dispatch<AnyAction>, O extends ActionListenerOptions>
  extends MiddlewareAPI<D, S> {
  stopPropagation: WhenFromOptions<O> extends 'before' ? () => void : undefined;
  unsubscribe(): void;
}

/**
 * @alpha
 */
export type ActionListener<A extends AnyAction, S, D extends Dispatch<AnyAction>, O extends ActionListenerOptions> = (
  action: A,
  api: ActionListenerMiddlewareAPI<S, D, O>
) => void;

export interface ActionListenerOptions {
  /**
   * Determines if the listener runs 'before' or 'after' the reducers have been called.
   * If set to 'before', calling `api.stopPropagation()` from the listener becomes possible.
   * Defaults to 'before'.
   */
  when?: When;
}

export interface AddListenerAction<
  A extends AnyAction,
  S,
  D extends Dispatch<AnyAction>,
  O extends ActionListenerOptions
> {
  type: 'actionListenerMiddleware/add';
  payload: {
    type: string;
    listener: ActionListener<A, S, D, O>;
    options?: O;
  };
}

/**
 * @alpha
 */
export const addListenerAction = createAction(
  'actionListenerMiddleware/add',
  function prepare(
    typeOrActionCreator: string | TypedActionCreator<string>,
    listener: ActionListener<any, any, any, any>,
    options?: ActionListenerOptions
  ) {
    const type =
      typeof typeOrActionCreator === 'string'
        ? typeOrActionCreator
        : (typeOrActionCreator as TypedActionCreator<string>).type;

    return {
      payload: {
        type,
        listener,
        options
      }
    };
  }
) as BaseActionCreator<
  {
    type: string;
    listener: ActionListener<any, any, any, any>;
    options: ActionListenerOptions;
  },
  'actionListenerMiddleware/add'
> & {
  <C extends TypedActionCreator<any>, S, D extends Dispatch, O extends ActionListenerOptions>(
    actionCreator: C,
    listener: ActionListener<ReturnType<C>, S, D, O>,
    options?: O
  ): AddListenerAction<ReturnType<C>, S, D, O>;

  <S, D extends Dispatch, O extends ActionListenerOptions>(
    type: string,
    listener: ActionListener<AnyAction, S, D, O>,
    options?: O
  ): AddListenerAction<AnyAction, S, D, O>;
};

interface RemoveListenerAction<A extends AnyAction, S, D extends Dispatch<AnyAction>> {
  type: 'actionListenerMiddleware/remove';
  payload: {
    type: string;
    listener: ActionListener<A, S, D, any>;
  };
}

/**
 * @alpha
 */
export const removeListenerAction = createAction(
  'actionListenerMiddleware/remove',
  function prepare(
    typeOrActionCreator: string | TypedActionCreator<string>,
    listener: ActionListener<any, any, any, any>
  ) {
    const type =
      typeof typeOrActionCreator === 'string'
        ? typeOrActionCreator
        : (typeOrActionCreator as TypedActionCreator<string>).type;

    return {
      payload: {
        type,
        listener
      }
    };
  }
) as BaseActionCreator<
  { type: string; listener: ActionListener<any, any, any, any> },
  'actionListenerMiddleware/remove'
> & {
  <C extends TypedActionCreator<any>, S, D extends Dispatch>(
    actionCreator: C,
    listener: ActionListener<ReturnType<C>, S, D, any>
  ): RemoveListenerAction<ReturnType<C>, S, D>;

  <S, D extends Dispatch>(type: string, listener: ActionListener<AnyAction, S, D, any>): RemoveListenerAction<
    AnyAction,
    S,
    D
  >;
};

/**
 * @alpha
 */
export function createActionListenerMiddleware<S, D extends Dispatch<AnyAction> = Dispatch>() {
  type ListenerEntry = ActionListenerOptions & {
    listener: ActionListener<any, S, D, any>;
    unsubscribe: () => void;
  };

  type ListenerEntryWithMatcher = ListenerEntry & {
    matcher: MatchFunction<any>;
  };

  const listenerMap: Record<string, Set<ListenerEntry> | undefined> = {};
  const matcherListeners = new Set<ListenerEntryWithMatcher>();

  const middleware: Middleware<
    {
      (action: Action<'actionListenerMiddleware/add'>): Unsubscribe;
    },
    S,
    D
  > = api => next => action => {
    if (addListenerAction.match(action)) {
      const unsubscribe = addListener(action.payload.type, action.payload.listener, action.payload.options);

      return unsubscribe;
    }
    if (removeListenerAction.match(action)) {
      // @ts-ignore
      removeListener(action.payload.type, action.payload.listener);

      return;
    }
    // @ts-ignore
    const listeners = listenerMap[action.type];
    let matchedMatcherListeners: ListenerEntry[] = [];

    if (matcherListeners.size > 0) {
      matchedMatcherListeners = Array.from(matcherListeners).filter(entry => {
        return entry.matcher(action);
      });
    }

    if (listeners || matcherListeners.size > 0) {
      const allListeners = Array.from(listeners ?? []).concat(matchedMatcherListeners);
      const defaultWhen = 'after';
      let result: unknown;
      for (const phase of ['before', 'after'] as const) {
        for (const entry of allListeners) {
          if (phase !== (entry.when || defaultWhen)) {
            continue;
          }
          let stoppedPropagation = false;
          let currentPhase = phase;
          let synchronousListenerFinished = false;
          entry.listener(action, {
            ...api,
            stopPropagation() {
              if (currentPhase === 'before') {
                if (!synchronousListenerFinished) {
                  stoppedPropagation = true;
                } else {
                  throw new Error('stopPropagation can only be called synchronously');
                }
              } else {
                throw new Error(
                  'stopPropagation can only be called by action listeners with the `when` option set to "before"'
                );
              }
            },
            unsubscribe: entry.unsubscribe
          });
          synchronousListenerFinished = true;
          if (stoppedPropagation) {
            return action;
          }
        }
        if (phase === 'before') {
          result = next(action);
        } else {
          return result;
        }
      }
    }
    return next(action);
  };

  type Unsubscribe = () => void;

  function addStringListener<T extends string, O extends ActionListenerOptions>(
    type: T,
    listener: ActionListener<Action<T>, S, D, O>,
    options?: O
  ): Unsubscribe {
    const listeners = getListenerMap(type);
    let entry = findListenerEntry(listeners, listener);

    if (!entry) {
      entry = {
        ...options,
        listener,
        unsubscribe: () => listeners.delete(entry!)
      };

      listeners.add(entry);
    }

    return entry.unsubscribe;
  }

  function addMatcherListener<MA extends AnyAction, M extends MatchFunction<MA>, O extends ActionListenerOptions>(
    matcher: M,
    listener: ActionListener<MA, S, D, O>,
    options?: O
  ): Unsubscribe {
    let entry = findListenerEntry(matcherListeners, listener) as ListenerEntryWithMatcher | undefined;

    if (!entry) {
      entry = {
        ...options,
        listener,
        matcher,
        unsubscribe: () => matcherListeners.delete(entry!)
      };

      matcherListeners.add(entry);
    }

    return entry.unsubscribe;
  }

  type GuardedType<T> = T extends (x: any) => x is infer T ? T : never;

  function addListener<C extends TypedActionCreator<any>, O extends ActionListenerOptions>(
    actionCreator: C,
    listener: ActionListener<ReturnType<C>, S, D, O>,
    options?: O
  ): Unsubscribe;
  // eslint-disable-next-line no-redeclare
  function addListener<T extends string, O extends ActionListenerOptions>(
    type: T,
    listener: ActionListener<Action<T>, S, D, O>,
    options?: O
  ): Unsubscribe;
  // eslint-disable-next-line no-redeclare
  function addListener<MA extends AnyAction, M extends MatchFunction<MA>, O extends ActionListenerOptions>(
    matcher: M,
    listener: ActionListener<GuardedType<M>, S, D, O>,
    options?: O
  ): Unsubscribe;
  // eslint-disable-next-line no-redeclare
  function addListener(
    typeOrActionCreator: string | TypedActionCreator<any>,
    listener: ActionListener<AnyAction, S, D, any>,
    options?: ActionListenerOptions
  ): Unsubscribe {
    if (typeof typeOrActionCreator === 'string') {
      return addStringListener(typeOrActionCreator, listener, options);
    } else if (typeof typeOrActionCreator.type === 'string') {
      return addStringListener(typeOrActionCreator.type, listener, options);
    } else {
      const matcher = typeOrActionCreator as unknown as MatchFunction<any>;
      return addMatcherListener(matcher, listener, options);
    }
  }

  function getListenerMap(type: string) {
    if (!listenerMap[type]) {
      listenerMap[type] = new Set();
    }
    return listenerMap[type]!;
  }

  function removeListener<C extends TypedActionCreator<any>>(
    actionCreator: C,
    listener: ActionListener<ReturnType<C>, S, D, any>
  ): boolean;
  // eslint-disable-next-line no-redeclare
  function removeListener(type: string, listener: ActionListener<AnyAction, S, D, any>): boolean;
  // eslint-disable-next-line no-redeclare
  function removeListener(
    typeOrActionCreator: string | TypedActionCreator<any>,
    listener: ActionListener<AnyAction, S, D, any>
  ): boolean {
    const type = typeof typeOrActionCreator === 'string' ? typeOrActionCreator : typeOrActionCreator.type;

    const listeners = listenerMap[type];

    if (!listeners) {
      return false;
    }

    let entry = findListenerEntry(listeners, listener);

    if (!entry) {
      return false;
    }

    listeners.delete(entry);
    return true;
  }

  function findListenerEntry(entries: Set<ListenerEntry>, listener: Function): ListenerEntry | undefined {
    for (const entry of entries) {
      if (entry.listener === listener) {
        return entry;
      }
    }
  }

  return Object.assign(middleware, { addListener, removeListener }, {} as WithMiddlewareType<typeof middleware>);
}
```

markerikson:
I forgot to mention it earlier, but I _did_ publish a standalone version of this middleware as https://www.npmjs.com/package/@rtk-incubator/action-listener-middleware a few days ago. We'd appreciate feedback on whether this API is working out!

markerikson:
I just collected a list of relevant existing discussions and comments over in the corresponding Discussions thread:

https://github.com/reduxjs/redux-toolkit/discussions/1648#discussioncomment-1598002

I'm going to go through this again tomorrow and extract a list of bullet points for things we ought to nail down specifically.

markerikson:
I've gone through the issue/PR threads for our WIP "action listener middleware", collected a list of outstanding API design questions that need to be solved, and summarized them along with my own suggestions:

https://github.com/reduxjs/redux-toolkit/discussions/1648#discussioncomment-1599805

Please give us feedback on these questions!

markerikson:
This PR is stale now that a copy of the middleware has been merged into the repo and published as a standalone `@rtk-incubator/action-listener-middleware` package, so I'm going to close this.

Please see the discussion in https://github.com/reduxjs/redux-toolkit/discussions/1648 for a change log of the temporary package and provide us feedback!

PR #549, opened 2020-05-10
add middlewareBuilder notation with type-curried arguments to co…

phryneas:
This adds the option to pass a function instead of the middleware array as `middleware` option on `configureStore`.

This function receives a `middlewareBuilderApi` argument of the shape
```ts
type MiddlewareBuilderApi<S> = {
  defaultMiddleware: ThunkMiddlewareFor<S>[]
  getDefaultMiddleware: CurriedGetDefaultMiddleware<S>
}
```

Essentially that state-type-currying allows to do something like 
```ts
      reducer: reducerA,
      middleware: ({ getDefaultMiddleware }) =>
        [
          otherMiddleware,
          ...getDefaultMiddleware({ thunk: false })
        ] as const
    })
```
which would before have been
```ts
      reducer: reducerA,
      middleware: [
          otherMiddleware,
          ...getDefaultMiddleware<RootState, { thunk: false }>({ thunk: false })
        ] as const
    })
```

Also, it is now not necessary to know the `RootState` type, which makes the use of custom middlewares possible in combination with the object-type `reducer` notation in TS.

(Also, while touching those files anyway, I moved the separated "3.4 tests" back into normal test suite, as we're not supporting pre-3.4 TS any more)


markerikson:
What's the reasoning for passing in _both_ `getDefaultMiddleware` and `middleware`?

phryneas:
I figured it might be the most common use case to just use the default middleware and extend it like

`middleware: ({defaultMiddleware}) => [ ...defaultMiddleware, logger ]`, but that some people also might want to change the options, so I added both.

But we can go the "only `getDefaultMiddleware`" route as well.

markerikson:
Yeah, I get the thought of just handing them the array, but let's assume that if they're providing a callback here that they're probably going to want to customize something.  don't think it's too much work to have them call it here.

phryneas:
Okay, I cut it down to just the `curriedGetDefaultMiddleware`.

Issue #552, opened 2020-05-11
A tale of type archeology: configureStore middleware typings, the spread operator and Typescript


phryneas:
I might make a blog article out of this on a later date, but for now I'll use this issue to document some weird behavior, important findings and the possible road to a fix.

Triggered by @msutkowski, I went a bit on a hunt.

"What is the issue", you ask? Typing the `middleware` parameter of `configureStore`, I answer.
The following works just fine:

```ts
const myCustomMiddleware: Middleware<(_: "foo" )=> "bar"> = () => next => action => {
  if (action === "foo") return "bar"
  return next(action);
}

const store = configureStore({
  reducer: () => ({}),
  middleware: [myCustomMiddleware, ...getDefaultMiddleware()]
})


const shouldBeBar: "bar" = store.dispatch("foo")
const shouldBeBaz: "baz" = store.dispatch(function thunk(){ return "baz" as const})
```

Whereas, when we turn the parameters around and move the spread to the front, it doesn't:

```ts
const myCustomMiddleware: Middleware<(_: "foo" )=> "bar"> = () => next => action => {
  if (action === "foo") return "bar"
  return next(action);
}

const store = configureStore({
  reducer: () => ({}),
  middleware: [...getDefaultMiddleware(), myCustomMiddleware] // => readonly Middleware<{}, AnyAction, Dispatch<AnyAction>>[]
})

// these two will error out
const shouldBeBar: "bar" = store.dispatch("foo")
const shouldBeBaz: "baz" = store.dispatch(function thunk(){ return "baz" as const})
```

[...]

PR #559, opened 2020-05-15
Adding correctly typed `` prepend`  and `` concat`  to the array…

phryneas:
This should solve #552 - opening as a draft PR because I want to re-review this tomorrow myself.

I streamed coding this live and it's available at https://www.youtube.com/watch?v=uQVJ6Aw3tho :smile: 

Also, this is based of #549, so once that is merged, this is probably more readable.


phryneas:
Okay, moved the `MiddlewareArray` to `utils.ts`, exported it & added documentation. I believe this might be ready for review now.

markerikson:
Tried merging #609, but that resulted in merge conflicts for this branch somehow. Let's just go with this one instead.

elizabeth-dev:
Hi, maybe this isn't the best place to discuss this (sorry for that), but it's a bit awkward that the docs reflect the reference for versions that aren't published yet (I was trying to understand why the library typings didn't include the correct type definition for `middleware`, when I discovered that passing a function was a recent addition. Maybe a bit of Murphy's law that I went to check specifically that part of the docs just today).

Btw, thank you both for your work on this library.

phryneas:
@elizabeth-dev Sorry for that - I'm just finishing the docs on #610 and we'll hopefully soon release the next version.

@markerikson This is actually a good point. Maybe we should deploy the docs from another branch? Might be finnicky though :/

markerikson:
Hrm. Yeah, sorry about that.  

We do plan to release 1.4 in the next couple days or so, so hopefully the mismatch won't last long.

But yeah, for future work, let's try to do integration branches for new features + docs.

PR #570, opened 2020-05-20
Prevent unhandled promises in createAsyncThunk

### LINKED ISSUE #567: Abort createAsyncThunk Uncaught (in promise)
I found a bug (or a feature...), the promise.abort console a Uncaught (in promise) when the promise wasn't actually called due to condition in createAsyncThunk.

``` typescript
    // same probleme with useEffect
    useLayoutEffect(() =>
    {
        const promise = dispatch(getData());

        return () =>
        {
            promise.abort('Unmount on pending');
        };
    }, [dispatch]);
```

```typescript
export const getData = createAsyncThunk<Data, undefined, { state: RootState }>(
    'apiData/getData',
    async () => DataServices.getData(),
    {
        condition: (_, { getState }) =>
        {
            const { data } = getState();
            return data.getDataStatus !== Status.Fulfilled;
        },
    },
);
```

If my condition return false and then we don't do the cycle of createAsyncThunk, my promise exist and do the abort, but the abort console redux-toolkit.esm.js:1345 Uncaught (in promise) {name: "AbortError", message: "Unmount on pending"}.

I have other way to handle this but this way is cleaner. Is it "normal" and I should just handle myself the condition and unmount in the useEffect or is it a bug and the abort should not be triggered or console the Uncaught ?

msutkowski:
- Fixes #567 

Note: There is currently not a good way to test that `unhandledRejection` errors happen or not. This PR introduces a method that will 'crash' the test runner if you happen to forget to resolve a promise.

Things tried before going to the 'crash the test runner' route: 
- using `process.on('unhandledRejection', {})` to use a callback function, and check that it is fired.
- logging in various ways and test with `getLog()`.

After some research, it seems that jest doesn't handle process.on in the test file itself, which is why it's moved into the config. See: https://github.com/facebook/jest/issues/5620

### DIFF
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -339,6 +339,7 @@ If you want to use the AbortController to react to \`abort\` events, please cons
             options.condition &&
             options.condition(arg, { getState, extra }) === false
           ) {
+            abortedPromise.catch(() => {})
             throw {
               name: 'ConditionError',
               message: 'Aborted due to condition callback returning false.'
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -326,9 +326,12 @@ If you want to use the AbortController to react to \`abort\` events, please cons
         )
       )
 
+      let started = false
       function abort(reason?: string) {
-        abortReason = reason
-        abortController.abort()
+        if (started) {
+          abortReason = reason
+          abortController.abort()
+        }
       }
 
       const promise = (async function() {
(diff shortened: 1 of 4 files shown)


src/createAsyncThunk.ts
```diff
@@ -339,6 +339,7 @@ If you want to use the AbortController to react to \`abort\` events, please cons
             options.condition &&
             options.condition(arg, { getState, extra }) === false
           ) {
+            abortedPromise.catch(() => {})
```
phryneas, line 342: I know this was initially my idea but I'm debating a bit if it would not be more future-proof, if we did something like
```diff

+    let started = false;
      function abort(reason?: string) {
+      if (started) {
        abortReason = reason
        abortController.abort()
+      }
      }

// ....


          }
+         started = true;
          dispatch(pending(requestId, arg)
```

msutkowski: Agreed on this.

PR #574, opened 2020-05-20
Export return type of createAsyncThunk

### LINKED ISSUE #571: Return type of createAsyncThunk
New issue as per #486.

Using essentially the same example from that issue:

```ts
type ThunkApiConfig = {
    dispatch: MyDispatch:
    rejectValue: MyErrorType;
};

export function createAppThunk<Returned, ThunkArg = void>(
    type: string,
    payloadCreator: AsyncThunkPayloadCreator<Returned, ThunkArg>
): AsyncThunk<Returned, ThunkArg> {
    return createAsyncThunk<Returned, ThunkArg, ThunkApiConfig>(type, payloadCreator);
}
```

In this toy example, `AsyncThunk` (not a real type at the moment) would be the type of the return value of `createAsyncThunk`.  As far as I can tell, there is no way to reference this type in a declaration.  (`ReturnType<typeof createAsyncThunk>` doesn't work because `createAsyncThunk` is generic.)

In this toy example the compiler can infer the type from the function body and so you can omit the return type declaration entirely, but this inference is not always possible.

For example, consider this function definition using overloads:

```ts
export function createApiThunk<Returned, ThunkArg>(
    type: string,
    api: (arg: ThunkArg) => Promise<Returned>
): AsyncThunk<Returned, ThunkArg>;

export function createApiThunk<Returned, ThunkArg, ApiResult>(
    type: string,
    api: (arg: ThunkArg) => Promise<ApiResult>,
    transform: (arg: ApiResult) => Returned
): AsyncThunk<Returned, ThunkArg>;

export function createApiThunk<Returned, ThunkArg, ApiResult>(
    type: string,
    api: (arg: ThunkArg) => Promise<ApiResult>,
    transform?: (arg: ApiResult) => Returned
): AsyncThunk<ApiResult | Returned, ThunkArg> {
    return createAsyncThunk<Returned, ThunkArg, ThunkApiConfig>(type,
        (arg, { rejectWithValue }) => api(arg)
            .then(value => transform ? transform(value) : value)
            .catch((err: MyErrorType) => rejectWithValue(err))
    );   
}
```

In this case, the function overload definitions can't get an inferred return type from the function body, so omitting them would yield a return type of `any`.  Getting strict typing then requires the type `AsyncThunk` to be available.

smrq:
Resolves #571.

### DIFF
--- a/etc/redux-toolkit.api.md
+++ b/etc/redux-toolkit.api.md
@@ -60,6 +60,14 @@ export interface ActionReducerMapBuilder<State> {
 export type Actions<T extends keyof any = string> = Record<T, Action>;
 
 // @public
+export type AsyncThunk<Returned, ThunkArg, ThunkApiConfig extends AsyncThunkConfig> = AsyncThunkActionCreator<Returned, ThunkArg, ThunkApiConfig> & {
+    pending: AsyncThunkPendingActionCreator<ThunkArg>;
+    rejected: AsyncThunkRejectedActionCreator<ThunkArg, ThunkApiConfig>;
+    fulfilled: AsyncThunkFulfilledActionCreator<Returned, ThunkArg>;
+    typePrefix: string;
+};
+
+// @public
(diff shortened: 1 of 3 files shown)


mKontakis:
Very useful, looking forward of that being merged.

h0tw4t3r:
Can #579 be affected by this?

phryneas:
> Can #579 be affected by this?

Completely different issue. I explained it over in #579.

PR #581, opened 2020-05-25
Avoid unnecesary prototype loop

WebReflection:
As the whole purpose of the `while (Object.getPrototypeOf(proto) !== null)` loop seems to counter validate there was nothing else to loop, I believe it's a performance benefit, as well as smaller code, to simply check the prototype never more than 2 levels, with a graceful fallback in case the first check resulted into `null` already:
  * the amount of `Object.getPrototypeOf` is fixed as `2` and never more
  * there is less code to shrink/optimize for bundlers or minifiers

### DIFF
--- a/src/isPlainObject.ts
+++ b/src/isPlainObject.ts
@@ -7,12 +7,9 @@
  * @returns {boolean} True if the argument appears to be a plain object.
  */
 export default function isPlainObject(value: unknown): value is object {
-  if (typeof value !== 'object' || value === null) return false
-
-  let proto = value
-  while (Object.getPrototypeOf(proto) !== null) {
-    proto = Object.getPrototypeOf(proto)
-  }
-
-  return Object.getPrototypeOf(value) === proto
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    Object.getPrototypeOf(Object.getPrototypeOf(value) || 0) === null
+  )
 }


src/isPlainObject.ts
```diff
@@ -7,12 +7,9 @@
...
-  let proto = value
-  while (Object.getPrototypeOf(proto) !== null) {
-    proto = Object.getPrototypeOf(proto)
-  }
-
-  return Object.getPrototypeOf(value) === proto
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    Object.getPrototypeOf(Object.getPrototypeOf(value) || 0) === null
```
WebReflection, line 13: P.S. the `0` could be just a literal `{}` instead, it'd work the exact same.

markerikson:
This was originally implemented by @timdorr and we copied it from the Redux core.

I'm not inclined to change it unless there's truly a specific need, but I'll let him weigh in on it.

timdorr:
Yeah, I'm not sure I really see the benefit here. Thanks, but this is fine as-is.

WebReflection:
> I'm not sure I really see the benefit here

is performance a concern of this project?

this brings _O(1)_ vs _O(n)_ on the plate, so it's faster, but also smaller.

every class based object, or `Object.create(other)` one, will crawl more than twice the prototype for no reason, and copy-pasting the initial solution shouldn't block this from improving, right?

timdorr:
_Theoretically_, this runs in linear time, but _practically_ it's constant time. The vast majority of users are passing in plain objects where this check is run. Those will have a single level prototype chain, meaning this `while` loop will run once and essentially become an `if` statement. 

Regardless, we've been over this a number of times. The previous conclusion is this code isn't in the hot path of execution, so perfecting the execution time isn't critical and doesn't net any visible benefit. Even if it was, the compiler is going to optimize it anyway and it won't end up being more than a few opcodes.

WebReflection:
> The vast majority of users are passing in plain objects where this check is run. Those will have a single level prototype chain, meaning this while loop will run once and essentially becomes an if statement.

current code runs `Object.getPrototypeOf` *at least* twice or more, my code runs in never more than twice.

> this code isn't in the hot path of execution, so perfecting the execution time isn't critical and doesn't net any visible benefit

to me it's a matter of logic. The `while` loop is redundant, and the `if` could be considered an edge case, covered by `|| 0` in this proposed change.

> the compiler is going to optimize it anyway and it won't end up being more than a few opcodes.

if it's not critical path I understand, but the current code execute more instructions than it needs.

Metaphorically speaking, why would you prefer this code:
```js
let current = i;
while (i--);
return (i + 1) === current;
```

instead of this?
```js
return --i === 0;
```

anyway, your projects, your rules, I just personally welcome simple improvements, when no regressions whatsoever are presented, only benefits (even if logical, or theoretical).

Best Regards

src/isPlainObject.ts
secf4ult: I think it lacks the check for `Object.create(null)`.

PR #586, opened 2020-05-29
Docs: Use configureStore in createAction example

mattjcheah:
In the Basic Tutorial, `configureStore` is introduced with an example before `createAction`. This implies that the app example at this point would use `configureStore` instead of `createStore`.

However in the `createAction` example, it still uses `createStore`. This change updates the docs to use `configureStore` in the `createAction` example since `configureStore` was introduced earlier.


markerikson:
Thanks!

PR #595, opened 2020-06-08
Allow undefined return type for createAsyncThunk options

### LINKED ISSUE #594: Confirm createAsyncThunk's condition return type?
Hi all - encountering an issue with the return type of `condition` method. [The docs state](https://redux-toolkit.js.org/api/createAsyncThunk#canceling-before-execution) that you only need to return false if the condition is met. There isn't any info about also having to return something outside of that condition.

So I'm getting the following type error when only returning false:
`Type 'false | undefined' is not assignable to type 'boolean'.`

The solution is either return something outside of the condition...

```js
condition: (userId, { getState, extra }) => {
  if (someCondition) {
    return false;
  }
  return true;
}
```

Or update the typings to set the return type as `boolean | undefined` so you don't need to return something. 

I can contribute a fix to the types or add some clarification to the docs, just not sure what's the intended behavior 🙂  For now I'm returning true outside the condition.

melanieseltzer:
Closes #594

### DIFF
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -199,12 +199,12 @@ interface AsyncThunkOptions<
    * A method to control whether the asyncThunk should be executed. Has access to the
    * `arg`, `api.getState()` and `api.extra` arguments.
    *
-   * @returns `true` if the asyncThunk should be executed, `false` if it should be skipped
+   * @returns `false` if it should be skipped
    */
   condition?(
     arg: ThunkArg,
     api: Pick<GetThunkAPI<ThunkApiConfig>, 'getState' | 'extra'>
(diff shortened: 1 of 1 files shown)


markerikson:
Thanks! I'll try to get out another release in the next day or so.

markerikson:
Okay, okay, so "next day or so" turned into "13 days".  Sorry!

But happy to say that https://github.com/reduxjs/redux-toolkit/releases/tag/v1.4.0 is out now :)

melanieseltzer:
Ah awesome! 👏 Many thanks!

markerikson:
Sure, and appreciate the detailed issue and PR!

PR #597, opened 2020-06-11
Upgrade immer

dudiharush:
[immer 7](https://github.com/immerjs/immer/releases)


dudiharush:
Conflicts

phryneas:
We have this on our todo list, but it will require some internal modifications. So stay tuned, it will probably come in a few days time.

PR #604, opened 2020-06-12
Bump immer to v7, export current util, add usage docs

msutkowski:
- [x] Review comment about adding an `isDraftable` check inside `createReducer` from Discord

  > and mweststrate pinged me about adding an isDraftable check to createReducer.

- [x] Figure out how to best warn about a non-draftable state

- [x] Review selector usage inside of a reducer and references to 'leaking outside of the draft' - https://github.com/immerjs/immer/issues/561, would need reproduction in a test

### DIFF
--- a/docs/api/createReducer.md
+++ b/docs/api/createReducer.md
@@ -141,16 +141,17 @@ const todosReducer = createReducer([], {
 It's very common for a developer to call `console.log(state)` during the development process. When using either `createSlice` or `createReducer`, you will need to use the [`current`](./otherExports#current.md) utility that we re-export from the [`immer` library](https://immerjs.github.io/immer). This awesome utility allows us to peek into the Proxy while it's being modified.
 
 ```ts
+// todosSlice.js
+import { createSlice, current } from '@reduxjs/toolkit'
+
 const slice = createSlice({
   name: 'todos',
   initialState: [{ id: 1, title: 'Example todo' }],
   reducers: {
-    updateTodo: (state, action) => {
+    addTodo: (state, action) => {
       console.log('before', current(state))
-      state.push(action)
-      console.log('after push', current(state))
-      state = state.filter(todos => todo.id !== 1)
-      console.log('final, after filtering one', current(state))
+      state.push(action.payload)
+      console.log('after', current(state))
     }
   }
 })
--- a/docs/api/createReducer.md
+++ b/docs/api/createReducer.md
@@ -135,3 +135,23 @@ const todosReducer = createReducer([], {
   }
 })
 ```
+
+## Debugging your state
+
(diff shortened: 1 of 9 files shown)


msutkowski:
This should be good to go. Adding a few additional tests that explicitly expect edge cases on return values. Pending an update here: https://github.com/immerjs/immer/pull/617

package.json
```diff
@@ -58,7 +58,7 @@
     "src"
   ],
   "dependencies": {
-    "immer": "^6.0.1",
+    "immer": "^7.0.0",
```
phryneas, line 61: Not that it would be *really* necessary, but want to make this a 7.0.1 just to be more explicit? (& update the package.lock, which will be necessary for the tests to pass) 
Other than that, I think it's ready to go :)

markerikson: yep, exactly

msutkowski: updated!

docs/api/createReducer.md
```diff
@@ -135,3 +135,24 @@ const todosReducer = createReducer([], {
   }
 })
 ```
+
+## Debugging your state
+
+It's very common for a developer to call `console.log(state)` during the development process. When using either `createSlice` or `createReducer`, you will need to use the [`current`](./otherExports#current.md) utility that we re-export from the [`immer` library](https://immerjs.github.io/immer). This awesome utility allows us to peek into the Proxy while it's being modified.
```
markerikson, line 141:
```suggestion
It's very common for a developer to call `console.log(state)` during the development process. However, browsers display Proxies in a format that is hard to read, which can make console logging of Immer-based state difficult.

When using either `createSlice` or `createReducer`, you may use the [`current`](./otherExports#current.md) utility that we re-export from the [`immer` library](https://immerjs.github.io/immer). This utility creates a separate plain copy of the current Immer `Draft` state value, which can then be logged for viewing as normal.
```

docs/api/otherExports.md
```diff
@@ -156,6 +156,23 @@ console.log(nanoid())
 
 The default immutable update function from the [`immer` library](https://immerjs.github.io/immer/), re-exported here as `createNextState` (also commonly referred to as [`produce`](https://immerjs.github.io/immer/docs/produce))
 
+### `current`
+
+The `current` function from the [`immer` library](https://immerjs.github.io/immer/), which takes a snapshot of the current state of a draft and finalizes it (but without freezing). Current is a great utility to print the current state during debugging, and the output of current can also be safely leaked outside the producer.
```
markerikson, line 161:
```suggestion
[The `current` function](https://immerjs.github.io/immer/docs/current) from the [`immer` library](https://immerjs.github.io/immer/), which takes a snapshot of the current state of a draft and finalizes it (but without freezing). Current is a great utility to print the current state during debugging, and the output of `current` can also be safely leaked outside the producer.
```

PR #610, opened 2020-06-13
add `addMatcher` to builder notations & `actionMatchers` argumen…

phryneas:
Still experimenting with some types & this will need some additional documentation, but I already want to put this through the CI.

The idea behind this is to enable usages such as this:

```ts
type Status = 'pending' | 'rejected' | 'fulfilled';

// Type guard function
function isPendingAction(action: AnyAction): action is PendingAction {
  return action.type.endsWith('/pending') && 'requestId'  in action.meta
};

const thunkStateSlice = createSlice({
  name: "thunkState",
  initialState: {} as {[requestId: string]: Status},
  reducers: {
    reset(){ return {} }
  },
  extraReducers: builder => builder
    // kind of "wildcard-y" extraReducer with a guard function - action is automatically inferred as `PendingAction` due to the signature of `isPendingAction`
    .addMatcher(isPendingAction, (state, action) => {
      state[action.meta.requestId] = 'pending'
    })
})
```

### DIFF
--- a/docs/api/createSlice.md
+++ b/docs/api/createSlice.md
@@ -94,6 +94,8 @@ createSlice({
 
 Instead of using a simple object as `extraReducers`, you can also use a callback that receives a `ActionReducerMapBuilder` instance.
 
+This builder notation is also the only way to add matcher reducers and default case reducers to your slice.
+
 ```typescript
 const incrementBy = createAction<number>('incrementBy')
 
--- a/docs/api/createSlice.md
+++ b/docs/api/createSlice.md
@@ -16,18 +16,27 @@ and automatically generates action creators and action types that correspond to 
 
 ```ts
 function createSlice({
+      // A name, used in action types
+    name: string,
+        // The initial state for the reducer
+    initialState: any,
     // An object of "case reducers". Key names will be used to generate actions.
     reducers: Object<string, ReducerFunction | ReducerAndPrepareObject>
-    // The initial state for the reducer
-    initialState: any,
-    // A name, used in action types
-    name: string,
-    // An additional object of "case reducers". Keys should be other action types.
+    // An additional object of "case reducers", where the keys should be other
+    // action types, or a "builder callback" function used to add more reducers
     extraReducers?:
     | Object<string, ReducerFunction>
     | ((builder: ActionReducerMapBuilder<State>) => void)
 })
 ```
+
+### `initialState`
+
+The initial state value for this slice of state.
+
+### `name`
+
+A string name for this slice of state. Generated action type constants will use this as a prefix.
 
 ### `reducers`
 
@@ -42,17 +51,38 @@ descriptive names.
 This object will be passed to [`createReducer`](./createReducer.md), so the reducers may safely "mutate" the
 state they are given.
 
+```js
+const counterSlice = createSlice({
+  name: 'counter',
+  initialState: 0,
+  reducers: {
+    increment: state => state + 1
+  }
+})
+// Will handle the action type `'counter/increment'`
+```
+
 #### Customizing Generated Action Creators
 
-If you need to customize the creation of the payload value of an action creator by means of a [`prepare callback`](./createAction.md#using-prepare-callbacks-to-customize-action-contents), the value of the appropriate field of the `reducers` argument object should be an object instead of a function. This object must contain two properties: `reducer` and `prepare`. The value of the `reducer` field should be the case reducer function while the value of the `prepare` field should be the prepare callback function.
+If you need to customize the creation of the payload value of an action creator by means of a [`prepare callback`](./createAction.md#using-prepare-callbacks-to-customize-action-contents), the value of the appropriate field of the `reducers` argument object should be an object instead of a function. This object must contain two properties: `reducer` and `prepare`. The value of the `reducer` field should be the case reducer function while the value of the `prepare` field should be the prepare callback function:
 
-### `initialState`
-
-The initial state value for this slice of state.
-
-### `name`
-
-A string name for this slice of state. Generated action type constants will use this as a prefix.
+```js
+const todosSlice = createSlice({
+  name: 'todos',
+  initialState: [],
+  reducers: {
+    addTodo: {
+      reducer: (state, action) => {
+        state.push(action.payload)
+      },
+      prepare: text => {
+        const id = nanoid()
+        return { payload: { id, text } }
+      }
+    }
+  }
+})
+```
 
 ### `extraReducers`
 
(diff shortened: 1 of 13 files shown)


src/createReducer.ts
```diff
@@ -97,15 +116,21 @@ export function createReducer<S>(
...
       ? executeReducerBuilderCallback(mapOrBuilderCallback)
-      : mapOrBuilderCallback
+      : [mapOrBuilderCallback]
 
   return function(state = initialState, action): S {
-    const caseReducer = actionsMap[action.type]
+    const caseReducer =
+      actionsMap[action.type] ||
+      actionMatchers
+        .concat(builderActionMatchers)
```
markerikson: Silly micro-optimization: can we do the  `concat` once outside of the returned function?

src/createReducer.test.ts
```diff
@@ -172,6 +274,121 @@ describe('createReducer', () => {
...
+      expect(reducer(undefined, decrementBy(1))).toEqual({
+        numberActions: 1,
+        stringActions: 0
+      })
+      expect(reducer(undefined, concatWith('foo'))).toEqual({
+        numberActions: 0,
+        stringActions: 1
+      })
+    })
+    test('prefers explicit reducer cases over actionMatchers', () => {
```
markerikson, line 322: Question on the possible semantics: what if we want to have _both_ the exact match _and_ the wildcard match run?  Is that a desirable behavior?

phryneas: We could even go so far as to run all matchers, but all that would be pretty un-reducer-like behaviour for my intuition.

But generally, I'm open for discussion there.

The question would be in which order they run.

src/createReducer.ts
phryneas: Good micro-point. Will do.

src/createReducer.test.ts
markerikson: https://github.com/redux-utilities/reduce-reducers is totally a thing, and in fact I'd even talked about adding it to RTK at some point.  Presumably they'd run in the order they were defined.

src/createReducer.test.ts
```diff
@@ -106,6 +106,105 @@ describe('createReducer', () => {
...
+    const initialState = { numberActions: 0, stringActions: 0 }
+    const numberActionsCounter = {
+      matcher: numberActionMatcher,
+      reducer(state: typeof initialState) {
+        state.numberActions = state.numberActions * 10 + 1
+      }
+    }
+    const stringActionsCounter = {
+      matcher: stringActionMatcher,
+      reducer(state: typeof initialState) {
```
markerikson, line 137: I don't see any test cases that use `action` here. What's the inferred action type like, if any?

src/createReducer.test.ts
```diff
@@ -204,6 +303,103 @@ describe('createReducer', () => {
...
+        builder.addMatcher(incrementBy.match, state => {
+          state.numberActions += 100
+        })
+      )
+      expect(reducer(undefined, incrementBy(1))).toEqual({
+        numberActions: 100,
+        stringActions: 0
+      })
+    })
+    test('calling addMatcher followed by addCase should result in an error in development mode', () => {
```
markerikson, line 391: Is this test obsolete now?

type-tests/files/mapBuilders.typetest.ts
```diff
@@ -50,5 +50,27 @@ function expectType<T>(t: T) {
...
+    builder
+      .addCase(
+        'increment',
+        (state, action: ReturnType<typeof increment>) => state
+      )
+      .addMatcher(decrement.match, (state, action) => {
+        expectType<ReturnType<typeof decrement>>(action)
+      })
+
+    // addMatcher().addCase() is not possible as `addCase` should be removed after calling `addMatcher`
```
markerikson, line 69: Are we actually still doing this?

markerikson: Eh, I see that we are, okay.

src/createReducer.test.ts
phryneas: Over in the typetest in https://github.com/reduxjs/redux-toolkit/pull/610/files#diff-8e07add1ff60e559fa564decce5eb329R66

Here, it would be `PayloadAction<string>` due to the declaration of `stringActionMatcher`.

src/createReducer.test.ts
phryneas: Given that we always run the matcher reducers after the case reducer, we should still enforce this.

phryneas:
And we have `addDefaultCase`. I'd leave it at that for now. @msutkowski if you can take a look at docs later, that would be great - otherwise I'll try & write some docs tomorrow.

docs/api/createSlice.md
```diff
@@ -94,6 +94,8 @@ createSlice({
 
 Instead of using a simple object as `extraReducers`, you can also use a callback that receives a `ActionReducerMapBuilder` instance.
 
+This builder notation is also the only way to add matcher reducers and default case reducers to your slice.
```
msutkowski, line 97: I think this will need to be expanded on and/or include links to those concepts. We should also consider updating examples in `createSlice` to show the new properties.

msutkowski:
I went through the docs and found some more places that need to be updated with this PR. I also added some notes about the Advanced Tutorial that can get moved into its own issue.

### Quick start
- `createReducer()`
  - [ ] Update definition

### Basic tutorial:
- Introducing: `createReducer`
  - Could be updated, but not necessary
- Summary: createReducer
  - [ ] Update `createReducer` definition
  
### Intermediate tutorial:
 - #Summary
   - [ ] Remove or update this copy: 
    > Coming soon: an Advanced Tutorial, where we'll look at how to use RTK in an app that does async data fetching and uses TypeScript.

### Advanced tutorial

**General notes:** Being that the tutorials link to specific commits in a repo, we should consider rewriting this page and pointing to a codesandbox instead of commits. I think the commit-linking approach locks us into examples that could easily become outdated. If we keep this as an example of a 'Migration guide', this would become a non-issue with a new tutorials section.

#### Changes without a new Advanced Tutorial:
- #Writing Thunks in Redux Toolkit
  - [ ] Update copy: 
    > However, Redux Toolkit does not currently provide any special functions or syntax for writing thunk functions. In particular, they cannot be defined as part of a createSlice() call. You have to write them separate from the reducer logic.
    
#### Changes with a new Advanced Tutorial:
   - [ ] Update to use `createAsyncThunk`?
   - [ ] Use `createEntityAdapter`
   - [ ] `useAppDispatch` instead of `useDispatch` in the TS example
   - [ ] show a generic reducer loading state with `addMatchers` by toggling value on /pending|rejected|fulfilled
   - [ ] show debugging actions with `addDefaultCase`

### Usage guide
 - #Async Requests with createAsyncThunk
   - [ ] Add paragraph about how cancellation is handled?
 - #Normalizing with normalizr
   - [ ] Add a note about `normalizr` typically being called in the action creator so that reducers can easily extract it's desired data _(this question comes up often in discord)_


### Usage with Typescript
 - #createReducer
   - [ ] Show additional properties (addMatchers/addDefaultCase) and a relevant example

markerikson:
@msutkowski : copied your list there over a new issue in #628

PR #612, opened 2020-06-16
Update broken link in immutable invariant middleware

msutkowski:
- Updates a broken link.

_Side note: Checked every other link in the repo and they're all good._


markerikson:
Yep. I hate that our docs URLs have changed over time :(

PR #621, opened 2020-06-18
Fixed multiple updates with same id in updateMany

### LINKED ISSUE #619: `updateMany` ignores early updates if later updates have the same ID
The `updateMany` action generated by the `createEntityAdapter` function is not working as intended when there are multiple updates to the same entity.

Consider the following code:

```ts
const slice = createSlice({
  name: 'books',
  initialState: booksAdapter.getInitialState(),
  reducers: {
    bookAdd: adapter.addOne,
    bookUpdates: adapter.updateMany,
  },
});

// configure store and add books ...

store.dispatch(
  slice.actions.bookUpdates([
    { id: 1, changes: { title: 'New Title 1' } },
    { id: 2, changes: { title: 'New Title 2' } },
    { id: 1, changes: { author: 'New Author 1' } },
    { id: 2, changes: { author: 'New Author 2' } },
  ])
);
```

The outcome of that last statement should be that entity `1` has a title of "New Title 1" _and_ a new author of "New Author 1".  But what actually happens is that the `author` is changed but not the title.

The [docs](https://redux-toolkit.js.org/api/createEntityAdapter) say, "If `updateMany()` is called with multiple updates targeted to the same ID, they will be merged into a single update, with later updates overwriting the earlier ones."  That's a little ambiguous.  It could mean that the later updates should be merged with earlier updates (not conflicting if the fields are different), or that the later updates would completely replace earlier updates with the same ID.  It's doing the latter, but should do the former.

See the codesandbox here:

https://codesandbox.io/s/redux-toolkit-updatemany-test-verzl?file=/src/App.tsx

jakeboone02:
Fixes #619.

### DIFF
--- a/src/entities/fixtures/book.ts
+++ b/src/entities/fixtures/book.ts
@@ -1,6 +1,7 @@
 export interface BookModel {
   id: string
   title: string
+  author?: string
 }
 
 export const AClockworkOrange: BookModel = Object.freeze({
@@ -16,4 +17,10 @@ export const AnimalFarm: BookModel = Object.freeze({
 export const TheGreatGatsby: BookModel = Object.freeze({
   id: 'tgg',
   title: 'The Great Gatsby'
+})
+
+export const TheHobbit: BookModel = Object.freeze({
+  id: 'th',
+  title: 'The Hobbit',
+  author: 'J. R. R. Tolkien'
 })
--- a/src/entities/unsorted_state_adapter.test.ts
+++ b/src/entities/unsorted_state_adapter.test.ts
@@ -4,7 +4,8 @@ import {
   BookModel,
   TheGreatGatsby,
   AClockworkOrange,
-  AnimalFarm
+  AnimalFarm,
+  TheHobbit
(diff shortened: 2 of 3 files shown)


jakeboone02:
@phryneas - actually the build is failing because this repo doesn't support optional chaining yet (tests pass, but build fails).  Should I rewrite to remove the optional chaining syntax, or configure the build to support it?  I don't currently know how to do the latter, but would prefer it to removing the syntax as it's much cleaner.

phryneas:
@jakeboone02 argh, good point, I only looked over the code. 

Please remove the optional chaining. We are running the tests for every TypeScript versions down to 3.5 and as we are running those against the source code, not the compiled product, the source code has to be compatible with 3.5.

markerikson:
Yep, looks good, thanks!

This will likely go out in 1.4.0 once we get #610 wrapped up.

markerikson:
Out in https://github.com/reduxjs/redux-toolkit/releases/tag/v1.4.0

PR #623, opened 2020-06-18
Add section to usage guide on working with non-serializable data

cloeper:
Adds
* Section to usage guide on working with non-serializable data
* Updates error message thrown when encountering non-serializable values to include a link to the new section

### DIFF
--- a/docs/usage/usage-guide.md
+++ b/docs/usage/usage-guide.md
@@ -84,7 +84,7 @@ import { configureStore } from '@reduxjs/toolkit'
 import rootReducer from './reducers'
 
 const store = configureStore({
-  reducer: rootReducer
+  reducer: rootReducer,
 })
 
 export default store
@@ -99,8 +99,8 @@ import postsReducer from './postsReducer'
 const store = configureStore({
   reducer: {
     users: usersReducer,
-    posts: postsReducer
-  }
+    posts: postsReducer,
+  },
 })
 ```
 
@@ -120,7 +120,7 @@ export default function configureAppStore(preloadedState) {
     reducer: rootReducer,
     middleware: [loggerMiddleware, ...getDefaultMiddleware()],
     preloadedState,
-    enhancers: [monitorReducersEnhancer]
+    enhancers: [monitorReducersEnhancer],
   })
 
   if (process.env.NODE_ENV !== 'production' && module.hot) {
@@ -165,7 +165,7 @@ function todosReducer(state = [], action) {
 
         return {
           ...todo,
-          completed: !todo.completed
+          completed: !todo.completed,
         }
       })
     }
@@ -196,7 +196,7 @@ const todosReducer = createReducer([], {
   REMOVE_TODO: (state, action) => {
     // Can still return an immutably-updated value if we want to
     return state.filter((todo, i) => i !== action.payload.index)
-  }
+  },
 })
 ```
 
@@ -273,7 +273,7 @@ Most action creators are very simple. They take some parameters, and return an a
 function addTodo(text) {
   return {
     type: 'ADD_TODO',
-    payload: { text }
+    payload: { text },
   }
 }
 ```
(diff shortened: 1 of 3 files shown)


cloeper:
So, there's a number of failing tests on master. About 10 tests fail, all look like this:

```
TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
src/configureStore.ts:178:5 - error TS2345: Argument of type 'DeepPartial<S>' is not assignable to parameter of type 'PreloadedState<S>'.
```

My changes have nothing to do with that, so I'm unsure how to proceed.

markerikson:
Hmm. Looking at build failures listed in Travis, they specifically seem to be about the snapshots for the serializability middleware. Not immediately seeing any TS-related issues.

markerikson:
Also, why is that "fixes" commit full of formatting changes? Are we flip-flopping Prettier versions between people's installs or something?

cloeper:
> Also, why is that "fixes" commit full of formatting changes? Are we flip-flopping Prettier versions between people's installs or something?

I wish I could tell you. I've run `yarn format` so I would assume that would resolve any issue.

cloeper:
Also, I updated the snapshots, but I have no idea why the tests aren't passing in travis. The only tests that don't pass for me locally are the ones I mentioned earlier.

 This repo just isn't behaving as I would expect it to and I don't have the time to figure it out. Feel free to take the changes from my commits and adding them in a different PR.

phryneas:
There was a rogue space. I pushed a fix, let's see if it works now.

markerikson:
I think we at least ought to have a specific example of setting up with Redux-Persist, because A) it's a legit example of customizing the behavior, and B) it's come up fairly frequently in the issues.

msutkowski:
@markerikson There for sure is with `react-redux-firebase` as they add Timestamp instances into the store and other `instance` types in almost every action. I've made a few contributions to that and `redux-firestore` and plan on opening a PR for their 4.0 release to address the serializable as all of that can be handled rather simply by calling `instance.toString()` or `instance.toJSON()` depending on the firebase instance type.

Either way, an example of the configuration needed for those libs looks like this:

```ts
import {
    configureStore,
    getDefaultMiddleware,
  } from "@reduxjs/toolkit";
  import {
    getFirebase,
    actionTypes as rrfActionTypes,
  } from "react-redux-firebase";
  import { constants as rfConstants } from "redux-firestore";
  import rootReducer from "./rootReducer";
  import { useDispatch } from "react-redux";

  const extraArgument = {
    getFirebase,
  };
  export type DefaultThunkExtra = typeof extraArgument;
  
  const middleware = [
    ...getDefaultMiddleware({
        serializableCheck: {
          ignoredActions: [
              // just ignore every redux-firebase and react-redux-firebase action type
            ...Object.keys(rfConstants.actionTypes).map(
              (type) => `${rfConstants.actionsPrefix}/${type}`
            ),
            ...Object.keys(rrfActionTypes).map((type) => `@@reactReduxFirebase/${type}`),
          ],
          ignoredPaths: ["firebase", "firestore"],
        },
        thunk: {
          extraArgument,
        },
      }),
  ] as const;
  
    reducer: rootReducer,
    middleware,
  });

  export type AppDispatch = typeof store.dispatch;
  export const useAppDispatch = () => useDispatch<AppDispatch>();
  
  export default store;
  ```

_*Note: there is more typing that needs to be done specifically for the reducers they ship to be TS compatible with RTK, but is probably a separate issue. I haven't seen many people mention that, but a PR I had opened for the problem is hopefully getting merged in their 3.6 release*_

Looking at that, we should probably add another property of `ignoreActionPrefix` for these scenarios _(or some other partial matcher)_. I believe `redux-websocket` still has this same type of issue, although a more limited # of action types. I'm sure there's even more that I'm not aware of 🤷

markerikson:
I'll go ahead and merge this, and do some more edits on the docs content from there.  Thanks!

markerikson:
Live in https://github.com/reduxjs/redux-toolkit/releases/tag/v1.4.0

PR #631, opened 2020-06-22
allow to return undefined for null-states in caseReducers

phryneas:
This was just brought up in the chat. `null` is non-draftable, but we should accept the reducer not returning anything in that specific case & not modify the state (keeping it `null`).
This might be the case where a reducer has the state `S | null` and just skips state modification while `draft` is `null`.

### DIFF
--- a/src/createReducer.test.ts
+++ b/src/createReducer.test.ts
@@ -274,6 +274,20 @@ describe('createReducer', () => {
         `"A case reducer on a non-draftable value must not return undefined"`
       )
     })
+    test('allows you to return undefined if the state was null, thus skipping an update', () => {
+      const reducer = createReducer(null as number | null, builder =>
+        builder.addCase(
+          'decrement',
+          (state, action: { type: 'decrement'; payload: number }) => {
+            if (typeof state === 'number') {
+              return state - action.payload
+            }
+          }
+        )
+      )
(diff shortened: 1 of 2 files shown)


markerikson:
For whatever reason I'm not entirely grokking the bug here, even after having read through the discussion in the channel.

Why is `return state` an issue when the state is null?

phryneas:
`state` is `null` => `state` is not "draftable", so we don't run the reducer with `immer`.
In that case we have to prevent the reducer from returning `undefined`, because that would set the state to `undefined`. Which is why usually, we'd throw an error to tell the user that they *really*  should return something.

The problem in the chat was that the type of the state was *sometimes* an object (draftable, takes the `immer` path) and *sometimes* `null` (non-draftable, does not take the `immer` path).


So if you have a caseReducer like this, for a state of `S|null`:
```
caseReducer(state, action) {
  if (state && action.payload.something == "foo") {
    state.something = action.payload.otherThing
  }
}
```
That would make sense from the point of someone writing that reducer - but sometimes it would take the `produce` path and sometimes not, causing an error.

So we'd have to handle the `undefined` return value extra for the case that the `state` was `null` to begin with - and assume that it just would imply "no not modify the state" instead of "reset it to `undefined`" or "throw an error".

markerikson:
Awright, if you say this works right, I believe you :)

PR #632, opened 2020-06-22
Update compressed-size action

msutkowski:
Hoping this fixes the action as it fails often.

- Use @v2 instead of @v2-beta
- Remove fetch-depth as the default is 1

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -7,9 +7,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2-beta
-        with:
-          fetch-depth: 1
+      - uses: actions/checkout@v2
       - uses: preactjs/compressed-size-action@v1
         with:
           repo-token: '${{ secrets.GITHUB_TOKEN }}'


phryneas:
:trophy:

markerikson:
So one issue we've seen is that it doesn't show sizes for PRs from forks or something like that.

Supposedly https://github.com/kuldeepkeshwar/size-plugin-bot works better because it's a bot and not an action, or so the author told me.

Should we give that a shot?

msutkowski:
@markerikson It looks like the action runs, but you have to click 'view details' to see the info instead of it posting a comment to the PR itself. 

In addition to what you linked, here are two alternatives by popularity:
1. ~~https://github.com/siddharthkp/bundlesize~~ - Being replaced by https://github.com/bundlewatch/bundlewatch#i-use-bundlesize-why-should-i-switch-to-bundlewatch ?
2. https://github.com/ai/size-limit

markerikson:
Yeah, you can definitely click into the "Details" on PRs, but half the point is to get it as an auto-updated comment :)

Worth a shot to compare them.

PR #645, opened 2020-07-01
Docs Idea: Showing a simple TS example of passing args to a matcher func

msutkowski:
I'm not sure if there is a lot of value here, but we currently don't show this kind of example anywhere and this could potentially save someone time. 

For a more realistic generic use case: https://codesandbox.io/s/rtk-14-addmatcher-counter-test-l11mt?file=/src/features/counter/counterSlice.ts:2037-2492

Let me know what you think.

### DIFF
--- a/docs/usage/usage-with-typescript.md
+++ b/docs/usage/usage-with-typescript.md
@@ -234,8 +234,12 @@ createReducer({ value: 0 }, builder =>
 Additionally, you may want to pass arguments to a matcher function to create reusable helpers. A pattern for that would look like:
 
 ```ts
-const isScopedNumberValueAction = (scope: string) => (action: AnyAction): action is PayloadAction<{ value: number }> => {
-  return action.type.startsWith(scope) && typeof action.payload.value === 'number'
+const isScopedNumberValueAction = (scope: string) => (
+  action: AnyAction
+): action is PayloadAction<{ value: number }> => {
+  return (
+    action.type.startsWith(scope) && typeof action.payload.value === 'number'
+  )
 }
 
 const slice = createSlice({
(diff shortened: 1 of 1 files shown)


docs/usage/usage-with-typescript.md
```diff
@@ -231,6 +231,25 @@ createReducer({ value: 0 }, builder =>
 })
 ```
 
+Additionally, you may want to pass arguments to a matcher function to create reusable helpers. A pattern for that would look like:
+
+```ts
+const isScopedNumberValueAction = (scope: string) => (action: AnyAction): action is PayloadAction<{ value: number }> {
```
phryneas, line 237:
```suggestion
const isScopedNumberValueAction = (scope: string) => (action: AnyAction): action is PayloadAction<{ value: number }> => {
```

phryneas:
This could even go a step farther, something like

```ts
const isScopedValueAction = <T>(
    scope: string,
    payloadValidator: (payload: any) => payload is T
  ) => (action: AnyAction): action is PayloadAction<T> => {
    return action.type.startsWith(scope) && payloadValidator(action.payload)
  }

  const isNumber = (v: any): v is number => {
    return typeof v === 'number'
  }

  const isScopedNumberValueAction = (scope: string) =>
    isScopedValueAction(scope, isNumber)

  const slice = createSlice({
    name: 'counter',
    initialState: { value: 0 },
    reducers: {},
    extraReducers: builder => {
      builder.addMatcher(
        isScopedNumberValueAction('counter/'),
        (state, action) => {
          state.value += action.payload
        }
      )
    }
  })
```

But I'm not sure if both of these additions are too much for the "usage with TypeScript" section. Those might be better suited as "recipes".

msutkowski:
@phryneas I actually like showing your full example, to be honest. We know that people like to copy the example code and then do extra 'interesting' implementations with generics that normally result in open issues or discord chatter - and this addresses that. Regarding the recipes, we should probably put together an outline of that what should be included. I've been curating a lot of different ways that things can be typed, but I'm not sure about the best way to organize any of that. The only real issues that consistently pop up seem to be about making `createAsyncThunk`, `createSlice`, or `createEntityAdapter` super generic, so I'm not sure how much would need to be included for recipes.

msutkowski:
Closing this due to #807 potentially being merged and have more examples.

PR #656, opened 2020-07-17
Use consistent middleware option type

asdf123101:
The default type of the `middleware` option in `configureStore` is a [tuple](https://github.com/reduxjs/redux-toolkit/blob/master/src/configureStore.ts#L126). However, the type of the `middleware` option in `ConfigureStoreOption` is an [array](https://github.com/reduxjs/redux-toolkit/blob/master/src/configureStore.ts#L48). As a result, if someone type `configureStore` partially, such as `configureStore<State>(...)`. The default tuple type is used to type the `middleware` option and an error of tuple and array mismatching is generated: 
```
Property '0' is missing in type 'MiddlewareArray<ThunkMiddleware<any, AnyAction, null> | ThunkMiddleware<any, AnyAction, undefined> | Middleware<{}, any, Dispatch<AnyAction>>>' but required in type '[ThunkMiddleware<any, AnyAction, null> | ThunkMiddleware<any, AnyAction, undefined>]'.
```
This error was not seen before 775da05287100afc363fa346196e3ef865402851.
Although the doc says `configureStore` doesn't need typing now, changing the type of the `middleware` option in `configureStore` from tuple to array can ensure type consistency and avoid this kind of meaningless error when upgrading the package.

### DIFF
--- a/src/configureStore.test.ts
+++ b/src/configureStore.test.ts
@@ -70,11 +70,11 @@ describe('configureStore', () => {
 
   describe('given custom middleware', () => {
     it('calls createStore with custom middleware and without default middleware', () => {
-      const thank: redux.Middleware = _store => next => action => next(action)
-      expect(configureStore({ middleware: [thank], reducer })).toBeInstanceOf(
+      const thunk: redux.Middleware = _store => next => action => next(action)
+      expect(configureStore({ middleware: [thunk], reducer })).toBeInstanceOf(
         Object
       )
-      expect(redux.applyMiddleware).toHaveBeenCalledWith(thank)
+      expect(redux.applyMiddleware).toHaveBeenCalledWith(thunk)
       expect(devtools.composeWithDevTools).toHaveBeenCalled()
       expect(redux.createStore).toHaveBeenCalledWith(
         reducer,
@@ -86,14 +86,14 @@ describe('configureStore', () => {
 
   describe('middleware builder notation', () => {
     it('calls builder, passes getDefaultMiddleware and uses returned middlewares', () => {
-      const thank = jest.fn((_store => next => action =>
+      const thunk = jest.fn((_store => next => action =>
         'foobar') as redux.Middleware)
 
       const builder = jest.fn(getDefaultMiddleware => {
         expect(getDefaultMiddleware).toEqual(expect.any(Function))
         expect(getDefaultMiddleware()).toEqual(expect.any(Array))
 
-        return [thank]
+        return [thunk]
       })
 
       const store = configureStore({ middleware: builder, reducer })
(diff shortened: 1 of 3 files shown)


src/configureStore.test.ts
```diff
@@ -70,11 +70,11 @@ describe('configureStore', () => {
 
   describe('given custom middleware', () => {
     it('calls createStore with custom middleware and without default middleware', () => {
-      const thank: redux.Middleware = _store => next => action => next(action)
-      expect(configureStore({ middleware: [thank], reducer })).toBeInstanceOf(
+      const thunk: redux.Middleware = _store => next => action => next(action)
```
asdf123101, line 73: I'm not sure if this 'thank' is intentional or a type. If this is intentional I will roll back my changes.

phryneas:
Sorry, I'm combing back to this only now.

It's a bit difficult to make a good decision here, but I believe the current behavior is correct.

Generally: `configureStore` should not be typed manually at all. 
Those generics should all be inferred, including the `State` generic.
This is due to the fact that TypeScript cannot infer a few generics and have some others specified.

So, as you noticed, `configureStore<State>` equals `configureStore<State, AnyAction, [ThunkMiddlewareFor<S>]>`.

Doing this is fine for the most basic case, something like
```ts
  configureStore<number>({
    reducer: () => 0
  })
```
but as soon as you specify middlewares, like
```ts
  configureStore<number>({
    reducer: () => 0,
    middlewares: [myCustomMiddleware] // this will error out for good
  })
```
neither the default value `[ThunkMiddlewareFor<S>]` nor the change you are suggesting `ThunkMiddlewareFor<S>[]` will actually match this here and it will error. This is a good thing as I explain later.

The only situation where your suggestion would help would be if someone were to specify the `thunk` middleware multiple times, like
```ts
  configureStore<number>({
    reducer: () => 0,
    middlewares: [thunkMiddleware, thunkMiddleware, thunkMiddleware]
  })
```
and I can't think of a situation where that might make any sense.

Now, why are these types so restrictive here, you might ask?

It's because, depending on the *exact type* of each single middleware in the middleware array, the `Dispatch` type is inferred.

So if your `myCustomMiddleware` were typed `Middleware<{ (action: {type: "foo"}) => "bar" }, any, {}>`, the signature of `(action: {type: "foo"}) => "bar"` would be added to `store.dispatch`.

This can't just happen if we fall back to something permissive like `Middleware[]` or `ThunkMiddleware[]` I specified above.

So, bottom line: for this to work, you either need to use *no generic argument* or *all generic arguments, very accurately*.

asdf123101:
Thanks for your detailed explanation. I'm closing this PR.

Issue #652, opened 2020-07-15
ES Modules for browsers


jonjharrison:
Would be great to add support for browser-friendly ES modules if possible - currently they don't load when loaded as modules from the browser because of the `process.env.NODE_ENV` references.

Looks like this was solved in the core redux package by adding an additional rollup [config](https://github.com/reduxjs/redux/blob/b08cf38b7bbf2dcb3f4668493514ca1ecc121737/rollup.config.js#L75), would it be possible to have something similar for toolkit?

markerikson:
Sure, happy to accept a PR on that.

msutkowski:
I did some digging around regarding this, and it seems like there is no easy way to get `tsdx` to do this at this time. I can swap out the regular ESM build with this browser-specific version like redux does but I've been unable to get multiple variations of the ESM build to happen. Ideally, we could just pass in a new custom build format _(like 'mjs')_ and have it process the defined config for the format, but `tsdx` restricts the build to it's supported formats as it supplies defaults. There is some more context in this issue: https://github.com/formium/tsdx/issues/631, but it seems like the `patch-package` solution might be a little bit of a 🐰 🕳️. Maybe someone else can figure this out :)

markerikson:
So this is a bit confusing to me. We apparently merged #658 a year ago, and that should have been in the 1.5.0 and 1.5.1 releases. But, I don't see any `.esm.production` artifacts in either published package.

So, I'm going to see if we can actually fix this in #957 , now.

markerikson:
Should be resolved by #957 , and will be out in 1.6.

PR #658, opened 2020-07-19
Use 'system' format option for ES browser builds

msutkowski:
Addresses #652.

This is 'hacky', but is the only way I can figure out how to get this to work with the current `tsdx` limitations.

Uses the unused `system` build format and just overrides the output for ES modules in the browser ~~with the same config Redux uses. Confirmed that the build does work by converting the Redux UMD example to use `configureStore`/`createSlice`.~~

Update: I spent some more time reviewing the `tsdx` source, and the default configuration for any build is basically identical to the config Redux used, so I just simplified things. A working repo of the build from this last commit can be seen here: https://codesandbox.io/s/rtk-as-esm-browser-module-1lvrs?file=/index.html

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1202,6 +1202,87 @@
       "integrity": "sha512-518yewjSga1jLdiLrcmpMFlaba5P+50b0TWNFUpC+SL9Yzf0kMi57qw+bMl+rQ08cGqH1vLx4eg9YFUbZXgZ0Q==",
       "dev": true
     },
+    "@rollup/plugin-node-resolve": {
+      "version": "8.4.0",
+      "resolved": "https://registry.npmjs.org/@rollup/plugin-node-resolve/-/plugin-node-resolve-8.4.0.tgz",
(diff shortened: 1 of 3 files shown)


phryneas:
Just gave it a test, seems to be working well. Dependencies also seem to be bundled, at least `immer` seems to do it's job.

Some HTML to test locally, for the lazy ^^

```html
<html>
    <head>
        <script type="module">
            import {configureStore, createReducer } from './redux-toolkit.esm.development.mjs';

            const reducer = createReducer([], {
                test1(state, action) {
                    state.push(1)
                },
                test2(state, action) {
                    state.push(2)
                }
            }, 
            [], 
            (state, action) => (console.log(action), state))

                reducer
            })

            const s1 = store.getState()
            store.dispatch({type: "test1"})
            const s2 = store.getState()
            store.dispatch({type: "test2"})
            const s3 = store.getState()

            console.log({s1,s2,s3})
        </script>
    </head>
    <body>
    </body>
</html>
```

msutkowski:
If it's approved all around, I'll drop a quick reference to this build by the UMD docs

markerikson:
Do we want to get this out with 1.5?

phryneas:
Probably a good idea :)

PR #666, opened 2020-07-22
added missing type property to docs

pmonty:
Added `type` property to the three locations that were missing it.


markerikson:
Thanks, and congrats! :)

PR #669, opened 2020-07-23
Docs: add missing imports to the Redux-Persist usage guide

leventebalogh:
### The problem 
In the example for how to use it with `redux-persist` there were some missing imports from `@reduxjs/toolkit`.

**Link to the docs page:** https://redux-toolkit.js.org/usage/usage-guide#use-with-redux-persist

```javascript
// The missing imports:
import { configureStore, getDefaultMiddleware } from '@reduxjs/toolkit'
```

### The solition
Added the missing imports to the usage-guide docs.


markerikson:
Thanks!

PR #691, opened 2020-08-12
Export createStructuredSelector

t1gor:
Add `createStructuredSelector` to exported types as this allows not to install `reselect` manually

### DIFF
--- a/etc/redux-toolkit.api.md
+++ b/etc/redux-toolkit.api.md
@@ -9,6 +9,7 @@ import { ActionCreator } from 'redux';
 import { AnyAction } from 'redux';
 import { default as createNextState } from 'immer';
 import { createSelector } from 'reselect';
+import { createStructuredSelector } from 'reselect';
 import { current } from 'immer';
 import { DeepPartial } from 'redux';
 import { Dispatch } from 'redux';
@@ -169,6 +170,8 @@ export interface CreateSliceOptions<State = any, CR extends SliceCaseReducers<St
     name: Name;
     reducers: ValidateSliceCaseReducers<State, CR>;
 }
+
+export { createStructuredSelector }
 
 export { current }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,6 +3,7 @@ export * from 'redux'
 export { default as createNextState, Draft, current } from 'immer'
 export {
   createSelector,
+  createStructuredSelector,
   Selector,
   OutputParametricSelector,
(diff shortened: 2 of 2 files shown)


t1gor:
Guys, there seems to be a problem with latest TS version (on Travis - https://travis-ci.org/github/reduxjs/redux-toolkit/builds/717252824?utm_source=github_status&utm_medium=notification), but I don't think this is related to pull-request. Could you please take a look?

phryneas:
Yeah, the fix for that is already included in #637 - just ignore it for now, it's the upcoming TS RC anyways ;)

t1gor:
@phryneas ok :) could you please merge my PR then?

phryneas:
Not gonna blindly merge and currently at work ;)

Also, I'll wait what @markerikson has to say on this one.

markerikson:
Appreciate the PR, but I'm actually going to say no to this one for now, for a couple reasons.

We're not entirely trying to blindly re-export _everything_ from the libs we depend on.  Admittedly, we _do_ re-export everything form the Redux core, but that's because this is a wrapper around that core.  We re-export _some_ stuff from Immer.

I've never been keen on using `createStructuredSelector`, personally, for whatever reason.  With the migration from `connect/mapState` to `useSelector`, I think there's a lot fewer use cases where you might even want to use it.


Strictly speaking, you can probably get away with doing `import {createStructuredSelector} from "reselect"` even if you didn't list it, due to how JS package managers work these days.  But, it _is_ really easy to add your own dependency on `reselect` if necessary.

t1gor:
> We're not entirely trying to blindly re-export _everything_ from the libs we depend on. Admittedly, we _do_ re-export everything form the Redux core, but that's because this is a wrapper around that core. We re-export _some_ stuff from Immer.

This is clear and understandable.

> ... personally, for whatever reason. ...

Your words, not mine. It's a library you're creating, right? So use-cases might differ. We use `createStructuredSelector` when we need to map a bunch of smaller selectors to a large chunk of state. This just makes things easier to read and maintain afterwards. Just saying.

> Strictly speaking, you can probably get away with doing `import {createStructuredSelector} from "reselect"` even if you didn't list it, due to how JS package managers work these days. But, it _is_ really easy to add your own dependency on `reselect` if necessary.

Well, yes, but that would result in a hidden dependency. In case you would then decide to switch from `reselect` to smth. else we would be in trouble.

In any case - thanks for creating a great library 👍

PR #696, opened 2020-08-16
docs: createAsyncThunk: Handling Thunk Errors

phryneas:
Just a quick one as we really had nothing on error handling in there. Suggestions on anything that's missing very welcome :)


markerikson:
Started to make an edit as a suggestion, decided to rework it more :)  Will do so and push.

docs/api/createAsyncThunk.md
```diff
@@ -246,6 +246,43 @@ const onClick = () => {
...
+  name?: string
+  message?: string
+  stack?: string
+  code?: string
+}
+```
+
+will have been removed from it.
+
+If you want to pass out additional information or information in another shape, instead of `throw`ing it, you can use the `api.rejectWithValue` method instead.
```
msutkowski, line 266:
```suggestion
In the scenario that you're able to make use of a specific error shape with known properties in your reducers or components, you should use the `thunkAPI.rejectWithValue` method instead of `throw`ing an error.
```

msutkowski: Clarity & context suggestion. We also reference `thunkAPI` everywhere else, so we should at least keep that standard.

Additional text that may be useful at the end of this: "This is particularly common when expecting field-level validation errors from an API."

markerikson:
Just pushed an update.  How's this read now?  Here to the next section:

https://deploy-preview-696--redux-starter-kit-docs.netlify.app/api/createAsyncThunk#handling-thunk-results

msutkowski:
Looks good to me

PR #704, opened 2020-08-21
fix unwrapResult behaviour

### LINKED ISSUE #701: Can UnwrapResult prefer payload error?
In unwrapResult, we throw returned.error, which for custom handlers is always "Rejected"  Can we throw payload.error if it exists before throwing returned.error?  (Is there a way to overwrite returned.error that I didn't see in rejectWithValue?)

```javascript
export function unwrapResult<R extends ActionTypesWithOptionalErrorAction>(
  returned: R
): PayloadForActionTypesExcludingErrorActions<R> {
  if ('error' in returned) {
    throw returned.error
  }
  return (returned as any).payload
}
```

phryneas:
This fixes #701 while roughly keeping current behaviour.

@stridera, wanna take a look?

I'll highlight behavioural differences in the code.


phryneas:
I'm merging this into a new v1.5.0-integration branch now.

PR #714, opened 2020-08-31
Clarify docs on useDispatch origin

aethelz:
It might not be obvious where did useDispatch hook comes from, could be worth it to clarify it with import statement.


markerikson:
Thanks!

PR #724, opened 2020-09-12
Update Docusaurus metadata and styles

markerikson:


Shrugsy:
@markerikson I noticed that the 'Notes' cards on dark mode have some color issues, presumably as a result of these changes.  

e.g. just above this section: https://redux-toolkit.js.org/api/createSlice#the-builder-callback-api-for-extrareducers

2020-09-16 19_30_42-createSlice _ Redux Toolkit

PR #732, opened 2020-09-16
Fix Docusaurus style - Fix dark theme blockquote color contrast

Shrugsy:
Removes the black text color from dark theme blockquotes in order to let it fall back to `--ifm-font-base-color`.  

Issue mentioned previously [here](https://github.com/reduxjs/redux-toolkit/pull/724#issuecomment-693297620) on PR #724


markerikson:
Good enough for now, although we really ought to switch over to using the new "admonitions" instead of these blockquotes.

Issue #734, opened 2020-09-23
[Feature request] Allow to re-throw error from createAsyncThunk


mhienle:
In case an error occurs in the payloadCreator of createAsyncThunk, createAsyncThunk catches these errors and then dispatches rejected-actions. I understand that this is a sensible default behavior.

However, one may wish to have createAsyncThunk throw the error additionally to dispatching the rejected-action in some scenarios, e.g. to trigger side effects (like showing a toast, cf. https://github.com/reduxjs/redux-toolkit/issues/520) or to prevent further work from being done. Currently this is not possible or at least requires additional work for a thunk action using the thunk action creator returned by createAsyncThunk.

Therefore I suggest to add a (global and/or local) configuration option to rethrow the error in case an error occurred inside the payload creator.

leventebalogh:
I think it would be useful as well, so created a PR for it:
https://github.com/reduxjs/redux-toolkit/pull/736

**For the time being you can do something similar to detect rejections on the call-site:**
```javascript
const finalAction = await dispatch(asyncThunk(args));
const isRejected = finalAction.type.endsWith('/rejected');

if (isRejected) {
  // ...
}
```

markerikson:
The `isRejected` check isn't needed.  That's the purpose of [the `unwrapResult` utility](https://redux-toolkit.js.org/api/createAsyncThunk#unwrapping-result-actions), which will convert the result action back into either the successful value or the thrown error.

Given that `unwrapResult` exists, I'm not really sure this change is needed.

leventebalogh:
Thanks for the info @markerikson, I completely missed that! 🙏 

Yea, in that case I am not sure it's needed either. Probably would make it a bit easier to handle these, but `unwrapResult` can do the job as well actually. Do you suggest closing the PR?

leventebalogh:
What we could do IMHO is updating the `Return Value` part of the `createAsyncThunk()` docs to raise a bit more attention on this feature.

markerikson:
Yeah, I definitely appreciate you taking the initiative to file a PR, but I'd say that `unwrapResult` covers the use case sufficiently for now.

Would definitely appreciate any docs update PRs you feel are appropriate to make this more clear.

leventebalogh:
No worries mate, it happens. At least I was digging through half of the codebase (the wrong half) and now I know it better, haha.

markerikson:
[...]

PR #736, opened 2020-09-23
Feature: allow re-throwing errors from createAsyncThunk()

leventebalogh:
**Related issue:** https://github.com/reduxjs/redux-toolkit/issues/734

This would allow re-throwing the original error, and would make it possible to act on the promise returned by the action.

This PR doesn't change the default way how `createAsyncThunk()` works, so is backward compatible.

```javascript
const asyncThunk = createAsyncThunk('test', () => {
    throw new Error('sample error');
}, {
    throwError: true
});

// ...

try {
    await dispatch(asyncThunk())
} catch (e) {
    // This will throw an error now that can be acted on in some exceptional scenarios
    // e = Error('sample error')
}
```


leventebalogh:
Closed because of: https://github.com/reduxjs/redux-toolkit/issues/734#issuecomment-697605719

Issue #751, opened 2020-10-06
TypeScript types for `serialize` property of `EnhancerOptions` seem wrong


blaiz:
Hello,

According to https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#serialize the `serialize` property can be assigned an object with the following properties: `options`, `replacer`, `reviver`, `immutable` and `refs`.

However, at https://github.com/reduxjs/redux-toolkit/blob/a1e8becbb2e4653d721bdc1d2363a45562c526a2/src/devtoolsExtension.ts#L39 the type definition for `serialize` is defined as follows:
```ts
  serialize?:
    | boolean
    | {
        date?: boolean
        regex?: boolean
        undefined?: boolean
        error?: boolean
        symbol?: boolean
        map?: boolean
        set?: boolean
        function?: boolean | Function
      }
```
which looks to me like it's skipping one level and it's defining the types for the `options` properties directly inside `serialize`. Seems like it should look more like:
```ts
  serialize?:
    | boolean
    | {
        options: | boolean
      | {
          date?: boolean
          regex?: boolean
          undefined?: boolean
          error?: boolean
          symbol?: boolean
          map?: boolean
          set?: boolean
          function?: boolean | Function
        },
      replacer: () => {}, // with some proper types here
      reviver: () => {} // etc. with the other properties listed above
      }
```

Thoughts?
If you agree, I can try and do some research on what the type definition for these missing properties would be and whip up a PR perhaps.

Thanks

markerikson:
Hmm.  Yeah, I copy-pasted some typedefs over, and it's entirely possible I screwed that up somewhere :)  If you could look into that, I'd appreciate it.  Thanks!

blaiz:
Sounds good. I found some types at https://www.github.com/reduxjs/redux-devtools/tree/master/packages%2Fredux-devtools-serialize%2Fsrc%2Fimmutable%2Fserialize.ts

I'll take a closer look tomorrow.

markerikson:
Fixed by #752 .

PR #752, opened 2020-10-06
fix(typescript): fix `serialize` types

blaiz:
Fix for #751

The `serialize` property can be an object which can have many different properties. Those properties are described at https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#serialize The current type only mentions what is inside the `options` property, and puts it directly inside `serialize`, which doesn't match the docs so I'm updating the types so they match the docs more closely, as well as the source code at https://github.com/reduxjs/redux-devtools/blob/master/packages/redux-devtools-serialize/src/immutable/index.ts

I tested this locally in my own project by linking the npm package and initializing the store with:
```ts
  reducer: webReducers,
  middleware: [thunk, sagaMiddleware],
  devTools: {
    actionSanitizer,
    stateSanitizer,
    serialize: {
      replacer: (_key, value) => {
        if (value instanceof Element) {
          return "DOMElement";
        }
        if (
          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
          // @ts-ignore
          window.google?.maps &&
          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
          // @ts-ignore
          value instanceof window.google.maps?.MVCObject
        ) {
          return "GoogleMaps";
        }
        return value;
      },
    },
  },
});
```

I used to have a TS error for `replacer` and I don't anymore, proving that this fixes the issue.

### DIFF
--- a/src/devtoolsExtension.ts
+++ b/src/devtoolsExtension.ts
@@ -28,25 +28,27 @@ export interface EnhancerOptions {
    */
   maxAge?: number
   /**
-   * - `undefined` - will use regular `JSON.stringify` to send data (it's the fast mode).
-   * - `false` - will handle also circular references.
-   * - `true` - will handle also date, regex, undefined, error objects, symbols, maps, sets and functions.
-   * - object, which contains `date`, `regex`, `undefined`, `error`, `symbol`, `map`, `set` and `function` keys.
-   *   For each of them you can indicate if to include (by setting as `true`).
-   *   For `function` key you can also specify a custom function which handles serialization.
-   *   See [`jsan`](https://github.com/kolodny/jsan) for more details.
+   * See detailed documentation at http://extension.remotedev.io/docs/API/Arguments.html#serialize
    */
   serialize?:
     | boolean
     | {
-        date?: boolean
-        regex?: boolean
-        undefined?: boolean
-        error?: boolean
-        symbol?: boolean
-        map?: boolean
-        set?: boolean
-        function?: boolean | Function
+        options?:
+          | boolean
+          | {
+              date?: boolean
+              regex?: boolean
+              undefined?: boolean
+              error?: boolean
+              symbol?: boolean
+              map?: boolean
+              set?: boolean
+              function?: boolean | Function
+            }
+        replacer?: (key: string, value: unknown) => unknown
+        reviver?: (key: string, value: unknown) => unknown
+        immutable?: unknown
+        refs?: unknown[]
       }
   /**
    * function which takes `action` object and id number as arguments, and should return `action` object back.


package.json
```diff
@@ -33,6 +33,7 @@
     "axios": "^0.19.2",
     "console-testing-library": "^0.3.1",
     "eslint-config-react-app": "^5.0.1",
+    "immutable": "^4.0.0-rc.12",
```
blaiz, line 36: I pulled in a new dev dependency to use its type. Let me know if you'd rather not add one more dependency and maybe we can just use a type of `any` or `unknown` above or something, although it'll be less precise.

src/devtoolsExtension.ts
```diff
@@ -28,25 +29,27 @@ export interface EnhancerOptions {
...
   maxAge?: number
   /**
-   * - `undefined` - will use regular `JSON.stringify` to send data (it's the fast mode).
-   * - `false` - will handle also circular references.
-   * - `true` - will handle also date, regex, undefined, error objects, symbols, maps, sets and functions.
-   * - object, which contains `date`, `regex`, `undefined`, `error`, `symbol`, `map`, `set` and `function` keys.
-   *   For each of them you can indicate if to include (by setting as `true`).
-   *   For `function` key you can also specify a custom function which handles serialization.
-   *   See [`jsan`](https://github.com/kolodny/jsan) for more details.
+   * See detailed documentation at https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#serialize
```
blaiz, line 32: The docs are so long that I felt like it wasn't worth copy/pasting the entire thing into this jsdoc and felt that it'd be more useful to just provide a link to the original docs

src/devtoolsExtension.ts
```diff
@@ -28,25 +29,27 @@ export interface EnhancerOptions {
...
   maxAge?: number
   /**
-   * - `undefined` - will use regular `JSON.stringify` to send data (it's the fast mode).
-   * - `false` - will handle also circular references.
-   * - `true` - will handle also date, regex, undefined, error objects, symbols, maps, sets and functions.
-   * - object, which contains `date`, `regex`, `undefined`, `error`, `symbol`, `map`, `set` and `function` keys.
-   *   For each of them you can indicate if to include (by setting as `true`).
-   *   For `function` key you can also specify a custom function which handles serialization.
-   *   See [`jsan`](https://github.com/kolodny/jsan) for more details.
+   * See detailed documentation at http://extension.remotedev.io/docs/API/Arguments.html#serialize
```
blaiz, line 32: The docs are so long that I felt like it wasn't worth copy/pasting the entire thing into this jsdoc and felt that it'd be more useful to just provide a link to the original docs

phryneas: We should do that the other way around. 
We can include Jsdoc comments directly into the documentation using "links" like `
[summary,remarks](docblock://mapBuilders.ts?token=ActionReducerMapBuilder.addMatcher)` in the documentation markdown files.
That way they will stay in sync - but they will also be fully available in autocomplete popups without an extra link.

package.json
phryneas: Having this just as a dev dependency will make bundlers like yarn2 with pnp complain that a dependency is missing. Correct would probably be an optional peer dependency.

Also, this might be just imported as a type directly by using 
```diff
-immutable?: typeof Immutable
+immutable?: import('immutable')
```
down there.
That way it will never be a runtime require - which might prevent tree-shaking.

blaiz: Sounds good. I'll add it as an optional dependency (couldn't find anything specific to optional *dev* dependency).

Also `immutable?: import('immutable')` did not work so I changed it to `immutable?: typeof import('immutable')`, otherwise TS complains:
Screen Shot 2020-10-06 at 1 18 02 PM

src/devtoolsExtension.ts
blaiz: I'm confused. We do not own this documentation. The `serialize` property is just a passthrough to the Redux Devtools extension, so they own the docs, hence why I put a link to their docs here.

package.json
phryneas: Optional *peer* dependency - it's not a dev dependency, because RTK's dev dependencies are only installed when developing against RTK. But you want it to point against the right package if RTK and immutable are installed while *using* RTK.

So it should be a "real" dependency, but the loosest form of that.
At least that's how I understood the usage of that - @markerikson might know more though.

See https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md - it's fairly new but should be supported by the bigger bundlers by now.

src/devtoolsExtension.ts
phryneas: Ah, good point. I had somehow assumed that we had our own docs on that like some other re-exports we do.

blaiz: Oh interesting. Yeah the docs here just point at the docs there. From https://redux-toolkit.js.org/api/configureStore#devtools it goes to https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#windowdevtoolsextensionconfig (which is the link I originally used in this comment but they have an official website, so you could also update this link in the docs).

package.json
blaiz: Oh interesting. I had never heard of that. I'll take a look.

blaiz: Now that it's a peer dependency, TS complains because it's not installed locally. Do you know how to fix that part properly?
Screen Shot 2020-10-06 at 2 35 25 PM

blaiz:
Anything new here @phryneas ?

phryneas:
Sorry, I'm swamped right now. I'll try to take another look at this tomorrow or the day after.

blaiz:
Sounds good. Understandable. Thank you!

package.json
```diff
@@ -43,6 +43,14 @@
...
     "typings-tester": "^0.3.2"
   },
+  "peerDependencies": {
+    "immutable": "^4.0.0"
+  },
+  "peerDependenciesMeta": {
+    "immutable": {
+      "optional": true
+    }
+  },
```
phryneas, line 53:
```suggestion
```
Let's just leave this out. As we are generally encouraging against using immutablejs, we should probably not mention it here.

src/devtoolsExtension.ts
```diff
@@ -28,25 +28,27 @@ export interface EnhancerOptions {
...
+              undefined?: boolean
+              error?: boolean
+              symbol?: boolean
+              map?: boolean
+              set?: boolean
+              function?: boolean | Function
+            }
+        replacer?: (key: string, value: unknown) => unknown
+        reviver?: (key: string, value: unknown) => unknown
+        immutable?: typeof import('immutable')
```
phryneas, line 50:
```suggestion
        immutable?: unknown
```
same reasoning as above: someone who wants to use this will know what to do, but we won't encourage it

phryneas (changes requested):
This seems to be in-line with https://github.com/zalmoxisus/redux-devtools-extension/pull/731. As for the `immutable` reference. Thinking about this another time, it's probably better if we just `unknown` that. If someone is using immutable, they'll know what to do, bu given than we advise to use immer instead of immutable, it will probably just confuse people if we add it as a dependency, even if only as an optional peer dependency.

Apart from that, it looks solid to me. @markerikson  what do you think?

package.json
blaiz: Yeah that makes sense. I definitely wasn't sure whether bringing in this dependency just for one type was a good idea or not. `unknown` is fine.

blaiz:
@phryneas my apologies, I know you're busy but I feel like this is small and very low risk since it only touches types. I've made the changes we talked about. Does it look good now?

phryneas:
Yup, I'm sorry. A lot on my mind. Let's get this merged.

PR #760, opened 2020-10-11
fix types of DevTools EnhancerOptions.serialize

dartess:
docs:
serialize: boolean or object which contains: options, replacer, reviver, ...
https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#serialize

types:
serialize: boolean or options.

seems like types is wrong


dartess:
double 752, sorry

PR #768, opened 2020-10-21
Use main reactiflux invite

vcarl:
The old invite was putting joiners into an error state due to the Code of Conduct gate


markerikson:
Done.  Other two Redux docs sites too?

vcarl:
I opened the above, but am not sure about the other spot you mentioned, mind pointing me at it?

PR #784, opened 2020-10-31
add `requestId` property to the promise returned by `createAsyncThunk`

phryneas:
I noticed this to be lacking while experimenting with https://github.com/phryneas/rtk_simple-query - @markerikson is this okay with you?


phryneas:
Hmm. I'd even go one step further and also add `arg` as a property of that promise. It's available and it's actually useful. Updating the PR.

markerikson:
Sure, seems like it might be useful.

PR #795, opened 2020-11-07
Export isPlainObject

msutkowski:
Just exports `isPlainObject` to consume with other RTK-based libs. I don't think this needs to be documented under `Other exports` in the docs, but I can add it if necessary.


phryneas:
@markerikson  Given that you assumed we were already exporting that one, I guess this one is fine with you? ;)

PR #804, opened 2020-11-15
Exposed Immer's PatchListener

Sun-2:
Recently I've been experimenting more with the Toolkit, including using it on the server, syncing state between two stores via WebSockets, and writing a debugging tool with easy diffing due to Immer patches.

I must say at the moment I'm absolutely not sure how useful other people would find it, but would you, by any chance, consider exposing it via RTK's `createSlice` and `createReducer` API?

### DIFF
--- a/src/createReducer.ts
+++ b/src/createReducer.ts
@@ -1,4 +1,9 @@
-import createNextState, { Draft, isDraft, isDraftable } from 'immer'
+import createNextState, {
+  Draft,
+  isDraft,
+  isDraftable,
+  PatchListener
+} from 'immer'
 import { AnyAction, Action, Reducer } from 'redux'
 import {
   executeReducerBuilderCallback,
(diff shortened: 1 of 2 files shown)


markerikson:
Mmm... my first question is, what _is_ this thing, and what's the actual use case for it?

I'm very hesitant to add any new APIs we have to support indefinitely without good reason to do so.  This isn't a hard "no", but I'd need a lot more info on potential use cases and how many users this would benefit in practice.

Sun-2:
It's an array of steps that tells how Immer modifies the state tree.

Example:
```
[
    {"op": "replace", "path": ["profile"], "value": {"name": "Noa", "age": 6}},
    {"op": "add", "path": ["tags", 3], "value": "kiddo"}
]
```

[Link to Immer docs about patches](https://immerjs.github.io/immer/docs/patches)

> Remember that one of the main goals of RTK is to simplify the "80% sweet spot" of common Redux use cases, and I'm not sure this sufficiently qualifies.

Then it certainly doesn't quality, so I'll just stop here. I'll just use my forked version where I need it, I guess.

PR #806, opened 2020-11-16
Documentation: Matching Utilities - isAnyOf, isAllOf

douglas-treadwell:
This is a documentation only PR, for the `matchers` feature branch.

This adds a `Matchers` section under `API Reference` in the sidebar, with `isAllOf` and `isAnyOf` for now.

The documentation shows some simple examples.  More complex use cases may be specific to the user, but if there are some common ones I'd be happy to add them.

Let me know what you think.  I'll be happy to make improvements.


markerikson:
My first thought is that these are similar enough we really only need one docs page.  Maybe call it "Matching Utilities" or something along those lines?  After all, the only real difference between these functions is the boolean operation being performed.

douglas-treadwell:
@markerikson That sounds good.

What would you suggest after this expands to include more functions like `isAsyncThunkAction`, `isFulfilled`, `isRejected`, `isPending` and `isRejectedWithValue`?

Perhaps one item under `Matchers` called `isAnyOf / isAllOf` (or alternatively, `Simple Matchers`), and another called `Async Matchers`?

douglas-treadwell:
Do you have any thoughts about the content of the documentation?  Are the examples sufficient?

docs/api/isAllOf.mdx
```diff
@@ -0,0 +1,180 @@
...
+          ...state,
+          isInteresting: true,
+        }
+      }
+    });
+});
+```
+
+With isAllOf, we can use matchers to reorganize that code into a more canonical form.
+And more importantly, it enables more complex cases to be handled straightforwardly.
```
msutkowski, line 84:
```suggestion
With `isAllOf`, we can use matchers to drastically simplify the `loadingReducer`.
```

docs/api/isAllOf.mdx
```diff
@@ -0,0 +1,180 @@
...
+  isSpecial: false,
+  isInteresting: false,
+}
+
+const request = createAsyncThunk('request', () => {
+  return {
+    isSpecial: true,
+    isInteresting: true,
+  };
+});
```
msutkowski, line 125:
```suggestion
// ... omitted
```

docs/api/isAllOf.mdx
```diff
@@ -0,0 +1,180 @@
...
+
+const loadingReducer = createReducer(initialState, builder => {
+  builder
+    .addMatcher(isAllOf(request.fulfilled, isSpecial), (state, action) => {
+      return {
+        ...state,
+        isSpecial: true,
+      }
```
msutkowski, line 133:
```suggestion
      state.isSpecial = true;
```

docs/api/isAllOf.mdx
```diff
@@ -0,0 +1,180 @@
...
+    })
+    .addMatcher(isAllOf(request.fulfilled, isInteresting), (state, action) => {
+      return {
+        ...state,
+        isInteresting: true,
+      }
```
msutkowski, line 139:
```suggestion
      state.isInteresting = true;
```

docs/api/isAllOf.mdx
```diff
@@ -0,0 +1,180 @@
...
+    });
+})
+```
+
+## As a TypeScript TypeGuard
```
msutkowski, line 144:
```suggestion
## As a TypeScript Type Guard
```

docs/api/isAnyOf.mdx
```diff
@@ -0,0 +1,111 @@
...
+    .addCase(request.pending, (state, action) => ({
+      isLoading: true
+    }))
+    .addMatcher(requestFinished, (state, action) => ({
+      isLoading: false
+    }))
+})
+```
+
+## As a TypeScript TypeGuard
```
msutkowski, line 75:
```suggestion
## As a TypeScript Type Guard
```

docs/api/isAllOf.mdx
```diff
@@ -0,0 +1,180 @@
...
+  isInteresting: true;
+}
+
+function isSpecial(action: PayloadAction<Data>): action is PayloadAction<Special> {
+  return action.payload.isSpecial;
+}
+
+function isInteresting(action: PayloadAction<Data>): action is PayloadAction<Interesting> {
+  return action.payload.isInteresting;
+}
```
msutkowski, line 171: I think once you merge these into one page, there will be no need to show these same types & type guards this many times. I feel like the bloat takes away from where the focus should be below.

docs/api/isAnyOf.mdx
```diff
@@ -0,0 +1,111 @@
...
+    .addCase(request.pending, (state, action) => ({
+      isLoading: true
+    }))
+    .addCase(request.fulfilled, (state, action) => ({
+      isLoading: false
+    }))
+    .addCase(request.rejected, (state, action) => ({
+      isLoading: false
+    }))
+})
```
msutkowski, line 46: I would consider keeping your `return` uniform throughout the examples - some folks are definitely going to get tripped up between implicit and explicit returns as well as mutable vs immutable updates. Being that we push `immer` so much, I'd recommend we just show:

```js
(state, action) => {
   state.isLoading = false;
}
```

msutkowski (commented):
Just some small feedback. As always, I'd wait on Mark to make any changes :)

douglas-treadwell:
> Just some small feedback. As always, I'd wait on Mark to make any changes :)

Thanks for the review @msutkowski!

docs/api/isAnyOf.mdx
douglas-treadwell: Good idea.  Thanks!

docs/api/matching-utilities.mdx
```diff
@@ -7,11 +7,21 @@ hide_title: true
 
 # Matching Utilities
 
-Redux Toolkit exports a couple of type-safe utility matchers that you can leverage to simplify your code.
+Redux Toolkit several type-safe utility matchers that you can leverage to simplify your code.
```
phryneas, line 10:
```suggestion
Redux Toolkit exports several type-safe utility matchers that you can leverage to simplify your code.
```

docs/api/matching-utilities.mdx
```diff
@@ -7,11 +7,21 @@ hide_title: true
...
+
+### General Purpose
 
 - [`isAllOf`](#isallof) - returns true when **all** conditions are met
 - [`isAnyOf`](#isanyof) - returns true when at least one of the conditions are met
 
+### `createAsyncThunk`-specific matchers
```
phryneas, line 17:
```suggestion
### `createAsyncThunk`-specific matchers

All these matchers can either be called with one or more thunks as arguments, in which case they will return a matcher function for that condition and thunks, or with one actions, in which case they will match for any thunk action with said condition.
```

docs/api/matching-utilities.mdx
```diff
@@ -29,6 +39,67 @@ It will return a type guard function that returns `true` if _all_ of the provide
...
+A higher-order function that returns a type guard function that may be used to check whether an action was created by [`createAsyncThunk`](./createAsyncThunk).
+
+```ts title="isAsyncThunkAction usage"
+import { isAsyncThunkAction } from '@reduxjs/toolkit'
+import { request, request2 } from '@virtual/matchers'
+
+if (isAsyncThunkAction(request, request2)) {
+  // These are all async thunk actions, handle the case.
+}
+```
```
phryneas, line 53:
```suggestion
```ts title="isAsyncThunkAction usage"
import { isAsyncThunkAction } from '@reduxjs/toolkit'
import { requestThunk1, requestThunk2 } from '@virtual/matchers'

const isARequestAction = isAsyncThunkAction(requestThunk1, requestThunk2);

function handleRequestAction(action: AnyAction) {
  if (isARequestAction(action)) {
    // action is either an action dispatched by `requestThunk1` or `requestThunk2`
  }
}
```
```

Comment (phryneas) [reply docs/api/matching-utilities.mdx:53]: the same pattern would apply to the others.

**Comment by phryneas** [comment]
Other usage examples would be
```js
if (isRejectedAction(action)) {
  // ...
}
```

and
```js
extraReducers: builder => {
  builder.addMatcher(isFulfilledAction(requestThunk1, requestThunk2), action => {
    // ...
  })
}
```

docs/api/matching-utilities.mdx
```diff
@@ -0,0 +1,229 @@
+---
+id: matching-utilities
+title: Matching Utilities
+sidebar_label: Matching Utilities
+hide_title: true
+---
+
+# Matching Utilities
+
+Redux Toolkit exports several type-safe utility matchers that you can leverage to simplify your code.
```
markerikson, line 10:
```suggestion
Redux Toolkit exports several type-safe action matching utilities that you can leverage when checking for specific kinds of actions.  These are primarily useful for the `builder.addMatcher()` cases in `createSlice` and `createReducer`, as well as when writing custom middleware.
```

docs/api/matching-utilities.mdx
```diff
@@ -0,0 +1,229 @@
...
+
+### General Purpose
+
+- [`isAllOf`](#isallof) - returns true when **all** conditions are met
+- [`isAnyOf`](#isanyof) - returns true when at least one of the conditions are met
```
markerikson, line 15:
```suggestion
- [`isAnyOf`](#isanyof) - returns true when **at least one of** the conditions are met
```

markerikson (changes requested):

docs/api/matching-utilities.mdx
```diff
@@ -0,0 +1,257 @@
...
+    }
+  }
+)
+
+// This has unnecessary complexity
+const loadingReducer = createReducer(initialState, builder => {
+  builder.addCase(isSpecialAndInterestingThunk.fulfilled, (state, action) => {
+    if (isSpecial(action)) {
+      state.isSpecial = true
+    } else if (isInteresting(action)) {
```
phryneas, line 197: Due to the `else` here, both examples actually behave differently.
In the matcher case, both execute.

msutkowski: Updated. I just copy and pasted that from the original and didn't bother to check 🤦

PR #807, opened 2020-11-16
AsyncThunk Matchers

douglas-treadwell:
AsyncThunk-related functionality: `isAsyncThunkAction`, `isFulfilled`, `isRejected`, `isPending`.

Does not yet include `isRejectedWithValue`.  Also, `isFulfilled`, `isRejected`, `isPending` do not yet accept async thunk parameters to narrow the resulting matcher, but `isAsyncThunkAction` does.

This is a Draft PR because I still need to add the functionality above and type tests.

Feedback is appreciated!

### DIFF
--- a/src/matchers.test.ts
+++ b/src/matchers.test.ts
@@ -142,7 +142,7 @@ describe('isAllOf', () => {
 
 describe('isPending', () => {
   test('should return false for a regular action', () => {
-    const action = createAction<string>('a')
+    const action = createAction<string>('action/type')('testPayload');
 
     expect(isPending(action)).toBe(false)
   })
@@ -166,7 +166,7 @@ describe('isPending', () => {
 
 describe('isRejected', () => {
   test('should return false for a regular action', () => {
-    const action = createAction<string>('a')
+    const action = createAction<string>('action/type')('testPayload');
 
     expect(isRejected(action)).toBe(false)
   })
@@ -190,7 +190,7 @@ describe('isRejected', () => {
 
 describe('isFulfilled', () => {
   test('should return false for a regular action', () => {
-    const action = createAction<string>('a')
+    const action = createAction<string>('action/type')('testPayload');
 
     expect(isFulfilled(action)).toBe(false)
   })
@@ -214,7 +214,7 @@ describe('isFulfilled', () => {
 
 describe('isAnyAsyncThunkAction', () => {
   test('should return false for a regular action', () => {
-    const action = createAction<string>('a')
+    const action = createAction<string>('action/type')('testPayload');
 
     expect(isAnyAsyncThunkAction(action)).toBe(false)
   })
--- a/src/matchers.test.ts
+++ b/src/matchers.test.ts
@@ -212,32 +212,14 @@ describe('isFulfilled', () => {
   })
 })
 
-describe('isAnyAsyncThunkAction', () => {
+describe('isAsyncThunkAction', () => {
   test('should return false for a regular action', () => {
     const action = createAction<string>('action/type')('testPayload');
 
-    expect(isAnyAsyncThunkAction(action)).toBe(false)
+    expect(isAsyncThunkAction()(action)).toBe(false)
   })
 
-  test('should return true for any async thunk action', () => {
-    const thunk = createAsyncThunk<string>('a', () => 'result')
-
-    const pendingAction = thunk.pending('fakeRequestId')
-    expect(isAnyAsyncThunkAction(pendingAction)).toBe(true)
-
-    const rejectedAction = thunk.rejected(
-      new Error('rejected'),
-      'fakeRequestId'
-    )
-    expect(isAnyAsyncThunkAction(rejectedAction)).toBe(true)
-
-    const fulfilledAction = thunk.fulfilled('result', 'fakeRequestId')
-    expect(isAnyAsyncThunkAction(fulfilledAction)).toBe(true)
-  })
-})
-
-describe('isAsyncThunkAction', () => {
-  test('should return true for any async thunk if no arguments were provided', () => {
+  test('should return true for any async thunk action if no arguments were provided', () => {
     const thunk = createAsyncThunk<string>('a', () => 'result')
     const matcher = isAsyncThunkAction()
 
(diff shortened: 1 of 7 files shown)


douglas-treadwell:
@phryneas I changed the name of the proposed `asyncThunkType` to `requestStatus` to match `requestId`.  (Because `requestId` could be thought of as a corresponding `asyncThunkId`.)

src/matchers.test.ts
```diff
@@ -131,3 +139,147 @@ describe('isAllOf', () => {
     ).toBe(false)
   })
 })
+
+describe('isPending', () => {
+  test('should return false for a regular action', () => {
+    const action = createAction<string>('a')
```
phryneas, line 145:
```suggestion
    const action = createAction<string>('action/type')('testPayload')
```
Otherwise you test the action creator, not the action.

src/matchers.test.ts
```diff
@@ -131,3 +139,147 @@ describe('isAllOf', () => {
...
+        new Error('rejected'),
+        'fakeRequestId'
+      )
+      expect(matchAC(rejectedAction)).toBe(expected)
+
+      const fulfilledAction = thunk.fulfilled('result', 'fakeRequestId')
+      expect(matchAC(fulfilledAction)).toBe(expected)
+    }
+
+    testAllActions(thunkA, true)
```
phryneas, line 281: This *could* also be done with test.each
```
test.each([
  [thunkA, true],
  [thunkB, true],
  [thunkC, false]
])('test %s for result %s', (thunk, expectedResult) => {
  // ...
})
```
but that's only a suggestion, this is fine by me too.

src/matchers.ts
```diff
@@ -56,3 +62,146 @@ export function isAllOf<Matchers extends [Matcher<any>, ...Matcher<any>[]]>(
...
+  | UnknownAsyncThunkFulfilledAction
+
+/**
+ * A type guard function that matches any async thunk action.
+ *
+ * @param action
+ *
+ * @internal
+ */
+export function isAnyAsyncThunkAction(
```
phryneas, line 145: I think this could be part of `isAsyncThunkAction`, when `isAsyncThunkAction` is called with only one argument that is not of type function.

src/matchers.ts
```diff
@@ -56,3 +62,146 @@ export function isAllOf<Matchers extends [Matcher<any>, ...Matcher<any>[]]>(
...
+    for (const asyncThunk of restAsyncThunks) {
+      matchers.push(
+        asyncThunk.pending,
+        asyncThunk.rejected,
+        asyncThunk.fulfilled
+      )
+    }
+
+    const combinedMatcher = isAllOf(isAnyAsyncThunkAction, isAnyOf(...matchers))
+    return combinedMatcher(action)
```
phryneas, line 205: This seems a bit long. I *think* this should do it as well, right?

```suggestion
    const matchers: [Matcher<any>, ...Matcher<any>[]] = []

    for (const asyncThunk of asyncThunks) {
      matchers.push(
        asyncThunk.pending,
        asyncThunk.rejected,
        asyncThunk.fulfilled
      )
    }

    const combinedMatcher = isAnyOf(...matchers)
    return combinedMatcher(action)
```

phryneas (changes requested):
That looks very promising!

I've left a few remarks, but nothing major :)

src/matchers.ts
douglas-treadwell: @phryneas The problem was that `const matchers: [Matcher<any>, ...Matcher<any>[]] = []` was failing type checking because [] does not have at least one element.  Specifically, `"Property '0' is missing in type [] but required in ...`.

And if I change the annotation to `const matchers: Matcher<any>[]`, then `isAnyOf(...matchers)` fails type checking because at least one argument is required.

I'm also not completely satisfied with how the code looks.  It's organized this way to pass type checking without using any casts.  Maybe there's a more straightforward way?

phryneas: Honestly, I'm totally fine with a
`const matchers: [Matcher<any>, ...Matcher<any>[]] = [] as unknown`
there. After all, we know better than the compiler.

src/matchers.ts
douglas-treadwell: I don't mind making the change, but I wonder if it's more efficient to reuse the same function?  Also, pulling it out and giving it a name may make the code easier for a reader to understand.  What do you think?

douglas-treadwell: It's actually not too bad if I move `isAnyAsyncThunkAction` inside `isAsyncThunkAction`.

```ts
  if (asyncThunks.length === 0) {
    return (action: any) => hasExpectedRequestMetadata(action, [
      'pending',
      'fulfilled',
      'rejected'
    ])
  }
```

I'll include that change shortly.

douglas-treadwell:
As of the latest commit, this PR includes:

AsyncThunk-related functionality: `isAsyncThunkAction`, `isFulfilled`, `isRejected`, `isPending` as HOCs.  Tests and type-tests are included, although I just noticed some bugs in my recently added tests.

I'm working on `isRejectedWithValue` and fixing the tests for the above HOCs.

douglas-treadwell:
I've also added `isRejectedWithValue`.  Ready for re-review when you have time.

src/matchers.test.ts
douglas-treadwell: Fixed.

phryneas:
I'll take a look after work. 

In general: do you think we can get this finished this weekend?
I'm happy to chime in & do everything that you feel you won't get to, just tell me.

It would just be great to still get this into v1.5 - and get v1.5 out at the end of the weekend, if possible :)

douglas-treadwell:
> I'll take a look after work.
> 
> In general: do you think we can get this finished this weekend?
> I'm happy to chime in & do everything that you feel you won't get to, just tell me.
> 
> It would just be great to still get this into v1.5 - and get v1.5 out at the end of the weekend, if possible :)

After this PR, I think the only missing piece is the documentation.

I'm available for at least one day to work on this over the weekend.  That's likely enough time to attempt a fix of anything you'd like improved with this PR, and make an attempt at documentation for all the recently added functions.

However, if you know what you want the documentation to say, or if you find it easier to fix some areas of concern in the PR directly rather than communicate about it, I certainly won't be offended.  On the contrary, I wouldn't want progress to be delayed and would be happy for any assistance getting this finished over the weekend.

phryneas:
This looks very good. Thank you a lot!

We'll take it from here - right now we have a chat of three super-motivated maintainers going on who want to get this out and are already splitting up responsibilities ;)

PR #811, opened 2020-11-18
move "development" TS version to TS3.9, use `@ts-expect-error` in CI whenever possible

phryneas:
let's see how this works out


phryneas:
This seems to work well and shaves 30s test time of all TS versions >= 3.9

PR #812, opened 2020-11-18
createAsyncThunk: serializeError option

phryneas:
This *would* address two requests from #792:

1. adding a `serializeError` option to `createAsyncThunk`
2. the option to call `await dispatch(async(3, { unwrap: true }))` and return it with the unwrapped result and otherwise throw an error. So, essentially an inlined version of `unwrapResult`.

I say *would*  because I'm still not a fan of Nr 2. Even less, after writing all this, tbh.
The amount of code (and especially types) required to support a second way of doing something we already have in the lib makes me uneasy.
Yes, it might be slightly "nicer". But only slightly. And it's a lot of code. And a lot of new behaviour (especially type-wise) that we have to commit to keep running.

So my personal suggestion would be to keep 1. (even though it also does something comparable to `rejectWithValue` as well)  and not merge 2 (no matter the amount of work I put into this).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -69,6 +69,12 @@ jobs:
         run: |
           sed -i -e 's|@reduxjs/toolkit": \["./src"\]|@reduxjs/toolkit": ["."]|' ./type-tests/files/tsconfig.json
 
+      - name: Prefix `freeze` re-export for pre-3.7 TS versions with @ts-ignore
+        if: ${{ matrix.ts < 3.7 }}
+        run: |
+          sed -i -e "/import .* freeze .* from 'immer'/s/^/\/\/ @ts-ignore\n/" dist/typings.d.ts
+          sed -i -e "/export .* freeze .* from 'immer'/s/^/\/\/ @ts-ignore\n/" src/index.ts
+
       - name: Use typings-tester for old TS versions
         if: ${{ matrix.ts < 3.9 }}
         run: |
--- a/src/createAsyncThunk.test.ts
+++ b/src/createAsyncThunk.test.ts
@@ -597,6 +597,37 @@ describe('conditional skipping of asyncThunks', () => {
       })
     )
   })
+
+  test('serializeError implementation', async () => {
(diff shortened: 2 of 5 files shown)


phryneas:
Also pinging @nathggns on that one. If you really see any value in this, try to convince me :)

.github/workflows/tests.yml
```diff
@@ -69,6 +69,12 @@ jobs:
         run: |
           sed -i -e 's|@reduxjs/toolkit": \["./src"\]|@reduxjs/toolkit": ["."]|' ./type-tests/files/tsconfig.json
 
+      - name: Prefix `freeze` re-export for pre-3.7 TS versions with @ts-ignore
+        if: ${{ matrix.ts < 3.7 }}
+        run: |
+          sed -i -e "/export .* freeze .* from 'immer'/s/^/\/\/ @ts-ignore\n/" dist/typings.d.ts
```
markerikson, line 75: 🤣

markerikson (approved):

.github/workflows/tests.yml
```diff
@@ -69,6 +69,12 @@ jobs:
         run: |
           sed -i -e 's|@reduxjs/toolkit": \["./src"\]|@reduxjs/toolkit": ["."]|' ./type-tests/files/tsconfig.json
 
+      - name: Prefix `freeze` re-export for pre-3.7 TS versions with @ts-ignore
+        if: ${{ matrix.ts < 3.7 }}
+        run: |
+          sed -i -e "/import .* freeze .* from 'immer'/s/^/\/\/ @ts-ignore\n/" dist/typings.d.ts
```
msutkowski, line 75: 🤯

nathggns:
Thanks for this — sorry I couldn't get around to writing this myself after our conversations. I think this is going to make quite a big difference to removing a lot of complexity in some of our code and I appreciate it.

OmarShehata:
Should `serializeError` be documented on here: https://redux-toolkit.js.org/api/createAsyncThunk ?

Issue #680, opened 2020-08-05
Autocomplete bug on vscode when using action in prepare/reducer format 


LoiKos:
I get an issue with vscode autocomplete when using slice with actions in reducer/prepare format.

```js
const todosSlice = createSlice({
...
  reducers: {
    addTodo: {
      reducer(state, action) {
        const { id, text } = action.payload;
        state.push({ id, text, completed: false });
      },
      prepare(text) {
        return { payload: { text, id: nextTodoId++ } };
      },
    },
    ...
  },
});

export const actions = todosSlice.actions;
```

I have build a small [project](https://github.com/LoiKos/intellisense-bug)  to demonstrate. 

Inside the src/features/todos you have the slice that include a "basic" function addTodo and a version with the format reducer/prepare. If you use the reducer/prepare version you shoudn't be able to access to autocomplete in other file from the folder (src/features/todos) but it just work well if you use the other version.

You still could see autocomplete with 'abc' symbols if you already use the function but that's all.

phryneas:
Can you validate if it works when you annotate the method parameters?

```js
const todosSlice = createSlice({
...
  reducers: {
    addTodo: {
	  /** @param {import('@reduxjs/toolkit').PayloadAction<{text: string, id: number}>} action */
      reducer(state, action) {
        const { id, text } = action.payload;
        state.push({ id, text, completed: false });
      },
      /** @param {string} text */
      prepare(text) {
        return { payload: { text, id: nextTodoId++ } };
      },
    },
    ...
  },
});
```

LoiKos:
Tested on [project](https://github.com/LoiKos/intellisense-bug) and works fine !

phryneas:
Yeah. So our TypeScript support is so good that it even works for JavaScript (which is why you are getting all that autocomplete), but there are some edge cases where our TypeScript cannot catch on everything that JavaScript code without types does.
I'll try to find a solution for this, but no guarantees. Writing TS to support JS is kind of bonkers ;)

LoiKos:
> Yeah. So our TypeScript support is so good that it even works for JavaScript (which is why you are getting all that autocomplete)
 
Yeah it's pretty awesome ! 

>I'll try to find a solution for this, but no guarantees. Writing TS to support JS is kind of bonkers ;)

Thank you if a fix is possible, it would be perfect :).

sweetliquid:
When I used prepare in Typescript, a type error occurred. Is it related to this issue?

[...]

PR #814, opened 2020-11-19
change default ThunkArg of cAT to OptionalUnknown

phryneas:
Since TypeScript 4.0, the following example will cause an error when checking JavaScript files (with allowJs & checkJs):
```js
const thunk = createAsyncThunk('', arg => {})
thunk()
// @ts-expect-error Expected 0 arguments, but got 1.ts(2554)
thunk('something')
```
The only way around that without breaking too much explicit existing behaviour is this type "OptionalUnknown".
On the flip side, using a payloadcreator without a defined argument, like
```ts
const thunk = createAsyncThunk('', () => {})
```
will now generate a thunk that can optionally be called with any first argument opposed to a thunk that would not accept any argument before.

This is not nice, but there's not much way around it without significantly changing `createAsyncThunk` types that are public. (And even then I'm not sure if I could find a better solution).
This is a change as well, but only one that will lead to more relaxed typechecking, so nobody upgrading should get any errors.

@markerikson, opinions on this?

### DIFF
--- a/etc/redux-toolkit.api.md
+++ b/etc/redux-toolkit.api.md
@@ -129,7 +129,7 @@ export function createAction<P = void, T extends string = string>(type: T): Payl
 export function createAction<PA extends PrepareAction<any>, T extends string = string>(type: T, prepareAction: PA): PayloadActionCreator<ReturnType<PA>['payload'], T, PA>;
 
 // @public (undocumented)
-export function createAsyncThunk<Returned, ThunkArg = void, ThunkApiConfig extends AsyncThunkConfig = {}>(typePrefix: string, payloadCreator: AsyncThunkPayloadCreator<Returned, ThunkArg, ThunkApiConfig>, options?: AsyncThunkOptions<ThunkArg, ThunkApiConfig>): AsyncThunk<Returned, ThunkArg, ThunkApiConfig>;
+export function createAsyncThunk<Returned, ThunkArg = OptionalUnknown, ThunkApiConfig extends AsyncThunkConfig = {}>(typePrefix: string, payloadCreator: AsyncThunkPayloadCreator<Returned, ThunkArg, ThunkApiConfig>, options?: AsyncThunkOptions<ThunkArg, ThunkApiConfig>): AsyncThunk<Returned, ThunkArg, ThunkApiConfig>;
 
 // @public (undocumented)
 export function createEntityAdapter<T>(options?: {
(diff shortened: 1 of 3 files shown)


markerikson:
So the only issue here is TS compat with `checkJs` files?

That doesn't seem like it's worth us spending much effort on, tbh - I'd rather have better types in TS files.

phryneas:
Might be affecting JS typehints in VSCode in general as well. My testing on that has been a little unconclusive.

markerikson:
I'm just really hesitant to do any changes that weaken the actual TS support for the purposes of semi-typed-JS interop :)

phryneas:
Hey, have you become me now, or what? :D

It's not *that* bad, so I'd *kinda* be okay with it:
image

But I'll try and investigate if this also occurs like, in a blank CRA JS installation.

phryneas:
Okay, blank CRA JS seems unaffected. 
So, make a draft PR out of this and see if multiple real people are actually affected by this? 
So far, I've only seen the case in one stackoverflow question.

phryneas:
Just some more experimentation:

* this also happens when adding `// @ts-check` to a .js file
* this also happens when adding a `jsconfig.json` with
```json
{
  "compilerOptions": {
    "moduleResolution": "node",
    "target": "es6",
    "checkJs": true
  }
}
```
to the project.

This can be fixed in the code by providing a type for the argument via
```js
const thunk2 = createAsyncThunk(
  "",
  /**  @param arg {string} */
  (arg) => {
    console.log(arg);
  }
);
// @ts-expect-error Expected 1 arguments, but got 0.ts(2554)
thunk2();
thunk2("something");
```

Especially the `jsconfig.json` bit is something that I would actually for a chunk of projects in the wild to use.
But probably by people who want the additional type safety - so maybe it would even be good for them to use it with that type annotation?

So we could alternatively go the way of *documenting* this. But I wouldn't really know where to put this.

phryneas:
This also goes for #680, which we can't really fix in types. We also cannot detect if the current file is being run in TS or JS and act accordingly.

markerikson:
I'm still not really clear on what the status of this one is :)

phryneas:
Could be cool, could be uncool, not really sure how many people it would affect. 
We should probably at least add documentation on how to get around it with JSDoc at some point.
Either way: no need to rush a decision for 1.5

PR #815, opened 2020-11-19
[entityAdapter] handle draft states in selectors

### LINKED ISSUE #606: Idea: wrapping createSelector
Just wanted to get a log of some ideas here in regards to: https://github.com/immerjs/immer/issues/561 and the original issue it references: https://github.com/reduxjs/redux-toolkit/issues/452

@phryneas and I were speaking earlier and brainstormed a few ideas for how this issue could be addressed. Currently, once we bump to v1.4 _(or whichever has immer 7.0.1)_, a user could solve this themselves by doing something like `selectorA(current(state))` inside of a reducer. Being that we use immer everywhere, we could provide a wrapped version of `createSelector` and/or entity adapter selectors that does this automatically for a user. 

Basically, something that would achieve this:
`const wrappedEntitySelector = (value) => isDraft(value) ? entitySelector(current(value)) : entitySelector(value)`

@phryneas I might've forgotten part of that convo, so please correct me if I'm missing something

phryneas:
Closes #606.

@markerikson what are your thoughts on exporting `createDraftSafeSelector`?

### DIFF
--- a/docs/api/createSelector.mdx
+++ b/docs/api/createSelector.mdx
@@ -53,7 +53,7 @@ executed on the same object - but `safe2` will actually be different from `safe1
 because the safe selector detected that it was executed on a `immer` draft object and recalculated using the current
 value instead of returning a cached value.
 
-Using it with non-draft states, this selector will still be caching, but using it with draft values it will
+When used with plain state values, the selector will still memoize normally based on the inputs.  But, when used with Immer draft values, the selector will err on the side of recalculating the results, just to be safe.
 err on the side of recalculating, just to be safe.
 
 All selectors created by `entityAdapter.getSelectors` will be such "draft safe" selectors by default.
--- a/docs/api/createSelector.mdx
+++ b/docs/api/createSelector.mdx
@@ -22,10 +22,18 @@ For more details on using `createSelector`, see:
 
 # `createDraftSafeSelector`
 
-Besides re-exporting `createSelector`, RTK also exports a wrapped version of `createSelector` named `createDraftSafeSelector` that allows you to create selectors that
-can be used in the context of `createReducer` and `createSlice` reducers with mutable logic.
+In general, we recommend against using selectors inside of reducers:
 
-So assume the following code:
+- Selectors typically expect the entire Redux state object as an argument, while slice reducers only have access to a specific subset of the entire Redux state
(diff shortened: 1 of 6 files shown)


markerikson:
That I could live with.

Y'know, I'm really starting to lose track of all the stuff we _do_ export from RTK... :)

phryneas:
Note to self: export that thing *and* add that to the docs. Tomorrow :)

That TODO list for 1.5.0 is getting shorter and shorter though :)

phryneas:
@markerikson any comments on the docs for this one? :)

docs/api/createSelector.mdx
```diff
@@ -19,3 +19,41 @@ For more details on using `createSelector`, see:
 > **Note**: Prior to v0.7, RTK re-exported `createSelector` from [`selectorator`](https://github.com/planttheidea/selectorator), which
 > allowed using string keypaths as input selectors. This was removed, as it ultimately did not provide enough benefits, and
 > the string keypaths made static typing for selectors difficult.
+
+# `createDraftSafeSelector`
+
+Besides re-exporting `createSelector`, RTK also exports a wrapped version of `createSelector` named `createDraftSafeSelector` that allows you to create selectors that
+can be used in the context of `createReducer` and `createSlice` reducers with mutable logic.
```
msutkowski, line 26: Should we explicitly state extra reducers as well?

docs/api/createSelector.mdx
```diff
@@ -19,3 +19,41 @@ For more details on using `createSelector`, see:
 > **Note**: Prior to v0.7, RTK re-exported `createSelector` from [`selectorator`](https://github.com/planttheidea/selectorator), which
 > allowed using string keypaths as input selectors. This was removed, as it ultimately did not provide enough benefits, and
 > the string keypaths made static typing for selectors difficult.
+
+# `createDraftSafeSelector`
+
+Besides re-exporting `createSelector`, RTK also exports a wrapped version of `createSelector` named `createDraftSafeSelector` that allows you to create selectors that
```
msutkowski, line 25: Do we need to explain what/why in regards to immer Drafts or link to what isDraftable even is? I think the description you have in the code would actually work well here for clarity.

msutkowski (commented):
Looks good! I just have a couple questions on descriptions

markerikson (changes requested):

PR #816, opened 2020-11-21
move CI from travis to github actions

phryneas:
Travis is ...flaky... recently, to say it friendly.

This would move the CI from travis to GH actions.
Also, it makes some changes:

* Linting, runtime tests and build now happen with only the project's current TypeScript version.
* The `dist` folder from that build is then thrown against all currently supported TS versions and only the TypeScript-tests are run with that. (before, the type-tests would run against `src`)

Time for all tests combined is down to 3 minutes with that change.

@markerikson what do you think, should I pull the trigger on this? We could run it side-by-side with travis for a while.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,82 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  build:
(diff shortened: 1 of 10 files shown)


markerikson:
Fine by me!

PR #821, opened 2020-11-27
Update Immer to 8.x

### LINKED ISSUE #809: Immer 8
Looks like Immer 8 just came out, and it switches to always auto-freezing in production.

It also looks like Immer 7 was slower than 6, but switching auto-freezing on actually _improves_ perf.

We need to look into this and figure out when/how to upgrade.  Can we do this in an RTK minor?

- https://github.com/immerjs/immer/releases/tag/v8.0.0
- https://github.com/immerjs/immer/issues/687
- https://github.com/immerjs/immer/issues/649
- https://github.com/immerjs/immer/issues/681

markerikson:
Fixes #809 .

This PR:

- Bumps `shelljs` to `0.8.4` to fix some annoying import warnings
- Adds a test that specifically imports Immer with `NODE_ENV === 'production'`, then tries to mutate state from a reducer and expects an error.  Under Immer 7, that test fails because it's an unfrozen POJO and the mutation runs without error
- Updates Immer to 8.x. Since Immer 8 autofreezes in prod, the test throws an error as expected and now passes.

My one concern atm is the Immer docs advice to consider "pre-freezing large data ahead of time":

https://immerjs.github.io/immer/docs/freezing

do we need to note this ourselves somewhere? should we be re-exporting `freeze()` or something?

### DIFF
--- a/docs/tutorials/basic-tutorial.md
+++ b/docs/tutorials/basic-tutorial.md
@@ -9,7 +9,7 @@ hide_title: true
 
 Welcome to Redux Toolkit ! This tutorial will show you the basic functions that are included with Redux Toolkit (also known as "RTK" for short).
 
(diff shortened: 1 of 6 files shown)


markerikson:
Uh... CI blew up on this push with some unrelated TS errors.  Something about the integration branch, maybe?

msutkowski (approved):

PR #822, opened 2020-11-27
overload: call thunk matchers with actions 

phryneas:
adds the ability to call
```js
isRejected(action)
```
in addition to the existing behaviour
```js
isRejected()(action)
isRejected(thunk, ...moreThunks)(action)
```


markerikson:
Code seems fine, but TS doesn't like some unused variables or something.

Issue #826, opened 2020-11-29
TypeScript error when trying to recreate Slice example from doc (defining-the-initial-state-type)


paulintrognon:
I tried to recreate [this example](https://redux-toolkit.js.org/usage/usage-with-typescript#defining-the-initial-state-type)

The reducers are not shown in the example. So I wanted to try by myself, but I keep getting TypeScript errors in my `loadData` reducer:
```ts
import { createSlice, PayloadAction } from '@reduxjs/toolkit'

type SliceState = { state: 'loading' } | { state: 'finished'; data: string }

const initialState: SliceState = { state: 'loading' }

createSlice({
  name: 'test1',
  initialState,
  reducers: {
    // TS Error: "Type '"finished"' is not assignable to type '"loading"'"
    loadData: (state, action: PayloadAction<string>) => ({
      ...state,
      state: 'finished',
      data: action.payload,
    }),
  },
})
```

How can I get this kind of SliceState type to work? Is it possible to include an example of a working reducer in the documentation?

Cheers

phryneas:
Yeah, this is unfortunately a problem in newer TS versions as TS will do a code flow analysis and decide on it's own that `initialState` will always be of type `{ state: 'loading' }`, because it never gets reassigned.

You can do `const initialState = { state: 'loading' } as SliceState` instead, as it's more expicit & TS will honor that.

Thank you for pointing that out - I'll fix up the docs.

Issue #735, opened 2020-09-23
createSlice reducer cannot properly infer the state type.


cxc421:
I try to define the type of initial state to `createSlice()`, and according to the [document](https://redux-toolkit.js.org/usage/usage-with-typescript#defining-the-initial-state-type):

> The standard approach is to declare an interface or type for your state, create an initial state value that uses that type, and pass the initial state value to createSlice. You can also use the construct initialState: myInitialState as SliceState.

I copied the code snippet to the [CodeSandbox](https://codesandbox.io/s/rtk-ts-test-f492j?file=/src/index.ts), but it seems like the reducer cannot correctly infer the state type.

![image](https://user-images.githubusercontent.com/18439296/93988678-97759a80-fdbb-11ea-875f-3f2964181a03.png)

markerikson:
Looks like it has something to do with the way you've defined the union of the two possible states.

This isn't an RTK-specific issue, so I'm going to close this. The issue tracker here on GitHub is reserved for bug reports and feature requests. For usage questions (which is what I believe this is), please use Stack Overflow or Reactiflux where there are a lot more people ready to help you out. Thanks!

phryneas:
This is an unfortunate TypeScript "feature" that is a bit unfortunate:

```ts
function identity<T>(t: T) {return t}


type SliceState = { state: "loading" } | { state: "finished"; data: string };

let initialState: SliceState = { state: "loading" };

// is inferred as  { state: "loading"; }
const derivedValue = identity(initialState)
```

TS actually does some code flow analysis and sometimes "tightens" types that you yourself declared wider.

Doing
```ts
const initialState = { state: "loading" } as SliceState;
```

would work as well here, as it is an implicit cast.

We might need to check our docs for this in a few places.

cxc421:
Got it, thank you!

cefn:
I'm concerned this documentation guides people towards the practice of routinely bypassing the compiler. This will create runtime errors in production since they are bypassing Typescript for no obvious reason. 

Consider the following simplified example which will compile fine, but break in production when following the updated guidance (and you can see it break running in the Typescript playground).

```typescript
interface UsersState {
  error: string | null | undefined;
  entities: Record<string, string>;
}

const initialState = {
  error: null,
  // entities here omitted by mistake
} as UsersState; // tell the compiler to ignore your mistake

console.log(`Number of entities is ${Object.keys(initialState.entities).length}`); // Error: Cannot convert undefined or null to object
```

[...]

PR #827, opened 2020-11-29
explicitly cast `initialState` in docs examples

phryneas:
see #826 and #735 - copying from the latter why this is neccessary:

This is a TypeScript "feature" that is a bit unfortunate and that I've noticed in recent versions:

```ts
function identity<T>(t: T) {return t}


type SliceState = { state: "loading" } | { state: "finished"; data: string };

let initialState: SliceState = { state: "loading" };

// is inferred as  { state: "loading"; }
const derivedValue = identity(initialState)
```

TS actually does some code flow analysis and sometimes "tightens" types that you yourself declared wider.

Doing
```ts
const initialState = { state: "loading" } as SliceState;
```

would work as well here, as it is an implicit cast.

We might need to check our docs for this in a few places.

_Originally posted by @phryneas in https://github.com/reduxjs/redux-toolkit/issues/735#issuecomment-697637009_

### DIFF
--- a/docs/api/createAsyncThunk.mdx
+++ b/docs/api/createAsyncThunk.mdx
@@ -574,10 +574,10 @@ interface UsersState {
   entities: Record<string, User>
 }
 
-const initialState: UsersState = {
+const initialState = {
   entities: {},
   error: null
-}
+} as UsersState
 
 const usersSlice = createSlice({
   name: 'users',
--- a/docs/api/createReducer.mdx
+++ b/docs/api/createReducer.mdx
@@ -52,7 +52,7 @@ const increment = createAction('counter/increment')
 const decrement = createAction('counter/decrement')
 const incrementByAmount = createAction<number>('counter/incrementByAmount')
 
-const initialState: CounterState = { value: 0 }
+const initialState = { value: 0 } as CounterState
 
 const counterReducer = createReducer(initialState, builder => {
   builder
--- a/docs/api/createSlice.mdx
+++ b/docs/api/createSlice.mdx
@@ -22,7 +22,7 @@ interface CounterState {
   value: number
 }
 
-const initialState: CounterState = { value: 0 }
+const initialState = { value: 0 } as CounterState
 
 const counterSlice = createSlice({
   name: 'counter',
--- a/docs/api/matching-utilities.mdx
+++ b/docs/api/matching-utilities.mdx
@@ -174,10 +174,10 @@ interface ExampleState {
   isInteresting: boolean
 }
 
-const initialState: ExampleState = {
+const initialState = {
   isSpecial: false,
   isInteresting: false
-}
+} as ExampleState
 
 export const isSpecialAndInterestingThunk = createAsyncThunk(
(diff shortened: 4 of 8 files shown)


msutkowski (approved):
LGTM. I'm also doing this same thing in the rtk-query examples already 👍

phryneas:
completely forgot to merge this one :facepalm:

polaroidkidd:
quick question: does this mean that if I declare my state as suggested I'll have to declare every possible state permutation?

phryneas:
@polaroidkidd only if you want your state to accept only very specific shapes.
If you are content with `{ state: string; data?: string };`, just use that.
But if you want `data` to only be accessible when `state` is `"finished"`, you'd need to go for a union type. But that's nothing redux-specific at all, that's just TypeScript.

polaroidkidd:
I'm aware that my following question is typescript specific, but I'm banking on @phryneas's generosity here ^^

I have the following code:

```typescript
import { createSlice, PayloadAction } from "@reduxjs/toolkit";

export type ErrorSlice =
  | { showErrorModal: false }
  | { showErrorModal: true; errorTitel: string; errorBody: string };

const initialState = {
  showErrorModal: false,
  errorBody: null,
  errorTitel: null
} as ErrorSlice;

const { reducer: uiErrorReducer, actions: uiErrorActions } = createSlice({
  name: "ui/error",
  initialState: initialState,
  reducers: {
    hideErrorModal: () => initialState,
    showErrorModal: (
      state,
      action: PayloadAction<{ errorTitel: string; errorBody: string }>
    ) => ({ showErrorModal: true, ...action.payload }),
  },
});

export { uiErrorActions, uiErrorReducer };
```  


I am wondering why TS isn't throwing an error for declaring  
```typescript 
const initialState = {
  showErrorModal: false,
  errorBody: null,
  errorTitel: null
} as ErrorSlice;
```

as the `ErrorSlice` type does not declare a permutation that allows for `showErrorModal: false` and `errorBody: null` & `errorTitle: null`. I can even set the error title/Body to contain strings (`"example"`), but the initial state is accepted as is. 


What am I missing here? :thinking:

phryneas:
@polaroidkidd interfaces only describe a shape that should be matched, but don't restrict other properties from being present. If you want that, you have to 
```ts
export type ErrorSlice =
  | { showErrorModal: false; errorTitel: never; errorBody: never }
  | { showErrorModal: true; errorTitel: string; errorBody: string };
```

polaroidkidd:
oh wow! I wasn't aware of the `never` property. This will be super useful! Thank you very much!

danielo515:
Doesn't have typescript a way to define "exact" objects? Flow has this notation `{| prop: boolean |}` which will not allow extra props, just that.

phryneas:
No. https://www.typescriptlang.org/docs/handbook/type-compatibility.html#a-note-on-soundness

danielo515:
Thanks for the link.
What a shame

2donny:
we

2donny:
asd

sably:
i couldn't understand

PR #841, opened 2020-12-09
experiment: specific entityId type

phryneas:
This came up this morning in the chat with Shrugsy: entityType is never fixed to whatever is actually returned by the `selectId` function.

This **would** do that. But at the added type complexity, I guess it won't be worth it. Nonetheless, I gave it a try.

I guess this is one of those "I'll make a PR to document that we tried it and it didn't work out" PRs that we can point to in the future if someone asks for it.

Unless you see it differently and want to get it in @markerikson ? It **has** some use, it's just questionable if that is worth the overhead.

### DIFF
--- a/src/entities/create_adapter.ts
+++ b/src/entities/create_adapter.ts
@@ -1,4 +1,10 @@
-import { EntityDefinition, Comparer, IdSelector, EntityAdapter } from './models'
+import {
+  EntityDefinition,
+  Comparer,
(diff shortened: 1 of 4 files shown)


markerikson:
Ew. Yeah, I don't think that's worth it atm.  (Not an unreasonable request, or a completely unreasonable implementation, but not something I think is justified in being added at this time.)

Issue #750, opened 2020-10-03
Setting "meta" on the action generated by "createAsyncThunk"


sanathsharma:
I have a couple of middlewares that get triggered by identifying the presence of few keys in the "meta" object on the action. "meta" of the action created by createAsyncThunk has few useful keys like "arg" & "requestId". 

Is there any way to add custom key-values to the "meta" object along side these keys?

markerikson:
At the moment, no, there isn't, beyond the fact that whatever you pass as an argument to the thunk action creator when you dispatch it will wind up as `action.meta.arg`.

What specifically are you needing to add to the `meta` field for those actions?

sanathsharma:
Thank you for your reply.

Here is the interface of the meta object that i generally use if I'm using createAction or a regular action.
```ts
interface Meta {
    throttle?: number;
    debounce?: number;
};
```

```ts
// createAction, simple example for how the action looks like for the middleware
const action = createAction( "someActionType", ( data:Record<string,any> ) => ( {
    payload: data,
    meta: {
        debounce: 1000
    }
} ) );
```
Debounce middleware is really useful when making a hit to search api end point with some debounce time, waiting for user to stop typing in a search field.

jstclair:
I was *just* looking for something similar, but not at the point of dispatch, but rather in `.rejected` - it would be nice to capture some additional information about the (in our case) axios request without having to wrap in a try / catch (for instance, request.url , etc.) 
In our app, we're capturing the existing  xx_FAILURE actions to send to our error logging service via custom middleware - I thought I could just replace that with `action.type.endsWith('/rejected') but `meta` and `error` don't provide the same information

sanathsharma:
The keys need not end up in fulfilled, pending, and rejected actions. I just need a way to halt or stop the action for getting dispatched. The `condition` options does the trick for throttling but not for debouncing.

```ts
```

sanathsharma:
The keys need not end up in fulfilled, pending, and rejected actions. Because i don't want to `throttle` or `debounce` fulfilled action or other actions. I want to control the api hit happening inside of `payloadCreator`. I get that, that is the reason for existence of `condition option`.

I can us it in below fashion to achieve throttling.

```ts
// throttle.tsx

// ----------------------------------------------------------------------------------------
// variables

const throttled: Record<string, boolean> = {};

// ----------------------------------------------------------------------------------------
// function

/**
 * throttle function to be applied on condition callback for createAsyncThunk
[...]

PR #844, opened 2020-12-10
Add an ability to pass custom meta object to the `createAsyncThunk`

stanislav-halyn:
This is a PR for the issue [#750](https://github.com/reduxjs/redux-toolkit/issues/750).

It allows to pass custom `additionalMeta` object to the options of `createAsyncThunk` and it'll be passed to the action's `meta` object as an `additionalMeta` field

### DIFF
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -217,6 +217,8 @@ interface AsyncThunkOptions<
   dispatchConditionRejection?: boolean
 
   serializeError?: (x: unknown) => GetSerializedErrorType<ThunkApiConfig>
+
+  additionalMeta?: { [key: string]: unknown }
 }
 
 export type AsyncThunkPendingActionCreator<
@@ -230,6 +232,7 @@ export type AsyncThunkPendingActionCreator<
     arg: ThunkArg
     requestId: string
     requestStatus: 'pending'
+    additionalMeta: { [key: string]: unknown }
   }
 >
 
@@ -248,6 +251,7 @@ export type AsyncThunkRejectedActionCreator<
     requestStatus: 'rejected'
     aborted: boolean
     condition: boolean
+    additionalMeta: { [key: string]: unknown }
   }
 >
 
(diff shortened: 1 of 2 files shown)


markerikson:
Per that thread, if we're going to consider this, it shouldn't be just a single object. It should be some kind of a callback function - basically the same idea as the `prepare` callback in our slices.

stanislav-halyn:
> Per that thread, if we're going to consider this, it shouldn't be just a single object. It should be some kind of a callback function - basically the same idea as the `prepare` callback in our slices.

Are there any arguments that should be passed to the meta ? We could pass the 'ThunkAPI' to allow users create meta based on the state etc.

src/createAsyncThunk.ts
```diff
@@ -263,6 +267,7 @@ export type AsyncThunkFulfilledActionCreator<
     arg: ThunkArg
     requestId: string
     requestStatus: 'fulfilled'
+    additionalMeta: { [key: string]: unknown }
```
phryneas, line 270: Also, these would need to be correctly typed - which might make the whole thing reaaally complicated though :(

stanislav-halyn: Agree, it might get complicated, but I think it's an essential feature, because right now I don't see any good ways to identify some particular action in a middleware.

For example, right now we are switching to redux toolkit and we have an ability to identify whether the action should show loader when loading or not. It's a global loader which is handled in a redux middleware, and right now I don't have any ideas of how to implement it with redux toolkit thunks, but changing meta, which is unfortunately impossible.


I could try to type it correctly

phryneas: Just to be sure here: you've seen the matchers and those don't do it for you?
https://redux-toolkit.js.org/api/matching-utilities

stanislav-halyn: Yes, I have seen them and they still don't help.

In current code base, we have actions that might have a meta data field 'showLoader'. The middleware fires special actions if there's 'showLoader: true'.

So, I doubt these could help me, because I fire the actions only if there's 'showLoader:true'

phryneas: > I could try to type it correctly

You can try, but please try to keep if as simple (while correct) as possible as these types here already tend to kinda explode. I'd suggest adding the return type for this this to `AsyncThunkConfig` which is probably the least intrusive way to implement this.

markerikson:
Superseded by #1083 .

PR #846, opened 2020-12-11
Remove unnecessary condition

sweetliquid:
For `extraReducers`, when it is a function it will not be `undefined`, when it is `undefined` we defined a default value for it. Therefore, the judment of `undefined` is redundant :)

https://github.com/reduxjs/redux-toolkit/blob/d5f1b65541c8b2a7b99b7e7ba5d34c6c3e2ac896/src/createSlice.ts#L257-L266

### DIFF
--- a/src/createSlice.ts
+++ b/src/createSlice.ts
@@ -259,9 +259,7 @@ export function createSlice<
     actionMatchers = [],
     defaultCaseReducer = undefined
   ] =
-    typeof options.extraReducers === 'undefined'
-      ? []
-      : typeof options.extraReducers === 'function'
+    typeof options.extraReducers === 'function'
       ? executeReducerBuilderCallback(options.extraReducers)
       : [options.extraReducers]
 


src/createSlice.ts
```diff
@@ -259,9 +259,7 @@ export function createSlice<
     actionMatchers = [],
     defaultCaseReducer = undefined
   ] =
-    typeof options.extraReducers === 'undefined'
-      ? []
-      : typeof options.extraReducers === 'function'
+    typeof options.extraReducers === 'function'
```
juzhiyuan, line 262: If options.extraReducers === undefined, then the result would be [undefined], right?

phryneas: It's a little "wrap-your-head-around" with this.

for `options.extraReducers === undefined`, this would result in 
```js
 const [ 
   extraReducers = {}, 
   actionMatchers = [], 
   defaultCaseReducer = undefined 
 ] = [undefined]
```
which would then take on the default value of `extraReducers = {}` at desctructuring.

So yeah, I believe the PR is correct, but I'd like a second opinion here @markerikson or @msutkowski

markerikson: Let's go with it.

markerikson:
Eh, this shoulda gone to `master`. manually merging back locally.

PR #855, opened 2020-12-23
Revert "Use main reactiflux invite"

nickmccurdy:
Reverts reduxjs/redux-toolkit#768

Follow up on https://github.com/reduxjs/react-redux/pull/1673

### DIFF
--- a/docs/introduction/quick-start.md
+++ b/docs/introduction/quick-start.md
@@ -62,6 +62,6 @@ The UMD package can be used as a [`<script>` tag](https://unpkg.com/@reduxjs/too
 
 ## Help and Discussion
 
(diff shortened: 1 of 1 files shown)


nickmccurdy:
CI is failing on the type test but I'm only modifying documentation

phryneas:
Yeah, it's only Travis.

On the topic itself though: would this channel-specific invite work? As far as I remember, when signing up for the server, all channels are locked and you have to check a checkbox over in announcments or so to gain access. Landing in another channel might be counterproductive?

nickmccurdy:
> Would this channel-specific invite work? As far as I remember, when signing up for the server, all channels are locked and you have to check a checkbox over in announcements or so to gain access. Landing in another channel might be counterproductive?

Yes. We've made our community channels public as we're now using [rule screening](https://support.discord.com/hc/en-us/articles/1500000466882-Rules-Screening-FAQ) instead of private channels with the Reaction Role bot

phryneas:
Ah, I missed that! Then let's get it in :)

PR #861, opened 2020-12-29
Unwrap result of async thunk so error throws to calling component

speakingcode:
Example "Requesting a user by ID, with loading state, and only one request at a time" should unwrap the async thunk promise result so that errors will throw and cause the toast to display.


markerikson:
Thanks!

PR #889, opened 2021-02-07
add "Basic Project Setup" section to TS docs, move TS docs to top level

phryneas:
This adds a "Basic Project Setup" section to the TS docs and also moves the quick start guide, TS docs & usage guide to the top level of the navigation.
The only two categories are now "Tutorials" and "API reference" while I directly uncollapsed the API reference part. What do you think @markerikson ?

### DIFF
--- a/docs/usage/usage-guide.md
+++ b/docs/usage/usage-guide.md
@@ -183,20 +183,21 @@ Notice that we specifically call `state.concat()` to return a copied array with 
 With `createReducer`, we can shorten that example considerably:
 
 ```js
-const todosReducer = createReducer([], {
-  ADD_TODO: (state, action) => {
-    // "mutate" the array by calling push()
-    state.push(action.payload)
-  },
-  TOGGLE_TODO: (state, action) => {
-    const todo = state[action.payload.index]
-    // "mutate" the object by overwriting a field
-    todo.completed = !todo.completed
-  },
-  REMOVE_TODO: (state, action) => {
-    // Can still return an immutably-updated value if we want to
-    return state.filter((todo, i) => i !== action.payload.index)
-  }
+const todosReducer = createReducer([], builder => {
+  builder
+    .addCase('ADD_TODO', (state, action) => {
+      // "mutate" the array by calling push()
+      state.push(action.payload)
+    })
+    .addCase('TOGGLE_TODO', (state, action) => {
+      const todo = state[action.payload.index]
+      // "mutate" the object by overwriting a field
+      todo.completed = !todo.completed
+    })
+    .addCase('REMOVE_TODO', (state, action) => {
+      // Can still return an immutably-updated value if we want to
+      return state.filter((todo, i) => i !== action.payload.index)
+    })
 })
 ```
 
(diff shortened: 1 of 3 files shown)


phryneas:
The next step after that would be a PR to https://redux.js.org/recipes/usage-with-typescript/ to add something like "This page describes a way to use vanilla redux with TypeScript. We do not recommend this though and instead recommend using our official redux toolkit instead, which will reduce the amount of required TypeScript typings significantly." to the first paragraph.

markerikson:
Hmm. I definitely like the addition of a "basic project setup" section and updating the core docs page. Not sure I like changing the categories structure.

Also, I'm concerned about suggesting that the typed hooks go exactly alongside the store, because that could lead to circular reference issues.  I think it would be good if we explicitly showed that the store code lives in a `app/store.ts` and the hooks in `app/hooks.ts`, to ensure safe imports.

phryneas:
The problem with the current structure is that the "Usage With TypeScript" section is not really discoverable without randomly clicking headers to open. Honestly, I would never intuitively think of clicking "Using Redux Toolkit" for TypeScript infos - and on redux.js.org, the same section is under "Recipes", which makes it even more confusing. 

I think we will save a lot of low-effort issues if we have that as a top-level item. And that would leave us with three categories with one item each.

As for the store & hooks, yeah, I can split those later.

markerikson:
For whatever reason I find myself not wanting to have individual pages at the top level, and both of these pages _are_ "usage guides".  Could we switch to having that category expanded by default?  Pretty sure there's a Docusaurus TOC option for that.

On the Redux core side, I'm about to create a "Using Redux" section and begin moving some of the "recipes" into that, so there will be a bit more consistency there.

website/sidebars.json
```diff
@@ -1,46 +1,63 @@
 {
```
phryneas, line 1: The sections should look like you want them to now.

docs/usage/usage-with-typescript.md
```diff
@@ -9,13 +9,44 @@ hide_title: true
...
 **If you encounter any problems with the types that are not described on this page, please open an issue for discussion.**
 
+## Basic Project Setup
+
+Using [configureStore](../api/configureStore.mdx) should not need any additional typings. You will, however, want to extract the `RootState` type and the `Dispatch` type.
+Also, you will want to create typed versions of the `useDispatch` and `useSelector` hooks for usage in your application.
+
+That said, a basic setup would look like this:
+
+```ts title="store.ts"
```
phryneas, line 23: split up into two files

docs/usage/usage-guide.md
```diff
@@ -183,20 +183,21 @@ Notice that we specifically call `state.concat()` to return a copied array with
...
-  TOGGLE_TODO: (state, action) => {
-    const todo = state[action.payload.index]
-    // "mutate" the object by overwriting a field
-    todo.completed = !todo.completed
-  },
-  REMOVE_TODO: (state, action) => {
-    // Can still return an immutably-updated value if we want to
-    return state.filter((todo, i) => i !== action.payload.index)
-  }
+const todosReducer = createReducer([], builder => {
```
phryneas, line 186: As we've had reports of confusion over "what is the recommended approach" with the usage guide still showing the object notation, I've update this part to builder notation.

docs/usage/usage-guide.md
```diff
@@ -230,30 +231,6 @@ updateValue(state, action) {
 
 Much better!
 
-### Defining Functions in Objects
```
phryneas, line 233: This paragraph doesn't make sense here any more. I've moved it down to `createSlice`, but I'm not 100% happy with the position there. Maybe we could also get it removed, because the Computed property notation is not really relavant any more and the rest needs fewer explanation?

phryneas (commented):
I've done some updates, @markerikson could you take another look?

docs/usage/usage-with-typescript.md
markerikson: Not seeing "two files" here - just the one "Usage with TS" page.  Did you forget to push something, or am I missing what you're referring to?

markerikson: Oh wait, you mean this _example_ is in two files now. got it.

PR #900, opened 2021-02-17
Docs: Update usage examples of customizing middleware

msutkowski:
- Prefers callback/concat usage overspreading `getDefaultMiddleware`

### DIFF
--- a/docs/usage/usage-guide.md
+++ b/docs/usage/usage-guide.md
@@ -109,7 +109,7 @@ Note that this only works for one level of reducers. If you want to nest reducer
 If you need to customize the store setup, you can pass additional options. Here's what the hot reloading example might look like using Redux Toolkit:
 
 ```js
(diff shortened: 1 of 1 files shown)


markerikson:
Totally unrelated side note:

Is there any way we can detect that a particular PR doesn't touch the docs, and skip all the TS checks?  That just seems... wasteful :)

msutkowski:
> Totally unrelated side note:
> 
> Is there any way we can detect that a particular PR doesn't touch the docs, and skip all the TS checks? That just seems... wasteful :)

We might be able to skip those checks on any PR that start with 'Docs' with a filter expression? https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#functions. I'm not 100% sure there, but that's the most likely thing I can find. I think in order to make that really work, we'd have to use a template to enforce standardization of PR names (at least, I think this is possible?)

PR #903, opened 2021-02-19
Upated tutorial with `createAsyncThunk` link

chomamateusz:
The tutorial says that `redux-toolkit` can't help with thunk, but as far as I know it is not true as `createAsyncThunk` function exists.


markerikson:
Obsolete due to #905 .

Issue #890, opened 2021-02-09
Bump immer dependency version to 8.0.1


koriner:
There is a current npm vulnerability for immer which is giving us warnings. Would it be possible to bump the version of immer?

Related patch on immer: https://github.com/immerjs/immer/commit/da2bd4fa0edc9335543089fe7d290d6a346c40c5

markerikson:
We already depend on `immer@^8.0.0`, so you just need to update your own installation.  

Dupe of #871, #874, #879 .

Issue #879, opened 2021-01-25
Prototype Pollution vulnerability in Immer 8.0.0


emanuelef:
Immer 8.0.0 seems to be vulnerable: https://github.com/immerjs/immer/issues/738
Would it be possible to update the dependency to 8.0.1 ?
Thanks

markerikson:
Already done because we depend on `immer@^8.0.0`.  Dupe of #871 and #874 .

Issue #874, opened 2021-01-22
[Security Vulnerability] - CVE-2020-28477 - immer package


pedrofernandes-hotmart:
CVE-2020-28477
nodejs-immer: prototype pollution may lead to DoS or remote code execution

"VulnerabilityID": "CVE-2020-28477",
 "PkgName": "immer",
 "InstalledVersion": "8.0.0",
 "FixedVersion": "8.0.1",
 "Layer": {
 "DiffID": "sha256:d5ebd5bd95b34339da24e0c1d7404f158e2c9aca695337aab1b41bf8e25e252f"
 },
"SeveritySource": "nvd",
"PrimaryURL": "https://avd.aquasec.com/nvd/cve-2020-28477",
"Title": "nodejs-immer: prototype pollution may lead to DoS or remote code execution",
"Description": "This affects all versions of package immer.",
"Severity": "HIGH",
"CVSS": {
"nvd": {
"V3Vector": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H",
"V3Score": 7.5
},

phryneas:
We depend on `"immer": "^8.0.0",` which means that when you install RTK and don't have `immer` installed yet, it will install the newest version which is without the vulnerability.
From there it's your job to keep your dependencies up-to-date, not ours.

pedrofernandes-hotmart:
Thanks for the explanation and reference @phryneas !  Got it!

Issue #871, opened 2021-01-20
Update immer to fix vuln


FrederikBolding:
Immer just released a new version that fixes a prototype pollution vuln.

https://github.com/immerjs/immer/commit/da2bd4fa0edc9335543089fe7d290d6a346c40c5

Is it possible to update the dependency?

markerikson:
Our dependency is on `immer@^8.0.0`, so there's nothing we need to change - you just need to reinstall things on your end / update your own lockfiles.

FrederikBolding:
Ah, my bad! We were using `1.4.0` which uses v7 of immer. Thanks for the heads up! 👍

markerikson:
Yep, we specifically bumped from Immer 7 to 8 in RTK 1.5:

https://github.com/reduxjs/redux-toolkit/releases/tag/v1.5.0

FrederikBolding:
Yeah, I figured! Solving the issue by bumping our Redux Toolkit dependency. 
Thanks for building a great library and sorry about the confusion! 😄

PR #909, opened 2021-02-22
fix: vulnerability in immer

i1skn:
https://www.npmjs.com/advisories/1603

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4602,9 +4602,9 @@
       "dev": true
     },
     "immer": {
-      "version": "8.0.0",
-      "resolved": "https://registry.npmjs.org/immer/-/immer-8.0.0.tgz",
-      "integrity": "sha512-jm87NNBAIG4fHwouilCHIecFXp5rMGkiFrAuhVO685UnMAlOneEAnOyzPt8OnP47TC11q/E7vpzZe0WvwepFTg=="
+      "version": "8.0.1",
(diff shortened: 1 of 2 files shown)


markerikson:
Strictly speaking, we don't need to do anything on our end. Our own dependency says "anything 8.0.0 or newer", and 8.0.1 is already newer (ref: #871, #874, #879 , #890 ).

It's up to users to update the transitive dependencies locally, because their lockfiles have already settled on 8.0.0.  And, any fresh installs of RTK will pick up 8.0.1 automatically.

But, I suppose it doesn't hurt to bump this here too.

markerikson:
So this PR is actually bugged, because it didn't update our own lockfile too.  Can you update the lockfile as well?

markerikson:
Thanks.  Will try to publish 1.5.1 later today or tomorrow, but as mentioned, you can resolve this on your end as well by just updating the transitive dependency yourself.

(Also, looking at the PR you linked: note that we recommend using `createSlice` as the default in most situations, rather than calling `createAction` directly, unless you really need to define standalone action types.)

sleric1024:
Hi @markerikson , May I know that when will the 1.5.1 version be published ?
Thanks.

markerikson:
Sorry, been busy and this has been a low priority.

As mentioned, strictly showing we didn't even need to make this change - we already depended on anything in the 8.x range, so you just need to update your own installed packages to pick up the newer Immer version.

PR #918, opened 2021-03-09
Docs: fix createAsyncThunk args typo

mgmarlow:
Correct `action.meta` value from `args` to `arg`.

This small typo always throws me off when I reference the docs since it comes before the formal model types 😛 .


msutkowski (approved):
:+1: Good catch!

PR #928, opened 2021-03-17
small language correction in docs

raphaeleidus:
"safely immutably" was grammatically incorrect
should either be "safe immutably" or "safely immutable"

### DIFF
--- a/docs/usage/immer-reducers.md
+++ b/docs/usage/immer-reducers.md
@@ -145,7 +145,7 @@ Writing immutable update logic by hand _is_ hard, and **accidentally mutating st
 
 [Immer](https://immerjs.github.io/immer/docs/introduction) is a library that simplifies the process of writing immutable update logic.
 
-Immer provides a function called `produce`, which accepts two arguments: your original `state`, and a callback function. The callback function is given a "draft" version of that state, and inside the callback, it is safe to write code that mutates the draft value. Immer tracks all attempts to mutate the draft value, and then replays those mutations with using their immutable equivalents to create a safe immutably updated result:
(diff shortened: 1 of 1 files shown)


docs/usage/immer-reducers.md
```diff
@@ -145,7 +145,7 @@ Writing immutable update logic by hand _is_ hard, and **accidentally mutating st
 
 [Immer](https://immerjs.github.io/immer/docs/introduction) is a library that simplifies the process of writing immutable update logic.
 
-Immer provides a function called `produce`, which accepts two arguments: your original `state`, and a callback function. The callback function is given a "draft" version of that state, and inside the callback, it is safe to write code that mutates the draft value. Immer tracks all attempts to mutate the draft value, and then replays those mutations with using their immutable equivalents to create a safely immutably updated result:
+Immer provides a function called `produce`, which accepts two arguments: your original `state`, and a callback function. The callback function is given a "draft" version of that state, and inside the callback, it is safe to write code that mutates the draft value. Immer tracks all attempts to mutate the draft value, and then replays those mutations with using their immutable equivalents to create a safe immutably updated result:
```
msutkowski, line 148:
```suggestion
Immer provides a function called `produce`, which accepts two arguments: your original `state`, and a callback function. The callback function is given a "draft" version of that state, and inside the callback, it is safe to write code that mutates the draft value. Immer tracks all attempts to mutate the draft value and then replays those mutations using their immutable equivalents to create a safe, immutably updated result:
```

@raphaeleidus Thanks for pointing this out! 

This copy is still strange to me no matter how it's written: 

> then replays those mutations using their immutable equivalents to create a safe, immutably updated result

I get the intent behind 'safe' is to say 'typically, mutating things is dangerous', but it should probably just be dropped _or_ a comma should be added as I'm proposing for a pause.

msutkowski: I know @markerikson has thoughts here about the intent and style here, so let's leave your suggestion in place _(`with` would still need to be removed)_, and see what we can come up with.

markerikson:
Thanks!

PR #940, opened 2021-03-25
Freeze initial state by passing through Immer

### LINKED ISSUE #939: Initial createReducer state is not frozen
Immer auto-freezes state when processed in a case reducer. However, we currently do not freeze the _initial_ state.  This means that it _is_ possible to mutate the initial state before any actions are dispatched.  Not sure if the mutation checker picks up on it or not.

Another issue is that I'm looking at updating the immutability and serialization checks to skip past frozen values for speed. If the initial state isn't frozen, we don't get the benefit there. Now, it's also likely that the initial state is almost empty so it won't take long to scan those values anyway, but I can't see a reason to _not_ freeze the initial state.

Fortunately, it looks like it's trivial to do this by adding:

```js
const frozenInitialState = createNextState(initialState, draft => {})
```

and using the frozen state instead.

markerikson:
This PR:

- Updates `createReducer` to freeze the initial state by running it through Immer

Fixes #939

### DIFF
--- a/src/createReducer.test.ts
+++ b/src/createReducer.test.ts
@@ -90,6 +90,16 @@ describe('createReducer', () => {
         'Cannot add property text, object is not extensible'
       )
     })
+
+    test('Freezes initial state', () => {
+      const initialState = [{ text: 'Buy milk' }]
+      const todosReducer = createReducer(initialState, {})
+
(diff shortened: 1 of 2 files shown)


markerikson:
Looks good except for TS Next checks

Issue #799, opened 2020-11-10
More advanced access patterns for `createEntityAdapter`?


lukeramsden:
Thank you for your work on redux-toolkit, it really is an amazing library.

I'm using `createEntityAdapter` to keep my data normalised and it's working very nicely, however, my use-case means not only do I need to access by ID, but also other fields. I'm currently having to maintain my own "indices" based on various fields of my entities.

My entity looks like this:

```ts
interface MyEntity {
	// primary key, handled nicely by createEntityAdapter
	id: string;
	// also handled by createEntityAdapter via sortComparer
	createdAt: string;

	//...content stuff, not relevant

	// the juicy bit
	parent: string | null;
}
```

I need to be able to quickly access a list of entities in one of 2 ways:

1. Where `parent` is `null`, for the index of top level entities with no parent
2. Where `parent` is equal to `x`, for the children of a certain entity

I'm currently maintaining my own two indices for this.
```ts
interface EntitiesState extends AsyncFetchReduxState {
  idsWhereParentIsNulll: EntityId[];
  idsByParent: Record<EntityId, EntityId[]>;
}
```

I then have to make sure these are both kept in sync wherever I'm updating the adapter state - not to mention keeping them sorted, which I haven't got round to yet.

```ts
// root entities, pretty simple
builder.addCase(fetchEntitiesWithNoParent.fulfilled, (state, action) => {
  state.status = 'succeeded';
  entitiesAdapter.upsertMany(state, action.payload);

  for (const entity of action.payload) {
    if (!state.idsWhereParentIsNulll.includes(entity.id)) {
      state.idsWhereParentIsNulll.push(entity.id);
    }
  }
});

// child entities, a little more complex
builder.addCase(fetchStoriesByParent.fulfilled, (state, action) => {
  entitiesAdapter.upsertMany(state, action.payload);

  const parent = action.meta.arg;
  const parentArrayExists = parent in state.idsByParent;
  if (!parentArrayExists) {
    state.idsByParent[parent] = [];
  }
  for (const entity of action.payload) {
    // push entity ID in to array
    const entityAlreadyInHashmap = state.idsByParent[parent].includes(
      entity.id,
    );

    if (!entityAlreadyInHashmap) {
      state.idsByParent[parent].push(entity.id);
    }
  }
});
```

[...]

PR #948, opened 2021-03-28
[DRAFT] Experiment: Add "indices" to entity adapter

markerikson:
Inspired by #799 , I'm trying to add some "indices" to `createEntityAdapter`. The idea is that you could define secondary sorting options beyond the standard sorting for `state.ids`, and the entity adapter will automatically keep those updated.

I initially was looking at `redux-indexers`, but decided that it wasn't quite a good fit.

I'm now playing with the idea of having the sorting get done inside of the `merge()` function in `sorted_state_adapter`.

Was having trouble getting the fields of the `indices` option to carry through, but @msutkowski just got that working in #946 .

Still got a lot of experimenting left to do.

### DIFF
--- a/etc/redux-toolkit.api.md
+++ b/etc/redux-toolkit.api.md
@@ -144,10 +144,7 @@ export function createAsyncThunk<Returned, ThunkArg = void, ThunkApiConfig exten
 export const createDraftSafeSelector: typeof createSelector;
 
 // @public (undocumented)
-export function createEntityAdapter<T>(options?: {
-    selectId?: IdSelector<T>;
-    sortComparer?: false | Comparer<T>;
-}): EntityAdapter<T>;
+export function createEntityAdapter<T>(options?: EntityDefinition<T>): EntityAdapter<T>;
 
 // @public
 export function createImmutableStateInvariantMiddleware(options?: ImmutableStateInvariantMiddlewareOptions): Middleware;
@@ -230,6 +227,8 @@ export interface EntityState<T> {
     entities: Dictionary<T>;
     // (undocumented)
     ids: EntityId[];
+    // (undocumented)
+    indices: Indices<T, IndexComparers<T>>;
 }
 
 // @public (undocumented)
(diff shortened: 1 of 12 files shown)


markerikson:
Got something kinda working here. Types still aren't right yet - it's showing fields in entityState.indices that don't actually exist. But the logic seems like it's mostly there.

sabarnix:
> ```javascript
> indices: {
>   usersOver40ByAge: {
>     sortComparer: (a, b) => a.age - b.age,
>     filter: user => user.age >= 40
>   }
> }
>```

I have a concerns with this. How to handle cases where filtering is dependent on some other key in state or url param. Let's say 40 (the filtering age) is user input and sort order is stored in query params. Another problem would be to determine when the dependent param has changed.

markerikson:
@sabarnix I think at that point you'd need to do the remaining filtering outside the adapter.

Assuming this is being used as a slice reducer, you only have access to the current slice state + action anyway, and at this point in the code, you don't even have access to the rest of the slice state.  So, _if_ we decide to support any filtering at all, I think it's going to be limited to just what you can do based on a single item itself.

dutzi:
Hey @markerikson, we have a use case for this: entities are "Segments" of shape `{ startTime: number; endTime: number }` that need to be stored sorted by both `startTime` and `endTime`.

Currently we use a selector for getting them sorted by `endTime`, and let RTK manage `startTime` sorting, but that means we re-sort the entire array, on updates (instead of pushing the updated items into a sorted array, efficiently)

I've been reading through this PR, to see if we could help, but didn't find any actionable items. 

Can you please give guidance, so we could help make this happen?

markerikson:
@dutzi tbh this is entirely backburnered right now.  We're focusing on finalizing RTK 1.9, and after that we'll be working on RTK 2.0 that drops back compat for IE11 and such.

The biggest issue here is that there isn't a good definition or requirements for what the desired behavior should be, and that also means that there isn't a valid implementation either.  This was a bit of an experiment, but definitely not ready for actual use.

If someone wants to think this topic through further and propose some API designs and behaviors, we can talk about it, and I'm definitely open to PRs.  But otherwise I don't see this happening any time soon.

dutzi:
Got it. Thanks for the quick response!

PR #949, opened 2021-03-28
fix PreloadedState type for TS 4.3

phryneas:


phryneas:
@markerikson this should fix our TS 4.3 problems & also improve the `preloadedState` typings - can you take a look please?

markerikson (approved):

PR #952, opened 2021-03-28
Extract common utils from both adapter files

markerikson:
This PR:

- Dedupes some of the logic in the `unsorted_state_adapter` / `sorted_state_adapter` files by extracting some common utils
- Fixes a leftover stray bit of commented out code in the serializability middleware


markerikson:
Amusing to note that switching `for..of` loops to `for(let i = 0` loops does drop the size a bit (-94b min+gz).  Will skip merging that, though - it'll be taken care of by creating a "modern JS" build.

PR #957, opened 2021-03-29
chore(build): move tsdx to esbuild

hardfist:
just try  what esbuild cann't be done for author libraries

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -1,5 +1,6 @@
 // @ts-check
 const { build, transform } = require('esbuild')
+const terser = require('terser')
 const rollup = require('rollup')
 const path = require('path')
 const fs = require('fs-extra')
@@ -75,9 +76,18 @@ async function bundle(options) {
     let code = result.outputText
     let mapping = mergedSourcemap
     if (minify) {
-      const transformResult = await transform(code, {
-        target: 'es5',
-        minify: true,
+      const transformResult = await terser.minify(code, {
+        sourceMap: true,
+        output: {
+          comments: false,
+        },
+        compress: {
+          keep_infinity: true,
+          pure_getters: true,
+          passes: 10,
+        },
+        ecma: 5,
+        toplevel: true,
       })
       code = transformResult.code
       mapping = transformResult.map
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -21,9 +21,12 @@ async function bundle(options) {
     sourcemap: 'inline',
     bundle: true,
     format: format === 'umd' ? 'esm' : format,
-    define: {
-      'process.env.NODE_ENV': JSON.stringify(env),
-    },
+    define:
+      format === 'esm'
+        ? {}
+        : {
+            'process.env.NODE_ENV': JSON.stringify(env),
+          },
     plugins: [
       {
         name: 'node_module_external',
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -11,21 +11,20 @@ const { extractInlineSourcemap, removeInlineSourceMap } = require('./sourcemap')
 const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms))
 const outputDir = path.join(__dirname, '../dist')
 async function bundle(options) {
-  const { format, minify, env, name } = options
+  const { format, minify, env, name, target } = options
   const result = await build({
     entryPoints: ['src/index.ts'],
     outfile: `dist/redux-toolkit${name}.js`,
     write: false,
-    target: 'es2015',
+    target: target ? target : 'es2015',
     sourcemap: 'inline',
     bundle: true,
     format: format === 'umd' ? 'esm' : format,
-    define:
-      format === 'esm'
-        ? {}
-        : {
-            'process.env.NODE_ENV': JSON.stringify(env),
-          },
+    define: env
+      ? {
+          'process.env.NODE_ENV': JSON.stringify(env),
+        }
+      : {},
     plugins: [
       {
         name: 'node_module_external',
@@ -68,12 +67,14 @@ async function bundle(options) {
         sourceMap: true,
         module:
           format !== 'cjs' ? ts.ModuleKind.ES2015 : ts.ModuleKind.CommonJS,
-        target:
-          name === '.modern' ? ts.ScriptTarget.ES2017 : ts.ScriptTarget.ES5,
+        target: name.includes('.modern')
+          ? ts.ScriptTarget.ES2017
+          : ts.ScriptTarget.ES5,
       },
     })
     const mergedSourcemap = merge(sourcemap, result.sourceMapText)
     let code = result.outputText
+    // TODO Is this used at all?
     let mapping = mergedSourcemap
     if (minify) {
       const transformResult = await terser.minify(code, {
@@ -152,24 +153,23 @@ async function main() {
     {
       format: 'cjs',
       name: '.cjs.development',
-      target: 'es2017',
       minify: false,
       env: 'development',
     },
     {
       format: 'cjs',
       name: '.cjs.production.min',
-      target: 'es2017',
       minify: true,
       env: 'production',
     },
+    // ESM, embedded `process`, ES5 syntax: typical Webpack dev
     {
       format: 'esm',
       name: '.esm',
-      target: 'es2017',
       minify: false,
       env: '',
     },
+    // ESM, embedded `process`, ES2017 syntax: modern Webpack dev
     {
       format: 'esm',
       name: '.modern',
@@ -177,25 +177,41 @@ async function main() {
       minify: false,
       env: '',
     },
+    // ESM, pre-compiled "dev", ES2017 syntax: browser development
+    {
+      format: 'esm',
+      name: '.modern.development',
+      target: 'es2017',
+      minify: false,
+      env: 'development',
+    },
+    // ESM, pre-compiled "prod", ES2017 syntax: browser prod
+    {
+      format: 'esm',
+      name: '.modern.production.min',
+      target: 'es2017',
+      minify: true,
+      env: 'production',
+    },
     {
       format: 'umd',
       name: '.umd',
-      target: 'es2017',
       minify: false,
       env: 'development',
     },
     {
       format: 'umd',
       name: '.umd.min',
-      target: 'es2017',
       minify: true,
       env: 'production',
     },
   ]
-  for (const options of buildTargets) {
-    bundle(options)
-  }
-  await sleep(3000) // hack, waiting file to save
+
+  // Run builds in parallel
+  const bundlePromises = buildTargets.map((options) => bundle(options))
+  await Promise.all(bundlePromises)
+
+  await sleep(500) // hack, waiting file to save
   await buildUMD()
   writeEntry()
   // addSubpath()
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -1,0 +1,148 @@
+const { build } = require('esbuild')
+const rollup = require('rollup')
+const path = require('path')
+const fs = require('fs-extra')
+const ts = require('typescript')
+const { fromJSON } = require('convert-source-map')
+const merge = require('merge-source-map')
+const { extractInlineSourcemap, removeInlineSourceMap } = require('./sourcemap')
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms))
+async function bundle(env, format) {
+  let envName = env
+  if (env === 'production') {
+    envName = 'production.'
+  } else if (format === 'cjs') {
+    envName = 'development.'
+  } else {
+    envName = ''
+  }
+  const result = await build({
+    entryPoints: ['src/index.ts'],
+    outfile: `dist/redux-toolkit.${format}.${envName}js`,
+    write: false,
+    target: 'es2015',
+    minify: env === 'production',
+    sourcemap: 'inline',
+    bundle: true,
+    format: format === 'umd' ? 'esm' : format,
+    define: {
+      'process.env.NODE_ENV': JSON.stringify(env),
+    },
+    plugins: [
+      {
+        name: 'node_module_external',
+        setup(build) {
+          build.onResolve({ filter: /.*/ }, (args) => {
+            if (format === 'umd') {
+              return
+            }
+            if (args.path.startsWith('.') || args.path.startsWith('/')) {
+              return undefined
+            } else {
+              return {
+                path: args.path,
+                external: true,
+              }
+            }
+          })
+          build.onLoad({ filter: /getDefaultMiddleware/ }, async (args) => {
+            if (env !== 'production' || format !== 'umd') {
+              return
+            }
+            const source = await fs.readFile(args.path, 'utf-8')
+            const defaultPattern = /\/\* PROD_START_REMOVE_UMD[\s\S]*?\/\* PROD_STOP_REMOVE_UMD \*\//g
+            const code = source.replace(defaultPattern, '')
+            return {
+              contents: code,
+              loader: 'ts',
+            }
+          })
+        },
+      },
+    ],
+  })
+
+  for (const chunk of result.outputFiles) {
+    const origin = chunk.text
+    const sourcemap = extractInlineSourcemap(origin)
+    const code = ts.transpileModule(removeInlineSourceMap(origin), {
+      compilerOptions: {
+        sourceMap: true,
+        module:
+          format === 'umd' ? ts.ModuleKind.ES2015 : ts.ModuleKind.CommonJS,
+        target: ts.ScriptTarget.ES5,
+      },
+    })
+    const mergedSourcemap = merge(sourcemap, code.sourceMapText)
+
+    await fs.writeFile(chunk.path, code.outputText)
+    await fs.writeJSON(chunk.path + '.map', mergedSourcemap)
+  }
+}
+/**
+ * since esbuild doesn't support umd, we use rollup to convert esm to umd
+ */
+async function buildUMD() {
+  // origin
+  const input = path.join(__dirname, '../dist/redux-toolkit.umd.js')
+  const instance = await rollup.rollup({
+    input: [input],
+    onwarn(warning, warn) {
+      if (warning.code === 'THIS_IS_UNDEFINED') return
+      warn(warning) // this requires Rollup 0.46
+    },
+  })
+  await instance.write({
+    format: 'umd',
+    name: 'redux-toolkit',
+    file: 'dist/redux-toolkit.umd.js',
+    sourcemap: true,
+  })
+  // minify
+  const input2 = path.join(__dirname, '../dist/redux-toolkit.umd.production.js')
+
+  const instance2 = await rollup.rollup({
+    input: [input2],
+    onwarn(warning, warn) {
+      if (warning.code === 'THIS_IS_UNDEFINED') return
+      warn(warning) // this requires Rollup 0.46
+    },
+  })
+  await instance2.write({
+    format: 'umd',
+    name: 'redux-toolkit',
+    file: 'dist/redux-toolkit.umd.min.js',
+    sourcemap: true,
+  })
+  try {
+    await fs.unlinkSync(input2)
+    await fs.unlinkSync(input2 + '.map')
+  } catch (err) {
+    // just ignore
+  }
+}
+async function writeEntry() {
+  await fs.writeFile(
+    'dist/index.js',
+    `'use strict'
+if (process.env.NODE_ENV === 'production') {
+  module.exports = require('./redux-toolkit.cjs.production.min.js')
+} else {
+  module.exports = require('./redux-toolkit.cjs.development.js')
+}`
+  )
+}
+async function addSubpath() {}
+async function main() {
+  for (const format of ['cjs', 'esm', 'umd']) {
+    for (const env of ['development', 'production']) {
+      bundle(env, format)
+    }
+  }
+  await sleep(2000) // hack, waiting file to save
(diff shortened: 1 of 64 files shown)


markerikson:
Hey, thanks for putting this together.

Unfortunately, we do still currently include a UMD build in our published packages, and I don't think ESBuild supports that at all.

hardfist:
> Hey, thanks for putting this together.
> 
> Unfortunately, we do still currently include a UMD build in our published packages, and I don't think ESBuild supports that at all.

done!, I just need to convert bundled esm to umd by rollup

markerikson:
So there's a couple other nuances to be aware of with our UMD builds.

We currently output both a dev and prod UMD build.

Not only is the prod UMD build minified, but we need to strip out a couple bits of code from `getDefaultMiddleware.ts` that are dev-mode only.  Currently, we use a `rollup-plugin-strip-code` to do that.

hardfist:
> So there's a couple other nuances to be aware of with our UMD builds.
> 
> We currently output both a dev and prod UMD build.
> 
> Not only is the prod UMD build minified, but we need to strip out a couple bits of code from `getDefaultMiddleware.ts` that are dev-mode only. Currently, we use a `rollup-plugin-strip-code` to do that.

only form umd production ? what about esm&cjs production?

markerikson:
Uh... just UMD, I think? :)  It's `getDefaultMiddleware.ts`, specifically.

That file's gone through a few different changes over time - you can see the changes in the git history.

hardfist:
> Uh... just UMD, I think? :) It's `getDefaultMiddleware.ts`, specifically.
> 
> That file's gone through a few different changes over time - you can see the changes in the git history.

I use esbuild plugin to handle this, I think it's ok now, but I do not know how to test

scripts/build.js
```diff
@@ -0,0 +1,102 @@
...
+            }
+          })
+          build.onLoad({ filter: /getDefaultMiddleware/ }, async (args) => {
+            if (env !== 'production' || format !== 'umd') {
+              return
+            }
+            const source = await fs.readFileSync(args.path, 'utf-8')
+            const defaultPattern = /\/\* PROD_START_REMOVE_UMD[\s\S]*?\/\* PROD_STOP_REMOVE_UMD \*\//g
+            const code = source.replace(defaultPattern, '')
+            // if (sourceMap !== false && sourcemap !== false) {
```
hardfist, line 46: need to handle sourcemap

markerikson:
SO FYI, I _think_ the "replace chunk of code" behavior could be accomplished using ESBuild's "define" replacement capability.

Specifically, if we set up the definitions such that `PROD_START_REMOVE_UMD */` and `/* PROD_STOP_REMOVE_UMD` were replaced with nothing, that would leave these code blocks commented out.

Also: it looks like the use of `tsc` is outputting a bunch of individual JS files to `dist`.  Playing with your branch locally, I think we want: 

`"emitDeclarationOnly": true,`

markerikson:
So there's one other point of concern here.

I believe ESBuild only outputs "modern" JS ( https://github.com/evanw/esbuild/issues/297 ).  Right now, our CommonJS and ESM builds are back-compiled to cover ES5 and IE11, as far as I know. 

I'm all in favor of dropping IE11/ES5-only support in general, but I'm not sure we can pull the trigger on that just yet. I don't know what the right semver rules are for that sort of thing.

hardfist:
i check the code length to ensure works，and check the require result of umd bundle

hardfist:
> So there's one other point of concern here.
> 
> I believe ESBuild only outputs "modern" JS ( [evanw/esbuild#297](https://github.com/evanw/esbuild/issues/297) ). Right now, our CommonJS and ESM builds are back-compiled to cover ES5 and IE11, as far as I know.
> 
> I'm all in favor of dropping IE11/ES5-only support in general, but I'm not sure we can pull the trigger on that just yet. I don't know what the right semver rules are for that sort of thing.

i normally using ts transpilemodule to transform esbuild bundle result to target es5，if you prefer i can change this

markerikson:
Yeah, that was gonna be my next train of thought - transpiling the built output itself in order to enable running okay in IE11, vs transpiling the _input_ files straight to ES5.

So here's two other things that would be useful to sort out.

First, right now TSDX is generating three files for CJS: `redux-toolkit.cjs.develeopment.js`, `redux-toolkit.cjs.production.js`, and an `index.js` file that conditionally imports one of those two:

https://unpkg.com/browse/@reduxjs/toolkit@1.5.1/dist/

Right now this PR is generating the dev and prod files, but the dev file is missing the `.development` portion of the name, and we don't have the `index.js` file being generated.

Also: part of the reason why we're wanting to revamp the build system is that we're going to need to support additional entry points.  In particular, we're going to be adding support for:

```js
import {thing1} from "@reduxjs/toolkit/query"
import {thing2} from "@reduxjs/toolkit/query/react"
```

Any chance you could put together an example of how that might work here?

btw, really do appreciate you working with me on fleshing out this idea :)  I was skeptical at first, but it does seem like this could work.

hardfist:
add additional entry points seems to need use NodeJS 12.7+ package exports, which typescript does not supports yet, https://github.com/microsoft/TypeScript/issues/33079, and package subpath exports is actually an breaking change

markerikson:
Hmm. We seem to have something working for multiple entry points over in the RTK Query repo, although admittedly we haven't published a real version with that build config - just the auto builds per PR.

hardfist:
add index.js entry points, down-level to es5, remove extra .js files

hardfist:
> Hmm. We seem to have something working for multiple entry points over in the RTK Query repo, although admittedly we haven't published a real version with that build config - just the auto builds per PR.

I check the RTK query docs, didn't find any docs about additional entry points, could you give me an example, so I can do the same things here.

scripts/build.js
```diff
@@ -0,0 +1,131 @@
...
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms))
+async function bundle(env, format) {
+  const result = await build({
+    entryPoints: ['src/index.ts'],
+    outfile: `dist/redux-toolkit.${format}.${
+      env === 'production' ? 'production.' : ''
+    }js`,
+    write: false,
+    target: 'es2015',
+    minify: env === 'production',
```
hardfist, line 15: todo sourcemap merge

markerikson:
The ultimate plan here is to have three entry points:

- `"@reduxjs/toolkit"`: standard RTK APIs ( `configureStore, createSlice` )
- `"@reduxjs/toolkit/query"`: RTK Query UI-agnostic version of `createApi` plus utils
- `"@reduxjs/toolkit/query/react"`: React-specific version of `createApi` that auto-generates React hooks too

You can see some of that config here:

https://github.com/rtk-incubator/rtk-query/blob/783a2ab7bfa6c966bffa86e17fe55a7b2f250c7e/rollup.config.js

and here's a CI build from https://github.com/rtk-incubator/rtk-query/pull/177 so you can see the rough build output:

[rtk-query-pr-177.zip](https://github.com/reduxjs/redux-toolkit/files/6219572/rtk-query-pr-177.zip)

We'll need _something_ sorta like that in order to release 1.6 the way we want to, but we're open to exactly how that gets done.

hardfist:
> The ultimate plan here is to have three entry points:
> 
> * `"@reduxjs/toolkit"`: standard RTK APIs ( `configureStore, createSlice` )
> * `"@reduxjs/toolkit/query"`: RTK Query UI-agnostic version of `createApi` plus utils
> * `"@reduxjs/toolkit/query/react"`: React-specific version of `createApi` that auto-generates React hooks too
> 
> You can see some of that config here:
> 
> https://github.com/rtk-incubator/rtk-query/blob/783a2ab7bfa6c966bffa86e17fe55a7b2f250c7e/rollup.config.js
> 
> and here's a CI build from [rtk-incubator/rtk-query#177](https://github.com/rtk-incubator/rtk-query/pull/177) so you can see the rough build output:
> 
> [rtk-query-pr-177.zip](https://github.com/reduxjs/redux-toolkit/files/6219572/rtk-query-pr-177.zip)
> 
> We'll need _something_ sorta like that in order to release 1.6 the way we want to, but we're open to exactly how that gets done.

put an package in another package ,clever idea, I will try this method

hardfist:
@markerikson it seems that query && query/react is in another repo, then How should I put the repo code here, so you want to use monorepo to put rtk/query code into RTK?

markerikson:
We'll be migrating the RTK Query code over into this repo in the near future, but it's not ready yet. For now, could you temporarily just add a couple dummy source files and use those as the source for two new subpackages, just to see if we can get this to work?

Ideally, one subpackage would export a plain JS function, and the other subpackage would export both that first function and a custom React hook, because that is what we'll be doing later.

hardfist:
sourcemap merging is ok now(except the umd version), you can test it here, https://evanw.github.io/source-map-visualization/

hardfist:
@markerikson  , it seems working now
image

markerikson:
That is COOL! Excellent work!

I'm very happy to be proven wrong here :)

Busy with work atm, but I will try to pull this branch and evaluate the output in more detail over the next couple days.

Mmm... at some point we'd need to:

- completely remove all mentions of TSDX
- switch out the `build` run scripts to use this instead
- set up new `test` and `lint` commands that call Jest and ESLint directly
- solve whatever the other CI build failures are here.  (Looks like that may be `tsdx lint` trying to treat the `package.json` files as `.js` files, maybe?)

markerikson:
Okay, trying to do some further analysis on this.



First check: making sure the code is IE11-safe.  I used this ESLint config to run checks against the files in `dist`:

```js
// dist/.eslintrc.js
module.exports = {
  "extends": ["plugin:compat/recommended"],
  parserOptions: {
    ecmaVersion: 5
  },
  "env": {
    "browser": true
  },
  "rules": {
    "compat/compat": ["warn", "ie 11"]
  }
}
```

Combination of https://github.com/amilajack/eslint-plugin-compat to detect non-IE11-compatible APIs like Promises, and turning down the parser syntax version to fail on syntax issues.  I manually added some syntax like `function* myGenerator() {}` into a couple files to confirm that it found them.

All the files in the `dist` folder currently pass, which is great!

Having said that, we're going to want to start shipping a "modern JS" build here, so at least one of these is going to need to _not_ be backwards-transpiled.  We'll have to fiddle with the setup to enable that. But hey, back-compat is a great starting point :)

Also, I note that we're currently relying on `rollup` being there implicitly, which it is thanks to TSDX.  We'll have to add that explicitly once we remove TSDX (as I found out when I removed TSDX locally.)

Got a couple more things I want to poke at here, but looking great so far!

markerikson:
More observations:

- CJS is `redux-toolkit.cjs.production.js`, but being imported in `index.js` as `redux-toolkit.cjs.production.min.js` (missing `.min`)
- We didn't ship an `esm.production` build previously, under the assumption the dev build gets bundled and processed at build time
- The ESM build is actually supposed to keep the ESM `import/export` syntax - it just needs to get compiled down to ES5 compat syntax _other_ than that. It shouldn't be compiled into CommonJS
- The UMD files grew noticeable. UMD Dev went 97K -> 104K, and UMD Prod went 33K -> 48K. This isn't _good_ per se, but on the other hand I have to assume that anyone using our UMD builds isn't overly concerned about size - they're probably just loading a couple scripts.  Still, it would be good if we knew _why_ they got bigger and could shrink them back down.
- Previously we were generating a single combined `typings.d.ts` file that was 65K and had all of our types. Per the `files` list in `package.json`, that's what gets included when we publish. Can we get that back?

hardfist:
so the esm version will keep process.env.NODE_ENV untouched ?

markerikson:
Yep.  If I look at the ESM build from `master`, that file definitely still has the `process.env.NODE_ENV` checks inside.

hardfist:
it seems we still need two version of esm , esm target es5 and esm target es2017 ?

markerikson:
Yeah, the second one would be the "modern" build I'm referring to.

hardfist:
> More observations:
> 
> * CJS is `redux-toolkit.cjs.production.js`, but being imported in `index.js` as `redux-toolkit.cjs.production.min.js` (missing `.min`)
> * We didn't ship an `esm.production` build previously, under the assumption the dev build gets bundled and processed at build time
> * The ESM build is actually supposed to keep the ESM `import/export` syntax - it just needs to get compiled down to ES5 compat syntax _other_ than that. It shouldn't be compiled into CommonJS
> * The UMD files grew noticeable. UMD Dev went 97K -> 104K, and UMD Prod went 33K -> 48K. This isn't _good_ per se, but on the other hand I have to assume that anyone using our UMD builds isn't overly concerned about size - they're probably just loading a couple scripts.  Still, it would be good if we knew _why_ they got bigger and could shrink them back down.
> * Previously we were generating a single combined `typings.d.ts` file that was 65K and had all of our types. Per the `files` list in `package.json`, that's what gets included when we publish. Can we get that back?
> * I'd like to add an `.modern.js` build, per https://web.dev/publish-modern-javascript/
> 
> Perhaps rather than having a pair of nested `for` loops, it would be better to have an array of builds? Something like:
> 
> ```js
> const buildTargets = [
>   {format: 'cjs', name: '.cjs.development', target: 'es5', minify: false},
>   {format: 'cjs', name: '.cjs.production.min', target: 'es5', minify: true},
>   {format: 'esm', name: '.modern', target: 'es2017', minify: false},
>   // etc
> ]
> ```

except the umd size problem, other problem seems solved(I can not see the typings problem in my repo)

markerikson:
Yeah, I think this may almost be there. Any idea why the `cjs.production` build grew from 16K previously to 27K now?

ref: https://unpkg.com/browse/@reduxjs/toolkit@1.5.1/dist/

hardfist:
because tsc transform the minify result, I can fix this

hardfist:
It's so bad typescript does not support minify yet

hardfist:
@markerikson  because terser do some aggressive DCE than esbuild's minify, after switch minifier to terser, the size nears the same

phryneas:
> * Previously we were generating a single combined `typings.d.ts` file that was 65K and had all of our types. Per the `files` list in `package.json`, that's what gets included when we publish. Can we get that back?

With the tooling having moved forward and also `import type` now being a thing, that is not strictly necessary any more.

The only left benefit I can think of we currently have from the combined type build is that people cannot just import our internal types so they don't rely on implementation details. 
On the opposite side, people are constantly going to be asking for us to export type X and Y.

So by now, I'd be fine if we don't have that combined type file, because it's a hassle to get there and the tooling is still buggy to this day.

markerikson:
I'd like to try to keep the build output as similar as possible in this PR, if only because the plan is to use this for our next 1.6 minor version.

If we feel the combined types file isn't useful, we can look at dropping that in 2.0.

markerikson:
@hardfist that last "minify terser" commit accidentally included a couple CJS build artifacts in the repo root. I just amended that commit to remove them and force-pushed it up.

markerikson:
Been agonizing over the right combination of build formats to output.

Here's where I'm landing:

| Filename                               | Module | Syntax | `process.env` | Purpose                |
|----------------------------------------|--------|--------|---------------|------------------------|
| redux-toolkit.cjs.development.js       | CJS    | ES5    | 'development' | Node / dev             |
| redux-toolkit.cjs.production.min.js    | CJS    | ES5    | 'production'  | Node / prod            |
| redux-toolkit.esm.js                   | ESM    | ES5    | Embedded      | Bundler, legacy syntax |
| redux-toolkit.modern.development.js    | ESM    | ES2017 | 'development' | Browser module, dev    |
| redux-toolkit.modern.js                | ESM    | ES2017 | Embedded      | Bundler, modern syntax |
| redux-toolkit.modern.production.min.js | ESM    | ES2017 | 'production'  | Browser module, prod   |
| redux-toolkit.umd.js                   | UMD    | ES5    | 'development' | Browser script, dev    |
| redux-toolkit.umd.min.js               | UMD    | ES5    | 'production'  | Browser script, prod   |

I think this covers all our bases.  (There's also `index.js`, which dynamically requires either CJS dev/prod depending on `process.env`.)

From that, based on https://web.dev/publish-modern-javascript/ and https://github.com/developit/microbundle, I believe our `package.json` should be:

```js
{
  "main": "dist/index.js",            // CommonJS bundle
  "unpkg": "dist/redux-toolkit.umd.js",    // UMD bundle
  "module": "dist/redux-toolkit.esm.js",        // ES Modules bundle
  "esmodule": "dist/redux-toolkit.modern.js", // Modern bundle
  "types": "dist/types.d.ts"          // TypeScript typings
}
```

hardfist:
yes, exports definitely will break code under nodejs, but I'm little confused, is it really necesary to split file to `index.js|xxx.development|xxx.production` other than just ship an `index.js` which contains development and production code, since bundler will eventually treeshake the unused code according to process.env.NODE_ENV, and node.js project seems don't care so much about the size

markerikson:
So in theory, bundlers should be picking up the `.esm` file, with its `process.env` values still embedded. CJS is a fallback for Node.

The `.modern` build is ESM, but with modern ES2017 syntax.

However, per #652 , files with `process.env` don't work in a browser.  So, I added `.modern.development` and `.modern.production` files to support loading those in a browser directly (via Unpkg or similar).

markerikson:
@hardfist okay, I was able to install the CSB-built package into a local CRA project, but there's something buggy about the imports/exports in `redux-toolkit.esm.js`:

```
Failed to compile.

./node_modules/@reduxjs/toolkit/dist/redux-toolkit.esm.js
Attempted import error: 'Draft' is not exported from 'immer'.
```

I see _two_ sets of exports in that file.  One is around like 75, like this:

```js

import { enableES5 } from "immer";
export * from "redux";
import { default as default2, Draft as Draft2, current as current2, freeze, original, isDraft as isDraft4 } from "immer";
import { createSelector as createSelector2, Selector, OutputParametricSelector, OutputSelector, ParametricSelector } from "reselect";
// src/createDraftSafeSelector.ts
import { current, isDraft } from "immer";
import { createSelector } from "reselect";
```

Notice that a bunch of those are actually TS types, and this file should not be exporting those.

then, at the bottom, there's:

```js
export { Draft2 as Draft, MiddlewareArray, OutputParametricSelector, OutputSelector, ParametricSelector, Selector, ThunkAction, ThunkDispatch, configureStore, createAction, createAsyncThunk, createDraftSafeSelector, createEntityAdapter, createImmutableStateInvariantMiddleware, default2 as createNextState, createReducer, createSelector2 as createSelector, createSerializableStateInvariantMiddleware, createSlice, current2 as current, findNonSerializableValue, freeze, getDefaultMiddleware, getType, isAllOf, isAnyOf, isAsyncThunkAction, isDraft4 as isDraft, isFulfilled, isImmutableDefault, isPending, isPlain, isPlainObject, isRejected, isRejectedWithValue, nanoid, original, unwrapResult };
```

which is A) duplicate, and B) still exporting a bunch of TS types that should not be exported.

It's getting late here and I gotta head to bed. Can you figure out what's going on there and fix it? Thank you!

hardfist:
> #652

haha,react doesn't support esm yet, so the bundle actually use cjs version of react, the esm split make sense to me now, since to support bundleness scenario

hardfist:
I will check it , I just get up

markerikson:
I DEMAND THAT YOU FIX THIS NOW, BEFORE BREAKFAST!!!

:)

(seriously, again, appreciate all your efforts on this!)

hardfist:
It's kind of tricky here, since esbuild does not support shakes implicit type import|export, so we can change all type export to explicit type export or waiting esbuild to support this feature.

markerikson:
I'm fine with changing whatever type import/export declarations we need to make this work - please go ahead!

hardfist:
woops, it seems that api-extractor doesn't supports type export yet

hardfist:
I currently disable api-extractor and ship all dts files instead, and it passes the redux-typescript cra template test. api-extractor seems needs some hack.

hardfist:
after I upgrade api-extractor to latest version, it works now

hardfist:
Oh, I don't know you have to support typescript version below 3.8 which doesn't support type export|import, can we drop support for this typescript version ?

phryneas:
Given that by now 4.2 is out, that would still fall under "support the last 5 TS versions", so I'd be fine with it.

hardfist:
It's easy just like before, all we need to do is to use babel with the previous plugin to transform the bundled result, but I don't like this babel trick actually , I would prefer follow what typescript handle  error message

markerikson:
Yeah, not sure how you'd test that, but seems like a good idea. Anything we can do to better verify the contents of the prod builds is a good thing.

markerikson:
Awright. At this point we have:

- All of our original artifacts, approximately the same size
- The additional "modern" artifacts
- Sourcemaps
- Linting + formatting + tests
- Examples of generating subpackages
- Dropping TS < 3.8
- Updated GHA+CSB to Node 14

I've installed the CSB-built package locally into a CRA app and it runs, and I see the latest built `.esm` file in the DevTools source debugger.

Got at least one thumbs-up on Twitter that the CSB build runs.

I think this PR is at least good to merge in, and we can do any follow-up work in additional PRs if needed.

I'm going to go ahead and merge this.

@hardfist , THANK YOU for your hard work putting this together!

scripts/build.js
```diff
@@ -0,0 +1,220 @@
...
+      env: 'production',
+    },
+    // ESM, embedded `process`, ES5 syntax: typical Webpack dev
+    {
+      format: 'esm',
+      name: '.esm',
+      minify: false,
+      env: '',
+    },
+    // ESM, embedded `process`, ES2017 syntax: modern Webpack dev
```
Andarist, line 172: How a webpack user can use this bundle?

PR #959, opened 2021-03-29
Add check for undefined middleware in configureStore

### LINKED ISSUE #937: Check for undefined middleware on startup
It's possible for a user to return `undefined` from the `middleware` creation callback in `configureStore`, in which case there are no middleware at all. We should consider this a user error and throw when that happens.  (ref: https://stackoverflow.com/q/66781880/62937 )

Shrugsy:
Closes #937 

As per commit message:
```
* Throw an error when a middleware creation callback
  is provided to configureStore that does not return
  an array of middleware
```

Two new tests are added:
- explicitly passing `middleware: undefined` should use default middleware (same as excluding the option)
- passing `middleware: () => undefined` should be considered a user error and throw an error

Note: While issue #937 arose due to a middleware creation function that returns `undefined`, I've considered `"returning anything besides an array"` as a user error in this context.

### DIFF
--- a/src/configureStore.test.ts
+++ b/src/configureStore.test.ts
@@ -98,6 +98,23 @@ describe('configureStore', () => {
     })
   })
 
+  describe('given a middleware creation function that returns an array with non-functions', () => {
+    it('throws an error', () => {
+      const invalidBuilder = jest.fn(getDefaultMiddleware => [true] as any)
+      expect(() =>
+        configureStore({ middleware: invalidBuilder, reducer })
+      ).toThrow('each middleware provided to configureStore must be a function')
+    })
+  })
+
+  describe('given custom middleware that contains non-functions', () => {
+    it('throws an error', () => {
+      expect(() =>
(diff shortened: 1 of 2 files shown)


markerikson:
Nice. Two requests:

- We can make the check be only in dev mode, I think
- We might as well also verify that all the values in the array are functions and not `undefined`

Shrugsy:
> Nice. Two requests:
> 
> * We can make the check be only in dev mode, I think
> * We might as well also verify that all the values in the array are functions and not `undefined`

I've added new commits to address these. For checking if each middleware in the array is actually a function, I've made it perform that check regardless of whether a `middleware creation callback` was used (i.e. passing a bad array to the `middleware` property directly will throw an error).

Question: Is there any kind of policy about squashing commits on pull requests, or will maintainers handle that?

markerikson:
We've got "Squash and Merge" turned on for all the Redux repos. tbh I'm not exactly a fan, but eh :)

Issue #953, opened 2021-03-28
Rework RTK build tooling for 1.6 release


markerikson:
There's several changes we need to make to our build tooling:

- Currently we're on TSDX 0.11, which is outdated. Also, we're using Babel 7.7.4. 
- However, TSDX apparently doesn't work well enough for generating multiple entry points, which is something we're going to need for RTK Query support. RTKQ's repo already is just using plain Rollup: https://github.com/rtk-incubator/rtk-query/blob/next/rollup.config.js , and we're anticipating needing to replace our current TSDX setup with that instead
- We don't actually have support for optional chaining turned on, as I found out when I tried to use it and that threw an error
- I'd like to see if we can start applying the principles from "ship modern JS" at https://web.dev/publish-modern-javascript/ .  In particular, https://web.dev/publish-modern-javascript/#modern-with-legacy-fallback-and-esm-bundler-optimizations looks like the combo of settings we would want to use

Dropping TSDX may also be a prerequisite for enabling shipping extraction of error messages into error codes in a prod build, per https://github.com/reduxjs/redux/pull/3920#issuecomment-804526725 .

Sorta related to this, I note that there's a few places in our codebase where we use `for(const x of y)`, which seems like it expands a bit when transpiled.  We _could_ try to micro-optimize those, or we could potentially just improve the build tooling and ship more modern JS that doesn't back-compile that syntax.

PR #962, opened 2021-03-30
Use esbuild for bundling

lindskogen:
This is an initial draft for a solution to #953.

esbuild doesn't seem to support UMD builds which could be a dealbreaker.﻿

Also, there's nothing like `rollup-plugin-strip-code` for esbuild yet.

### DIFF
--- a/build.ts
+++ b/build.ts
@@ -1,0 +1,42 @@
+const fsPromises = require('fs/promises')
+
+const esbuild = require('esbuild')
+
+const createBuild = ({ format, env, minify = false }) => {
+  return esbuild.build({
+    entryPoints: ['src/index.ts'],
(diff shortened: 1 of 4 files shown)


markerikson:
Hey there, thanks for the PR! Appreciate you filing this.  Unfortunately, there's actually already an in-progress PR to convert our build system over to ESBuild over in #957 , which is close to complete.  I was actually just about to review it again to see if there's anything left to be done.

Given that, I'm afraid this probably needs to be closed. Sorry about that!  However, I'd definitely appreciate any help you could offer with some of the other open issues we're tracking for our 1.6 milestone:

https://github.com/reduxjs/redux-toolkit/milestone/3

PR #970, opened 2021-04-01
Add simpler unwrapping of createAsyncThunk promises

### LINKED ISSUE #775: Allow to unwrap async thunks automatically
It makes sense that the result of async thunk is wrapped by default and points made in #618 are valid. However, users could be allowed to change this behaviour if they need to reduce verbosity:

```js
try {
  result = unwrapResult(await dispatch(asyncAction()));
  ...
```

The suggestion is to add an option that will apply `unwrapResult` automatically:
```js
let asyncAction = createAsyncThunk(
  'asyncAction',
  ...,
  { unwrap: true }
);
```

### LINKED ISSUE #910: Throw error on rejection
Hi, I'm migrating my application from redux-thunk + redux-promise-middleware to redux-toolkit.

I have a problem that I cannot solve that concerns throwing the error in an asynchronous action.

The main difference with redux-promise-middleware is that if an error is generated in an asynchronous action, it dispatches the failed action and at the same time does not stop the error throw.
On the redux-toolkit instead the action is only rejected.
Reading the various issues on redux-toolkit, I deduced that adding the serializeError option and throwing the error, I did also receive the error on the dispatch of the action, however the reject action is not performed.

I'm going to write some code to explain the context:

```
class PostContainer extends React.Component {
  public componentDidMount() {
    try {
      this.props.preload();
    } catch(err) {
      console.log(err);
      // i'd like to receive here the error
    }
  }
}

const mapDispatchToProps = (dispatch: any) => ({
  preload() {
    await dispatch(preload());
    dispatch(somethingelse())
  }
});

export default connect(null, mapDispatchToProps)(PostContainer);

// action
const preload = createAsyncThunk('post/preload', async (_: void, { dispatch }) => {
  // if one of this two action throw an error
  await dispatch(getPosts());
  await dispatch(getUsers());
});

expost const postSlice = createSlice({
  name: 'post',
  initialState,
  reducers: {},
  extraReducers: (builder) => {
    // i'd like to receive the rejected action
    builder.addCase(prealod.rejected, (state, action) => {
      return {
        ...state,
        isError: true
      }
    })
  }
})

````

Is there a way to achieve this with redux-toolkit?
Maybe should i use the `unwrapResult` utility?
If this is the case, how can i implement in this scenario?
Thanks for you time.

### PR DESCRIPTION
**Pull request by markerikson**
This PR:

- Adds a `.unwrap()` function attached to promises returned from `createAsyncThunk`
- Merges back the changes from #846 into `master`, since they went into `next`

Fixes #775, fixes #910 .

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,10 +21,12 @@ jobs:
       - uses: c-hive/gha-npm-cache@v1
 
       - name: Install deps
-        run: npm ci --ignore-scripts
+        run: npm ci
 
       - name: Pack (including Prepare)
         run: npm pack
+        env:
+          NODE_ENV: 'production' # this doesn't actually matter, so just set it to production
 
       - uses: actions/upload-artifact@v2
         with:
--- a/etc/redux-toolkit.api.md
+++ b/etc/redux-toolkit.api.md
@@ -85,6 +85,7 @@ export type AsyncThunkAction<Returned, ThunkArg, ThunkApiConfig extends AsyncThu
     abort(reason?: string): void;
     requestId: string;
     arg: ThunkArg;
+    unwrap(): Promise<Returned>;
 };
 
 // @public
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,12 +5,12 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.5.5",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
-      "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
+      "version": "7.12.13",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.13.tgz",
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by markerikson** [comment]
Hmm.

@hardfist , for some reason I'm seeing this failure in the CI action:

```
> @reduxjs/toolkit@1.5.1 build-ci /home/runner/work/redux-toolkit/redux-toolkit
> node scripts/build.js && tsc && tsc -p src/query/tsconfig.json && api-extractor run

dir: /home/runner/work/redux-toolkit/redux-toolkit/dist
/home/runner/work/redux-toolkit/redux-toolkit/node_modules/esbuild/bin/esbuild:2
throw new Error(`esbuild: Failed to install correctly
^

Error: esbuild: Failed to install correctly

Make sure you don't have "ignore-scripts" set to true. You can check this with
"npm config get ignore-scripts". If that returns true you can reset it back to
false using "npm config set ignore-scripts false" and then reinstall esbuild.

If you're using npm v7, make sure your package-lock.json file contains either
"lockfileVersion": 1 or the code "hasInstallScript": true. If it doesn't have
either of those, then it is likely the case that a known bug in npm v7 has
corrupted your package-lock.json file. Regenerating your package-lock.json file
should fix this issue.
```

Any idea why we might be having an issue there?

hardfist:
oh esbuild currently doesnt support npm7 because of npm bug

hardfist:
https://github.com/npm/cli/issues/2606

hardfist:
you may need to use npm@6 before bug fixed

markerikson:
Apparently the issue was just that we had `--ignore-scripts` in our workflow file, so the binary wasn't getting downloaded.

hardfist:
> Apparently the issue was just that we had `--ignore-scripts` in our workflow file, so the binary wasn't getting downloaded.

i never expect this :)

PR #976, opened 2021-04-02
Add `idGenerator` option to `createAsyncThunk` (#743)

### LINKED ISSUE #743: Possibility of customising function that generates requestId in the action's meta of createAsyncThunk
Hi! :) First of all, let me express my gratitude that this library (toolkit) had already existed back in February/March when I first re-architected some project at my current company – it's been making my life so much easier since then, and fun to use – so a big THANK YOU! 🙏😇🙇🏻‍♂️🙇🏻‍♂️

Recently though, there is one aspect/requirement that I have to fulfil in my app, which is all the network requests have to be able to be tracked by **[uuidv4](https://www.npmjs.com/package/uuid#uuidv4options-buffer-offset)** (to conform with the existing logging/analysis infrastructure).

As I want to make use of the RTK built-in feature as much as possible, would it be possible to add a possibility to pass a custom function other than `nanoid`  (in my case, it'd be `uuidv4`) to generate the `action.meta.requestId` used by `createAsyncThunk`?

This is a rather simplistic construct of how my slice would look like:
```javascript
extraReducers: {
  [fetchSomething.pending]: (state, action) => {
    state.loading = true;
    state.pendingRequestIds.push(action.meta.requestId);
  },
},
```

I mean, currently, to satisfy this requirement, I'm making use of `action.meta.arg.requestId` and doing the wiring when dispatching: `dispatch(fetchSomething({ requestId: uuidv4() }))`. So I thought it would be nice and cleaner if this feature could be implemented per se.

Thank you.

Shrugsy:
Closes #743 

This PR adds an optional `idGenerator` option to the `createAsyncThunk` options to allow specifying how request IDs are generated rather than using nanoid.

### DIFF
--- a/docs/api/createAsyncThunk.mdx
+++ b/docs/api/createAsyncThunk.mdx
@@ -89,7 +89,7 @@ An object with the following optional fields:
 
 - `condition`: a callback that can be used to skip execution of the payload creator and all action dispatches, if desired. See [Canceling Before Execution](#canceling-before-execution) for a complete description.
 - `dispatchConditionRejection`: if `condition()` returns `false`, the default behavior is that no actions will be dispatched at all. If you still want a "rejected" action to be dispatched when the thunk was canceled, set this flag to `true`.
-- `idGenerator`: a function to use when generating the `requestId` for the request sequence. Defaults to use [nanoid](https://www.npmjs.com/package/nanoid).
+- `idGenerator`: a function to use when generating the `requestId` for the request sequence. Defaults to use [nanoid](./otherExports.mdx/#nanoid).
 
 ## Return Value
 
--- a/docs/api/createAsyncThunk.mdx
+++ b/docs/api/createAsyncThunk.mdx
@@ -89,6 +89,7 @@ An object with the following optional fields:
 
 - `condition`: a callback that can be used to skip execution of the payload creator and all action dispatches, if desired. See [Canceling Before Execution](#canceling-before-execution) for a complete description.
 - `dispatchConditionRejection`: if `condition()` returns `false`, the default behavior is that no actions will be dispatched at all. If you still want a "rejected" action to be dispatched when the thunk was canceled, set this flag to `true`.
+- `idGenerator`: a function to use when generating a 'request ID' per request instance. Defaults to use [nanoid](https://www.npmjs.com/package/nanoid).
 
 ## Return Value
 
--- a/docs/api/createAsyncThunk.mdx
+++ b/docs/api/createAsyncThunk.mdx
@@ -89,7 +89,7 @@ An object with the following optional fields:
 
 - `condition`: a callback that can be used to skip execution of the payload creator and all action dispatches, if desired. See [Canceling Before Execution](#canceling-before-execution) for a complete description.
 - `dispatchConditionRejection`: if `condition()` returns `false`, the default behavior is that no actions will be dispatched at all. If you still want a "rejected" action to be dispatched when the thunk was canceled, set this flag to `true`.
-- `idGenerator`: a function to use when generating a 'request ID' per request instance. Defaults to use [nanoid](https://www.npmjs.com/package/nanoid).
+- `idGenerator`: a function to use when generating the `requestId` for the request sequence. Defaults to use [nanoid](https://www.npmjs.com/package/nanoid).
 
 ## Return Value
 
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -225,7 +225,9 @@ export interface AsyncThunkOptions<
   serializeError?: (x: unknown) => GetSerializedErrorType<ThunkApiConfig>
 
   /**
-   * A function to use when generating a 'request ID' per request instance. Defaults to use nanoid.
+   * A function to use when generating the `requestId` for the request sequence.
+   *
+   * @default `nanoid`
    */
   idGenerator?: () => string
 }
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -223,6 +223,11 @@ export interface AsyncThunkOptions<
   dispatchConditionRejection?: boolean
 
   serializeError?: (x: unknown) => GetSerializedErrorType<ThunkApiConfig>
+
+  /**
+   * A function to use when generating a 'request ID' per request instance. Defaults to use nanoid.
+   */
+  idGenerator?: () => string
 }
 
 export type AsyncThunkPendingActionCreator<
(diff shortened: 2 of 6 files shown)


docs/api/createAsyncThunk.mdx
```diff
@@ -89,6 +89,7 @@ An object with the following optional fields:
 
 - `condition`: a callback that can be used to skip execution of the payload creator and all action dispatches, if desired. See [Canceling Before Execution](#canceling-before-execution) for a complete description.
 - `dispatchConditionRejection`: if `condition()` returns `false`, the default behavior is that no actions will be dispatched at all. If you still want a "rejected" action to be dispatched when the thunk was canceled, set this flag to `true`.
+- `idGenerator`: a function to use when generating the `requestId` for the request sequence. Defaults to use [nanoid](https://www.npmjs.com/package/nanoid).
```
msutkowski, line 92:
```suggestion
- `idGenerator`: a function to use when generating the `requestId` for the request sequence. Defaults to use [nanoid](https://redux-toolkit.js.org/api/other-exports#nanoid).
```

We don't use all of nanoid, so we should just link back specifically to what we inline and re-export

markerikson, line 92: :hand: Let's point this to `./otherExports.mdx#nanoid` instead, and maybe also tweak the wording of the `nanoid` paragraph to say it's the "default" instead of "automatically used"

src/createAsyncThunk.ts
```diff
@@ -392,7 +399,7 @@ If you want to use the AbortController to react to \`abort\` events, please cons
     arg: ThunkArg
   ): AsyncThunkAction<Returned, ThunkArg, ThunkApiConfig> {
     return (dispatch, getState, extra) => {
-      const requestId = nanoid()
+      const requestId = (options?.idGenerator || nanoid)()
```
markerikson, line 402: :hand: We can use nullish coalescing here instead of `||`, right?

markerikson (changes requested):

src/createAsyncThunk.ts
Shrugsy: How exactly are you thinking? As in `options?.idGenerator?.() || nanoid()`?

src/createAsyncThunk.ts
```diff
@@ -223,6 +223,13 @@ export interface AsyncThunkOptions<
   dispatchConditionRejection?: boolean
 
   serializeError?: (x: unknown) => GetSerializedErrorType<ThunkApiConfig>
+
+  /**
+   * A function to use when generating the `requestId` for the request sequence.
+   *
+   * @default `nanoid`
+   */
+  idGenerator?: () => string
```
msutkowski, line 232: Would like to propose alternative names such as `idFn`, `requestIdFn`, etc etc just make sure we're going with the best option. Any other suggestions?

src/createAsyncThunk.ts
markerikson: `(options?.idGenerator ?? nanoid)()`

just swapping `||` for `??`. same behavior, slightly more semantically correct.

msutkowski (commented):
Looks good, just have a few comments.

Shrugsy:
I'm struggling to understand why the `preact` step in the CI build is failing here: https://github.com/reduxjs/redux-toolkit/pull/976/checks?check_run_id=2254585946#step:3:265

>       src/createAsyncThunk.test.ts:706:9 - error TS2345: Argument of type '{ idGenerator: jest.Mock<string, []>; }' is not assignable to parameter of type 'AsyncThunkOptions<void, {}>'.  
>       Object literal may only specify known properties, and 'idGenerator' does not exist in type 'AsyncThunkOptions<void, {}>'.  
>  
>      706       { idGenerator }

As far as I'm aware I added `idGenerator` to the `AsyncThunkOptions` definition, including the api:

[createAsyncThunk.ts](https://github.com/reduxjs/redux-toolkit/pull/976/files#diff-72c9100d5dcb08a309727a42a354cd576dcf318c3d0820c98ad26ef2bf6392e6R232)  

[redux-toolkit.api.md](https://github.com/reduxjs/redux-toolkit/pull/976/files#diff-e7e6e821811a38e7b647f3f04a03476f93f4296096f9004c9e46564370dcea3eR95)  

Does anyone have any insight as to what might be causing this error? Have I missed something somewhere?

docs/api/createAsyncThunk.mdx
markerikson: Yeah, but the link should go to the sibling Markdown file instead of the website directly :)

markerikson:
I think I saw the compressed size check also failing on some TS type changes with one of my other PRs this week. Not sure what's causing that.

src/createAsyncThunk.ts
Shrugsy: Sorry just realised you said nullish coalescing (??). Definitely makes more sense, will do

docs/api/createAsyncThunk.mdx
msutkowski: i'm fired :joy:

src/createAsyncThunk.ts
markerikson: Eh, I'm find with `idGenerator`.

docs/api/createAsyncThunk.mdx
Shrugsy: Both points now addressed

src/createAsyncThunk.ts
Shrugsy: Now addressed

docs/api/createAsyncThunk.mdx
Shrugsy: Now addressed

src/createAsyncThunk.ts
Shrugsy: As a consumer I'd be fine with all suggestions so far, so I'm pretty indifferent as to what it's called

msutkowski:
This line in the cAT docs will also need to be updated:

`  - `requestId`: a unique string ID value that was automatically generated to identify this request sequence`

markerikson:
Eh, I think that line's fine. There's still a `requestId`, and it's still a string, and it's still automatically generated - it's just a question of what the string contents are.

Shrugsy:
Regarding the `requestId` description on the `cAT` docs: I think it is still accurate as it is. It *could* be extended to say something like "If an `idGenerator` function is provided, it uses that to generated the ID instead", but that would just essentially be repeating  what is in the 'Options' section shortly after it.

Regarding the `Compressed size` check failing - Is there anything I can do about that? As far as I can tell, the failure it's hitting just *shouldn't* be happening. The same test that fails there passes in every other check.

Regarding comment threads, do you folks prefer:
-  for PR creators to advise that they addressed it, and leave it to the 'comment thread starter' to actually confirm it was addressed and press 'resolve conversation'?
or
- for PR creators to decide for themselves that it was resolved, and press 'resolve conversation'?

docs/api/createAsyncThunk.mdx
Shrugsy: I realised I made a sentence for the `nanoid` description incorrect when I adjusted it, but tweaked it to make sense again this time.

markerikson:
Y'know what, this looks good. Let's put it in.

PR #977, opened 2021-04-03
Update `immer` to v9.0.1

### LINKED ISSUE #931: Update Immer to 9.x
Immer 9 just came out, per https://twitter.com/mweststrate/status/1373021543979171840 , and we'll want to include this in RTK 1.6.

Changelog at https://github.com/immerjs/immer/commit/2c2f30e1c7bda5a1902acb4548678434e18cae5d :

> The TypeScript typings of Immer have been rewritten, and support better inference now and are more accurate. The best practices are documented here: https://immerjs.github.io/immer/typescript. Please visit that page first when running into any compile errors after updating. 
> 
> The following changes have been made:
> 
> Immer can now infer the type of the `draft` for curried producers, if they are directly passed to another function / context from which the type can be inferred. So lines like these will now be strongly typed: `setTodo(produce(draft => { draft.done = true }))`, where `setTodo` is a React state updater (for example). Fixes #720.
> 
> Immer will now better respect the original types passed to a producer, and generally not wrap any returned types with `Immutable` automatically.
> 
> BREAKING CHANGE: It is no longer allowed to return `nothing` from a recipe if the target state doesn't accept `undefined`.
> 
> BREAKING CHANGE: It is no longer allowed to return arbitrary things from a recipe. Recipes should either return nothing, or something that is assignable to the original state type. This will catch mistakes with accidental returns earlier.

Shrugsy:
Closes #931 

As far as I can tell, nothing in RTK is affected by the immer upgrade.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5725,9 +5725,9 @@
       "dev": true
     },
     "immer": {
-      "version": "8.0.4",
-      "resolved": "https://registry.npmjs.org/immer/-/immer-8.0.4.tgz",
-      "integrity": "sha512-jMfL18P+/6P6epANRvRk6q8t+3gGhqsJ9EuJ25AXE+9bNTYtssvzeYbEd0mXRYWCmmXSIbnlpz6vd6iJlmGGGQ=="
+      "version": "9.0.1",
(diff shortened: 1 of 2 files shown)


markerikson:
I didn't think anything would be :)  Thanks!

PR #981, opened 2021-04-04
chore(build): improved build tools

hardfist:
I try to resolve the following problems of build tools
* sourcemap verify
* watch support 
* extract error message
* ...


scripts/register.js
```diff
@@ -0,0 +1,11 @@
+const esbuild = require('esbuild')
+const fs = require('fs')
+require.extensions['.ts'] = (mod, filename) => {
+  const ts = fs.readFileSync(filename, 'utf-8')
+  const { code } = esbuild.transformSync(ts, {
+    loader: 'ts',
+    target: 'es2017',
+    format: 'cjs',
+  })
+  mod._compile(code, filename)
```
markerikson, line 10: :speech_balloon: hah, is this basically a homegrown `ts-node` using ESBuild? Nice :)

markerikson (commented):
So what is the actual net effect of these changes besides converting the build file to TS?

scripts/register.js
hardfist: yes, but much faster than ts-node

hardfist:
@markerikson  I think umd production sourcemap problem should be fixed now

scripts/register.js
hardfist: There will be some complex logic, and I'm not comfortable with JS so I convert build file to TS

hardfist:
CI is green now, I may leave watch to another PR(I don't know whether you need the watch functionality since you already have tsdx watch)

markerikson:
Spent the evening playing around with the existing `build.js` trying to get it to spit out bundles for the new `/query` and `/query/react` entry points, and got it mostly working:

image

image

However, I don't have any of the changes that you put in here. I'm going to repoint this to `v1.6-integration`, and then merge it so I've got this to use as a baseline. Thanks!

markerikson:
Hey @hardfist . Now that I've merged this and am trying it out, I'm very confused by what's going on with the sourcemap stuff. Is some of this experimental / debug?  What's it supposed to be doing?

hardfist:
> Hey @hardfist . Now that I've merged this and am trying it out, I'm very confused by what's going on with the sourcemap stuff. Is some of this experimental / debug? What's it supposed to be doing?

somthing wrong with sourcemap？

markerikson:
I'm just trying to understand what these lines are doing:

https://github.com/hardfist/redux-toolkit/blob/8358c0cf3fab3d21812571a1794d5865ee97bd6c/scripts/build.ts#L110-L139

it seems like it's debug code or something - why are we running asserts in the middle of a build script?

hardfist:
> I'm just trying to understand what these lines are doing:
> 
> https://github.com/hardfist/redux-toolkit/blob/8358c0cf3fab3d21812571a1794d5865ee97bd6c/scripts/build.ts#L110-L139
> 
> it seems like it's debug code or something - why are we running asserts in the middle of a build script?

It's kind of test to make sure the we can use the generated sourcemap to recover location from bundle to origin sourcefile, maybe it should be put into an separate test file.

markerikson:
If it's not related to actual functionality, yeah.

FYI, I've got `build.ts` heavily modified on my local branch right now to support generating bundles for all three entry points (RTK, RTK-Query core, RTK-Query + React).  I'm hoping to at least put up a draft PR with that by tomorrow if not tonight, so probably best to hold off on any PRs that modify that file.

PR #986, opened 2021-04-07
No need for await return

zargold:
When using an `async` function await automatically runs on return see: eslint rule https://eslint.org/docs/rules/no-return-await


phryneas:
Theoretically correct, but having it like this doesn't hurt and enables people that copy & paste the example to modify it to edit additional transformation logic into the next few lines without having to think about the promise.

Many people going through that are still pretty new to Promises, so this kinda helps them around that, so personally I'm inclined to keep it like that.

What do you think @markerikson?

markerikson:
Yeah, if anything we could just make the fake response have a `.data` field so you have to await and then `return res.data`.

PR #989, opened 2021-04-09
Update quick-start.md

abd1rahmane:
No need to import the 'useState'  Hook.


markerikson:
Yeah, that was left over from copy-pasting the component from the `Counter` file in the CRA template, but removing the non-Redux logic for space.  Thanks!

PR #990, opened 2021-04-09
experiment: additional arguments into `selectState`

phryneas:
As requested by Daniel in chat.

It's not a lot of code, so it could be discussion-worthy.

Essentially, the idea is to be able to pass additional args into the selectors created by the entityAdapter.
Actually, this is currently already possible for everything except `selectById` due to how reselect works - but not supported by the types.

Example:

```ts
import { createEntityAdapter } from "@reduxjs/toolkit";

const adapter = createEntityAdapter();

export const { selectAll: getComponents } = adapter.getSelectors(
  // @ts-ignore
  (state, containerId) => state.page.containers[containerId].components
);

let components = adapter.getInitialState();
components = adapter.addOne(components, { id: 1, foo: "bar" });

const state = {
    page: {
      containers: [undefined, undefined, { components }]
    }
  };

//@ts-ignore
const x = getComponents(state, 2);

```

### DIFF
--- a/etc/redux-toolkit.api.md
+++ b/etc/redux-toolkit.api.md
@@ -222,17 +222,17 @@ export interface EntityAdapter<T> extends EntityStateAdapter<T> {
 export type EntityId = number | string;
 
 // @public (undocumented)
-export interface EntitySelectors<T, V> {
+export interface EntitySelectors<T, V, ExtraArgs extends any[] = []> {
     // (undocumented)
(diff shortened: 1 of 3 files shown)


markerikson:
Hmm. This still something that's necessary with Reselect 4.1?

markerikson:
Eh, I don't feel this is critical for 1.9.

PR #991, opened 2021-04-11
Docs: Export store in TS guide

msutkowski:
- Exports the store for copy&pastability


markerikson:
Yeah, really need to do the same for the core and React-Redux docs too :)

PR #992, opened 2021-04-11
Minor docs tweaks

Shrugsy:
Further expands on #991 by showing the store exported in other snippets that might be mimicked closely. Also tries to fix a couple of instances identified where the snippet highlighting did not appear to match up with the code being discussed.

Changes:
```
    - Show `store` as exported in relevant snippets
    - Fix incorrect highlighted section in docs regarding
      middleware usage with typescript
    - Convert majority of snippet highlighting methods to explicitly wrap
      lines as opposed to specifying line numbers
```


markerikson:
Thanks!

PR #1006, opened 2021-04-20
Migrate RTK Query repo contents into the RTK repo

markerikson:
This PR:

- Migrates the entire RTK Query repo contents into a temporary `/query` folder, including all files with their history, as a first step towards integrating the source and docs into the RTK repo.

I'm not even touching anything to do with running or building the code here. I just want all the files to be in the RTK repo so we can be moving all the pieces around.


markerikson:
- Tests don't explode now that I'm ignoring the `/query` folder entirely
- Turned off CSB-CI, Compressed Size, TS typechecks, and Netlify previews for now. This will affect the `v1.6-integration` branch, which is intentional until we get things mostly sorted out.  We'll want to turn them back on later.
- Had to rename the RTKQ `package.json` files to fix a Jest collision, although that might be revertable now that I'm ignoring `/query`

Next step is to try reworking our test setup so that we can have tests in nested `./tests` subfolders, then begin moving the RTKQ source into the correct locations.

PR #1011, opened 2021-04-21
Try to build bundles for RTK Query (core and React)

markerikson:
## VERY BUNDLED MUCH EXCITEMENT MAY MERGE NOW

I've managed to coerce the `build.ts` script into outputting the same bundle sets for the `/query` and `/query/react` dist folders.  Notes so far:

- Need to get `package.json` files for those entry points in the output, and they probably need to have proper keys for the bundle files (`module`, `main`, etc)
- Running `tsc` against `src/query/tsconfig.json` produces a copy of the RTK `.d.ts` files in a weird nested structure. Do we just need to run it once on `src`? Was thinking I maybe needed to run `tsc` programmatically for each one, but maybe not.
- Got API Extractor outputting reports for all three entry points if you manually call it and pass the config files. I was looking at calling it programmatically from within the build loop and adding `--local` and `--skipExtraction` args to `build.ts`
- Not at all ready to try doing a package + install into an app to try this out
- UMD for those entry points too?
- Had to add some dummy type imports into the codebase to avoid uses of `import('some-lib').SomeType`, per @phryneas's comment at https://github.com/microsoft/rushstack/issues/1050#issuecomment-631951991
- Haven't tried setting up tests whatsoever


markerikson:
Good news, bad news.

**Good news:**

- I've got the TS typedef output looking fine. Dropped the combined `.d.ts` files per suggestion from Lenz, and switched to compiling all of `src` with TSC in one shot. That gives us all the typedef files in the right structure.
- Got all of the bundle files showing up the way I want them, including even the UMD files.
- API Extractor is generating the right output for each entry point
- TS doesn't seem to support nested `exports` fields yet, but I was able to hack around that and get it to recognize `@reduxjs/toolkit/query/react` as a viable entry point via a `typesVersions` section per https://github.com/microsoft/TypeScript/issues/33079#issuecomment-702617758:

image

**Bad news:**

- I can't seem to get Webpack to recognize my nested entry points, and tbh I'm getting _very_ lost in this maze of keywords and options.

Here's what I've got right now:

```js
"main": "dist/index.js",
  "module": "dist/redux-toolkit.esm.js",
  "unpkg": "dist/redux-toolkit.umd.min.js",
  "types": "dist/index.d.ts",  
  "exports": {
    ".": {
      "types": "./dist/index.d.ts",
      "module": "./dist/redux-toolkit.esm.js",
      "default": "./dist/index.js"
    },
    "./query": {
      "module": "./dist/query/rtk-query.esm.js",
      "default": "./dist/query/index.js"
    },
    "./query/react": {
      "types": "./dist/query/react.d.ts",
      "module": "./dist/query/react/rtk-query-react.esm.js",
      "default": "./dist/query/react/index.js"
    }
  },
  "typesVersions": {
    "*": {
      "query": ["./dist/query/index.d.ts"],
      "query/react": ["./dist/query/react.d.ts"]
    }
  },
```

When I try to start up this CRA app, the result is:

```
Module not found: Can't resolve '@reduxjs/toolkit/query/react' in 'D:\Projects\redux\temp\rtkq-react\src\app'
```

I've dug through a dozen issues in random repos regarding export maps, and thus far I don't think I have any real idea what I'm doing.

Really going to need some advice from experts on this one.

markerikson:
Per guidance from @andarist , the problem was that Webpack 4 doesn't yet recognize export maps, so your package output has to have a folder structure that matches the intended import paths:

https://twitter.com/AndaristRake/status/1385120048818364419

I moved the legacy RTKQ repo folder from `query` to `query-old` to get that out of the way, tweaked the build script to move `dist/query` to `./query` as the final step, and then updated inclusion paths to pull that in... and it works!

image

image

I've re-enabled the CI checks I'd disabled and I'm going to go ahead and merge this.

We don't have tests for the RTKQ code working yet, but that may be the biggest piece we're missing.

PR #1013, opened 2021-04-23
Failing attempts at getting TS version selection working

markerikson:
## VERY WORKING MUCH QUERY SUCH VERSIONING

This is a completely failing attempt to get TS4.0 compat working, by creating a clone of `query/react/index.d.ts` named `indexTs40.d.ts`, swapping an import, and getting TS `typesVersions` detection working.

This is really built off of #1012 , so it's got all the changes there in terms of test updates, plus some reworking of the `src/query/react/` files that probably shoulda been over here.

I've been using `--traceResolution`, and I can see TSC finding the `typesVersion` field in `dist/query/react/package.json`... and then it just seems to ignore it. If I have a `types` field in there, it settles on that no matter what the version check is. If I don't have a `types` field, it ends up latching on to `main` and deriving an `index.d.ts` from that.

Frankly I have no idea what the path mappings in `typesVersions` are supposed to be. Individual filenames? Paths relative to repo root? Paths relative to this folder? Noooo idea.


markerikson:
Per Orta at https://twitter.com/orta/status/1385610364323237889 :

> Checked the compiler, you’re failing this containsPath check: https://github.com/microsoft/TypeScript/blob/master/src/compiler/moduleNameResolver.ts#L1386
> 
> Because your resolved path doesn’t live inside the tree structure provided by the module, could be a security check 
> 
> If it’s all internal, maybe move the matched dts into the package

So, will have to figure out a way around that.

markerikson:
After several more hours of hacking, this _appears_ to be working!

Lenz suggested that what we really needed was a `./src/query/react/versionedTypes` folder with its own `package.json`, and the two TS40/41 types files in there:

image

I reworked the packaging and build setup to have that. 

I've been using a standalone copy of the RTKQ React example app as my testbed.  Ran into some more issues with TSC not handling `typesVersions` correctly, but some use of `--traceResolutions` gave me the debug info I needed to finally get this all sorted out.

I've seen TS 4.0 correctly parse `typesVersions` for the `versionedTypes` import (and then generate 2.4MB of types errors because it didn't recognize the hooks the source was still exporting). After redoing the service files to "manually" export the hooks under their correct names, TSC 4.0 correctly compiled the app, and I was then able to actually build it successfully with CRA.

The resulting bundle looks like this:

image

That looks basically as I'd expect it: both RTK and RTKQ in the bundle, ESM format, seemingly valid sizes.

I did have to revert the earlier `ApiModule` restructuring that I did. I hadn't run the tests in a while, and suddenly TS was screaming loudly because something had broken.  All that is now in `query/react/index.ts`.

I _think_ this basically wraps up the actual code integration aspect, and we can now work on integrating the docs.

PR #1014, opened 2021-04-24
Add full types to useMutation's returned callback

Shrugsy:
`useMutation` returns a 'trigger function' as it's first item in the return tuple.  
This PR changes the signature of the value returned by that 'trigger function' to include all properties it actually has, including that it returns a `Promise`, not a plain object.

e.g.
#### Before - only `unwrap`
image

#### After - all properties
image


query-old/docs/api/created-api/hooks.md
```diff
@@ -91,18 +91,40 @@ The query arg is used as a cache key. Changing the query arg will tell the hook
...
 ```ts
-type MutationHook = () => [
-  (
-    arg: any
-  ) => {
-    unwrap: () => Promise<ResultTypeFrom<D>>;
-  },
-  MutationSubState<D> & RequestStatusFlags
-];
+type UseMutation<Definition> = () => [UseMutationTrigger<Definition>, UseMutationResult<Definition>];
```
phryneas, line 94: I don't know the best way to write this, but like this it is omitting `selectFromResult`

query-old/docs/api/created-api/hooks.md
```diff
@@ -91,18 +91,40 @@ The query arg is used as a cache key. Changing the query arg will tell the hook
...
+  endpointName?: string; // The name of the given endpoint for the mutation
+  error?: any; // Error result if present
+  fulfilledTimestamp?: number; // Timestamp for when the mutation was completed
+  isError: boolean; // Mutation is currently in an "error" state
+  isLoading: boolean; // Mutation has been fired and is awaiting a response
+  isSuccess: boolean; // Mutation has data from a successful call
+  isUninitialized: boolean; // Mutation has not been fired yet
+  originalArgs?: ArgTypeFrom<Definition>; // Arguments passed to the latest mutation call
+  startedTimeStamp?: number; // Timestamp for when the latest mutation was initiated
+  status: 'uninitialized' | 'pending' | 'fulfilled' | 'rejected'; // @deprecated - A string describing the mutation state
```
phryneas, line 122:
```suggestion
```

let's keep that one out since it's already deprecated and we will remove it

query-old/docs/api/created-api/hooks.md
```diff
@@ -91,18 +91,40 @@ The query arg is used as a cache key. Changing the query arg will tell the hook
...
-  ) => {
-    unwrap: () => Promise<ResultTypeFrom<D>>;
-  },
-  MutationSubState<D> & RequestStatusFlags
-];
+type UseMutation<Definition> = () => [UseMutationTrigger<Definition>, UseMutationResult<Definition>];
+
+type UseMutationTrigger<Definition> = (
+  arg: ArgTypeFrom<Definition>
+) => Promise<{ data: ResultTypeFrom<Definition> }> & {
```
phryneas, line 98:
```suggestion
) => Promise<{ data: ResultTypeFrom<Definition> | { error: unknown } }> & {
```
could be a bit more accurate thatn `unknown`, but definitely mention the error here

query-old/docs/api/created-api/hooks.md
```diff
@@ -91,18 +91,40 @@ The query arg is used as a cache key. Changing the query arg will tell the hook
...
+  arg: {
+    endpointName: string; // The name of the given endpoint for the mutation
+    originalArgs: ArgTypeFrom<Definition>; // Arguments passed to the mutation
+    track?: boolean; // Whether the mutation is being tracked in the store
+    startedTimeStamp: number; // Timestamp for when the mutation was initiated
+  };
```
phryneas, line 104: This is kinda "implementation detail". Could we mark that as `@internal` in the docblock and omit it here instead?

src/query/core/buildInitiate.ts
```diff
@@ -87,10 +87,85 @@ export type MutationActionCreatorResult<
...
+  }
+  /**
+   * A unique string generated for the request sequence
+   */
   requestId: string
+  /**
+   * A method to cancel the mutation promise. Note that this is not intended to prevent the mutation
+   * that was fired off from reaching the server, but only to assist in handling the response.
+   *
+   * Calling `abort()` prior to the promise resolving will make it throw with an error like:
```
phryneas, line 116: I actually have no idea what that would do, have you tested that? I assume it would resolve with `{ error: { name: 'AbortError', message: 'Aborted' }` unless you unwrapped it.

src/query/core/buildInitiate.ts
```diff
@@ -87,10 +87,85 @@ export type MutationActionCreatorResult<
...
+    originalArgs: QueryArgFrom<D>
+    /**
+     * Whether the mutation is being tracked in the store.
+     */
+    track?: boolean
+    /**
+     * Timestamp for when the mutation was initiated
+     */
+    startedTimeStamp: number
+  }
```
phryneas, line 107: Uh, nice :bug: catch!

src/query/core/buildInitiate.ts
```diff
@@ -87,10 +87,85 @@ export type MutationActionCreatorResult<
     >
   >
 > & {
-  arg: QueryArgFrom<D>
+  arg: {
```
phryneas, line 90:
```suggestion
  /** @internal */
  arg: {
```

src/query/core/buildInitiate.ts
```diff
@@ -87,10 +87,85 @@ export type MutationActionCreatorResult<
...
+   *   const payload = await addPost({ id: 1, name: 'Example' }).unwrap();
+   *   console.log('fulfilled', payload)
+   * } catch (error) {
+   *   console.error('rejected', error);
+   * }
+   * ```
+   */
   unwrap(): Promise<ResultTypeFrom<D>>
+  /**
+   * A method to manually unsubscribe from the mutation call
```
phryneas, line 167:
```suggestion
   * A method to manually unsubscribe from the mutation call, meaning it will be removed from cache after the usual caching grace period.
   The value returned by the hook will reset to `isUninitialized` afterwards.
```

phryneas (changes requested):

query-old/docs/api/created-api/hooks.md
Shrugsy: Yep sounds good to me

src/query/core/buildInitiate.ts
Shrugsy: Ah yeah my wording is definitely off. When I wrote it, in my head I was specifically thinking of the unwrapped promise, but that's not what I actually wrote.

I'll re-word it to make sense

src/query/core/buildInitiate.ts
```diff
@@ -79,18 +80,98 @@ type StartMutationActionCreator<
...
+  | {
+      error:
+        | BaseQueryError<
+            D extends MutationDefinition<any, infer BaseQuery, any, any>
+              ? BaseQuery
+              : never
+          >
+        | SerializedError
+    }
 > & {
```
Shrugsy, line 93: @phryneas I've tried fixing the type for the resolved value of the 'trigger' promise here, but the actual `error` type produced has an additional `undefined` type attached to it.

image

Can you make any sense out of that?

I've temporarily suppressed the error in order to allow codesandbox to build, in order to see how it behaved there:
https://codesandbox.io/s/vanilla-typescript-forked-duy39?file=/src/index.ts

Shrugsy: Ah it's just occurred to me; would that be caused by the scenario where the `extends MutationDefinition` ternary fails?

```ts
        | BaseQueryError<
            D extends MutationDefinition<any, infer BaseQuery, any, any>
              ? BaseQuery
              : never
          >
```

Shrugsy: I'm particularly stumped because it looks like `BaseQueryError` is explicitly intended to *remove* the case with `{ error: undefined }`
```ts
export type BaseQueryError<BaseQuery extends BaseQueryFn> = Exclude<
  UnwrapPromise<ReturnType<BaseQuery>>,
  { error: undefined }
>['error']
```

Shrugsy: I pushed a new version which explicitly removes `undefined` from the type returned by `BaseQueryError` like so:

```ts
Promise<
  | { data: ResultTypeFrom<D> }
  | {
      error:
        | Exclude<
            BaseQueryError<
              D extends MutationDefinition<any, infer BaseQuery, any, any>
                ? BaseQuery
                : never
            >,
            undefined
          >
        | SerializedError
    }
>
```

But I'm iffy about two things:
- Is `undefined` actually supposed to be there for some reason?
- If not, and I remove it doing the above, and someone defines their error type like: `{ error: undefined }`, then the promise type will be wrong (Because it will remove `undefined`)

query-old/docs/api/created-api/hooks.md
Shrugsy: Done, went with this option:

```ts
) => Promise<{ data: ResultTypeFrom<Definition> } | { error: BaseQueryError | SerializedError }> & {
```

src/query/core/buildInitiate.ts
Shrugsy: Re-worded to refer to 'error state' rather than saying it will 'throw'

src/query/core/buildInitiate.ts
Shrugsy: 🐛

query-old/docs/api/created-api/hooks.md
Shrugsy: I've added in a pseudo-type for `UseMutationStateOptions`, but it's definitely getting very crowded. What are your thoughts on removing the `Definition` pseudo-generic from this section completely? My thoughts are that it's already an inaccurate generic compared to the real types, and it probably clutters the explanation more than it clarifies anything.

src/query/core/buildInitiate.ts
Shrugsy: Also fwiw, it behaves exactly like you thought for the 'wrapped' promise

src/query/tests/queryFn.test.tsx
```diff
@@ -247,7 +247,7 @@ describe('queryFn base implementation tests', () => {
     {
       const thunk = mutationWithNeither.initiate('mutationWithNeither')
       const result = await store.dispatch(thunk)
-      expect(result.error).toEqual(
+      expect((result as any).error).toEqual(
```
Shrugsy, line 250: FYI this change is because `result` previously got inferred as `any`, but is now more accurately:

```ts
{ data: string; } | { error: string | SerializedError }
```

If anyone has a better idea for handling this than `any-casting` I'm all ears

Shrugsy: Decided on a better option than `as any`:
```ts
      expect('error' in result && result.error).toEqual(
```

markerikson (approved):
LGTM, but would like a :+1: from @phryneas or @msutkowski to confirm

query-old/docs/api/created-api/hooks.md
```diff
@@ -91,18 +91,50 @@ The query arg is used as a cache key. Changing the query arg will tell the hook
...
+
+Passing a completely empty object will prevent the hook from causing a re-render at all, e.g.
+```ts
+selectFromResult: () => ({})
 ```
+:::
 
 - **Returns**: a tuple containing:
-  - `trigger`: a function that triggers an update to the data based on the provided argument
+  - `trigger`: a function that triggers an update to the data based on the provided argument. The trigger function returns a promise with the properties shown above that may be used to handle the behaviour of the promise.
```
msutkowski, line 137: Your non-American behavior here is silly! 🤣

src/query/tests/buildHooks.test.tsx
```diff
@@ -804,6 +816,82 @@ describe('hooks tests', () => {
...
+
+      render(<User />, { wrapper: storeRef.wrapper })
+      expect(screen.queryByText(/An error has occurred/i)).toBeNull()
+      expect(screen.queryByText(/Successfully updated user/i)).toBeNull()
+      expect(screen.queryByText('Request was aborted')).toBeNull()
+
+      fireEvent.click(
+        screen.getByRole('button', { name: 'Update User and abort' })
+      )
+      await screen.findByText('An error has occurred updating user Banana')
```
msutkowski, line 891: Appreciate you keeping the 🍌 s around 💯

msutkowski (approved):

query-old/docs/api/created-api/hooks.md
```diff
@@ -91,18 +91,50 @@ The query arg is used as a cache key. Changing the query arg will tell the hook
...
+  originalArgs?: ArgTypeFrom<Definition>; // Arguments passed to the latest mutation call
+  startedTimeStamp?: number; // Timestamp for when the latest mutation was initiated
+};
+```
+
+:::tip
+
+The generated `UseMutation` hook will cause a component to re-render by default after the trigger callback is fired as it affects the properties of the result. If you want to call the trigger but don't care about subscribing to the result with the hook, you can use the `selectFromResult` option to limit the properties that the hook cares about.
+
+Passing a completely empty object will prevent the hook from causing a re-render at all, e.g.
```
phryneas, line 130: `selectFromResult` will trigger at least one re-render.

PR #1017, opened 2021-04-25
Expose `useLazyQuerySubscription` for individual endpoints

Shrugsy:
The typing for `QueryHooks` claimed to include `useLazyQuerySubscription`, which *appeared* to expose the hook for individual endpoints. However in the react module, it was never used, so something like `api.endpoints.getPosts.useLazyQuerySubscription` would show the corresponding type, but would actually be `undefined`.

This PR makes the react modules attach `useLazyQuerySubscription` to query endpoints, in line with the types.

### DIFF
--- a/src/query/react/module.ts
+++ b/src/query/react/module.ts
@@ -130,12 +130,14 @@ export const reactHooksModule = ({
           const {
             useQuery,
             useLazyQuery,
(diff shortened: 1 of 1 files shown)


phryneas (approved):
good catch!

PR #1018, opened 2021-04-25
Docs - RTKQ - Flesh out content on hooks api page

Shrugsy:
Fleshes out missing content for the RTK Query hooks api page. 

- Adds pseudo type signatures for each hook
- Adds brief descriptions of hook parameters/return values
- Adds brief hook descriptions
- Adds 'feature' points for each hook
- Adds a 'feature comparison' table for a quick eyeball comparison of what each hook does

Some descriptions do feel a bit brief, but it's more than nothing


Shrugsy:
@markerikson, @phryneas, @msutkowski; If anyone has some good descriptions for the missing/brief ones, please feel free to fill them in or post them here.

In particular, I wasn't confident in describing these accurately:
- `useQueryState` (description)
- `useQuerySubscription` (description)
- `useLazyQuerySubscription` (options + description)

Shrugsy:
> @markerikson, @phryneas, @msutkowski; If anyone has some good descriptions for the missing/brief ones, please feel free to fill them in or post them here.
> 
> In particular, I wasn't confident in describing these accurately:
> ~- `useQueryState` (description)~
> ~- `useQuerySubscription` (description)~
> 
> * `useLazyQuerySubscription` (options + description)

I've added short descriptions for `useQueryState` and `useQuerySubscription`.

However I'm not sure exactly what `useLazyQuerySubscription` does.

If a `QuerySubscription` hook is concerned with the automatic fetching lifecycle, and the `Lazy` hook is concerned with opting out of the automatic fetching lifecycle to perform fetches manually, what does `useLazyQuerySubscription` do?

I'm sure I'm just missing a key understanding for one or both of the hooks

markerikson:
My vague understanding just based on the chats and PRs is that the pieces are:

- "state": reads the data and re-renders
- "subscription": actually triggers a fetch of the data, and does so automatically
- "lazy": triggers a fetch of the data, but only when you tell it to manually
- "query": wraps matching "state" + "subscription" hooks

So, I _assume_ that `useLazyQuerySubscription` is "just a fetch of the data without reading it and re-rendering, but only triggers when you tell it to".

Shrugsy:
> So, I assume that useLazyQuerySubscription is "just a fetch of the data without reading it and re-rendering, but only triggers when you tell it to".

Ah that wording made it click for me, thanks heaps. I was too tied up trying to correlate `subscription` with 'automatic fetching'

markerikson:
Y'know, when you put it that way, it's almost like "subscription" is the wrong word to use for those hooks.  What would be better? "Fetch"? "Request"?  For that matter, once you trigger a lazy subscription, does all the normal timing/polling/whatever stuff kick in from there on out?  If so, maybe "subscription" is fine.

Shrugsy:
> For that matter, once you trigger a lazy subscription, does all the normal timing/polling/whatever stuff kick in from there on out? If so, maybe "subscription" is fine.

Yep, it's exactly like that. I think `subscription` is fitting. Getting all of that info together, `LazyQuerySubscription` definitely makes sense

phryneas:
@markerikson @Shrugsy  "subscription" is exactly that, but on another level than you think: the *component* subscribes to a *cache value in the store* to keep it from getting collected.

So yes, `useQuerySubscription` fetches data etc, but a normal `dispatch(initiate)` would do that too - but then it keeps it in the store and "unsubscribes" on unmount, so it's more about the component lifecycle synchronisation thing.
Same for `useLazyQuery`.

Shrugsy:
Thanks @phryneas, that's a good clarification. I think the updated hook descriptions should be pretty accurate now

src/query/react/buildHooks.ts
```diff
@@ -112,6 +124,19 @@ interface UseQuerySubscriptionOptions extends SubscriptionOptions {
   refetchOnMountOrArgChange?: boolean | number
 }
 
+/**
+ * A React hook that automatically triggers fetches of data from an endpoint, and 'subscribes' the component to the cached data.
```
msutkowski, line 128:
```suggestion
 * A React hook that automatically triggers a request for a given endpoint, and 'subscribes' the component to the cached data.
```

msutkowski: I think this copy would be slightly more readable and matches more closely to the `UseMutation` language below.

docs/api/rtk-query/created-api/hooks.md
```diff
@@ -78,87 +231,281 @@ type UseQueryResult<T> = {
...
+type UsePrefetch = (
+  endpointName: string,
+  options?: UsePrefetchOptions
+) => PrefetchCallback
+
+type UsePrefetchOptions =
+  | {
+      // If specified, only runs the query if the difference between `new Date()` and the last
+      // `fulfilledTimeStamp` is greater than the given value
+      ifOlderThan?: false | number
```
msutkowski, line 486: Should we note that this value is in seconds here?

msutkowski (commented):
This is awesome! Thanks for all of the work here. Only a question and one suggestion on my end :trophy:

docs/api/rtk-query/created-api/hooks.md
Shrugsy: Yep, good point. Done now

msutkowski (approved):
👍

PR #1019, opened 2021-04-25
Examples/rtk query

msutkowski:
WIP


src/query/examples/concepts-queries-deduping-caching/craco.config.js
```diff
@@ -0,0 +1,26 @@
...
+        ),
+        'react-dom$': resolveFrom(
+          path.resolve(__dirname, '../../../node_modules'),
+          'react-dom'
+        ),
+        'eslint-plugin-prettier': resolveFrom(
+          'node_modules',
+          'eslint-plugin-prettier'
+        ),
+        'babel-jest': resolveFrom('node_modules', 'babel-jest'),
```
msutkowski, line 22: Unfortunately, this doesn't work - even with specifying the resolutions. There is a conflict here between this and the babel-jest dep in the root for rtk itself. This is why `SKIP_PREFLIGHT_CHECK=true` is on :disappointed:

src/query/react/versionedTypes/index.ts
```diff
@@ -5,4 +5,4 @@
 // Overall, this setup allows us to selectively override the one
 // file that has any difference between 4.1 and earlier, without
 // having to ship two completely duplicate copies of our typedefs.
-export { HooksWithUniqueNames } from './ts41Types'
+export type { HooksWithUniqueNames } from './ts41Types'
```
msutkowski, line 8: I don't think this would be breaking, but we should check (@phryneas) - this was required for the vite example support, otherwise `isolatedModules` would have to change in the tsconfig.

```ts
r TS1205: Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.

8 export { HooksWithUniqueNames } from './ts41Types'
           ~~~~~~~~~~~~~~~~~~~~


Found 1 error.
```

src/query/examples/concepts-queries-deduping-caching/package.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "devDependencies": {
+    "@types/react": "^17.0.3",
+    "@types/react-dom": "^17.0.3",
+    "eslint-config-prettier": "^8.3.0",
+    "eslint-plugin-prettier": "^3.4.0",
+    "typescript": "^4.2.4"
+  },
+  "resolutions": {
+    "babel-jest": "^26.6.0"
```
msutkowski, line 24: See note above, this does not work, so `SKIP_PREFLIGHT_CHECK` is set.

docs/usage/rtk-query/queries.md
```diff
@@ -177,7 +177,7 @@ What you'll see below is this:
 Click the 'Add bulbasaur' button. You'll observe the same behavior described above until you click the 'Refetch' button on one of the components.
 :::
 
-<iframe src="https://codesandbox.io/embed/concepts-queries-deduping-caching-5qy3n?fontsize=12&hidenavigation=1&theme=dark"
+<iframe src="https://codesandbox.io/embed/github/reduxjs/redux-toolkit/tree/examples/rtk-query/src/query/examples/concepts-queries-deduping-caching?fontsize=12&hidenavigation=1&theme=dark"
```
msutkowski, line 180: This would get updated to something like `https://codesandbox.io/embed/github/reduxjs/redux-toolkit/tree/master/src/query/examples/concepts-queries-deduping-caching` once merged.

msutkowski:
Closing this out - there was no need for this approach as we're not pulling from the source.

PR #1022, opened 2021-04-26
export Result Definition types

dtwiers:
I'd like to use these to define common functions for `providesTags` in my client code

### DIFF
--- a/src/query/endpointDefinitions.ts
+++ b/src/query/endpointDefinitions.ts
@@ -80,7 +80,7 @@ export enum DefinitionType {
   mutation = 'mutation',
 }
 
-type GetResultDescriptionFn<
+export type GetResultDescriptionFn<
   TagTypes extends string,
   ResultType,
   QueryArg,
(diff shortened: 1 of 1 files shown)


phryneas (approved):

Shrugsy:
Unless I'm mistaken, these will need to be added to the exports in `src/query/index.ts` as well to actually import them from outside of the library

phryneas:
Yes and no. If we no longer go for a single-type-file-rollup, they can just be imported manually from `dist/ts`, which I would prefer for a lot of those types as they're still "kinda internal" (this one is more public, but doesn't feel "important enough" to be exported by index? I don't know :thinking: )

We should probably go through all exports everywhere and mark them as @public or @internal to mark a "stable API interface" and other types as "internal type, you can use it but we might change it".

dtwiers:
Hmm. My opinion, recognizing that I'm also not the maintainer - not usually a big fan of importing things specifying the `dist` folder if it's avoidable, though that's probably just my nature trying to avoid using unstable internal interfaces as a general rule. I can annotate with @internal and/or I can add it to index to re-export. What would you prefer? @Shrugsy @phryneas

phryneas:
I'll merge this in now and we'll discuss that along the way. Thanks for the contribution!

PR #1026, opened 2021-04-30
Invalidation tests

phryneas:
@Shrugsy I think this test describes the "what invalidates what when" behaviour pretty good - do you think you could make a table in the docs out of that?


Shrugsy:
> @Shrugsy I think this test describes the "what invalidates what when" behaviour pretty good - do you think you could make a table in the docs out of that?

Yep, I'll try to have a look into that tomorrow

PR #1027, opened 2021-04-30
add export for `copyWithStructuralSharing`, run CI against `dist`

phryneas:


phryneas:
image

This seems to uncover a build error. 
But the tests work so far, so I'll merge it in.

PR #1029, opened 2021-04-30
fix/default serialize query args

phryneas:
fixes https://github.com/rtk-incubator/rtk-query/issues/225


msutkowski (approved):

chihabhajji:
weird one, but only one nested property fails even tho everything is configured to default 
on swagger/postman i get 
image
but any time i try to access the message in the error object i get [Object object]

PR #1031, opened 2021-04-30
Docs - Add explanation for removeAll method for createEntityAdapter's CRUD functions

julianfssen:
Hi, thanks for the awesome library and documentation!

I was looking for a method that removes all entities in state objects created by `createEntityAdapter`. I thought this was not available as the current documentation does not explain the method as [seen here](https://redux-toolkit.js.org/api/createEntityAdapter#crud-functions).

Turns out I was wrong as there _is_ a `removeAll` method described in the [types](https://redux-toolkit.js.org/api/createEntityAdapter#return-value) section above.

I'm adding a line to explain the `removeAll` method to improve clarity for beginners like me that gloss over the docs and assume no such method exists.

Thanks again!


markerikson:
Thanks!

PR #1034, opened 2021-05-01
split up middleware, add OptionalPromise, add cache entry lifecycle

phryneas:
Background on the "cache entry lifecycle": https://github.com/rtk-incubator/rtk-query/issues/215


phryneas:
what's still left todo here code-wise is removing the existing endpoint lifecycle onStart/etc.

src/query/core/buildMiddleware/queryLifecycle.ts
```diff
@@ -40,11 +40,14 @@ export const build: SubMiddlewareBuilder = ({
           originalArgs
         )
 
+        const extra = mwApi.dispatch((_, __, extra) => extra)
```
markerikson, line 110: 🤣

phryneas:
Btw, Bikeshedding:

we need a discussion if `onCacheEntryAdded` and `onQuery` are appropriate names

src/query/core/buildMiddleware/queryLifecycle.ts
phryneas: Genius, isn't it? :D

I was like "the old API had it, so we mayybe need it here too?" and then "how on earth do I get that?". Yup.

phryneas:
And internal TODO:

 I'm thinking of adding a `updateCacheEntry` (a pre-bound version of `updateQueryResult`) as an addition to https://github.com/reduxjs/redux-toolkit/pull/1034/commits/e2e1fbc0cdc461da92d4a4f936e3b04c393c0daf#diff-37992e5ad65f35286d92ad23730058800d8a872da260317ff5e9459aa03433a8R17-R41

Also, maybe adding a `undo` property on the thunk return value of `updateQueryResult` so that usually, `patchQueryResult` doesn't need to be dispatched by hand

src/query/core/buildMiddleware/cacheLifecycle.ts
```diff
@@ -0,0 +1,241 @@
...
+}
+
+export const build: SubMiddlewareBuilder = ({
+  api,
+  reducerPath,
+  context,
+  queryThunk,
+  mutationThunk,
+}) => {
+  type CacheLifecycle = {
```
asherccohen, line 120: Can this be moved outside, or is there a reason it's declared here?

src/query/createApi.ts
```diff
@@ -277,6 +277,26 @@ export function buildCreateApi<Modules extends [Module<any>, ...Module<any>[]]>(
             }
             x.providesTags ??= x.provides
           }
+          if (x.onStart || x.onSuccess || x.onError) {
```
asherccohen, line 280: Nice, glad u added a message here!

src/query/core/buildMiddleware/cacheLifecycle.ts
phryneas: It's only used in a locally scoped variable in the next line and once more to declare a value to add to that variable.

Moving it outside would raise questions like "should this be exported" etc, just keeping it here just circumvents all that and will also 100% remove it on transpilation.

src/query/tests/optimisticUpdates.test.tsx
```diff
@@ -30,25 +30,19 @@ const api = createApi({
...
           api.util.updateQueryResult('post', id, (draft) => {
             Object.assign(draft, patch)
           })
-        ).inversePatches
-      },
-      onError({ id }, { dispatch, context }) {
-        dispatch(api.util.patchQueryResult('post', id, context.undoPost))
+        )
+        resultPromise.catch(undo)
       },
```
phryneas, line 46: It's way too late, but this diff is kinda worth it

src/query/core/buildMiddleware/cacheLifecycle.ts
asherccohen: Cool. Tx for the explanation.

PR #1035, opened 2021-05-01
simplify `selectFromResult` usage

phryneas:


phryneas:
So with that, there should be a lot less confusing about using `selectFromResult`, but it's a small breaking change

msutkowski (approved):
Updated the tests here to make sure this behavior is included and type-checked, and all looks good. I don't see any issue with the `useStore` here either.

markerikson:
Still only partly understand what all is going on here :)  but seems okay at first glance.

PR #1036, opened 2021-05-02
[WIP DRAFT] - Docs - add draft 'Cached data' concept page

Shrugsy:
Draft for a 'cached data' docs page. Intended to cover as much of the caching concept in a single page as is reasonable


docs/usage/rtk-query/cached-data.md
```diff
@@ -0,0 +1,516 @@
...
+
+  TODO: description of any other key concepts
+
+## Default cache handling behaviour
+
+TODO:
+
+## Cache tags
+
+RTK Query uses the concept of 'tags' to determine whether a mutation for one endpoint intends to 'invalidate' some data that was 'provided' by a query from another endpoint. When defining an `api`, `createApi` accepts an array of tag type names for the `tagTypes` property, which is a list of possible tag name options that the queries for the `api` could provide.
```
phryneas, line 28:
```suggestion
RTK Query uses the concept of 'tags' to determine whether a mutation for one endpoint intends to 'invalidate' some data that was 'provided' by a query from another endpoint.
If cache data is being invalidated, it will either refetch the providing query (if components are still using that data) or remove the data from the cache.

When defining an `api`, `createApi` accepts an array of tag type names for the `tagTypes` property, which is a list of possible tag name options that the queries for the `api` could provide.
```

docs/usage/rtk-query/cached-data.md
```diff
@@ -0,0 +1,516 @@
...
+        url: `posts/${body.id}`,
+        method: 'POST',
+        body,
+      }),
+    }),
+  }),
+})
+```
+
+Note that for the example above, the `id` is used where possible on a successful result. In the case of an error, no result is supplied, and we still consider that it has provided the general `'Post'` tag type rather than any specific instance of that tag.
```
phryneas, line 148: Hmm. I'm not 100% sure if providing the general `['Post']` tag makes any sense here. That would invalidate only if an endpoint would invalidate *all* posts by `invalidatesTags: ['Post']`.
I'd rather have a `LIST` id in both results as well.

phryneas: PS: reading further I why it is still missing here though

phryneas (commented):
I would change the wording of the first few paragraphs a bit.

In my mind, it's more like "the cache data for this endpoint provides these tags" not "this endpoint provides these tags to the cache".

docs/usage/rtk-query/cached-data.md
```diff
@@ -0,0 +1,516 @@
...
+
+# Cached Data
+
+A key feature of RTK Query is it's management of cached data. When data is fetched from the server, RTK Query will store the data in the redux store as a 'cache'. When an additional request is performed for the same data, RTK Query will provide the cached data rather than sending an additional request to the server.
+
+RTK Query provides a number of concepts and tools to manipulate the cache behaviour and adjust it to your needs.
+
+## Definitions
+
+- _tags_ - Items used to identify the data present in the cache. An individual `tag` has a `type`, represented as a `string` name, and an optional `id`, represented as a `string` or `number`.
```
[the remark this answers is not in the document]
Shrugsy: I've relocated the definitions from the 'Mutation' page into this 'Cached data' page with some minor tweaks/additions, and added a 'note' section to the 'Providing cache data' section to better clarify the intent & behaviour of tags with a more explicit explanation

docs/usage/rtk-query/cached-data.md
Shrugsy: Yeah the goal was to try to introduce it as a general concept piece by piece first rather than immediately going into 'This is how you should be using it'.

I've added another 'tip' block in the 'Providing cache data' section linking to the 'Advanced Invalidation' section which hopefully is clear enough.

docs/usage/rtk-query/cached-data.md
```diff
@@ -0,0 +1,747 @@
...
+     { type: 'Post', id: 3 },
+   ]
+   ```
+4. The `editPost` mutation is fired off to alter a particular post
+5. Upon completion, RTK Query internally registers that the `'Post'` tag is now invalidated, and removes the previously provided `'Post'` tags from the cache
+6. Since the `getPosts` endpoint has provided tags of type `'Post'` which now has invalid cache data, and the component is still subscribed to the data, the `/posts` request is automatically fired off again, fetching new data and registering new tags for the updated cached data
+
+For more granular control over the invalidated data, invalidated `tags` can have an associated `id` in the same manner as `providesTags`. This enables a distinction between 'any of a particular tag type' and 'a specific instance of a particular tag type'.
+
+The example below declares that the `editPost` mutation provides a specific instance of a `Post` tag, using the ID provided when calling the mutation function:
```
phryneas, line 255:
```suggestion
The example below declares that the `editPost` mutation invalidates a specific instance of a `Post` tag, using the ID passed in when calling the mutation function:
```

docs/usage/rtk-query/cached-data.md
```diff
@@ -0,0 +1,747 @@
...
+In order to provide stronger control over invalidating the appropriate data, you can use an arbitrary ID such a `'LIST'` for a given tag. See [Advanced Invalidation with abstract tag IDs](#advanced-invalidation-with-abstract-tag-ids) for additional details.
+:::
+
+### Invalidating cache data
+
+Each individual `mutation` endpoint can `invalidate` particular tags for existing cached data. Doing so enables a relationship between cached data from one or more `query` endpoints and the behaviour of one or more `mutation` endpoints.
+
+The `invalidatesTags` property on a `mutation` endpoint is used for this purpose.
+
+The example below declares that the `addPost` and `editPost` `mutation` endpoints `invalidate` any cached data with the `'Post'` tag, using the `invalidatesTags` property for a `mutation` endpoint:
```
phryneas, line 192:
```suggestion
Each individual mutation endpoint can `invalidate` particular tags for existing cached data. Doing so enables a relationship between cached data from one or more query endpoints and the behaviour of one or more mutation endpoints.

The `invalidatesTags` property on a mutation endpoint is used for this purpose.

The example below declares that the `addPost` and `editPost` mutation endpoints `invalidate` any cached data with the `'Post'` tag, using the `invalidatesTags` property for a mutation endpoint:
```
this was just too difficult to read in the end :)

docs/usage/rtk-query/cached-data.md
```diff
@@ -0,0 +1,747 @@
...
+      <PostDetail id={2} />
+      <PostDetail id={3} />
+    </div>
+  )
+}
+```
+
+**What to expect**
+
+When `addPost` is fired, it will only cause the `PostsList` to go into an `isFetching` state because `addPost` only invalidates the 'LIST' id, which causes `getPosts` to rerun (because it provides that specific id). So in your network tab, you would only see 1 new request fire for `GET /posts`. As the singular `getPost` queries have not been invalidated, they will not re-run as a result of `addPost`.
```
phryneas, line 625:
```suggestion
When `addPost` is fired, it will only cause the `PostsList` to go into an `isFetching` state because `addPost` only invalidates the `'LIST'` id, which causes `getPosts` to rerun (because it provides that specific id). So in your network tab, you would only see 1 new request fire for `GET /posts`. As the singular `getPost` queries have not been invalidated, they will not re-run as a result of `addPost`.
```

phryneas (approved):
Gonna add some minor changes & then merge this in :+1:

phryneas:
Ahh damn conflicts. The joys of concurrent PRs :sweat_smile: 

@Shrugsy if you have the resources could you please fix those up and maybe also add TS to the examples here? 
Otherwise, I'll try but I probably won't get to it before thursday.

Shrugsy:
> Ahh damn conflicts. The joys of concurrent PRs 
> 
> @Shrugsy if you have the resources could you please fix those up and maybe also add TS to the examples here?
> Otherwise, I'll try but I probably won't get to it before thursday.

No problem, I'll get it fixed up when I have some time

PR #1041, opened 2021-05-05
prevent badly typed `Middleware<any>` from casting `dispatch` to `any`

phryneas:
Happened to myself: had a cast to `Middleware<any>` somewhere in the tests and suddenly all our `dispatch` were `any`, too. We should catch that.


phryneas:
merging this after #1034 should also fix those type tests

PR #1044, opened 2021-05-10
Custom meta suggestion

trmjoa:
Just quickly put together as a draft / discussion point for https://github.com/reduxjs/redux-toolkit/issues/750
Typescript types are (as you see) not defined and its not really tested in any way. Consider it psudo code.


phryneas:
Could you take a look over in https://github.com/reduxjs/redux-toolkit/pull/1083 if that fulfills your needs?

trmjoa:
Sorry for the late reply. Yes that is exactly what I had in mind! Thank you!

PR #1056, opened 2021-05-15
feature: allow using skipSymbol as arg in hooks 

phryneas:
This still needs tests.

This will resolve
* https://github.com/rtk-incubator/rtk-query/issues/101
* https://github.com/rtk-incubator/rtk-query/issues/220
 
Thanks @omerman for the suggestion.

Naming is as always up for debate. @omerman suggested `skipToken` which should also fit the bill fine.

@markerikson opinions?


src/query/core/buildSelectors.ts
```diff
@@ -17,7 +17,32 @@ import {
...
+ *
+ * ```ts
+ * // codeblock-meta title="using skipSymbol instead" no-transpile
+ * useSomeQuery(arg ?? skipSymbol)
+ * ```
+ *
+ * If passed directly into a query or mutation selector, that selector will always
+ * return an uninitialized state.
+ */
+export const skipSymbol = Symbol('skip selector')
```
phryneas, line 43: This should describe what it does ;)

Issue #964, opened 2021-03-30
[RTK Query Migration] Fill out remaining not-yet-covered RTKQ docs content


markerikson:
Depends on #963 .

Once the RTKQ source/docs content is in this repo, we need to update the RTK docs to include that info. My plan is:

- New `RTK Query` subcategory under "API Reference", with `createApi` and friends
- New "RTK Query" subcategory under "Usage Guide", containing the content previously in the RTKQ "Concepts" section

Could also argue it should be its own top-level section in some way (maybe just the "Concepts" material, but not the API reference?)

We'll need to emphasize that this is all exported by the new `"@reduxjs/toolkit/query"` and `"@reduxjs/toolkit/query/react"` entry points, not the main `"@reduxjs/toolkit"` entry point.

This should probably be done on a 1.6 integration branch so that we don't publish it too early.

markerikson:
Primary work was done in #1015 .

Going to leave this open to cover fleshing out the rest of the missing content.

markerikson:
Missing pieces from discussion:

- Could use an "RTK Overview" page in addition to the Tutorial and Usage Guide areas
- "Comparison" page
- Explanation of why we don't do normalization (put that on "Comparison", maybe?)
- I think some of the hooks aren't fully documented yet
- No TypeScript usage guide and insufficient types docs for the APIs

Shrugsy:
Just so it doesn't get missed:

The 'bailing out of errors' section on 'Error Handling' is still TODO also: https://deploy-preview-1016--redux-starter-kit-docs.netlify.app/usage/rtk-query/error-handling#bailing-out-of-errors

markerikson:
Per a user:

> It might be worth adding at least a one-line explanation of "reducerPath" in the Getting Started introduction, because it wasn't immediately clear to me what it does and I questioned if it was really even necessary. This section is more helpful, although it could also maybe add just a little bit more detail: https://rtk-query-docs.netlify.app/api/createapi/#reducerpath

Shrugsy:
Both the "bailing out of errors" section & "Need to cover how to use a custom base query" are intended to be covered by #1057

Shrugsy:
> Show switching from existing thunk-based fetching logic to RTKQ

Intended to be covered by #1060

Shrugsy:
> Could use an "RTK Overview" page in addition to the Tutorial and Usage Guide areas

This is intended to be covered by #1066 
[...]

PR #1057, opened 2021-05-16
Docs - add `Customizing Queries` page

Shrugsy:
Related to #964 

- Adds a 'Customizing Queries' page showcasing how to use different `baseQuery` functions for `createApi`, as well as `queryFn` on individual endpoints.
- Relocates existing examples of custom `baseQuery` to this page.


phryneas:
This looks pretty good. We should probably also add `Meta` in here.

Essentially, `Meta` is information that is not stored in `data` by default, but can be returned from the `baseQuery` to be accessed in `transformResponse`.

It *was* also available in `onSuccess` and `onError`, but that is not the case any more.

We *could* think of re-adding that in the `onQuery` lifecycle, but that would require us to include it in the thunk actions and I kinda don't want to add it to the `payload`.

That would be a very fine opportunity to do #844 - we now even have our own use case for that!

Shrugsy:
> This looks pretty good. We should probably also add Meta in here.

Definitely, I completely glossed over that. I've added a Meta example, and also updated the return types for `fetchBaseQuery` to include it's `meta` property.

The `transformResponse` usage I've added for the Meta example is definitely very contrived, but I wanted to show using `meta` in some way

docs/usage/rtk-query/customizing-queries.mdx
```diff
@@ -406,3 +405,96 @@ const api = createApi({
...
+// highlight-start
+const metaBaseQuery: BaseQueryFn<
+  string | FetchArgs,
+  unknown,
+  FetchBaseQueryError,
+  {},
+  Meta & FetchBaseQueryMeta
+> = async (args, api, extraOptions) => {
+  const requestId = uuid()
+  const timestamp = Date.now()
```
phryneas, line 453: requestId and timestamp are a bit unfortunate examples as both of those are already existing - the requestId should even be available here as `api.requestId`.
I can't really think of a better example though :sweat_smile: 
So I guess let's leave it like this now and see once a better example comes up in chat or in an issue :laughing:

Shrugsy: Yep, I'm definitely not happy with that example but it's all I could think of :sweat_smile: 

It should definitely be replaced when someone comes up with a better usage for custom `meta`

phryneas: Hmm. In the context of the "retry" wrapper.. maybe the amount of retries?

Shrugsy: Hmm that's a good idea. I'll get some more of your thoughts on it after I have a think on it to try to come up with a feasible example snippet

PR #1059, opened 2021-05-16
RFC: add "merge" functionality

phryneas:
This might be useful in "streaming data" scenarios where a `.refresh` call should not remove old data, but rather just merge current values into the existing dataset, e.g. as a "re-sync".

Not 100% sure about the usefulness though, hence the "RFC" in the title.

### DIFF
--- a/packages/toolkit/src/query/core/buildSlice.ts
+++ b/packages/toolkit/src/query/core/buildSlice.ts
@@ -27,6 +27,7 @@ import { calculateProvidedByThunk } from './buildThunks'
 import type {
   AssertTagTypes,
   EndpointDefinitions,
+  QueryDefinition,
 } from '../endpointDefinitions'
 import type { Patch } from 'immer'
 import { applyPatches } from 'immer'
@@ -156,11 +157,16 @@ export function buildSlice({
             meta.arg.queryCacheKey,
             (substate) => {
               if (substate.requestId !== meta.requestId) return
+              const { merge = (x: any) => x } = definitions[
+                meta.arg.endpointName
+              ] as QueryDefinition<any, any, any, any>
               substate.status = QueryStatus.fulfilled
+              let newData = merge(payload, substate.data)
+
               substate.data =
                 definitions[meta.arg.endpointName].structuralSharing ?? true
-                  ? copyWithStructuralSharing(substate.data, payload)
-                  : payload
+                  ? copyWithStructuralSharing(substate.data, newData)
+                  : newData
               delete substate.error
               substate.fulfilledTimeStamp = meta.fulfilledTimeStamp
             }
--- a/packages/toolkit/src/query/core/buildSlice.ts
+++ b/packages/toolkit/src/query/core/buildSlice.ts
@@ -6,6 +6,7 @@ import {
   isAnyOf,
   isFulfilled,
   isRejectedWithValue,
+  createNextState,
 } from '@reduxjs/toolkit'
 import type {
   CombinedState as CombinedQueryState,
@@ -157,16 +158,37 @@ export function buildSlice({
             meta.arg.queryCacheKey,
             (substate) => {
               if (substate.requestId !== meta.requestId) return
-              const { merge = (x: any) => x } = definitions[
+              const { merge } = definitions[
                 meta.arg.endpointName
               ] as QueryDefinition<any, any, any, any>
               substate.status = QueryStatus.fulfilled
-              let newData = merge(payload, substate.data)
 
-              substate.data =
-                definitions[meta.arg.endpointName].structuralSharing ?? true
-                  ? copyWithStructuralSharing(substate.data, newData)
-                  : newData
+              if (merge) {
+                if (substate.data !== undefined) {
+                  // There's existing cache data. Let the user merge it in themselves.
+                  // We're already inside an Immer-powered reducer, and the user could just mutate `substate.data`
+                  // themselves inside of `merge()`. But, they might also want to return a new value.
+                  // Try to let Immer figure that part out, save the result, and assign it to `substate.data`.
+                  let newData = createNextState(
+                    substate.data,
+                    (draftSubstateData) => {
+                      // As usual with Immer, you can mutate _or_ return inside here, but not both
+                      return merge(draftSubstateData, payload)
+                    }
+                  )
+                  substate.data = newData
+                } else {
+                  // Presumably a fresh request. Just cache the response data.
+                  substate.data = payload
+                }
+              } else {
+                // Assign or safely update the cache data.
+                substate.data =
+                  definitions[meta.arg.endpointName].structuralSharing ?? true
+                    ? copyWithStructuralSharing(substate.data, payload)
+                    : payload
+              }
+
               delete substate.error
               substate.fulfilledTimeStamp = meta.fulfilledTimeStamp
             }
--- a/packages/toolkit/src/query/endpointDefinitions.ts
+++ b/packages/toolkit/src/query/endpointDefinitions.ts
@@ -275,6 +275,18 @@ export interface QueryExtraOptions<
   invalidatesTags?: never
 
   serializeQueryArgs?: SerializeQueryArgs<any>
+
+  /**
+   * Can be provided to merge the current cache value into the new cache value.
+   *
+   * Useful if you don't want a new request to completely override the current cache value,
+   * maybe because you have manually updated it from another source and don't want those
+   * updates to get lost.
+   */
+  merge?(
+    newCacheValue: ResultType,
+    currentCacheValue: ResultType | undefined
+  ): ResultType
 }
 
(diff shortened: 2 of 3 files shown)


aldreth:
@phryneas This is what I came to github to ask about! I currently use a thunk to merge new data into state & can't see a way of replicating that with RTK Query.

phryneas:
@aldreth Currently I'm at the point of "almost all usecases I can think of go against the idea of RTKQ itself and would probably be more dangerous that useful for almost everyone". But with enough good use cases I might change my mind. Can you elaborate a bit on your usecase?

aldreth:
@phryneas I am basically showing a list of events. The API, which I don't control, drops doesn't return today's events. I save the redux state into local storage, so if a user has previously visited the site I can merge new events into state and still show events for today.
I can see that it's an edge case - the right thing to do would be to update the API, but it's an local government thing, so I have no control over it.

phryneas:
In that case I'd recommed to use an `extraReducer` in a slice you have control over to save that data to the side. You should not persist an RTKQ cache slice.

ryota-murakami:
Is this will become default strategy of refetch()?
Actually common use case doe't affect this change i think.

phryneas:
@ryota-murakami no, this will not be the default - the default will still be replacing the last result. 
This will be an edge case feature that should be used with extreme care - for example in an endless scrolling situation it could easily lead to memory leaks if the user does not clean up old entries. The cache entry would just grow and grow.

Even for endless scrolling, it would still be advisable for almost all situations to use multiple `useQuery` hook calls, as that way RTK-Q can do the cleanup.

ryota-murakami:
@phryneas Thanks, that's make sense!

markerikson:
Here's an example of where this might be useful, from Replay's codebase, along with the possibility of customizing cache key generation.

Replay stores info on "source actors". I'm still kinda hazy on the difference between a "source" and a "source actor", but we have a reducer that stores an entry per "source actor" along with its async request status, and that entry has a few different fields, including details on "breakable lines" and "hit counts per line":

https://github.com/replayio/devtools/blob/61525ea3f097926435d6d03f94a7c9895cca71a8/src/devtools/client/debugger/src/reducers/source-actors.ts#L30-L44

We fetch the basic data up front, and then fill in more of that detail later based on user interactions (hovering over a line number fetches hit counts, etc).

The current fetching logic is really ugly and uses an internal abstraction called a "memoizableAction":

https://github.com/replayio/devtools/blob/61525ea3f097926435d6d03f94a7c9895cca71a8/src/devtools/client/debugger/src/actions/source-actors.ts#L85-L119

and the corresponding reducer logic then has to figure out:

- Do we have an entry for this ID?
- what's its status?
- okay, now use an "update resource" util to find that item and update these couple fields, including merging the hit counts data into an existing array:

https://github.com/replayio/devtools/blob/61525ea3f097926435d6d03f94a7c9895cca71a8/src/devtools/client/debugger/src/reducers/source-actors.ts#L195-L228

```ts
function updateBreakpointHitCounts(
  state: SourceActorsState,
  action: SetSourceActorBreakpointHitCountsAction
) {
  // @ts-ignore start/done mismatch
  const value = asyncActionAsValue<SetSourceActorBreakpointHitCountsAction["value"]>(action);
  if (value.state === "pending") {
    return state;
  }

  const { id: sourceId } = action;

  if (!hasResource(state, sourceId)) {
    return state;
  }

  const {
    breakpointHitCountLinesKnown: currentLinesKnown,
    breakpointHitCounts: currentBreakpointHitCounts,
  } = {
    ...state.values[sourceId],
  };

  return updateResources(state, [
    {
      breakpointHitCountLinesKnown: mergeRanges([
        ...(currentLinesKnown || []),
        { max: action.value.max, min: action.value.min },
      ]),
      breakpointHitCounts: [...(currentBreakpointHitCounts || []), ...action.value.hits],
      id: sourceId,
    },
  ]);
}
```

This all _feels_ like code that could potentially be replaced by RTK Query endpoint handling... except that we don't have a way to merge in more data over time.

markerikson:
I've rebased this PR and would like to finalize it. But, I don't think the behavior in this initial implementation is what we'd want.

Right now it's:

```ts
              const { merge = (x: any) => x } = definitions[
                meta.arg.endpointName
              ] as QueryDefinition<any, any, any, any>
              substate.status = QueryStatus.fulfilled
              let newData = merge(payload, substate.data)

              substate.data =
                definitions[meta.arg.endpointName].structuralSharing ?? true
                  ? copyWithStructuralSharing(substate.data, newData)
                  : newData
```

If you're supplying a `merge` option, that means you're now trying to mutate `substate.data`.  which means that at that point `newData` and `structuralSharing` are a moot point, right?

I feel like if you've supplied a `merge` callback, it's now your responsibility to do the update work, and we should make that mutually exclusive with the replacement or structural sharing behavior.

markerikson:
I successfully updated the logic to look like this:

```ts
if (merge) {
  if (substate.data !== undefined) {
    // There's existing cache data. Let the user merge it in themselves.
    // We're already inside an Immer-powered reducer, and the user could just mutate `substate.data`
    // themselves inside of `merge()`. But, they might also want to return a new value.
    // Try to let Immer figure that part out, save the result, and assign it to `substate.data`.
    let newData = createNextState(
      substate.data,
      (draftSubstateData) => {
        // As usual with Immer, you can mutate _or_ return inside here, but not both
        return merge(draftSubstateData, payload)
      }
    )
    substate.data = newData
  } else {
    // Presumably a fresh request. Just cache the response data.
    substate.data = payload
  }
} else {
  // Assign or safely update the cache data.
  substate.data =
    definitions[meta.arg.endpointName].structuralSharing ?? true
      ? copyWithStructuralSharing(substate.data, payload)
      : payload
}
```

and added a couple tests.

This looks pretty good to me.

packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -156,11 +158,37 @@ export function buildSlice({
...
-                  ? copyWithStructuralSharing(substate.data, payload)
-                  : payload
+
+              if (merge) {
+                if (substate.data !== undefined) {
+                  // There's existing cache data. Let the user merge it in themselves.
+                  // We're already inside an Immer-powered reducer, and the user could just mutate `substate.data`
+                  // themselves inside of `merge()`. But, they might also want to return a new value.
+                  // Try to let Immer figure that part out, save the result, and assign it to `substate.data`.
+                  let newData = createNextState(
```
phryneas, line 172: The nested `createNextState` is not causing any problems? I do remember *something* vaguely that this had caused problems in the past.

packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -156,11 +158,37 @@ export function buildSlice({
...
+                    substate.data,
+                    (draftSubstateData) => {
+                      // As usual with Immer, you can mutate _or_ return inside here, but not both
```
phryneas, line 175: unnecessary `return` if you omit the method body. I don't know if minifiers would eliminate that, but why take the chance they don't? :)

packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -156,11 +158,37 @@ export function buildSlice({
...
+                      return merge(draftSubstateData, payload)
+                    }
+                  )
+                  substate.data = newData
+                } else {
```
phryneas, line 180: this case could be skipped if you test for `substate.data` up there together with the `if (merge) {` - the fallthrough to `copyWithStructuralSharing` would be totally fine

packages/toolkit/src/query/endpointDefinitions.ts
```diff
@@ -275,6 +275,25 @@ export interface QueryExtraOptions<
   invalidatesTags?: never
 
   serializeQueryArgs?: SerializeQueryArgs<any>
+
+  /**
+   * Can be provided to merge the current cache value into the new cache value.
+   * If supplied, no automatic structural sharing will be applied - it's up to
```
phryneas, line 281: :100: for mentioning this. I hope people read it.

packages/toolkit/src/query/tests/buildSlice.test.ts
```diff
@@ -29,6 +29,10 @@ const authSlice = createSlice({
 
 const storeRef = setupApiStore(api, { auth: authSlice.reducer })
 
+function delay(ms: number) {
```
phryneas, line 32: see `waitForFakeTimer` from `helpers.ts`

packages/toolkit/src/query/core/buildSlice.ts
markerikson: Yeah, Immer had some issues with nested drafts in earlier versions. That behavior has been made more consistent now.  I did some testing here with both mutate and return to make sure they work, and also confirmed it throws if you mutate + return.

PR #1060, opened 2021-05-16
Docs - add 'Migrating to RTK Query' page

Shrugsy:
Related to #964 

- Adds a ~~'Migrating from thunks'~~ 'Migrating to RTK Query' page

Note that realistically it is more like 'How to delete your thunk code and re-write it shorter with RTK Query'

### DIFF
--- a/docs/usage/rtk-query/migrating-from-thunks.mdx
+++ b/docs/usage/rtk-query/migrating-from-thunks.mdx
@@ -1,0 +1,312 @@
+---
+id: migrating-from-thunks
+title: Migrating from thunks
+sidebar_label: Migrating from thunks
+hide_title: true
+---
+
+# Migrating from thunks
+
+While RTK Query is not intended to be a replacement for thunks, it _is_ expected to cover a lot of overlapping behaviour that users may have previously used `createAsyncThunk` for, including caching purposes, and request lifecycle managment (e.g. `isUninitialized`, `isLoading`, `isError` states).
+
+In order to migrate from standalone thunks + reducers to RTK Query, the appropriate endpoints should be added to an RTK Query api, and the existing thunks & reducers deleted. This generally will not include much common code kept between the two, as they work differently and one will replace the other.
+
+## Examples
+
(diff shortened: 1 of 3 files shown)


markerikson:
A couple suggestions:

First, let's retitle it "Migrating to RTK".  I don't think we need to specifically show examples of switching from sagas and observables also, but let's keep the title more general.

Next, I'd suggest starting with a brief recap like:

> The most common use case for side effects in Redux apps is fetching data.  Redux apps typically use a tool like thunks, sagas, or observables to make an AJAX request, and [dispatch actions based on the results of the request](https://redux.js.org/tutorials/fundamentals/part-7-standard-patterns#async-request-status).  Reducers then listen for those actions to manage loading state and cache the fetched data.
> 
> RTK Query is purpose-built to solve the use case of data fetching.  While it can't replace all of the situations where you'd use thunks or other side effects approaches, using RTK Query should eliminate the need for most of that hand-written side effects logic.

Shrugsy:
Thanks @markerikson, I've made those changes

markerikson:
Content-wise, this doesn't really feel like a "migrating" page at the moment.

The examples section is more like "here's two separate examples, one without RTKQ and one with RTKQ".  I'd like to see more of a "here's our starting point with just RTK, and here's the individual steps I'd take to convert the RTK code to RTKQ", similar to what I have in https://redux.js.org/tutorials/fundamentals/part-8-modern-redux for going from handwritten Redux logic to RTK.

Shrugsy:
@markerikson I've re-written it with a more 'step-based' approach. Let me know your thoughts

Shrugsy:
Actually I've just realised this snippet is wrong (regarding the 'status' check)

```ts
    const response = await fetch(
      `https://pokeapi.co/api/v2/pokemon/${name}`
    ).then((res) => res.json())
    if (response.status < 200 || response.status >= 300) {
      return rejectWithValue(response)
    }
```
I'll fix that up later on

markerikson:
Looks fairly good. We could probably make more improvements, but at this point I just want to get some reasonable content in place for any remaining topics so we can launch.

PR #1066, opened 2021-05-18
Docs - add RTK Query content to `Getting Started` page

Shrugsy:
Related to #964 

Adds some additional detail regarding RTK Query to the 'Getting Started' page


markerikson:
Netlify build is failing due to busted links:

```
9:00:15 AM: Exhaustive list of all broken links found:
9:00:15 AM: 
9:00:15 AM: - On source page path = /introduction/getting-started:
9:00:15 AM:    -> linking to ./getting-started-rtk-query.mdx (resolved as: /introduction/getting-started-rtk-query.mdx)
```

markerikson:
Hmm. I know I had a couple vague thoughts in my head for what an "Overview" page would contain, but I'm not sure what those were atm.  I think I was picturing something that was more of a cross between the "RTKQ Tutorial" page and a "why should we do this data fetching stuff via RTKQ anyway?" writeup.

I don't think this is quite what I was picturing, but it's good material anyway.

I'd actually suggest we just consolidate the RTKQ-specific content here into the existing "Getting Started" page as an additional section - no need to repeat all the install info.

Shrugsy:
> I'd actually suggest we just consolidate the RTKQ-specific content here into the existing "Getting Started" page as an additional section - no need to repeat all the install info.

That's a good idea, I'll change it up like that

markerikson:
Looks better! Not sure "Installation" is the right word for that header, since we're not actually installing anything. Can't come up with a good alternate header phrase atm - maybe just drop that "Installation" header entirely and have that code block be directly after the preceding couple of paragraphs?

Shrugsy:
> Looks better! Not sure "Installation" is the right word for that header, since we're not actually installing anything. Can't come up with a good alternate header phrase atm - maybe just drop that "Installation" header entirely and have that code block be directly after the preceding couple of paragraphs?

Ah, I had the same reservations as you trying to come up with a more accurate header, but didn't even consider "no header" as an option. I was thinking it was *roughly* accurate in the sense that it is "here is how to install: it's already installed". But forgoing the header is definitely better

markerikson:
LGTM!

I do plan on doing a typos/editing pass over the docs before release, so I'm not doing much review on those atm.

PR #1071, opened 2021-05-19
Docs - expand explanation of cache lifetime and subscription behavior

Shrugsy:
Related to #964 

Expands on the explanation of cache lifetime and subscription behavior slightly, as well as adding an example scenario to assist with the explanation of cache lifetime & subscription behaviour

Yoinked from @markerikson's explanation here: https://www.reddit.com/r/reactjs/comments/nej4sv/redux_to_store_a_lot_of_data/gykyzms/


markerikson:
I approve of this example that I came up with :)

PR #1074, opened 2021-05-19
Docs - extend detail for `Queries` & `endpoints`

Shrugsy:
Related to #964 

- Add brief descriptions to the query hooks
- Add brief descriptions to the query hook options
- Add brief tip about `skipToken` on the `Conditional Fetching` page
- Shift 'Anatomy of an endpoint' into a larger section at end of page
- Flesh out property details & examples in 'anatomy of an endpoint'
- Miscellaneous order re-arranging
- Update `defaultSerializeQueryArgs` shown in docs as per actual version
- Update links to specific sections of endpoint anatomy where applicable


docs/api/rtk-query/createApi.mdx
```diff
@@ -237,7 +234,98 @@ export const { endpoints, reducerPath, reducer, middleware } = api
...
+
+[examples](docblock://query/createApi.ts?token=CreateApiOptions.reducerPath)
+
+### `serializeQueryArgs`
+
+[summary](docblock://query/createApi.ts?token=CreateApiOptions.serializeQueryArgs)
+
+Defaults to:
+
+```ts no-transpile
```
markerikson, line 255: :question: Do we really need to copy the entire implementation here? Seems like a written description of behavior would be sufficient.

markerikson (approved):

docs/api/rtk-query/createApi.mdx
Shrugsy: Hmm I agree. A written description would certainly cover it.

However, can you think of a theoretical example for a custom version of `serializeQueryArgs` someone might want to use rather than the default? If so, I think that would be beneficial as a snippet

Shrugsy:
@markerikson following on from https://github.com/reduxjs/redux-toolkit/issues/964#issuecomment-846342406

I've removed the 'query hook return types' & 'mutation hook trigger type' codeblocks, and instead replaced them with written descriptions/bullet points like so:

https://deploy-preview-1074--redux-starter-kit-docs.netlify.app/usage/rtk-query/queries#frequently-used-query-hook-return-values
https://deploy-preview-1074--redux-starter-kit-docs.netlify.app/usage/rtk-query/mutations#frequently-used-mutation-hook-return-values

Let me know if you think any of them probably wouldn't "frequently used" enough to warrant being there, or if you have any other recommendations

Shrugsy:
@akursat please let me know if you have any opinions on these changes, following your comment: https://github.com/reduxjs/redux-toolkit/issues/964#issuecomment-846290295

Note that both the query & mutation endpoint definitions are proposed to move to [`createApi#endpoints`](https://deploy-preview-1074--redux-starter-kit-docs.netlify.app/api/rtk-query/createApi#endpoints) as part of this PR

docs/api/rtk-query/createApi.mdx
Shrugsy: FYI I've replaced this with a written description, but I can't think of a realistic example someone would use to override it

markerikson:
Do it.

PR #1078, opened 2021-05-21
nest data structure for `cacheDataLoaded` and `queryFulfilled`

phryneas:
...to prepare for adding more information in the resolved/rejected values in future releases

~~This still needs docs updates, but the code could already be reviewed.~~
Nice, this was never mentioned in the docs :sweat_smile:


phryneas:
Let's get this in

PR #1079, opened 2021-05-23
Docs - restructure rtk-query docs organisation

Shrugsy:
Related to #964

Per feedback in the [twitter poll here](https://twitter.com/acemarke/status/1395023250212655105) and other discussions, restructures the RTK Query `Usage` & `API Reference` portions of the docs to reduce nesting, and make them more accessible, by adding a single top level `RTK Query` section, with `Using RTK Query` & `API Reference` as sub-sections within that


Shrugsy:
FYI the structure is proposed like so with `RTK Query` at the bottom:

image

Which when expanded, has `API Reference` expanded by default, but not `Using RTK Query`:  

image

If anyone has opinions on whether any of that should change regarding what is collapsed/expanded by default, please voice them

markerikson:
Mmm... I'd actually lean towards keeping all the API Ref sections collapsed or something, perhaps like this:

image

I also want to add the RTKQ "Overview" and "Comparison" pages within that top-level section, so I think it makes sense to keep the nested "Using" and "API Ref" sections closed.

markerikson:
Let's get this in and keep iterating.

Issue #924, opened 2021-03-12
`createAsyncThunk`: How can I either replace the `dispatch` call used, or attach meta data or a middleware for all actions?


jmccarrell-lmi:
I'd prefer to use the dispatch provided by `createAsyncThunk`, but it's too vanilla.

I have a dispatch that adds a meta tag, which is picked up by a middleware, which strips the tag, dresses up the action, and sends it over a websocket.

I'd like to have ALL dispatched actions [including pending, complete, etc.] to be dressed this way, but I'm not sure how I could do that using `createAsyncThunk`.

I currently have a hook to get my enhanced dispatch [which simply uses useDispatch and attaches meta data].
The simplest solution for me would be to provide that hook when i call `createAsyncThunk`. It would be leak react implementation into Redux, if it accepted a hook, so instead I'd probably need to abstract my hook...

Side note: I'm finally feeling the downside of hooks, how they deeply couple you with the React implementation. React was originally an agnostic stand-alone framework.

markerikson:
I'm not quite sure I follow what you're trying to do here. Can you show some sample code?

phryneas:
If you want to attach some meta to all your actions, the best place for that would be a middleware, as every action always goes through all middlewares before it reaches the store.

You can also wrap the store you pass into the `Provider` to have a different `dispatch`, so you would get that out of every `useDispatch` and `connect`, but not injected into a middleware like `thunk`. Also, it seems like a very unclean way of doing things.

What exactly are you trying to achieve here?

markerikson:
Seems like this is a dupe of #750 , best as I can tell.

markerikson:
Resolved by #1083 .

PR #1083, opened 2021-05-23
createAsyncThunk: support for `meta`

### LINKED ISSUE #750: Setting "meta" on the action generated by "createAsyncThunk"

phryneas:
Gotta add docs & more type tests for different scenarios on this one, but this would essentially be it.

This should close #750, #924, #1044

@markerikson can you take a look if this covers all your thoughts?


phryneas:
Damn, that failed fast :sweat_smile:

phryneas:
Added a bunch of docs for now, we can always add more later.

Issue #1080, opened 2021-05-23
Get response headers in onCacheEntryAdded?


TamasSzigeti:
Hi,

Thanking muchly for this great lib.
I've done some custom implementation of streaming updates in alpha.2, using onSuccess. In my api response I advertise the subscribe url in a http link header, and I have used the meta.response there to get it. 
Now I was excited to see first class support for streaming updates but I can see that the meta argument is unavailable in onCacheEntryAdded. Is there any way we could provide it in the callback, or can you recommend a replacement for this behaviour?

TIA

phryneas:
Unfortunately, it didn't make it into this beta. It's on the roadmap, which is why the response value of `cacheDataLoaded` is shaped like `{ data: ResultType }` (so that `meta` can be added there, too), but it requires bigger adjustments in `createAsyncThunk` to get this working correctly. I hope we can still get this into the 1.6 release, but I can't give any guarantees yet. Worst case it would be a 1.7 that would have a much shorter release cycle than our last few releases.

TamasSzigeti:
Thanks for the quick response! The problem is that the meta argument in onSuccess is now typed as undefined, so with the beta even my original solution seems to be broken. Is it maybe possible to at least recover the meta there the way it worked in alpha?

phryneas:
Unfortunately not, since `onSuccess` is now just polyfilled using `onCacheEntryAdded` for migration purposes. But good news: I'm about halfway through the required changes of `createAsyncThunk`, so assuming everything works out you'll maybe even today have a codesandbox build with `meta` support. Until that, you'll have stay on the old alpha, sorry!

TamasSzigeti:
That's amazing! Sure, no rush though, the alpha is stable enough and works pretty well with the custom streaming update implementation, I just wanted to make sure I will eventually be able to use the official callbacks. Thanks again!

markerikson:
Should be resolved by #1084 .

PR #1084, opened 2021-05-23
add baseQueryMeta to thunks, meta to lifecycle results

phryneas:
This contains #1083 and is a definite WIP, just want to see if it works for #1080

@TamasSzigeti could you give the CodeSandbox build of this a spin? Types are still missing, but `meta` is *on* there, so you might have to do an `any`-cast to get it out there atm. if working with TS

Install instructions at https://ci.codesandbox.io/status/reduxjs/redux-toolkit/pr/1084/

### DIFF
--- a/src/createAsyncThunk.ts
+++ b/src/createAsyncThunk.ts
@@ -366,7 +366,9 @@ export type AsyncThunkRejectedActionCreator<
     aborted: boolean
     condition: boolean
   } & (
-    | { rejectedWithValue: false }
+    | ({ rejectedWithValue: false } & {
+        [K in keyof GetRejectedMeta<ThunkApiConfig>]?: undefined
+      })
     | ({ rejectedWithValue: true } & GetRejectedMeta<ThunkApiConfig>)
   )
 >
--- a/src/query/baseQueryTypes.ts
+++ b/src/query/baseQueryTypes.ts
@@ -59,7 +59,7 @@ export type BaseQueryMeta<BaseQuery extends BaseQueryFn> = UnwrapPromise<
 
 export type BaseQueryError<BaseQuery extends BaseQueryFn> = Exclude<
   UnwrapPromise<ReturnType<BaseQuery>>,
-  { error: undefined }
+  { error?: undefined }
 >['error']
 
 export type BaseQueryArg<
--- a/src/query/core/buildInitiate.ts
+++ b/src/query/core/buildInitiate.ts
@@ -5,7 +5,12 @@ import type {
   QueryArgFrom,
   ResultTypeFrom,
 } from '../endpointDefinitions'
-import type { QueryThunkArg, MutationThunkArg } from './buildThunks'
+import type {
+  QueryThunkArg,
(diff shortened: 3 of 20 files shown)


TamasSzigeti:
Yes I'm using TypeScript and seems the parts relying on createAsyncThunk are having type errors now. Specifically in extraReducers where I'm using `builder.addCase(someThunk.fulfilled, (state, { payload }) => { … })`, it results in `payload` being typed as `unknown`.
I can try to spend more time on it tomorrow to understand it better.

phryneas:
@TamasSzigeti I have added a test for that over at https://github.com/reduxjs/redux-toolkit/pull/1083/commits/64a8c64d6cc2ca035d93aa55e79b92cb3534341c#diff-a11e9f94b91b0761740b0b3daf0f10c2b45a1881e554e229914dc23d146826e1R104 but I cannot really reproduce that behaviour. Can you please create a reproduction?

TamasSzigeti:
> @TamasSzigeti Can you please create a reproduction?

@phryneas I trimmed down my case, and it seems that the unknown payload comes about when I pass a second argument to the payloadCreator, whether I want to use `dispatch` or `getState`. When I leave that argument out, as the case in the tests, the return type is propagating just fine.

https://gist.github.com/TamasSzigeti/2ddade02d2e647d25d8810da04dba494

phryneas:
Thank you @TamasSzigeti !
That one was actually super tricky to find & solve, but I think I got it. Can you try it again?

Thanks for the report :+1:  :smiley:

TamasSzigeti:
> That one was actually super tricky to find & solve, but I think I got it. Can you try it again?

Nice, thank you! My case works perfectly having re-implemented it with `onCacheEntryAdded`. Meta is there, only the typing is missing.
Plus the serializability middleware is complaining about the meta field in the `api/executeQuery/fulfilled` action payload:
```
A non-serializable value was detected in an action, in the path: `meta.baseQueryMeta.request`
```

phryneas:
Yup, the types will be the next step :) Thanks for validating!

phryneas:
Aaand we have types.

TamasSzigeti:
Awesomeness, and it works perfectly, same types as with onSuccess before. One thing that might worth adding to the docs is a note about when can the meta come through as undefined. I guess it has to do with an overridden queryFn?

Thanking muchly, this is becoming nice and clean, very pleasant to work with.

phryneas:
@TamasSzigeti Those are always optional because I did not want to add multiple generic arguments here for something that is more of an "edge case" use.
Depending on your `baseQuery`, it might only return `meta` for successes, only for errors, or even none or different ones. I've put these into one "combined" generic argument to reduce complexity of the types (which already are insane anyways).

In the case of `fetchBaseQuery`, they should always be there with the maximum available information unless some unexpected error happened there and it actually crashed.

TamasSzigeti:
@phryneas Thanks, my interpretation was correct then. My point is that this could be useful to be added in the docs.

PR #1088, opened 2021-05-25
Docs - add usage with typescript page

Shrugsy:
Related to #964 

Adds a `usage with typescript` page.


Shrugsy:
RFC:  Is this location in the sidebar appropriate, or would this page warrant being up one level, in line with `RTK Query Overview` and `Comparison with Other Tools`?

image

Shrugsy:
RFC: Can anyone think of anything else missed that should be covered on this page?

markerikson:
I'm actually going to say we maybe oughta move it up to the upper level.

I know that's a bit inconsistent in terms of "usage" location with our page for the RTK core (and even the Redux core), but given how many people will likely want to use RTKQ + TS together, I'd rather not have it buried in the middle of the other pages.

PR #1092, opened 2021-05-26
Docs - split up cache pages

Shrugsy:
Splits up the 'Cached Data' page into 'Cache Behavior' and 'Automated Re-fetching'.

Note that the line is still a bit blurry between them: 'Cache Behavior' now has a section for 'Manipulating Cache Behavior', which includes a lot of things that would be considered as fitting 'automated re-fetching' as well.

However, the main goal is to have the 'automatic re-fetching via cache tags' concept on it's own page. Hopefully the two pages reference each other enough in the appropriate locations so that no relevant information gets missed.


msutkowski (approved):
Wooo, lots of reading material here! Looks good though and makes sense.

Issue #1099, opened 2021-05-28
Fix undefined error in quick-start tutorial


jonathanbutler7:
When following the `docs/tutorials/quick-start.md` tutorial, passing in`(state) => state.counter.value` to useSelector returns an error: 
```
TypeError: Cannot read property 'value' of undefined
```

However, passing in `(state) => state.counterReducer.value` resolves the error.

markerikson:
The tutorial should be correct as written. Based on the error you're describing, I'm pretty sure I know what the issue is in your code.  

The store setup _should_ look like:

```js
export default configureStore({
  reducer: {
    counter: counterReducer,
  },
})
```

Note that the variable name is `counterReducer`, but we're passing it as a field named `counter`.  That is why we'll end up with a root state field of `state.counter`.

If you're getting `state.counterReducer` in your own app, it's because you did something like this:

```js
export default configureStore({
  reducer: {
    counterReducer,
  },
})
```

That will create a root state field of `state.counterReducer`, which is not what the tutorial is asking you to do.

PR #1100, opened 2021-05-28
Fix undefined error in quick-start tutorial

jonathanbutler7:
# This PR is in reference to [issue 1099](https://github.com/reduxjs/redux-toolkit/issues/1099).

**Problem**: When following the `docs/tutorials/quick-start.md` tutorial, passing in`(state) => state.counter.value` to useSelector returns an error: 
```
TypeError: Cannot read property 'value' of undefined
```

**Solution**: Passing in `(state) => state.counterReducer.value` resolves the error.


markerikson:
Appreciate the PR, but looks like it's probably an error in your code.

PR #1101, opened 2021-05-29
Docs/Examples  move quick start examples to repo

Khartir:
I added the `ApiProider` to the advanced example and left the normal `Provider`/`store` in the basic example. Not sure if that is too early, what do you think?


msutkowski:
> I added the `ApiProider` to the advanced example and left the normal `Provider`/`store` in the basic example. Not sure if that is too early, what do you think?

@Khartir  Thanks for doing this work! I'd prefer to keep the `ApiProvider` usage restricted to its specific example as it isn't going to be the normal pattern, and we probably don't want to show mixed configs at this point in the getting started portion. It could imply that something else needs to be different here, which is not the case.

If you're okay with it, I'll go ahead and merge this and we can revisit in a cleanup pass once all of the examples are in. Also, are you planning on pushing up any more examples today? If not, I'm going to pick up where you've left off here. Thanks!

msutkowski:
> > I added the `ApiProider` to the advanced example and left the normal `Provider`/`store` in the basic example. Not sure if that is too early, what do you think?
> 
> @Khartir Thanks for doing this work! I'd prefer to keep the `ApiProvider` usage restricted to its specific example as it isn't going to be the normal pattern, and we probably don't want to show mixed configs at this point in the getting started portion. It could imply that something else needs to be different here, which is not the case.
> 
> If you're okay with it, I'll go ahead and merge this and we can revisit in a cleanup pass once all of the examples are in. Also, are you planning on pushing up any more examples today? If not, I'm going to pick up where you've left off here. Thanks!

Spoke with Lenz about this - we'll use the `ApiProvider` on the two getting started examples, and the rest will be the 'standard' config.

phryneas:
@msutkowski actually I meant it the other way round: `ApiProvider` in the advanced examples since we don't need to show the full setup all the time and `configureStore` in the basic examples where we really want to show how to do a setup.
But in the end I don't really mind either way :D

msutkowski:
> @msutkowski actually I meant it the other way round: `ApiProvider` in the advanced examples since we don't need to show the full setup all the time and `configureStore` in the basic examples where we really want to show how to do a setup.
> But in the end I don't really mind either way :D

I'm going to finish the last 2, and then we can do a cleanup pass. Swapping the setup will only take a minute in most cases 🎉

PR #1103, opened 2021-05-29
add `readonly` to array-type function parameters where mutation is not intended

phryneas:
That will allow all those functions to be called with array variables that were defined with `as const`.


markerikson:
Seems fine other than the fact that all the TS builds are now failing :)

phryneas:
Uh yeah. Tests to be fixed tomorrow apparently

markerikson:
I patched up the entity adapter type tests - duplicated the lines that were breaking, and updated each pair to have one `ReadonlyArray<>` and one `// @ts-expect-error`.

All TS tests now pass except for `ts@next`.

PR #1109, opened 2021-05-30
Add GraphQL example

msutkowski:


examples/query/react/graphql/src/mocks/db.ts
```diff
@@ -0,0 +1,71 @@
...
+        data: {
+          posts
+        } as { posts: Post[] },
+        per_page,
+        page,
+        total_pages: Math.ceil(db.post.count() / per_page),
+        total: db.post.count(),
+      })
+    )
+  }),
```
msutkowski, line 69: This seems strange to me, but unfortunately, even though `@mswjs/data` supports pagination on the autogenerated handlers, it doesn't return any pagination information so we have to do this type of thing.

msutkowski (commented):
@phryneas If you get time to patch this so it's a more realistic implementation, that'd be great. This seems close, but wonky :sweat:

phryneas:
We *could* ass graphql codegen to generate the types here and provide a more "realistic" example... lemme try

examples/query/react/graphql/src/app/services/posts.ts
```diff
@@ -0,0 +1,79 @@
...
+export const api = createApi({
+  baseQuery: graphqlBaseQuery({
+    baseUrl: "/graphql"
+  }),
+  endpoints: (builder) => ({
+    getPosts: builder.query<GetPostsResponse, number | void>({
+      query: (page) => ({
+        body: gql`
+          query GetPosts($page: Int, $per_page: Int) {
+            posts(page: ${page}, per_page: 10) {
```
phryneas, line 49:
```suggestion
            posts(page: $page, per_page: 10) {

```

PR #1111, opened 2021-05-30
more advanced graphql example with code generation

phryneas:
This could either be integrated or maybe also just "it's own example".

It adds graphql codegen and a generated introspection schema for the apollo vscode extension. I maybe got carried away a little bit, but this would be a more "realistic" example.
This requires https://github.com/mswjs/data/pull/96 to get merged first and I do agree with @msutkowski by now that this should probably expose a more [relay-like](https://relay.dev/graphql/connections.htm#sec-undefined.PageInfo.Fields) pagination, but that should probably be solved over in https://github.com/mswjs/data.


msutkowski:
This looks good to me, but I think this should probably be kept as a 2nd example as mentioned (`graphql-with-codegen`) as it introduces a few more concepts to understand. I'll see what I can do to get the `msw` PR moved along.

PR #1114, opened 2021-05-30
🐛 Fix bad type for prefetch thunk action creator

Shrugsy:
- Fixes the `endpointName` arg being typed as `never`

#### Description

When using the `prefetch` thunk action creator manually (not the generated `usePrefetch` hook), the first argument `endpointName` is typed as `never`.

#### Expected behaviour

The type for `endpointName` should allow one of the `queryKeys` based on the endpoints defined on the corresponding api.

#### Version

1.6.0-beta.0

#### Reproduction example

https://codesandbox.io/s/rtkq-apiutilprefetch-type-bug-ksc27?file=/src/api.ts:460-519

```
Argument of type 'string' is not assignable to parameter of type 'never'.ts(2345)
```

### DIFF
--- a/src/query/core/buildThunks.ts
+++ b/src/query/core/buildThunks.ts
@@ -193,7 +193,7 @@ export function buildThunks<
   reducerPath: ReducerPath
   context: ApiContext<Definitions>
   serializeQueryArgs: InternalSerializeQueryArgs
-  api: Api<BaseQuery, EndpointDefinitions, ReducerPath, any>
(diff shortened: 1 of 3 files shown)


Shrugsy:
cool, I didn't know that the codesandbox bot creates a build using a sandbox from the description

phryneas:
Yup, CSB is *really* useful :)

Thanks a ton for the fix - let's get this in!

Issue #1023, opened 2021-04-26
Final RTKQ alpha feedback, fixes, and tweaks


markerikson:
Gotta resolve any issues that get reported during the alpha, and apply any final changes left over from the old repo.

mgcrea:
Can't wait to test RTK-Query in my new projects!

Minor note, in the changelog the import path is `import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/react'` when it looks like it should be `import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react'`.

phryneas:
@mgcrea  Thanks, updated it.

markerikson:
Yeah, we need to go through and add visibility attributes to all types so we stop getting extraction warnings.

markerikson:
Running list of merged PRs since `alpha.2` for reference:

- #1026: add export for `copyWithStructuralSharing`
- #1029: fix/default serialize query args
- #1022: export Result Definition types
- #1035: simplify `selectFromResult` usage
- #1038: fix MiddlewareArray for new build
- #1039: remove last traces of TSDX
- #1042: Docs - RTKQ - Convert codeblocks to TS & enable codeblock transpilation
- #1036: Docs - add 'Cached data' concept page
- #1034: split up middleware, add OptionalPromise, add cache entry lifecycle
- #1041: prevent badly typed `Middleware<any>` from casting `dispatch` to `any`
- #1049: Docs - add Streaming Updates page
- #1053: add a per-endpoint keepUnusedDataFor option
- #1051: remove alpha.2 deprecation fallbacks

phryneas:
*just some minor tweaks* :laughing: 

Still missing on that list: 
* #1056 allow using skipSymbol as arg in hooks
* add a `merge` (or `mergeWithLastResult` option to endpoints)
* add a docs page w/ an example for infinite scrolling (actually, without using `merge`, that's only gonna be useful in a "streaming" context)

bravew:
First of all, excellent functionalities for RTK query, start refactoring our codes by adopting alpha.2, can't wait the official release. 
A quick question ,Is there any way to change baseUrl after calling createApi?
In our app we allow user to change the server address which is stored in the redux state, and will be used in following REST API call. It seems that baseUrl for RTKQ can't be modified after initialization.   
Thanks!

StefanBRas:
[...]

PR #1115, opened 2021-05-30
📝 Add dynamic base url example snippet

Shrugsy:
Related to #1023 

Adds an example snippet for using a dynamically generated base url for a baseQuery, based on the snippet by @StefanBRas from https://github.com/reduxjs/redux-toolkit/issues/1023#issuecomment-850176487

I opted to use the explicit `BaseQueryFn` type here, but it's probably worth adding an alternate version using `BaseQueryEnhancer` in the future too.

I'm open to convincing as to whether there are advantages for either side regarding `BaseQueryFn` vs doing something like `...[args, api, extraOptions]: Parameters<typeof baseQuery>`

### DIFF
--- a/docs/rtk-query/usage/customizing-queries.mdx
+++ b/docs/rtk-query/usage/customizing-queries.mdx
@@ -343,6 +343,7 @@ import {
   FetchBaseQueryError,
 } from '@reduxjs/toolkit/query'
 import { tokenReceived, loggedOut } from './authSlice'
+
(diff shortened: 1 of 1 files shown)


msutkowski (approved):
I think this looks good. Regarding the alternative types and what to show, I'd like to see if @phryneas has any specific feedback there. 

In my own projects I type them as you've shown _(so far, at least)_ :tada:

Issue #1090, opened 2021-05-25
Type bug: Query's `createApi` throws annotation error when `declaration: true` 


colemars:
Error:
```
The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.
```

Reproduced:

RTK example repo with tsconfig updated to set `declaration: true`
https://codesandbox.io/s/heuristic-ellis-49zgc?file=/src/services/api.ts

colemars:
https://github.com/microsoft/TypeScript/issues/43817

may be useful

kasparkallas:
I was going to create a new issue but it's too related:
`fakeBaseQuery` with tsconfig's `"declaration": true` causes `TS2527: The inferred type of '***' references an inaccessible 'unique symbol' type. A type annotation is necessary.` 

To reproduce, just use the `fakeBaseQuery` example from: https://redux-toolkit.js.org/rtk-query/usage-with-typescript

phryneas:
Unless the TypeScript team finally gives a *good* explanation when & why *exactly* those errors occur and how to prevent them (apart from exporting *every single type*, which is not a valid solution) I'm sorry there is not much we can do about that.

kasparkallas:
> Unless the TypeScript team finally gives a _good_ explanation when & why _exactly_ those errors occur and how to prevent them (apart from exporting _every single type_, which is not a valid solution) I'm sorry there is not much we can do about that.

FYI, I copy-pasted `fakeBaseQuery` implementation into my own codebase, the error(s) went away and everything compiled, including React hooks. I'll update if anything comes up.

PR #1116, opened 2021-05-31
remove `Id` to force lazy evaluation of Api type

phryneas:
This should fix https://github.com/reduxjs/redux-toolkit/issues/1090
@colemars could you please give the codesandbox build a try?

### DIFF
--- a/src/query/apiTypes.ts
+++ b/src/query/apiTypes.ts
@@ -4,7 +4,7 @@ import type {
   EndpointDefinition,
   ReplaceTagTypes,
 } from './endpointDefinitions'
-import type { UnionToIntersection, Id, NoInfer } from './tsHelpers'
+import type { UnionToIntersection, NoInfer } from './tsHelpers'
 import type { CoreModule } from './core/module'
 import type { CreateApiOptions } from './createApi'
 import type { BaseQueryFn } from './baseQueryTypes'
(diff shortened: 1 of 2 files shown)


colemars:
Checked it out locally and looks good 👍 

Thanks @phryneas for digging into this. Seemed like a pretty nasty/opaque error.

phryneas:
Thanks for the report & for checking! :smiley:

colemars:
@phryneas came across an issue when building and attempting to export the api with destructured hook exports. Not sure if related to this.

Build error:
> semantic error TS2339: Property 'useGetPokemonByNameQuery' does not exist on type 'Api<BaseQuer
yFn<string | FetchArgs, unknown, FetchBaseQueryError, {}, FetchBaseQueryMeta>, { getPokemonByName: QueryDefinition<string, BaseQueryFn<string | FetchArgs, unknown, FetchBaseQueryError, {}, FetchBaseQueryMeta>, never, any, string>; }, "api", never, unique symbol | unique symbol>'.

Failing repo: https://stackblitz.com/edit/node-tgjkay?file=src%2Fapi.ts

Can build it there as well to check out the error. 

Can be avoided by just not destructuring the created api. But destructured hook exports are currently in the codegen's schema.

PR #1124, opened 2021-06-02
✅ Update configureStore test

Shrugsy:
- Use implicit undefined middleware in preparation
  for the `strictOptionalProperties` TS option
  https://github.com/microsoft/TypeScript/pull/43947


Shrugsy:
This was intended to pre-emptively prepare for the `--strictOptionalProperties` TS flag upcoming for v4.4.0, but I'm not sure what the best way is to handle 'wrapper' functions that pass on optional properties to another function, when that function has optional properties that don't explicitly include undefined.

Better to leave it for someone else, or deal with it later

PR #1133, opened 2021-06-06
📝 Extend onQueryStarted documentation

Shrugsy:
* add explicit query lifecycle example
  * add hover tooltips + example for `onQueryStarted`
  * increase consistency of fake example endpoints
  
  In general: The example should make it much clearer how to perform `onStart`/`onSuccess`/`onError` side-effects for a query

### DIFF
--- a/docs/rtk-query/usage/optimistic-updates.mdx
+++ b/docs/rtk-query/usage/optimistic-updates.mdx
@@ -35,12 +35,12 @@ const api = createApi({
   tagTypes: ['Post'],
   endpoints: (build) => ({
     getPost: build.query<Post, number>({
-      query: (id) => `posts/${id}`,
+      query: (id) => `post/${id}`,
       providesTags: ['Post'],
     }),
     updatePost: build.mutation<void, Pick<Post, 'id'> & Partial<Post>>({
       query: ({ id, ...patch }) => ({
-        url: `posts/${id}`,
+        url: `post/${id}`,
         method: 'PATCH',
         body: patch,
       }),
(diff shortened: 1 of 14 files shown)


docs/rtk-query/usage/optimistic-updates.mdx
```diff
@@ -35,12 +35,12 @@ const api = createApi({
   tagTypes: ['Post'],
   endpoints: (build) => ({
     getPost: build.query<Post, number>({
-      query: (id) => `post/${id}`,
+      query: (id) => `posts/${id}`,
```
phryneas, line 38: I think most CRUD implementations go against the singular more than the plural, so I'd keep it that way. Or are we getting inconsistent somehow?

phryneas:
Generally, :+1: 
I'm not sure about the singular-plural thing, but if we have plural everywhere else, go for it :)

docs/rtk-query/usage/optimistic-updates.mdx
Shrugsy: I saw some with singular, some with plural, and I more or less flipped a coin in my head to decide which one to change them all to. Either version works for me, so I'd rather change them all to singular if that's more commonly used.

Shrugsy: I've changed them to use singular `'post'` and `'post/:id'` rather than `'posts'`/`'posts/:id` where fitting when dealing with a singular resource, except for the examples. There are likely still 'plural' instances for singular resources for other endpoints too, but I didn't want this to get out of hand for what was only intended to be a small side-change initially 😅

markerikson: tbh _I've_ always preferred using plural myself, and I think that's what I did in the Essentials tutorial.

but obviously this is bikeshedding and doesn't actually matter in the end :)

PR #1134, opened 2021-06-06
Yarn workspaces

phryneas:
Yeah, sidetrack of the day.
This needs work for the CI, but package building, docs and examples already work **and it's so much nicer**


phryneas:
Changing this to a normal PR as it seems to otherwise skip the size action and I want to see if that works

.yarnrc.yml
```diff
@@ -0,0 +1,13 @@
...
+  - path: .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
+    spec: '@yarnpkg/plugin-workspace-tools'
+  - path: .yarn/plugins/@yarnpkg/plugin-compat.cjs
+    spec: '@yarnpkg/plugin-compat'
+
+yarnPath: .yarn/releases/yarn-berry.cjs
+
+# use "global cache" so when switching between branches, packages will not be removed from cache so they will not have to be redownloaded
+globalFolder: './.yarn/global'
+enableGlobalCache: true
```
merceyz, line 13:
```suggestion
enableGlobalCache: true
```

You can have `enableGlobalCache: true` without the `globalFolder` config so it uses the default one and shares it between all projects on your machine

Default paths to the cache:
```
Linux   - ~/.yarn/berry/cache
Windows - ~/AppData/Local/Yarn/Berry/cache
```

.yarn/plugins/@yarnpkg/plugin-compat.cjs
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  name: `@yarnpkg/plugin-compat`,
+  factory: (require) => {
+    // we are not using PNP and want to use `typescript@next` in CI without hassle
+    // dummy implementation to override the built-in version of this plugin
```
merceyz, line 5: In `>=3.0.0-rc.1` the patch is optional and wont stop the install if it fails

```
yarn set version 3.0.0-rc.5
```

.github/workflows/tests.yml
```diff
@@ -18,21 +22,21 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
 
-      - uses: c-hive/gha-npm-cache@v1
+      - uses: actions/cache@v2
+        with:
+          path: .yarn/global
+          key: yarn
```
merceyz, line 28:
```suggestion
          key: yarn-${{ hashFiles('yarn.lock') }}
          restore-keys: |
            yarn-
```

.github/workflows/tests.yml
```diff
@@ -51,23 +55,28 @@ jobs:
...
```
merceyz, line 61:
```suggestion
          key: yarn-${{ hashFiles('yarn.lock') }}
          restore-keys: |
            yarn-
```

.github/workflows/tests.yml
```diff
@@ -88,29 +97,32 @@ jobs:
...
```
merceyz, line 103:
```suggestion
          key: yarn-${{ hashFiles('yarn.lock') }}
          restore-keys: |
            yarn-
```

merceyz (commented):
Some minor notes after skimming over the PR

.github/workflows/tests.yml
phryneas: Thanks!
But wouldn't that make it impossible to share the cache between builds where a dependency changed?

I was kinda aiming for "share the global cache folder between all CI runs", but I might be seeing this perfectly wrong.

merceyz: It will try the exact key first then if there is no match it will take whatever cache was last created

https://github.com/actions/cache/blob/0638051e9af2c23d10bb70fa9beffcad6cff9ce3/README.md#inputs

belgattitude: My 2 cents, I would not use 'enableGlobalCache: true', let the recommended cache folder (.yarn/cache) 

I use this action with yarn 3... based on https://github.com/actions/cache/blob/main/examples.md#node---yarn-2

```yml
      # Note: Determine the cache folder(s) to include

      - name: Get yarn cache directory path
        id: yarn-cache-dir-path
        run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
      
      - name: Yarn cache
        uses: actions/cache@v2
        id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
        with:
          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
          key: yarn-cache-folder-os-${{ runner.os }}-node-${{ env.node-version }}-${{ hashFiles('**/yarn.lock') }}
          restore-keys: |
            yarn-cache-folder-os-${{ runner.os }}-node-${{ env.node-version }}-
            yarn-cache-folder-os-${{ runner.os }}-
```

merceyz: The cache is identical between OS and node version so you don't need to include those in the key

belgattitude: Totally, habit I took completely optional.

.yarn/plugins/@yarnpkg/plugin-compat.cjs
phryneas: I'll stay on the stable version right now, but noted - I'll remove it once we switch over :+1:

package.json
```diff
@@ -36,7 +36,7 @@
     "prettier": "^2.2.1"
   },
   "resolutions": {
-    "react-scripts": "patch:react-scripts@npm:4.0.2#.yarn/patches/react-scripts@npm:4.0.2.patch",
+    "react-scripts": "patch:react-scripts@npm:4.0.2#.yarn/patches/react-scripts__npm:4.0.2.patch",
```
merceyz, line 39: The `@` isn't a problem, it's the `:` it has issues with

phryneas:
Let's get this into master :rocket:

package.json
```diff
@@ -32,12 +32,18 @@
...
   "resolutions": {
     "react-scripts": "patch:react-scripts@npm:4.0.2#.yarn/patches/react-scripts@npm:4.0.2.patch",
     "react-redux": "npm:7.2.4",
     "react": "npm:17.0.2",
     "react-dom": "npm:17.0.2"
+  },
+  "scripts": {
+    "build:examples": "yarn workspaces foreach --include '@reduxjs/*' --include '@examples-query-react/*' -vtp run build",
+    "build:docs": "yarn workspace docs run build",
+    "build:package": "yarn workspace @reduxjs/toolkit run build"
```
merceyz: Fun fact: scripts containing `:` are global so you could define these two in their respective workspaces and you can call them from anywhere

phryneas: Uuuh, TIL. Thanks!
But I think for visibility we'll keep them there. Personally I just always look into the package.json what scripts are available

merceyz: Understandable, another fun fact: `yarn run` will print all the scripts you can run, though I see it doesn't include global scripts from other workspaces yet

PR #1165, opened 2021-06-10
📝 Fix broken apollo-boost link

### LINKED ISSUE #1155: Broken link in doc
URL: https://redux-toolkit.js.org/introduction/getting-started

Text with the broken link: 
> We can't solve every use case, but in the spirit of create-react-app and **[apollo-boost](
https://www.apollographql.com/blog/announcements/frontend/zero-config-graphql-state-management)**, we can try to provide some tools that abstract over the setup process and handle the most common use cases, as well as include some useful utilities that will let the user simplify their application code.

Broken link:  
https://www.apollographql.com/blog/announcements/frontend/zero-config-graphql-state-management

Shrugsy:
Fixes #1155

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ The **Redux Toolkit** package is intended to be the standard way to write Redux 
 - "I have to add a lot of packages to get Redux to do anything useful"
 - "Redux requires too much boilerplate code"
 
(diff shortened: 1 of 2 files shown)


phryneas (approved):
Generally, apollo-boost seems like it is discontinued:

> apollo-boost
>
> The Apollo Boost project is now retired, because Apollo Client 3.0 provides a similarly straightforward setup. We recommend removing all apollo-boost dependencies and modifying your ApolloClient constructor as needed.

( from https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/#updating-imports )

But I think in this case it lends good to the explanation, so I'm all for leaving it in with the updated link :)

PR #1214, opened 2021-06-25
Clarify mutation loading / fetching lack of distinction

nickbabcock:
My assumption is that this sentence should say that there is no distinction
between loading and fetching considering that only `isLoading` is exposed
(unlike `useQuery` which exposes `isFetching` as well)


phryneas:
Very good catch, thank you!

PR #1216, opened 2021-06-25
Fix RTKQ migration doc syntax

nickbabcock:
The hanging code tick was causing formatting issues. Also unticked the names
of the project as it doesn't appear they are ticked in other places in the
docs.


markerikson:
Thanks!

PR #1218, opened 2021-06-26
Add test cases for basic example project

dggsax:
My goal with this PR is to add test cases for all of the RTK-Query example projects in the repo in an attempt to share the lessons and advice I've gotten with regards to testing.

For now, I've just added some test cases to the basic project. If the pattern that I am using looks good (in terms of utilization of [msw](https://mswjs.io/), the change I made to `store.ts` so each rendered instance of `<App />` has it's own instance of the store, etc.) then I'll proceed with adding test cases for as many of the examples as I can!


examples/query/react/basic/src/tests/App.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+      </Provider>
+    )
+
+    await waitFor(() => {
+      screen.getByText('Loading...')
+    })
+
+    await waitFor(() => {
+      screen.getByText('Oh no, there was an error')
+    })
```
Shrugsy, line 72:
```suggestion
    screen.getByText('Loading...')

    await screen.findByText('Oh no, there was an error')
```

Loading should be shown immediately, so no need to await.

Since you're only waiting for the error text to show and nothing else, it's simpler to use `findBy*`

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find

examples/query/react/basic/src/tests/App.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+    await screen.getByText('Loading...')
+
+    await waitFor(() => {
+      screen.getByText('bulbasaur')
+
+      const img = screen.getByAltText('bulbasaur') as HTMLImageElement
+      expect(img.src).toBe(
+        'https://raw.githubusercontent.com/PokeAPI/sprites/master/sprites/pokemon/shiny/1.png'
+      )
+    })
```
Shrugsy, line 46:
```suggestion
    await screen.findByRole('heading', { name: /bulbasaur/i })
    const img = screen.getByRole('img', {  name: /bulbasaur/i}) as HTMLImageElement
    expect(img.src).toBe(
      'https://raw.githubusercontent.com/PokeAPI/sprites/master/sprites/pokemon/shiny/1.png'
    )
```

Where possible, *byRole queries are typically preferred, especially compared to *byAltText

https://testing-library.com/docs/queries/about/#priority

Since the image should exist as soon as you've got the 'bulbasaur' heading there, you don't need to have that part wrapped in the `waitFor`. This makes it clearer which bits you're waiting for, and which bits should be ready together

examples/query/react/basic/src/store.ts
```diff
@@ -1,11 +1,13 @@
...
 
-export const store = configureStore({
-  reducer: {
-    [pokemonApi.reducerPath]: pokemonApi.reducer,
-  },
-  // adding the api middleware enables caching, invalidation, polling and other features of `rtk-query`
-  middleware: (getDefaultMiddleware) =>
-    getDefaultMiddleware().concat(pokemonApi.middleware),
-})
+export const createStore = (): Store => {
```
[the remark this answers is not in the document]
dggsax: That's good to know, thank you!

examples/query/react/basic/src/tests/App.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+  afterAll(() => server.close())
+
+  it('handles good response', async () => {
+    render(
+      <Provider store={createStore()}>
+        <App />
+      </Provider>
+    )
+
+    await screen.getByText('Loading...')
```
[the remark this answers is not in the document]
dggsax: That's great to know. Thank you!

examples/query/react/basic/src/tests/App.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+  afterEach(() => server.resetHandlers())
+
+  afterAll(() => server.close())
+
+  it('handles good response', async () => {
+    render(
+      <Provider store={setUpStore()}>
+        <App />
+      </Provider>
+    )
```
phryneas, line 35: I would establish rendering with a wrapper here - maybe something like 

```jsx
// defined elsewhere
function renderWithStore(jsx: React.ReactNode){
  const store = setUpStore()
  function Wrapper({children}) {
    return <Provider store={store}>{children}</Provider>
  }
  return { store, ...render(jsx, { wrapper: Wrapper }) }
}
```

```suggestion
    renderWithStore(<App />)
```
If direct access to the store was needed, it could also be accessed with
```js
const { store } = renderWithStore(<App />)
```

examples/query/react/basic/src/tests/App.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    return res(ctx.json(mockApiResponse))
+  })
+)
+
+describe('App', () => {
+  beforeAll(() => server.listen())
+
+  afterEach(() => server.resetHandlers())
+
+  afterAll(() => server.close())
```
phryneas, line 28: I think all that could go into a `setupMockServer` function:
```suggestion
// maybe in a separate file?
function setupMockServer() {
  const server = setupServer(
    rest.get('https://pokeapi.co/api/v2/pokemon/bulbasaur', (req, res, ctx) => {
      const mockApiResponse = {
        species: {
          name: 'bulbasaur',
        },
        sprites: {
          front_shiny:
            'https://raw.githubusercontent.com/PokeAPI/sprites/master/sprites/pokemon/shiny/1.png',
        },
      }
      return res(ctx.json(mockApiResponse))
    })
  )
  beforeAll(() => server.listen())
  afterEach(() => server.resetHandlers())
  afterAll(() => server.close())
  return server;
}

  
const server = setupMockServer()
describe('App', () => {
```

examples/query/react/basic/src/store.ts
```diff
@@ -1,11 +1,13 @@
...
+export const setUpStore = () => {
+  return configureStore({
+    reducer: {
+      [pokemonApi.reducerPath]: pokemonApi.reducer,
+    },
+    // adding the api middleware enables caching, invalidation, polling and other features of `rtk-query`
+    middleware: (getDefaultMiddleware) =>
+      getDefaultMiddleware().concat(pokemonApi.middleware),
+  })
+}
```
phryneas, line 13: As the store should maybe also be used with non-RTKQ tests, we could already do a bit more setup here.

The question here is how far we go with that. 
Allowing for `preloadedState` would be great, but it requires that we call `combineReducers` ahead of time to get the correct type for that.

```suggestion
import {
  configureStore,
  combineReducers,
  PreloadedState,
} from '@reduxjs/toolkit'
import { pokemonApi } from './services/pokemon'

// Allow for a preloadedState.
// Doing so moves quite a few things around (we need the reducer scoped outside to get the RootState type which is required for the PreloadedState type), so I'm not 100% sure it's worth it, but many people would probably encounter the need for this.

const reducer = combineReducers({
  [pokemonApi.reducerPath]: pokemonApi.reducer,
})
export const setUpStore = (preloadedState?: PreloadedState<RootState>) => {
  return configureStore({
    reducer,
    // adding the api middleware enables caching, invalidation, polling and other features of `rtk-query`
    middleware: (getDefaultMiddleware) =>
      getDefaultMiddleware().concat(pokemonApi.middleware),
    preloadedState,
  })
}

// we should probably show how to get those types with this more testing-targeted setup
export type RootState = ReturnType<typeof reducer>
export type AppDispatch = ReturnType<typeof setUpStore>['dispatch']

```

phryneas (commented):
This goes into a very cool direction :)

I'm wondering if we should show a good setup that would allow to do some more "classic" testing with preloadedState and a store setup that also allows to get all the necessary types to setup "normal usage" (like typing the hooks according to https://redux-toolkit.js.org/tutorials/typescript#define-typed-hooks).
It would be more code and probably not 100% relevant in each example, but it would allow for better use in real world applications. 

Any opinions? @Shrugsy @markerikson @msutkowski

examples/query/react/basic/src/tests/App.test.tsx
msutkowski: Agreed, this should be in a separate file.

examples/query/react/basic/src/tests/App.test.tsx
msutkowski: Also agreed. We might want to consider moving `setupMockServer` into `setupTests`, and returning the server, renderWithProviders (or renderWithStore) as shown here. 

If the goal is to test all of the examples, we're going to need to include a memory router as well as chakra providers for everything but the Pokemon examples, so establishing that kind of pattern upfront is probably in our best interest :tada:

Shrugsy: we might as well go the rest of the way with this and have it accept `preloadedState`, as well as the standard `render` options.

I think we should try to make it consistent with the `Redux` 'writing tests' example here: https://redux.js.org/recipes/writing-tests#example-1

Although I'm more inclined to update the example there while we're at it after agreeing on a convention here.

Mainly the changes from here that I'd personally like to get as a convention are:
- use the same `setUpStore` function that the app itself uses (rather than calling `configureStore` again separately, with duplicated store setup code)
- *don't* export as a function called `render` overriding the one from `@testing-library/react`. I don't see the benefit to this as opposed to one with a new name like `renderWithStore` or `renderWithProviders`. It makes it less obvious what you're actually rendering

examples/query/react/basic/package.json
```diff
@@ -24,12 +26,13 @@
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build"
+    "build": "react-scripts build",
+    "test": "react-scripts test"
```
msutkowski, line 30:
```suggestion
    "test": "react-scripts test --runInBand"
```

msutkowski: `msw` recommends using `runInBand` to prevent the possibility of overlapping handlers when running in parallel. I'd rather show this as the default approach upfront at this time.

examples/query/react/basic/src/store.ts
msutkowski: Passing `preloadedState` also makes bolting on Storybook super easy :)

examples/query/react/basic/src/tests/App.test.tsx
dggsax: Absolutely, there's a pattern for creating a customized wrapper function that I use at work that a colleague helped me get set up nicely with the types. Let me try to get that put in first and I can push it and y'all can review and let me know what you think!

examples/query/react/basic/src/store.ts
dggsax: Sounds good! That's another pattern that I utilize at work, so I'll get that worked in. I definitely agree with being consistent and doing the best practices all around for all the test projects.

examples/query/react/basic/src/tests/App.test.tsx
dggsax: And yes @Shrugsy I would love to help contribute a typed example for that part of the redux docs https://redux.js.org/recipes/writing-tests#example-1 😁 So will poke in this conversation once I've added a `setupTests.ts` file

dggsax:
Also, slightly related but not entirely related to this PR, but have y'all heard of [Wallaby.js](https://wallabyjs.com/)? It's an incredible unit test running tool that integrates the testing experience entirely within VSCode (or [whatever IDE you prefer](https://wallabyjs.com/docs/tutorial/overview.html)). It's [free for use with OSS projects](https://wallabyjs.com/oss/) like this one! Just thought I would recommend and share 😀

image

With Wallaby, I haven't had to run `npm run test` once (except of course when I've made major changes to make sure that they run still via that way).

phryneas:
Wallaby looks interesting. I've been just using the Jest extension by Orta (of course he just about writes everything). That one is pretty amazing too, with direct "run test with breakpoints", inline test results etc.

examples/query/react/basic/src/tests/App.test.tsx
dggsax: Ok I've made some updates:
 - Moved the `msw` server setup into `./src/test/server` directory, containing an `index.ts` to setup the server and export it, and then a separate `serverHandlers.ts` file. For the basic project, this handlers file will have only one endpoint, but as you said in the more advanced projects things will get more complicated, so I figured putting the handlers in their own file would be fine. I used one of Kent Dodd's example projects as reference: https://github.com/kentcdodds/bookshelf/tree/main/src/test/server
 - Created a `setupTests.ts`, all this does is activate the mock server, but we can imagine this getting more complex for the other example projects.
 - Created a `render()` function that combined some of the feedback in other comments that I based off of the render function that I use internally for work. It exports the store as was suggested. I'll leave comments with some questions I have. If y'all think looks good, would be happy to start a separate PR where we can figure out a good example/setup to use with typings for https://redux.js.org/recipes/writing-tests#example-1
 - Updated `App.test.tsx` to use the necessary render function 😀

examples/query/react/basic/src/store.ts
```diff
@@ -1,11 +1,28 @@
...
-  },
-  // adding the api middleware enables caching, invalidation, polling and other features of `rtk-query`
-  middleware: (getDefaultMiddleware) =>
-    getDefaultMiddleware().concat(pokemonApi.middleware),
+const rootReducer = combineReducers({
+  [pokemonApi.reducerPath]: pokemonApi.reducer,
 })
+
+// We could do DeepPartial<RootState>? I don't know if preloadedState merges state or not
+export const setupStore = (initialState?: RootState) => {
```
dggsax, line 13: Can I just use `DeepPartial<RootState>` here? I I haven't had the chance to check if that will play nicely under the hoods with the `preloadedState` store parameter

examples/query/react/basic/src/test/server/index.ts
```diff
@@ -0,0 +1,7 @@
+import { setupServer } from 'msw/node'
+import { handlers } from './serverHandlers'
+
+const server = setupServer(...handlers)
+
+export * from 'msw'
+export { server }
```
dggsax, line 7: Would y'all want the server and the other `msw` methods exported here, or from `setupTests.ts`? Personally I think exporting it from here is nice because then in the test file the import is clear:
```typescript
import { server } from './test/server'
```

examples/query/react/basic/src/test/test-utils.tsx
```diff
@@ -0,0 +1,54 @@
...
+  // if we wanted to go ahead and set up the Chakra stuff and MemoryRouter, etc,
+  // all we'd have to do is update the return in Wrapper to:
+  //
+  //  <Provider store={store}>
+  //    <ChakraProvider>
+  //      <BrowserRouter>
+  //        <App />
+  //      </BrowserRouter>
+  //    </ChakraProvider>
+  //  </Provider>
```
dggsax, line 45: Do you want me to go ahead and put in Chakra and React-Router as dependencies for the basic project and get them hooked up here for the render function?

examples/query/react/basic/src/App.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { rest } from 'msw'
```
dggsax, line 1: As for the location of the `App.test.tsx` file, let me know what y'all would prefer I do. We can do the `__test__` directory in each directory where a file is present, usually I just put the test file right next to the component being tested for clickability.

examples/query/react/basic/src/store.ts
phryneas: No, should be `PreloadedState<RootState>` as that will only allow full slices to be present or `undefined`, but not only half a slice being present (which is nothing `combineReducers` would account for)

examples/query/react/basic/src/test/server/index.ts
phryneas: Looks good from here.

examples/query/react/basic/src/App.test.tsx
phryneas: Right next to it is a lot less confusing I think.

examples/query/react/basic/src/store.ts
dggsax: Sweet, doing that makes the `test-utils.tsx` setup easier and the typing for `IWithStoreOptions` more concise. Will fix in next commit.

examples/query/react/basic/src/test/test-utils.tsx
dggsax: @msutkowski based on your comment here https://github.com/reduxjs/redux-toolkit/pull/1218#discussion_r659173421 it seems like I should go ahead and set it up with the Chakra provider and MemoryRouter, that way copying the files to the other example projects is easier and consistent?

dggsax:
Also, if all looks good, part of me is thinking that it'll be a good idea to just check in this PR as an introduction of test cases for just the `./examples/query/react/basic/` example project, and then apply the same testing pattern to the rest of the example projects in another PR. 

What do y'all think? @phryneas @msutkowski  @Shrugsy - if all looks good for the basic project, I can go ahead and mark this PR as ready for review. If you think it'd be fine to do this as an all-in-one kinda deal, that also works for me 😀 

Thanks for the reviews and time so far, I've learned a lot in the process!

examples/query/react/basic/src/test/test-utils.tsx
Shrugsy: I would say to *not* put them here if it's not being used. The goal is just to have it set up so that adding the extra providers when they're required is next to no effort

Shrugsy:
> Also, if all looks good, part of me is thinking that it'll be a good idea to just check in this PR as an introduction of test cases for just the `./examples/query/react/basic/` example project, and then apply the same testing pattern to the rest of the example projects in another PR.
> 
> What do y'all think? @phryneas @msutkowski @Shrugsy - if all looks good for the basic project, I can go ahead and mark this PR as ready for review. If you think it'd be fine to do this as an all-in-one kinda deal, that also works for me 😀
> 
> Thanks for the reviews and time so far, I've learned a lot in the process!

I'm a big proponent of small, self-contained merge requests personally, so I'm all for getting this one polished, tied up & merged first independent of the others

examples/query/react/basic/src/store.ts
```diff
@@ -1,11 +1,28 @@
...
-    [pokemonApi.reducerPath]: pokemonApi.reducer,
-  },
-  // adding the api middleware enables caching, invalidation, polling and other features of `rtk-query`
-  middleware: (getDefaultMiddleware) =>
-    getDefaultMiddleware().concat(pokemonApi.middleware),
+const rootReducer = combineReducers({
+  [pokemonApi.reducerPath]: pokemonApi.reducer,
 })
+
+export const setupStore = (initialState?: PreloadedState<RootState>) => {
```
Shrugsy, line 13: grammatically speaking, when considering the function name as using verbs, `setUpStore` is more correct over `setupStore` (`set up` vs `setup`), but it's not a massive deal

e.g. 'We set up the store setup by calling `setUpStore`'

There might be some convention where people just use `setup` regardless, `msw` seems to have gone that route

examples/query/react/basic/src/store.ts
```diff
@@ -1,11 +1,28 @@
...
+  // Adding pokemonApi's middleware to default middleware for RTK-Query features
+  const middleWare = getDefaultMiddleware<RootState>().concat(
+    pokemonApi.middleware
+  )
```
Shrugsy, line 17: I would just use the callback notation within `configureStore` for `getDefaultMiddleware` rather than the exported version. You save an import, and don't need to give it the `RootState` type explicitly

examples/query/react/basic/src/store.ts
Shrugsy:
```suggestion
export const setupStore = (preloadedState?: PreloadedState<RootState>) => {
```

Might as well keep this consistent with `configureStore` & the types, and call this `preloadedState`

examples/query/react/basic/src/store.ts
```diff
@@ -1,11 +1,28 @@
...
+
+  return configureStore({
+    reducer: rootReducer,
+    middleware: middleWare,
+    preloadedState: initialState,
```
Shrugsy, line 22:
```suggestion
    preloadedState,
```

examples/query/react/basic/src/test/test-utils.tsx
```diff
@@ -0,0 +1,45 @@
...
+  store?: AppStore
+}
+
+function render(
+  ui: React.ReactElement,
+  {
+    initialState = {},
+    store = setupStore(initialState),
+    ...renderOptions
+  }: IWithStoreOptions = {}
```
Shrugsy, line 24:
```suggestion
interface ExtendedRenderOptions extends Omit<RenderOptions, 'queries'> {
  initialState?: PreloadedState<RootState>
  store?: AppStore
}

function render(
  ui: React.ReactElement,
  {
    initialState = {},
    store = setupStore(initialState),
    ...renderOptions
  }: ExtendedRenderOptions = {}
```

I think this name (`ExtendedRenderOptions`) is a little clearer. The `I` prefix isn't really necessary with typescript, but that's not a big deal.

Omitting 'store' explicitly in the name makes it retain accuracy if it gets extended in the future with a bunch of other options unrelated to the store

examples/query/react/basic/src/test/test-utils.tsx
```diff
@@ -0,0 +1,45 @@
...
+// This type interface extends the default options for render from RTL, as well
+// as allows the user to specify other things such as initialState, store. For
+// future dependencies, such as wanting to test with react-router, you can extend
+// this interface to accept a path and route and use those in a <MemoryRouter />
+interface IWithStoreOptions extends Omit<RenderOptions, 'queries'> {
+  initialState?: PreloadedState<RootState>
+  store?: AppStore
+}
+
+function render(
```
Shrugsy, line 18: @dggsax @markerikson @phryneas @msutkowski 

I'd like feedback from as many people as possible about these particular points:

1. I'd personally prefer to explicitly *not* call this `render`, and make it something like `renderWithProviders` instead

2. I'd also personally prefer to not re-export everything from `RTL` in this file - I don't think saving a line when importing in this case is worth the obfuscation about the actual source of your imports. At a glance in the testing file, you won't know which ones actually have customized behaviour or not when they're all imported from a custom `test-utils` file, and particularly when the ones that *are* customized have hi-jacked the original names.

FWIW for point 2, I've also seen cases before with jetbrains having problems retaining the type information when re-exported with that pattern

examples/query/react/basic/src/store.ts
phryneas: :100: we are actually planning on deprecating the `getDefaultMiddleware` import in favor of the callback notation.

examples/query/react/basic/src/test/test-utils.tsx
phryneas: Agree on both.
Especially since `hooks-testing-library` has similarly named exports with slightly On 2: differing functionality. So having the user import stuff from a source they don't know 100% might be counterproductive.

examples/query/react/basic/src/test/test-utils.tsx
dggsax: Sounds good, that makes sense.

examples/query/react/basic/src/test/test-utils.tsx
dggsax: I agree, thanks!

examples/query/react/basic/src/store.ts
dggsax: Ah, great catch thank you!

As for `setupStore()` vs. `setUpStore()` I can go either way, personally `setupStore()` just **feels** better but I am not known for my grammar. Looks like RTK does have it's [own method](https://redux-toolkit.js.org/rtk-query/api/setupListeners) `setupListeners()` so perhaps just follow the library for now?

examples/query/react/basic/src/test/test-utils.tsx
dggsax: So I just pushed a commit to rename it to `renderWithProviders()` as suggested.

As for part two for re-exporting from a library, then in that case I should probably go ahead and remove the line `export * from 'msw'` line from `./src/test/server/index.ts` to be consistent.

examples/query/react/basic/src/store.ts
dggsax: good to know!

examples/query/react/basic/src/store.ts
phryneas: good point.

Shrugsy (approved):
LGTM! 👍

.gitignore
```diff
@@ -18,6 +18,10 @@ build/
 typesversions
 .cache
 .yarnrc
+
+# Ignore package-lock.json from example projects
+examples/query/react/**/package-lock.json
+
```
phryneas, line 24:
```suggestion
```

This will not be necessary in master any more since we switched to yarn workspaces - there is only one lock file in the repo root now :)

examples/query/react/basic/src/store.ts
```diff
@@ -1,11 +1,20 @@
...
-  middleware: (getDefaultMiddleware) =>
-    getDefaultMiddleware().concat(pokemonApi.middleware),
+const rootReducer = combineReducers({
+  [pokemonApi.reducerPath]: pokemonApi.reducer,
 })
+
+export const setupStore = (preloadedState?: PreloadedState<RootState>) => {
+  return configureStore({
+    reducer: rootReducer,
+    middleware: (getDefaultMiddleware) =>
```
phryneas, line 12:
```suggestion
    reducer: rootReducer,
    // adding the api middleware enables caching, invalidation, polling and other features of `rtk-query`
        middleware: (getDefaultMiddleware) =>

```

examples/query/react/basic/src/store.ts
```diff
@@ -1,11 +1,20 @@
...
+      getDefaultMiddleware().concat(pokemonApi.middleware),
+    preloadedState,
+  })
+}
+
+export type RootState = ReturnType<typeof rootReducer>
+export type AppStore = ReturnType<typeof setupStore>
+export type AppDispatch = ReturnType<typeof setupStore>['dispatch']
```
phryneas, line 20:
```suggestion
export type AppStore = ReturnType<typeof setupStore>
export type AppDispatch = AppStore['dispatch']
```

examples/query/react/basic/src/test/test-utils.tsx
```diff
@@ -0,0 +1,32 @@
+import { render as rtlRender } from '@testing-library/react'
```
phryneas, line 1:
```suggestion
import { render } from '@testing-library/react'
```
I think this does not need to be renamed any more, right?

examples/query/react/basic/src/test/test-utils.tsx
```diff
@@ -0,0 +1,32 @@
...
+  {
+    preloadedState = {},
+    store = setupStore(preloadedState),
+    ...renderOptions
+  }: ExtendedRenderOptions = {}
+) {
+  function Wrapper({ children }: PropsWithChildren<{}>): JSX.Element {
+    return <Provider store={store}>{children}</Provider>
+  }
+  return { store, ...rtlRender(ui, { wrapper: Wrapper, ...renderOptions }) }
```
phryneas, line 29:
```suggestion
  return { store, ...render(ui, { wrapper: Wrapper, ...renderOptions }) }
```

phryneas (changes requested):
Just some minor nitpicks, generally a :+1:

examples/query/react/basic/src/store.ts
dggsax: For sure, it's a pretty important line 😁

phryneas:
Alright, let's get this in :)

PR #1219, opened 2021-06-26
👷 Remove defunct strictOptionalProperties refs

Shrugsy:
Should allow the `ts-next` task to pass again.

On TS-next, `strictOptionalProperties` has been renamed, and is now opt-in. As a result, the allowances previously made to account for it are no longer necessary, and currently cause an error due to the now non-existent property.

See https://github.com/microsoft/TypeScript/pull/44626


phryneas:
Thanks :)

PR #1223, opened 2021-06-26
Docs: Brave user warning + misc config

msutkowski:
- [x] Adds a notification for brave users to disable shields
- [x] Adds a dev:docs cmd to scripts
- [x] Updates `CONTRIBUTING.md` to reflect current dev process in the monorepo

### Misc notes:
We should reintroduce `lint` and `format` commands to the repo?

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,8 +45,8 @@
     "test": "yarn test:packages",
     "build:examples": "yarn workspaces foreach --include '@reduxjs/*' --include '@examples-query-react/*' -vtp run build",
     "build:docs": "yarn workspace website run build",
-    "dev:docs": "yarn workspace website run start",
     "build:packages": "yarn workspaces foreach --include '@reduxjs/*' --include '@rtk-query/*' --topological-dev run build",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -13,7 +13,6 @@
   },
   "homepage": "https://redux-toolkit.js.org",
   "directories": {
-    "doc": "docs",
```
msutkowski, line 16: This was blocking the install of dependencies for docs - so if you cloned this repo, ran `yarn` and then tried to start the docs, it'd error with a missing formik dependency.

msutkowski:
image

image

website/src/theme/DocPageWithBraveWarning/index.js
```diff
@@ -0,0 +1,235 @@
...
+            localStorage.setItem('brave-warning-dismissed', true)
+            setIsBrave(false)
+          }}
+        >
+          Dismiss
+        </button>
+      </div>
+    </div>
+  ) : null
+}
```
msutkowski, line 77: This is the only code that isn't the standard template that was `swizzle`d

markerikson:
Seems reasonable, and neat!

PR #1224, opened 2021-06-27
Clarify RTKQ mutation data after trigger

### LINKED ISSUE #1215: Undefined data in intermediate RTKQ mutation loading step
[The docs state](https://redux-toolkit.js.org/rtk-query/usage/mutations#frequently-used-mutation-hook-return-values):

> The "mutation result" is an object containing properties such as the latest `data` for the mutation request

What I'm observing is that after the initial mutation, subsequent mutations substitute in `undefined` for the latest data and I wanted to confirm that this is working as designed.

To give an example, I have a button that returns a new API key. It's possible for the user to click the button more than once to generate more than one API key.

```tsx
export const AccountContent: React.FC<{}> = () => {
  const [trigger, { data }] = myApi.endpoints.newApiKey.useMutation();
  console.log(data);
  return (
    <>
      { data?.api_key ? <p>{data?.api_key}</p> : null }
      <button onClick={() => trigger()}>Generate</button>
    </>
  )
}
```

The following is the output from logging the data after three buttons clicks.

```
undefined
"api_key"
undefined
"api_key"
undefined
"api_key"
```

And the element for holding the API key will flicker with every undefined result.

I had anticipated that if `isLoading` is true that the previous defined result would be returned if available, but it appears I am mistaken.

It kinda seems unintended as the docs don't seem to indicate that data, isLoading, and isUninitialized are mutually exclusive to each other (ie: if one is truthy then the others must be falsey). This issue does not seem to apply to queries, interestingly enough.

nickbabcock:
It was unclear to me that triggering a mutation request clears out
previous the previous response before the new one is finished, so I
added some docs to hopefully clarify this situation (and suggest a
workaround of component level caching, which is what I ended
us doing).

Closes #1215

### DIFF
--- a/docs/rtk-query/usage/mutations.mdx
+++ b/docs/rtk-query/usage/mutations.mdx
@@ -103,7 +103,7 @@ The "mutation result" is an object containing properties such as the latest `dat
 
 Below are some of the most frequently used properties on the "mutation result" object. Refer to [`useMutation`](../api/created-api/hooks.mdx#usemutation) for an extensive list of all returned properties.
 
-- `data` - The returned result if present. Any previous result will be undefined after triggering a mutation request, so consider component level caching if the previous response data is required for a smooth transition to new data.
+- `data` - The data returned from the latest trigger response, if present. If subsequent triggers from the same hook instance are called, this will return undefined until the new data is received. Consider component level caching if the previous response data is required for a smooth transition to new data.
 - `error` - The error result if present.
 - `isUninitialized` - When true, indicates that the mutation has not been fired yet.
 - `isLoading` - When true, indicates that the mutation has been fired and is awaiting a response.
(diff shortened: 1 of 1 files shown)


docs/rtk-query/usage/mutations.mdx
```diff
@@ -103,7 +103,7 @@ The "mutation result" is an object containing properties such as the latest `dat
 
 Below are some of the most frequently used properties on the "mutation result" object. Refer to [`useMutation`](../api/created-api/hooks.mdx#usemutation) for an extensive list of all returned properties.
 
-- `data` - The returned result if present.
+- `data` - The returned result if present. Any previous result will be undefined after triggering a mutation request, so consider component level caching if the previous response data is required for a smooth transition to new data.
```
Shrugsy, line 106:
```suggestion
- `data` - The data returned from the latest trigger response, if present. If subsequent triggers from the same hook instance are called, this will return undefined until the new data is received. Consider component level caching if the previous response data is required for a smooth transition to new data.
```

I think this wording is slightly more accurate, as it doesn't actually affect a previous result, it just doesn't return the previous result any more. It hopefully also adds a few more subtle clarifications

nickbabcock: Agreed. Thank you for the suggestion!

markerikson:
Thanks!

Issue #1203, opened 2021-06-22
Update all remaining code blocks to have TS/JS toggles


markerikson:
We have a number of code blocks that are still JS-only.  It would be useful to convert all of them to be toggleable TS/JS blocks instead.  As an example, the initial example in `createAsyncThunk`:

https://redux-toolkit.js.org/api/createAsyncThunk#overview

PrinceRajRoy:
Hi Mark, is this supposed to be done using the same remark tool?

phryneas:
In the long run we might switch to shiki-twoslash once they add a feature like that, but for now yes.

PrinceRajRoy:
@phryneas great! just a quick query, the `website` dir contains package-lock.json but the README directs to use `yarn` instead, is that intended?

phryneas:
I assume that is the default docusaurus README, just ignore it for now. We will switch over to a yarn monorepo soon anyways.

phryneas:
Also, you might run into memory problems on the main branch. Update remark-typescript-tools to 1.0.7 to fix that.

PrinceRajRoy:
@phryneas I'm trying to use the remark-typescript-tools on jsx snippets, but it doesn't seem to work. I have updated it to v1.0.7 before using, so I'm a bit confused here. Sorry if I'm missing something here 😅

Snippet

```ts title="index.tsx"
// file: App.tsx noEmit
import React from 'react'
export default function App() {
  return <div>...</div>
}

// file: app/store.ts noEmit
import { configureStore } from '@reduxjs/toolkit'

export default configureStore({
  reducer: {},
})

// file: index.tsx
import React from 'react'
import ReactDOM from 'react-dom'
import './index.css'
import App from './App'
// highlight-start
import store from './app/store'
import { Provider } from 'react-redux'
// highlight-end

ReactDOM.render(
  // highlight-next-line
  <Provider store={store}>
    <App />
  </Provider>,
  document.getElementById('root')
)
```

Works fine with "jsx": "react"

![1](https://user-images.githubusercontent.com/47452572/123157481-3719a380-d488-11eb-99b1-160b2ec25ecc.PNG)

Doesn't work with "jsx": "preserve"

![2](https://user-images.githubusercontent.com/47452572/123157560-4dbffa80-d488-11eb-9e51-dec6ecae8523.PNG)

phryneas:
[...]

PR #1230, opened 2021-06-29
Tutorials Section: Update code blocks with JS/TS toggle

PrinceRajRoy:
Kicking off #1203 

This PR updates the code snippets under Tutorials section to support JS/TS toggles


PrinceRajRoy:
@phryneas The remark tool [removes `React` imports](https://deploy-preview-1230--redux-starter-kit-docs.netlify.app/tutorials/typescript#use-typed-hooks-in-components) in JS version if jsx is not used in the file. Is there a to skip that?

phryneas:
Hmm. Not entirely sure to be honest. That could be a prettier thing maybe?

In this case - maybe just return a `<div>{/* rendering logic */}</div>`?

markerikson:
I'm not sure it makes sense to have the code samples in the "TS Quick Start" page be compiled to plain JS, given that the entire point is to have them in TS :)

PrinceRajRoy:
@markerikson I was thinking about it, then again the issue said All the snippets, so I thought that's included, anyways pushing the discarded changes 😅

markerikson:
look you're supposed to read my mind and realize that when I said "all" I actually meant "all of them that make sense" even though I didn't specify that and wasn't even thinking about this case at the time :)

markerikson:
looks good!

PR #1235, opened 2021-06-30
RTK Query: Handle non-json responses from a json response handler

### LINKED ISSUE #1234: RTKQ: Error Handling on 404 Not Found Response
If an API endpoint is called and it responds with a 404 Not Found response, it causes parsing issues with the error message. 
Probably a valid JSON is expected from the rtkq as response, but this cannot always be expected from the API endpoint in case of an error. At the end, I get the following error message and cannot make a comparison on `error.status === 404`, for example.

```json
{
  "name":"SyntaxError",
  "message":"Unexpected token N in JSON at position 0",
  "stack":"SyntaxError: Unexpected token N in JSON at position 0\n at JSON.parse (<anonymous>)\n at eval..."
}
```` 

Code example of the pokemon api by calling pokemon 'foo':
https://codesandbox.io/s/mystifying-mirzakhani-smx23?file=/src/App.tsx

Am I missing something in the error handling documentation or could this be a bug?

msutkowski:
- Fixes #1234

### DIFF
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -47,7 +47,18 @@ const handleResponse = async (
 
   if (responseHandler === 'json') {
     const text = await response.text()
-    return text.length ? JSON.parse(text) : undefined
+    // We run this check because some 404 endpoints may return a text/plain content type
+    // See: https://github.com/reduxjs/redux-toolkit/issues/1234
+
+    if (text.length) {
+      try {
+        return JSON.parse(text)
+      } catch (err) {
+        return text
+      }
+    }
+
+    return undefined
   }
 }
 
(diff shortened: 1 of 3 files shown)


phryneas:
Hmm, not sure on this to be honest. This would essentially make it succeed in case of a 200, right?

phryneas:
Isn't the problem here more that we are not sending an `Accept` header so the api can respond with whatever?

msutkowski:
> Isn't the problem here more that we are not sending an `Accept` header so the api can respond with whatever?

This is going to be an issue as any server could choose to ignore the accept headers. If you modify the example to include Accept headers, it'll still fail.

phryneas:
But then it is a bug on the server, isn't it?

But back on topic: I'm not sure how to exactly tackle this one *correctly*.

It should definitely lead into an error case, so a 200 with incorrect JSON should never lead to a "isSuccess" state.

For the error, it's a bit more difficult.

Imagine a scenario where a 418 error always returns valid JSON. The user expects that valid JSON and wants to access a nested property of it in the component. 
But now, we go into this fallback and suddenly, it's a `{ status: 418, data: "foo" }`. Now we will have crashing runtime code on that access.
Even worse if we start introducing stuff like error type generics or something.

Honestly, I don't really know what the best solution would be there. Is there any prior art?

msutkowski:
Better handled in #1250

PR #1236, opened 2021-06-30
docs(rtkq): update endpoints section

SimonGolms:
- replace `useAppDispatch` and `useAppSelector` as relative imports with absolute imports of `useDispatch` and `useSelector` from `react-redux`

### DIFF
--- a/docs/rtk-query/api/created-api/endpoints.mdx
+++ b/docs/rtk-query/api/created-api/endpoints.mdx
@@ -94,11 +94,11 @@ When dispatching an action creator, you're responsible for storing a reference t
 
 ```tsx title="initiate query example"
 import { useState } from 'react'
-import { useAppDispatch } from './store/hooks'
+import { useDispatch } from 'react-redux'
 import { api } from './services/api'
 
 function App() {
-  const dispatch = useAppDispatch()
+  const dispatch = useDispatch()
   const [postId, setPostId] = useState<number>(1)
 
   useEffect(() => {
@@ -121,11 +121,11 @@ function App() {
 
 ```tsx title="initiate mutation example"
 import { useState } from 'react'
-import { useAppDispatch } from './store/hooks'
+import { useDispatch } from 'react-redux'
 import { api, Post } from './services/api'
 
 function App() {
-  const dispatch = useAppDispatch()
+  const dispatch = useDispatch()
   const [newPost, setNewPost] = useState<Omit<Post, 'id'>>({ name: 'Ash' })
 
   function handleClick() {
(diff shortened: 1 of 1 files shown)


msutkowski:
@SimonGolms Is there a particular reason you're wanting to introduce this change? All of the recent docs changes reflect moving  to typed hooks for `useAppDispatch`/`useAppSelector` for a reason. If you're a TS user, this is going to be broken.

SimonGolms:
Hi @msutkowski, please correct me if I'm wrong, but I'm just updating the example code snippets ([1](https://redux-toolkit.js.org/rtk-query/api/created-api/endpoints#example), [2](https://redux-toolkit.js.org/rtk-query/api/created-api/endpoints#example-1)) in the appropriate docs section and not any CodeSandbox examples, where I absolutely see your point about TS?
For me as an end user, the explicit use of `useAppDispatch/useAppSelector` was quite confusing, especially since it was not obvious to me what else was behind the functions. It took quite a long time and search in the source code until I knew that I can also simply use `useDispatch/useSelector` here.

msutkowski:
> Hi @msutkowski, please correct me if I'm wrong, but I'm just updating the example code snippets ([1](https://redux-toolkit.js.org/rtk-query/api/created-api/endpoints#example), [2](https://redux-toolkit.js.org/rtk-query/api/created-api/endpoints#example-1)) in the appropriate docs section and not any CodeSandbox examples, where I absolutely see your point about TS?
> For me as an end user, the explicit use of `useAppDispatch/useAppSelector` was quite confusing, especially since it was not obvious to me what else was behind the functions. It took quite a long time and search in the source code until I knew that I can also simply use `useDispatch/useSelector` here.

This is a tricky line for us because we prefer to show TS usage in these examples - that's why those snippets have types in them. If a TS user were to copy and paste the snippets with the changes you have here as you did, they'd have more confusing issues that would ultimately link them here: https://redux-toolkit.js.org/usage/usage-with-typescript#getting-the-dispatch-type. In this case, the user would need the correct `Dispatch` type, which is why those hooks are shown here. We could consider annotating every single place with `// if you're a js user, you can just useDispatch and useSelector'`, but I'm not sure if that's necessary. If you used the search bar and put in 'useAppDispatch', it would've linked you to the getting the dispatch type section.

@markerikson or @Shrugsy might have an opinion on this. The thing I care most about is consistency, and if all new tutorials and TS examples show the pattern of exporting typed hooks, we should probably keep things as is.

SimonGolms:
Oh, I just looked at the source code sections and how the documentation is created in detail. Pretty cool by the way and my change makes not much sense in this situation - sorry for the inconvenience.

msutkowski:
> Oh, I just looked at the source code sections and how the documentation is created in detail. Pretty cool by the way and my change makes not much sense in this situation - sorry for the inconvenience.

Not an inconvenience at all! We want to make sure the docs are as clear as possible, so we're always open to feedback :)

PR #1249, opened 2021-07-04
Fix the example code in "Quick start" tutorial

antego:
VSCode shows the error in the `store.js` file
```
Exported variable 'store' has or is using name 'CounterState' from external module ".../src/client/features/counter/counterSlice" but cannot be named.
ts(4023)
```

Managed to fix this by exporting the CounterState in the `features/counter/counterSlice.js`.


markerikson:
Thanks!

PR #1250, opened 2021-07-04
improve error handling of fetchBaseQuery

### LINKED ISSUE #1234: RTKQ: Error Handling on 404 Not Found Response

phryneas:
Fixes #1234 

It will still cause an error, but a graceful one without information to be lost. This is probably the best way to tackle that.

Also I added handling for "fetch" failing, e.g. due to connection loss.


msutkowski (approved):
I think this looks great 💪 . We'll just need to update docs to reflect these changes, but that's not a blocker for getting this in.

ether-btc:
Closing as not planned. Summary of status:

**PR #1250 (improve error handling of fetchBaseQuery):**
- Status: ✅ MERGED into master on July 4, 2021
- Author: phryneas
- Fixes: #1234
- Added graceful error handling for fetch API failures (e.g. connection loss)
- Size impact: +0.95–1.56% across affected packages, deemed acceptable

This PR was fully resolved and merged years ago. No further action needed.

PR #1251, opened 2021-07-04
log unhandled errors during request to console

phryneas:
This logs an error message to the console in the case an unhandled error is thrown during a request.

This should happen in user-provided code, either in
* `transformResponse`
* `baseQuery`
* `fetchBaseQuery`: in the `prepareHeaders` or `validateStatus` option.

As is will lead to `providesTags`/`invalidatesTags`not being taken into account for that request, we probably should warn this to the console with a very red warning.

### DIFF
--- a/packages/toolkit/src/query/core/buildThunks.ts
+++ b/packages/toolkit/src/query/core/buildThunks.ts
@@ -313,6 +313,15 @@ export function buildThunks<
       if (error instanceof HandledError) {
         return rejectWithValue(error.value, { baseQueryMeta: error.meta })
       }
+      if (
+        typeof process !== 'undefined' &&
+        process.env.NODE_ENV === 'development'
+      ) {
+        console.log(`An unhandled error occured processing a request for the endpoint ${arg.endpointName}.
+In the case of an unhandled error, no tags will be "provided" or "invalidated".`, error)
+      } else {
+        console.log(error)
+      } 
       throw error
     }
   }
--- a/packages/toolkit/src/query/core/buildThunks.ts
+++ b/packages/toolkit/src/query/core/buildThunks.ts
@@ -317,10 +317,10 @@ export function buildThunks<
         typeof process !== 'undefined' &&
         process.env.NODE_ENV === 'development'
       ) {
-        console.log(`An unhandled error occured processing a request for the endpoint ${arg.endpointName}.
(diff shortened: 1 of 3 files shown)


packages/toolkit/src/query/core/buildThunks.ts
```diff
@@ -313,6 +313,15 @@ export function buildThunks<
       if (error instanceof HandledError) {
         return rejectWithValue(error.value, { baseQueryMeta: error.meta })
       }
+      if (
+        typeof process !== 'undefined' &&
+        process.env.NODE_ENV === 'development'
+      ) {
+        console.log(`An unhandled error occured processing a request for the endpoint ${arg.endpointName}.
```
msutkowski, line 320:
```suggestion
        console.warn(`An unhandled error occured processing a request for the endpoint ${arg.endpointName}.
```

packages/toolkit/src/query/core/buildThunks.ts
```diff
@@ -313,6 +313,15 @@ export function buildThunks<
...
+In the case of an unhandled error, no tags will be "provided" or "invalidated".`, error)
+      } else {
+        console.log(error)
```
msutkowski, line 323:
```suggestion
        console.error(error)
```

msutkowski (changes requested):
I think we should use either `console.warn` or `console.error` here. I personally prefer using `warn` when providing a message like this though.

packages/toolkit/src/query/core/buildThunks.ts
phryneas: Actually I think both of these should be `console.error`. Dunno what I was (not) thinking there.

msutkowski: You put the intention in the description though! 🏆

phryneas:
Since the message is just an addition to the error itself, I've gone for the error.

msutkowski:
Can you get a test to run for this by any chance? I gave it a shot, and couldn't get this to log. I tried `throw`ing in all mentioned places.

phryneas:
Apparently I'm too stupid for a `console.log` PR... I'll try in the evening again.

phryneas:
It works in the tests and in the CSB build so I'll assume there was some hiccup on your side and get this merged :)
image

Issue #1130, opened 2021-06-04
@typescript-eslint/unbound-method error while implementing Redux-Observable Epic


jonparkdev:
https://redux-toolkit.js.org/api/createAction#with-redux-observable

At my company we are building a strictly typed web application using react+redux+observables.  Following the documentation linked above we ran into a small bug where the ```actionCreator.match``` within the ```filter()``` was highlighting a tslinting error described by the rule ```@typescript-eslint/unbound-method```

Investigating the typed interface:  

```
declare interface BaseActionCreator<P, T extends string, M = never, E = never> {
    type: T;
    match(action: Action<unknown>): action is PayloadAction<P, T, M, E>;
}
```

```match``` is typed as a "class member function" which I believe is causing the error. 

If the issue is approved, I can send a PR with the solution of typing  ```match``` as a function valued property instead:

```
 match: (action: Action<unknown>) => action is PayloadAction<P, T, M, E>;
```

**Comment by phryneas** [comment]
Uhh, while we would probably accept that, honestly, I'd take that up to eslint. 

That is the [method definition](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions) syntax and has nothing to do with classes in the slightest - and making any assumptions on how something is implemented based on that is, honestly, pretty problematic, as TypeScript itself does not make this distinction.

Like, TypeScript as underlying language does not make this distinction - this is all just the - honestly, weird - interpretation of the author of an eslint rule.

All of these are 100% interchangeable from TypeScript's perspective:

```ts
class X {
  foo(){ return "x" }
}

const classInstance = new X()

const methodDefinitions = {
  foo() {
    return "x"
  }
}

const arrowFn = {
  foo: () => "x"
}


{
  let x: X;
  x = classInstance
  x = methodDefinitions
  x = arrowFn
}

{
  let x: typeof methodDefinitions;
  x = classInstance
  x = methodDefinitions
  x = arrowFn
}

{
  let x: typeof arrowFn;
  x = classInstance
  x = methodDefinitions
  x = arrowFn
}
```

markerikson:
Going to go ahead and close this because I don't think it's specifically actionable on our end, but if you feel the types can be improved in some way, please go ahead and submit a PR.

jonparkdev:
@phryneas  That makes sense, thank you for clarifying.

PR #1263, opened 2021-07-07
improve typing for unbound functions

ajcrites:
This updates the creator APIs to more accurately express types for the returned function objects. For example, `match():` would be treated as a bound function relying on 'this', but it is now updated to `match: () =>`. This will not trigger the `@typescript-eslint/unbound-method` rule when passing `actions.act
ionName.match`.

### DIFF
--- a/packages/toolkit/src/createAction.ts
+++ b/packages/toolkit/src/createAction.ts
@@ -83,7 +83,7 @@ export type _ActionCreatorWithPreparedPayload<
  */
 interface BaseActionCreator<P, T extends string, M = never, E = never> {
   type: T
-  match(action: Action<unknown>): action is PayloadAction<P, T, M, E>
+  match: (action: Action<unknown>) => action is PayloadAction<P, T, M, E>
 }
 
 /**
--- a/packages/toolkit/src/createAsyncThunk.ts
+++ b/packages/toolkit/src/createAsyncThunk.ts
@@ -235,10 +235,10 @@ export type AsyncThunkAction<
   | ReturnType<AsyncThunkFulfilledActionCreator<Returned, ThunkArg>>
   | ReturnType<AsyncThunkRejectedActionCreator<ThunkArg, ThunkApiConfig>>
 > & {
-  abort(reason?: string): void
+  abort: (reason?: string) => void
   requestId: string
   arg: ThunkArg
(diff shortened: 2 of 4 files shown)


phryneas:
I'm still pretty much of the opinion that the ESLint authors do actually not understand how TypeScript works. This rule is not matching up with anything the language TypeScript does.

Please check my comment here:
https://github.com/reduxjs/redux-toolkit/issues/1130#issuecomment-854930442

There *are* differences between the two notations, but `this` is *not* one of them on the TypeScript level.

[Quoting the TypeScript technical lead Ryan Cavanaugh:](https://twitter.com/SeaRyanC/status/1402666193710850048)
> Two differences: The second one is bivariant on its parameter types, and can have additional overloads added via interface augmentation

ajcrites:
@phryneas after looking into it further, it seems like the preferred way of doing this would be the `this: void` annotation for the functions that don't use `this` which would still allow keeping the method syntax. That may have other ramifications though.

phryneas:
Honestly is is just insane that some ESLint author thinks they add additional meaning to a programming language and now unrelated upstream libraries have to conform to their weird interpretation to not hurt their users. :slightly_frowning_face: 

I gotta sleep over this, but right now I'm honestly not really happy about either of those changes.

ajcrites:
That's okay... I could be wrong on this, but the inability to augment the interface also prevents a local type definition from using the `this: void` annotation as well. If the _only_ value that updating the typings here is to allow conforming to that rule, then maybe it isn't worth it. I think the rule itself is fairly innocuous since it just provides a warning that the method _could_ require binding to the host object but there is no way to know.

I'm not the best TypeScripter, but in `createAction`, for example, the match property is assigned to the object as a function and is not part of a method declaration on the object. I think that this would make the `match: () =>` syntax more accurate.

phryneas:
Well, given that the changes themselves are innocent enough I'll go ahead and merge this in.

I still don't agree with the reason for it and would rather see that ESLint rule deprecated (or at least adjusted, even though I don't see how that intent could reasonably be done in a *correct* way), but given my previous interactions with ESLint, I don't see that happening.

And since I don't want to have this discussion again and again... Let's just get it in :sweat_smile:

Thanks for the contribution - sorry for me being so grumpy :slightly_smiling_face:

PR #1267, opened 2021-07-08
🐛 Prevent closure over `fetch`

Shrugsy:
Defaulting `fetchFn = fetch` keeps it around in a closure.
This change avoids that, allowing `window.fetch`
to be mocked/monkey-patched more easily

Also updates the lockfile

### DIFF
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -27,6 +27,12 @@ export interface FetchArgs extends CustomRequestInit {
   validateStatus?: (response: Response, body: any) => boolean
 }
 
+/**
+ * A mini-wrapper that passes arguments straight through to `window.fetch`.
+ * Avoids storing `window.fetch` in a closure, in order to permit mocking/monkey-patching.
+ */
+const defaultFetchFn = (...args: Parameters<typeof window.fetch>) => window.fetch(...args);
+
 const defaultValidateStatus = (response: Response) =>
   response.status >= 200 && response.status <= 299
 
@@ -118,7 +124,7 @@ export type FetchBaseQueryMeta = { request: Request; response: Response }
 export function fetchBaseQuery({
   baseUrl,
   prepareHeaders = (x) => x,
-  fetchFn = fetch,
+  fetchFn = defaultFetchFn,
   ...baseFetchOptions
 }: FetchBaseQueryArgs = {}): BaseQueryFn<
   string | FetchArgs,
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -31,7 +31,8 @@ export interface FetchArgs extends CustomRequestInit {
  * A mini-wrapper that passes arguments straight through to `window.fetch`.
  * Avoids storing `window.fetch` in a closure, in order to permit mocking/monkey-patching.
  */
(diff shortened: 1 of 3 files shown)


packages/toolkit/src/query/fetchBaseQuery.ts
```diff
@@ -27,6 +27,12 @@ export interface FetchArgs extends CustomRequestInit {
   validateStatus?: (response: Response, body: any) => boolean
 }
 
+/**
+ * A mini-wrapper that passes arguments straight through to `window.fetch`.
+ * Avoids storing `window.fetch` in a closure, in order to permit mocking/monkey-patching.
+ */
+const defaultFetchFn = (...args: Parameters<typeof window.fetch>) => window.fetch(...args);
```
msutkowski, line 34: Should we check for `window` here and warn on it? The [docs already mention handling this in an SSR env](https://redux-toolkit.js.org/rtk-query/api/fetchBaseQuery), but it's the only feedback I have.

msutkowski (approved):

packages/toolkit/src/query/fetchBaseQuery.ts
Shrugsy: Yeah, that's a good idea. I've put it immediately within the `fetchBaseQuery` creation call though, rather than when a fetch is attempted. Caught quicker, and spams less

packages/toolkit/src/query/fetchBaseQuery.ts
```diff
@@ -27,6 +27,13 @@ export interface FetchArgs extends CustomRequestInit {
   validateStatus?: (response: Response, body: any) => boolean
 }
 
+/**
+ * A mini-wrapper that passes arguments straight through to `window.fetch`.
+ * Avoids storing `window.fetch` in a closure, in order to permit mocking/monkey-patching.
+ */
+const defaultFetchFn = (...args: Parameters<typeof window.fetch>) =>
+  window.fetch(...args)
```
phryneas, line 35:
```suggestion
const defaultFetchFn: typeof fetch = (...args) =>
  fetch(...args)
```

Given that we had it without the `window` reference before, let's just keep it like that.

Shrugsy: No problem, changed as per the above, and tweaked the docstring slightly.

The reason I explicitly changed it previously was that I've experienced a setup where Sinon's [useFakeTimers](https://sinonjs.org/releases/latest/fake-timers/) will mock `window.setTimeout` but not `setTimeout`. But given that I haven't found the exact reason for that, or how version specific or test setup specific the behaviour is, I can't assume that will occur elsewhere.

packages/toolkit/src/query/fetchBaseQuery.ts
```diff
@@ -127,6 +134,11 @@ export function fetchBaseQuery({
   {},
   FetchBaseQueryMeta
 > {
+  if (typeof window === 'undefined' && fetchFn === defaultFetchFn) {
+    console.warn(
+      'Warning: `window` is not available. Please supply a custom `fetchFn` property to use `fetchBaseQuery` on SSR environments.'
+    )
+  }
```
phryneas, line 141:
```suggestion
  if (typeof fetch === 'undefined' && fetchFn === defaultFetchFn) {
    console.warn(
      'Warning: `fetch` is not available. Please supply a custom `fetchFn` property to use `fetchBaseQuery` on SSR environments.'
    )
  }
```

phryneas:
I'll wait for the tests to pass and then get this in.

PR #1270, opened 2021-07-08
Feature/rtkq ssr

barnabasJ:
This is the Proof of Concept for the `getDataFromTree` implementation mentioned in this discussion  #1184. Please let me know what you think.


markerikson:
Going to page @ephem to see if he's got any thoughts here.

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -501,6 +501,34 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+      if (
+        typeof api.ssr === 'object' &&
+        !(
+          api.ssr[api.reducerPath] &&
+          api.ssr[api.reducerPath][name] &&
+          api.ssr[api.reducerPath][name][JSON.stringify(stableArg)]
+        )
+      ) {
+        // ssr mode is on and there is no promise yet
+        const promise = dispatch(
```
Ephem, line 513: Just leaving this comment to point out to anyone not used to SSR that this dispatch is the necessary "hack" to make this whole approach work. Since `useEffect` doesn't run on the server, when in ssr mode this PR would make the dispatch fire directly in render instead, breaking the no observable side effects in render rule. The new SSR renderer in React 18 will be pretty likely to break/break with this approach.

Because of how SSR works pre React 18, this is pretty much the only way to do it if you want to support fetching data from within the component tree, instead of only ahead of rendering. So while not an uncommon approach, I wanted to clearly mark out the "hacky part" to others. 😄

(I'll leave some more thoughts and comments on the linked issue instead of here on the PR since I think they fit better there)

PR #1272, opened 2021-07-09
[docs:query] add useQuery() manual call example at overview pt2

ryota-murakami:
I do exact same change to https://github.com/reduxjs/redux-toolkit/pull/1257.
Actually same code example used to in 2 different pages, I illustrate each page link.

Screen Shot 2021-07-09 at 15 18 43

Thank you for your review!


phryneas:
Thanks :)

PR #1277, opened 2021-07-09
naive SSR & suspense support

phryneas:
This still requires a way to HYDRATE on the server, but could be a start to build on & discuss

### DIFF
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -5,19 +5,8 @@ import type {
   QueryArgFrom,
   ResultTypeFrom,
 } from '../endpointDefinitions'
-import type {
-  QueryThunkArg,
-  MutationThunkArg,
-  QueryThunk,
-  MutationThunk,
-} from './buildThunks'
-import type {
-  AnyAction,
-  AsyncThunk,
-  ThunkAction,
-  SerializedError,
-} from '@reduxjs/toolkit'
-import { unwrapResult } from '@reduxjs/toolkit'
+import type { QueryThunk, MutationThunk } from './buildThunks'
+import type { AnyAction, ThunkAction, SerializedError } from '@reduxjs/toolkit'
 import type { QuerySubState, SubscriptionOptions, RootState } from './apiState'
 import type { InternalSerializeQueryArgs } from '../defaultSerializeQueryArgs'
 import type { Api } from '../apiTypes'
@@ -191,18 +180,31 @@ export function buildInitiate({
   mutationThunk: MutationThunk
   api: Api<any, EndpointDefinitions, any, any>
 }) {
+  const runningQueries: Record<string, Promise<unknown> | undefined> = {}
+  const runningMutations: Record<string, Promise<unknown> | undefined> = {}
+
   const {
     unsubscribeQueryResult,
     unsubscribeMutationResult,
     updateSubscriptionOptions,
   } = api.internalActions
-  return { buildInitiateQuery, buildInitiateMutation }
+  return {
+    buildInitiateQuery,
+    buildInitiateMutation,
+    getRunningOperationPromises,
+  }
+
+  function getRunningOperationPromises() {
+    return Object.values(runningQueries)
+      .concat(Object.values(runningMutations))
+      .filter((t): t is Promise<unknown> => !!t)
+  }
 
   function middlewareWarning(getState: () => RootState<{}, string, string>) {
     if (process.env.NODE_ENV !== 'production') {
       if ((middlewareWarning as any).triggered) return
-      const registered = getState()[api.reducerPath]?.config
-        ?.middlewareRegistered
+      const registered =
+        getState()[api.reducerPath]?.config?.middlewareRegistered
       if (registered !== undefined) {
         ;(middlewareWarning as any).triggered = true
       }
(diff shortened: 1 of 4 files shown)


packages/toolkit/src/query/core/buildInitiate.ts
```diff
@@ -219,93 +224,117 @@ Features like automatic cache collection, automatic refetching etc. will not be
...
-      const statePromise = Object.assign(
-        thunkResult.then(() =>
-          (api.endpoints[endpointName] as ApiEndpointQuery<any, any>).select(
-            arg
-          )(getState())
-        ),
-        {
-          arg,
-          requestId,
+    const queryAction: StartQueryActionCreator<any> =
```
phryneas, line 227: yay prettier... :facepalm: 
please activate that checkbox there :smile: 
image

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -502,7 +523,7 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
 
       const promiseRef = useRef<QueryActionCreatorResult<any>>()
 
-      useEffect(() => {
+      usePossiblyImmediateEffect((): void | undefined => {
```
Ephem, line 526: Since I'm not that familiar with RTK Query I'm just skimming this PR from my (SSR) perspective. I'm not sure how subscriptions work here (and don't have time to dig) so I'll go ahead and ask some questions. 😄

I'm not sure about converting this entire effect to be immediate. Specifically I notice that there is an effect cleanup calling `promiseRef.current?.unsubscribe()` on line 559. This is never called on the server, will that cause problems? (You mentioned hydration issues related to this in a Twitter-thread, is there anything else that might be problematic like memory leaks on the server?) I'm guessing `dispatch(initiate(...))` is what sets up that subscription? Is there a way to start a fetch and get a promise back without setting up a subscription and can that be used instead with `sideEffectsInRender`?

About the Suspense parts, who knows what that will look like when Suspense for data fetching is finished, but at a glance it looks like this could work in the current versions.

(Aside: I also noticed this effect will be called in every render in Suspense mode, but looks like it's basically a noop if nothing has changed so maybe that is fine? Might be unexpected and brittle if adding things to `usePossiblyImmediateEffect`s in the future though?)

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -451,11 +457,26 @@ type GenericPrefetchThunk = (
...
+  },
 }: {
   api: Api<any, Definitions, any, any, CoreModule>
   moduleOptions: Required<ReactHooksModuleOptions>
 }) {
+  const usePossiblyImmediateEffect: (
+    effect: () => void | undefined,
+    deps?: DependencyList
+  ) => void =
+    unstable__sideEffectsInRender || unstable__suspense
```
Ephem, line 476: Is the separate `unstable__sideEffectsInRender` needed here, maybe it could just be "if you want to fetch inside of components on the server, you use Suspense"?

phryneas: I wanted to enable a "fetch in render, but do not throw" use case here, so I'd prefer to keep them separated.

packages/toolkit/src/query/react/buildHooks.ts
phryneas: Having those subscriptions added in SSR might actually be helpful.

I'm thinking of a "hydrate" action (or other helper) that first restores the state from the server and then unsubscribes all subscriptions that were part of that "initial data". 
That would start all those "the last component unsubscribed" timers for cache cleanup while components could "re-claim" those cache entries themselves by subscribing to them.

In reality though, that might be quite a lot of actions flying around at once, so I'll have to experiment around with this a bit.

phryneas: As for the `usePossiblyImmediateEffects` being potentially brittle: I think we'll need to do something during render in every possible suspense scenario I can think of, so there is probably no way around that. I've just stuck this to the `useEffects` where it's necessary (also those without a cleanup) and hope that we can get away with it, at least until other patterns surface. It's also a marker that we have to handle these effects with extra care.

phryneas: Memory leaks on the server should probably not be a problem, if the Redux store is disposed of correctly. (If nothing unsubscribes, there will not even be any timers around for cache collection.)

It might be a good practice to clean up thoroughly in a SSR scenario by doing a `store.dispatch(api.util.resetApiState())` after the render is complete though, to make double sure really no timeouts etc. are sticking around. 

We should definitely add that in a SSR documentation.

packages/toolkit/src/query/react/buildHooks.ts
Ephem: Ah, I get it, I missed that this wasn't throwing! As I mentioned in the issue I'm wary of that use case personally and don't see any benefit over using the Suspense paradigm, but that's up to you of course and this _is_ clearly marked as unstable. 😄

packages/toolkit/src/query/react/buildHooks.ts
Ephem: > That would start all those "the last component unsubscribed" timers for cache cleanup

Since I'm not familiar with what composes "a subscription" in RTK Query terms I'm just asking naïve questions here, but couldn't this be done for all queries in the store at hydration anyway? Does there need to be a previous subscription for this to work? (I get why that might make implementation easier though)

The reason I'm asking is that since components only render once in the same render on the server, subscriptions are usually a funky mental model conceptually there and most libraries avoid them for that reason to avoid having to deal with cleaning them up (something that often leaks into userland). To me it would seem unexpected if I inspected the store state after rendering on the server and I saw component subscriptions still lingering around. Doesn't mean avoiding subscriptions on the server is necessarily the right call here of course, just trying to provide context. 😄

> It's also a marker that we have to handle these effects with extra care

I think the naming is great and doing things in render is just how Suspense works so that part is great, the potentially unexpected part for me was just that the dependency array had no meaning in some cases. One should always code so effects works without the dep array anyway, so that's no big deal.

> Memory leaks on the server should probably not be a problem

Great! Just wanted to mention it as something to look out for.

phryneas: A "subscription" in this case is more like a "reference counter". It's the information "hey, there is a component out there, displaying that data, please don't remove it from the cache". (There can be more information attached like "this component requests for the data to be polled", but I feel that's not of major importance here). 
So you might be reading more into the word that it actually means :slightly_smiling_face: 

I agree that we could do stuff like "never subscribe in the first place", but if we go this route of restoring a "live cache snapshot", it would also enable users of local storage persistence to rehydrate. 
While I personally think that most times, cache data shouldn't be persisted and if you want to have that kind of cache, just set the right HTTP headers and let the browser handle it, I'm sure there are some edge cases around where data has to be persisted - and having a rehydration tool around for that would also be handy.

It might also come down to removing subscriptions before restoring data, if it is contained in the snapshot. I guess I'll explore many options there. That is all a bit muddy before I get to it :sweat_smile:

Ephem: Ah, I was starting to suspect that a "subscription" in RTK Query was a way to _track_ subscriptions/dependents rather than actually _subscribing_ since that seems like something you would do through regular Redux means. In React Query there's the concept of "observers" which seems very similar. This does make things fall into place, thanks!

In fact, I'm pretty sure we've at some point had a bug in React Query where data fetched by prefetch was in some cases prematurely thrown away because it had no observers or something like that, so providing a way of fetching data without counting that as an "observer" also has tradeoffs.

I guess a different way to put it is, when you support imperative fetching with cache eviction, you either:

* Register that as an "dependent" and require the developer to later "unregister" that manually (the hydration tool you talk about)
* Don't register a "dependent", but now the data is immediately stale/scheduled for cache eviction, or you place it in the cache for infinite time, or you need to find some other way to deal with it

I guess it boils down to supporting imperative fetching in libraries where "query lifetime" is a central concept is kind of a mismatch that requires working around one way or the other. 😄

phryneas:
So there is kind of a demo over at https://github.com/phryneas/ssr-experiments/blob/main/nextjs-blog/pages/pokemon/%5Bid%5D.tsx

Quite hacky still (especially the hydration), but it works for SSG and SSR with next.

phryneas:
And we have some hydration. It's still wonky, because the HYDRATE action does not seem to trigger the middleware on the client side, but I'm gonna look into that at a later time.

Code in question: https://github.com/phryneas/ssr-experiments/blob/126262fcce7f8b788df3670dbc061362c65b6b9e/nextjs-blog/lib/pokemonApi.ts#L8-L12

Seems to get triggered on the server, but not on the client?

ryota-murakami:
@phryneas I'm sorry about I have no answer for your question 🙇‍♂️ Just put I thought about SSR/Next.js/`getServerProps()`
 
Next.js don't run SSR except visitor's first time access on the website.  
Open the page by SSR once, after that Next.js switching Client Side Rendering every internal page transition to fetch splitting each page chunk and get api called JSON response from `getServerSideProps()` when click `<Link href="/foo">`.

It's Next.js specific thing but seems tricky that `getServerSideProps()` behave just like API proxy when called with CSR.

phryneas:
Correction: it works perfectly, I just had a stale next cache and after deleting the `.next` folder it does what I expected.

So I guess I'm waiting for feedback on this one, if anyone wants to do some experimentation.

Also, I would tacke a suspense example next, but don't have a lot of time at the moment. If anyone want to spin up a suspense example to try it out - or a SSR example with another framework than next, I'd be very grateful.

ryota-murakami:
@phryneas Congratulations!
Will the Draft PR merge soon? or waiting for someone make example based on this PR branch Redux Toolkit build?

I think current SSR hydration flow has wasted process too much, so I wish Suspense fetch support publish soon.
Server Component also.

phryneas:
@ryota-murakami I probably won't merge this until I am sure the api here will work well with

* next SSR/SSG :heavy_check_mark: 
* the current hacky suspense for data fetching we have
* rehydration of redux-persist
* other frameworks such as gatsby

essentially, we have only one shot at doing this right.

So for experimentation, please just grab youself a [CodeSandbox build](https://ci.codesandbox.io/status/reduxjs/redux-toolkit/pr/1277/) and get going with that :)

ryota-murakami:
@phryneas sure, thanks for that information!

ishowman:
What about release a @next version for experiment feature? Waiting for Suspense support.

phryneas:
@ishowman install the CodeSandbox build [install instructions here](https://ci.codesandbox.io/status/reduxjs/redux-toolkit/pr/1277)

Look at the example repo:

* set up normally with next-redux-wrapper
* set up rehydration as in 
https://github.com/phryneas/ssr-experiments/blob/126262fcce7f8b788df3670dbc061362c65b6b9e/nextjs-blog/lib/pokemonApi.ts#L8-L12
* for SSG, fire all queries and await `getRunningOperationPromises()` in `getStaticProps` as in https://github.com/phryneas/ssr-experiments/blob/126262fcce7f8b788df3670dbc061362c65b6b9e/nextjs-blog/pages/pokemon/ssg/%5Bname%5D.tsx#L25-L38
* for SSR, fire all queries and await `getRunningOperationPromises()` in `getServerSideProps` as in https://github.com/phryneas/ssr-experiments/blob/126262fcce7f8b788df3670dbc061362c65b6b9e/nextjs-blog/pages/pokemon/ssr/%5Bname%5D.tsx#L11-L24

packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
```diff
@@ -61,14 +58,45 @@ export const build: SubMiddlewareBuilder = ({ reducerPath, api, context }) => {
...
+              state.config
+            )
+          }
+        }
+
+        // this part could have it's own middleware, but for now it lives here
+        const extracted = context.extractRehydrationInfo?.(action)
+        if (extracted) {
+          mwApi.dispatch(api.internalActions.rehydrate(extracted))
+        }
```
phryneas, line 83: TODO: move into reducer

packages/toolkit/src/query/core/buildInitiate.ts
```diff
@@ -309,14 +347,21 @@ Features like automatic cache collection, automatic refetching etc. will not be
...
+        const ret = Object.assign(returnValuePromise, {
           arg: thunkResult.arg,
           requestId,
           abort,
           unwrap: thunkResult.unwrap,
           unsubscribe: reset,
           reset,
         })
+
+        runningMutations[requestId] = ret
```
phryneas, line 359: TODO: additionally set & unset in `fixedCacheKey`

phryneas: TODO: make double sure this tracks the **latest** promise, not the first

packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -108,6 +102,9 @@ export function buildSlice({
   >
 }) {
   const resetApiState = createAction(`${reducerPath}/resetApiState`)
+  const rehydrate = createAction<ReturnType<typeof reducer>>(
+    `${reducerPath}/rehydrate`
+  )
```
phryneas, line 107: TODO: remove rehydrate action, add (memoized) `.addMatcher` using `extractRehydrationInfo` instead.

packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -233,6 +240,8 @@ export function buildSlice({
             substate.error = (payload ?? error) as any
           })
         })
+      // mutation state will not be rehydrated as it could never be retrieved by a component
+      // .addCase(rehydrate, (draft, action) => {})
```
phryneas, line 244: TODO: rehydrate when `fixedCacheKey` is used

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -618,7 +633,7 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
         }
       }, [])
 
-      return useMemo(
+      const ret = useMemo(
```
phryneas, line 636: TODO undo change

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -695,7 +712,9 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
         }
       }, [arg, trigger])
 
-      return useMemo(() => [trigger, arg], [trigger, arg])
+      const ret = useMemo(() => [trigger, arg] as const, [trigger, arg])
+
+      return ret
```
phryneas, line 717: undo code change

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -841,10 +860,12 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
       )
 
-      return useMemo(
-        () => [triggerMutation, finalState],
+      const ret = useMemo(
+        () => [triggerMutation, finalState] as const,
         [triggerMutation, finalState]
       )
+
+      return ret
```
phryneas, line 868: undo code change

phryneas (commented):
I'm leaving myself some TODOs here to also make it work with `fixedCacheKey`. I hope I'll get around to doing these changes tomorrow.

alibidjandy:
Hi 
excuse me just developed a single project with next.js and RTK-query everything is fine but after ctrl + u there's not something as a html , Head , etc 
just did step by step of your way of manage Next.js with RTK query 
could you plz have a look at my project on github :
[Github](https://github.com/alibidjandy/single-store-page/blob/main/pages/products/index.tsx)
this is the main page I'd like to see its information on source code
and live view of my project :
[Netlify](https://suspicious-varahamihira-ae0993.netlify.app/products/7)

PR #1281, opened 2021-07-10
Add URL that was requested to all FetchBaseQueryError responses

dggsax:
As a wrapper of `fetch`, I believe it would be valuable to append the URL that was requested to the three different possible error responses defined in `FetchBaseQueryError` because with a library like RTK-Query it will be handy to have easy access to the fully-form URL that was made as part of the request, including the URL search parameters. While the use of the `url` property on the error object would admittedly not be used as much in the UI, there are cases where displaying that information and having access to it is very useful.

I understand that there are considerations of impacts in the bundle size, how can I calculate/determine the impact of this change on the bundles?

Just curious to see what y'all think about this. Thanks!


phryneas:
Hmm.   
The problem is, the more we add here, the more is also required for errors returned by `fetchFn`, as `fetchFn` has to return the same error format as the underlying `baseQuery`.  
We could get around that by adding another `"UNKNOWN"` or `"OTHER"` status code that would not need any of those information.

But even then, honestly I don't think that this is information many people would actually want to access in their components.

Accessing it in a middleware? 
Sure. The full `response` object is already in `action.meta.baseQueryMeta.response` and can be accessed by middleware. Same for the request.
Accessing it in lifecycle?  
The full response is, as part of `meta`, already accessible in `onQueryStarted` by awaiting `queryFulfilled`. Same for the request.  
But accessing it in a component? Unlike the `statusText`, the url seems like a very internal thing that you probably would not want do display to any end user ever.

dggsax:
I think I get what you're saying, however I hope I can make an argument for this change that isn't "I want to make something more convenient for me" and instead "I want to make it more convenient for everyone".

One of the most impressive aspects of RTK-Query is the fact that the entire end-to-end data fetching experience from the perspective of building a user interface comes out of the box and is accessible with one hook, removing the need to manage that state information that controls the UI in custom redux logic. Such as having a piece of state to track when loading and fetching data (super easy now with the `isFetching` variable returned by the hook), I no longer need to think about integrating and managing this logic in a reducer and controlling that state with complex thunks.

The missing piece for my team and our projects in this experience of adopting RTK-Query is the `error` property that you can utilize from generated query hook: for the error's `status`, `statusText`, and `data`, what was the URL that was queried? One thing that we hope to be able to do is have a component that can just take that `error` object as a prop and figure out how to render it depending on the `NODE_ENV`, the user using the site, as well as whether they've specified in the settings that they want to see the full error information. Most importantly, as we need to support our customers, is the magical "copy to clipboard" functionality that we advise our customers to use so it's really easy for them to share debugging details with us to help get them unblocked. Having easy access to the URL for the query that was made is the missing piece of this puzzle, and for 20 bytes I hope the argument can be made that it will be valuable to add and worth it.

In the meantime, it's great to know about the middleware options for accessing that crucial piece of information and I will explore it, but it kind of feels like a step backwards from the "all-in-one" hook experience that there is that makes the loading, fetching, and "here's the data" logic integrated and easy to consume and understand with RTK-Query now.

Let me know what you think, if you still think it's not worthy of adding it to the types, no worries and thanks for the consideration! 🙂

phryneas:
The thing is: usually that information is *not* exposed in any way to the user, and stuff like error reporting usually happens transparently in the back, not with user interaction.
 
I get that in your case that information might be useful, but I *really* believe that this is an extreme edge case and not something common. If it comes up more often, I might reconsider, but so far you are the first person to ever mention a use case even remotely like this.

I'm also open if @markerikson, @msutkowski or @Shrugsy see this differently, but from my experience I don't think this will be used by many.

In the meantime, there is a documented way of adding that information to all errors for you, without having to add it to the library: [wrapping your `baseQuery`](https://redux-toolkit.js.org/rtk-query/usage/customizing-queries#automatic-re-authorization-by-extending-fetchbasequery):


```js
const fetchBaseQueryWithUrl = (...args) => {
  const bq = fetchBaseQuery(...args)
  return (...args) => {
    const result = bq(...args)
    if (result.error?.code) {
      return { ...result, error: { ...result.error, url: result.meta.request.url }}
    }
    return result
  }
}
```

msutkowski:
I'd personally lean towards the wrapper like you've shown instead of baking it in - its super to cheap to implement and a more flexible solution long term

Shrugsy:
I'm open to this being another [customizing queries](https://redux-toolkit.js.org/rtk-query/usage/customizing-queries#examples---basequery) example such as `Altering result data by wrapping fetchBaseQuery`. We do have the [automatic re-authorization by extending fetchBaseQuery](https://redux-toolkit.js.org/rtk-query/usage/customizing-queries#automatic-re-authorization-by-extending-fetchbasequery) example that @phryneas pointed out, but this could be a clearer example of "how to wrap fBQ, and tweak the result".  That's also arguably more useful to the userbase than making the change into fBQ directly.

I'm not very convinced as to the usefulness of this change into fBQ itself at the moment. @dggsax your anecdotal case about this being the missing piece doesn't hold weight, given that it's a simple matter to use a custom wrapped fBQ with the functionality you want. You will still have the 'all in one' hook by doing that

dggsax:
Thanks for the feedback and perspectives all, I agree that turning this into a more fleshed-out example of extending the `fetchBaseQuery` to do more things will serve more in the long run as opposed to doing what I was proposing.

After some finagling, I think I got an example working that I like so far, let me know what y'all think, and I can help with adding it to the docs for examples of customizing queries

```typescript
/**
 * Type definition for extended BaseQuery with url
 * in the error response
 */
type ExtendedBaseQuery = BaseQueryFn<
  string | FetchArgs,
  unknown,
  FetchBaseQueryError & { url?: string }
>;

/**
 * Create a new baseQuery instance used for the createApi method
 * from RTK-Query
 *
 * @param baseUrl the base URL
 * @returns An instance of fetchBaseQuery extended to include URL information in the error response
 */
export const createBaseQuery = (baseUrl: string): ExtendedBaseQuery => {
  const validateBaseUrl = new RegExp(/\/.+/gm);
  if (!validateBaseUrl.test(baseUrl)) {
    throw new Error("Base url must start with a forward slash");
  }

  // Initialize baseQuery with fetchBaseQuery
  const baseQuery = fetchBaseQuery({
    baseUrl:
      process.env.NODE_ENV === "production"
        ? process.env.PUBLIC_URL + baseUrl
        : baseUrl,
    prepareHeaders: async (headers) => {
      const token = await authProvider.getApiToken();
      headers.set("Authorization", `Bearer ${token.accessToken}`);
      return headers;
    },
  });

  // Construct base query function that will
  // include the request/response URL in the error
  const baseQueryWithUrl: ExtendedBaseQuery = async (
    args,
    api,
    extraOptions,
  ) => {
    const result = await baseQuery(args, api, extraOptions);

    if (result.error !== undefined) {
      return {
        ...result,
        error: {
          ...result.error,
          url: result.meta?.request.url,
        },
      };
    }
    return result;
  };

  return baseQueryWithUrl;
};

// ... and then some examples
const fooApi = createApi({
  ...,
  baseQuery: createBaseQuery("/api/foo");
  ...
});

const barApi = createApi({
  ...,
  baseQuery: createBaseQuery("/api/bar"),
  ...
});
```

One thing I definitely like about the wrapper approach is it helped me figure out the next challenge I was getting too which was making sure all the API's did the same `prepareHeaders` logic

phryneas:
That one sure looks nice :)

>  all the API's did the same prepareHeaders logic

Ohh, spider senses tingling. Just to make sure: as long as those are not completely independent external servers (which I doubt since you access them all through `process.env.PUBLIC_URL`) we *really* recommend that you create only *one* api with *all* the endppoints! (You can still code-split by using `injectEndpoints` like in the "Code Splitting" docs page)

We had initially not stressed that enough in the docs, but by now it should be plastered at about 4 or 5 different places.

dggsax:
Oooh, I had totally forgotten about that concept when reading the documentation a couple weeks ago, I remember being pretty stoked about that principle since we use a lot of external API's in our tool. 

In that case I'll probably make the `createBaseQuery` example take a `FetchBaseQueryArgs`, and then start experimenting with the `injectEndpoints` functionality. Will fix this when I get home later and see if there's any updates I want to make to the wrapper!

PR #1304, opened 2021-07-15
undo adding `statusText` since it is empty in HTTP/2

phryneas:
I wasn't aware of this, but HTTP/2 does not send a `statusText`, so it would be an empty string, depending on host configuration. Better leave it out before it confuses users - also, no sense in keeping it in when it is going away in the long run.

https://github.com/whatwg/fetch/issues/599

### DIFF
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -66,7 +66,6 @@ export type FetchBaseQueryError =
        */
       status: number
       data: unknown
-      statusText: string
     }
   | {
       /**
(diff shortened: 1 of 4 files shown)


packages/toolkit/src/query/fetchBaseQuery.ts
```diff
@@ -86,10 +85,18 @@ export type FetchBaseQueryError =
...
     }
+  | {
+      /**
+       * * `"CUSTOM_ERROR"`:
+       *   A custom error type that you can return from your `fetchFn` where another error might not make sense.
+       **/
+      status: 'CUSTOM_ERROR'
+      data?: unknown
+      error: string
+    }
```
phryneas, line 99: also adding this, which will probably come in handy for users of `queryFn`

msutkowski (approved):
Looks fine to me... and that's an interesting change, but I get it!

Issue #1303, opened 2021-07-15
RTKQ: IE11 Immer dependency issue.


komar94:
Hello,

I have a problem with getting RTKQ to work on IE11. I called `enableES5()` before doing anything with the store, but IE11 still throws this exception: 

`SCRIPT5022: [Immer] The plugin for 'ES5' has not been loaded into Immer. To enable the plugin, import and call `enableES5()` when initializing your application.`

I'm attaching a repository where the issue is recreated: https://github.com/komar94/redux-toolkit-issue

Playing with this in our development environment lead me to believe that there is something wrong with api middleware I tried to comment it out and error was gone, but I did not manage to confirm that with this example.

Do you have any idea what we did wrong?

Thank you.

markerikson:
In theory, you shouldn't have to call `enableES5` at all - we do it automatically already.

Ironically, there was a similar discussion issue filed just earlier today:

https://github.com/reduxjs/redux-toolkit/discussions/1302

We did make a change to when we call that in RTK 1.6, so it's possible that we ended up with a regression in behavior.

komar94:
hey @markerikson - thanks for very quick reply. 

thanks to your reply in #1302, I tried putting `enableES5()` back to `index.ts`. That resolves the problem in the repository I shared. 

Funny thing is that same steps in app created with `create-react-app` does not have that problem. Could there be that webpack 4.x is doing something different with imports than 5.x and that causes the problem?

Do you think this is something you can put back to what was before? I also summited a PR with the change.

markerikson:
The goal was to try to eliminate the extra 1KB in case people _aren't_ using one of our reducers.

That said... given that `createReducer` is at the core of both `createSlice` and RTK Query, I would have to imagine that almost everyone using RTK _is_ using that eventually, so we could probably justify the size hit that way.

When we do move on to an RTK 2.0, we'll be dropping that `enableES5` call and it'll be up to end users to do it if they need to support IE11.

markerikson:
Should be fixed in https://github.com/reduxjs/redux-toolkit/releases/tag/v1.6.1

PR #1305, opened 2021-07-15
put `enableES5()` back to index.ts

komar94:
resolve issue with immer mentioned in #1303 and #1302


markerikson:
Yep, seems like the simplest resolution here.

Issue #1315, opened 2021-07-16
RTKQ: File Uploading


ybelenko:
I use `queryFn` parameter to send `multipart/form-data`:
```js
endpoints: (build) => ({
  uploadFile: build.mutation({
    async queryFn(file, _queryApi, _extraOptions, fetchWithBQ) {
      // upload with multipart/form-data
      const formData = new FormData();
      formData.append('file', file);
      const response = await fetchWithBQ({
        url: '/files',
        method: 'POST',
        body: formData,
      }, _queryApi, _extraOptions);
      if (response.error) throw response.error;
      return response.data ? { data: response.data } : { error: response.error };
    },
  }),
}),
```

It works totally correct, but I have arguments serialization error in console:
```console
[Error] A non-serializable value was detected in the state, in the path: `api.mutations.EB07i6POmA2Xe9YPh5gpb.originalArgs`. Value: (2)
File {name: "message_saved.mp3", lastModified: 1625855746000, webkitRelativePath: "", size: 46478, type: "audio/mpeg", …}
"
Take a look at the reducer(s) handling this action type: api/executeMutation/rejected.
(See https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state)"
	(anonymous function) (main.js:5794)
	measureTime (main.js:5553)
	(anonymous function) (main.js:5789)
	(anonymous function) (main.js:5717)
	(anonymous function) (main.js:6546)
	step (main.js:5447)
	rejected (main.js:5486)
	promiseReactionJob
```

I tried to overwrite `serializeQueryArgs` option in `createApi` but it doesn't help. Can't even catch that request over there. It's weird that I see other GET requests inside `serializeQueryArgs`, but not that POST file mutation.
```js
serializeQueryArgs: (args) => {
  const { endpointName, queryArgs } = args;
  // try to overwrite uploadFile endpoint serialization
  if (endpointName === 'uploadFile') {
    const { lastModified, name, size, type } = queryArgs;
    const payload = JSON.stringify({
      lastModified, name, size, type,
    });
    return `${endpointName}(${payload})`;
  }

  // Sort the object keys before stringifying, to prevent useQuery({ a: 1, b: 2 }) having a different cache key than useQuery({ b: 2, a: 1 })
  return `${endpointName}(${JSON.stringify(queryArgs, (key, value) =>
    isPlainObject(value)
      ? Object.keys(value)
          .sort()
[...]

PR #1318, opened 2021-07-17
remove `originalArgs` from the mutation slice

### LINKED ISSUE #1209: Suggestion: Expose tools to make our own normalized query
I know making a  normalized cache is hard, but I wish there is way to directly  interact with cache and how cache is read/write 
, used

I have no experience in implementing cache so yeah... take it with a grain of salt xd

Proposal : say we want to normalize getPost and getPostByStuff, both returns array of posts which might contains duplicates

```javascript
 
  import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query'
  
  const api = createApi({
    baseQuery: fetchBaseQuery({
      baseUrl: '/',
    }),
    tagTypes: ['Post'],
    normalizedTags:["Post"]
    endpoints: (build) => ({
      getPost: build.query({
        // note: an optional `queryFn` may be used in place of `query`
        query: (id) => ({ url: `post/${id}` }),
        // Pick out data and prevent nested properties in a hook or selector
        transformResponse: (response) => response.data,
        providesTags: (result, error, id) => [{ type: 'Post', id }],
        // The 2nd parameter is the destructured `QueryLifecycleApi`
        async onQueryStarted(
          arg,
          {
            dispatch,
            getState,
            extra,
            requestId,
            queryFulfilled,
            getCacheEntry,
            updateCachedData,
          }
        ) {},
        // The 2nd parameter is the destructured `QueryCacheLifecycleApi`
        async onCacheEntryAdded(
          arg,
          {
            dispatch,
            getState,
            extra,
            requestId,
            cacheEntryRemoved,
            cacheDataLoaded,
            getCacheEntry,
            updateCachedData,
          }
        ) {},
      }),
      
      getPostBySomething: builder.query({

      normalizeTags:{"POST": (resultFromQuery, cache)=>{
       // a function that returns a array of reference 
			
      //  merge cache outselves
      
      //cache {id: POST , id: POST , newID:Post }
      // ref  {someUniqueKey: [ids]}
	}    
} 


})
      
    }),

  })

```

phryneas:
Fixes #1209 #1315

This needs a few more tests and we should generally discuss how we communicate it:

Technically, it is a breaking change (removes a store property what would have been returned by a selector), but it is a necessary bugfix.

Generally, the information removed is still available as
* property on the promise returned by `dispatch`
* part of the thunk action meta
* return value of the thunk (I could keep that using the promise approach described above)

so it has very little impact.

Are we okay putting this out as a patch release @markerikson?

### DIFF
--- a/packages/toolkit/src/query/core/apiState.ts
+++ b/packages/toolkit/src/query/core/apiState.ts
@@ -175,7 +175,6 @@ export type QuerySubState<D extends BaseEndpointDefinition<any, any, any>> = Id<
 >
 
 type BaseMutationSubState<D extends BaseEndpointDefinition<any, any, any>> = {
-  originalArgs?: QueryArgFrom<D>
   data?: ResultTypeFrom<D>
   error?:
     | SerializedError
@@ -202,7 +201,6 @@ export type MutationSubState<D extends BaseEndpointDefinition<any, any, any>> =
     } & WithRequiredProp<BaseMutationSubState<D>, 'error'>)
   | {
       status: QueryStatus.uninitialized
-      originalArgs?: undefined
       data?: undefined
       error?: undefined
       endpointName?: string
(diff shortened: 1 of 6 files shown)


markerikson:
Yeah, fine by me. What else do we need before merging?

PR #1358, opened 2021-07-26
Update doc to reflect that default middleware is not replaced when using "getDefaultMiddleware"

njradford:
Seen in the final example in this section: https://redux-toolkit.js.org/api/immutabilityMiddleware#createimmutablestateinvariantmiddleware

I think this line may have just been pasted from the prior example in the section, but isn't true when using `getDefaultMiddleware`

This whole line can also be removed. Whichever seems best to you all.

### DIFF
--- a/docs/api/immutabilityMiddleware.mdx
+++ b/docs/api/immutabilityMiddleware.mdx
@@ -120,7 +120,7 @@ import exampleSliceReducer from './exampleSlice'
 
 const store = configureStore({
   reducer: exampleSliceReducer,
-  // Note that this will replace all default middleware
+  // Note that this will NOT replace all default middleware
   middleware: (getDefaultMiddleware) =>
     getDefaultMiddleware({
       immutableCheck: {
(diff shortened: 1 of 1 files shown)


markerikson:
You could sort of argue the phrasing both ways, really.

In a sense, you _are_ replacing the _default_-default middleware with whatever is being returned here.  It's just that in this example, we are still calling `gDM()` and including its list in the return value.

Perhaps a better comment here might be:

`// This replaces the original default middleware with the customized versions`

njradford:
> You could sort of argue the phrasing both ways, really.
> 
> In a sense, you _are_ replacing the _default_-default middleware with whatever is being returned here. It's just that in this example, we are still calling `gDM()` and including its list in the return value.
> 
> Perhaps a better comment here might be:
> 
> `// This replaces the original default middleware with the customized versions`

Makes sense to me. Updated!

markerikson:
Thanks!

PR #1371, opened 2021-07-31
Improve wording for the "Typing query and mutation `endpoints`" section of the docs

ivansvlv:
…on of the docs

- added a case when `query` doesn't have parameter
- added a case when `query` has an optional parameter


phryneas:
Looks good, thanks :)

PR #1374, opened 2021-08-02
Docs: fixed typo in rtk-query docs

jbsouvestre:
There's a `test` in the rtk-query docs, stumbled into it while looking for "test" in the docs search


markerikson:
Thanks!

PR #1436, opened 2021-08-23
Pagination and cache invalidation

maruhgar:
Add a section to specify that when we cache paginated queries, in case of deletion, we should delete not just the cache of that particular id, but the LIST as well to ensure that the total count is correct as well as to avoid the same record in adjacent pages.

### DIFF
--- a/docs/rtk-query/usage/pagination.mdx
+++ b/docs/rtk-query/usage/pagination.mdx
@@ -76,6 +76,33 @@ const PostList = () => {
 };
 ```
 
+## Pagination and Caching
+
+If you have caching enabled for your paginated queries, it maybe a good idea to invalidate the `LIST` cache as well when you do a `delete` mutation.
+
+```ts title="src/features/services/post.ts
+// Or from '@reduxjs/toolkit/query' if not using the auto-generated hooks
+import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react'
+import { Post } from './types'
+
+export const postApi = createApi({  
+  reducerPath: 'postsApi',  
+  baseQuery: fetchBaseQuery({ baseUrl: '/' }),  
+  tagTypes: ['Posts'],  
+  endpoints: (build) => ({
+    deletePost: build.mutation<{ success: boolean; id: number }, number>({ 
+      query(id) {        
+        return {          
+          url: `post/${id}`,          
+          method: 'DELETE',        
+        }      
+      },      
+      // Invalidates queries that subscribe to this Post `id` as well as the `LIST`.      
+      invalidatesTags: (result, error, id) => [{ type: 'Posts', id }, 
+                                               { type: 'Posts, 'LIST'}],    
+   }),
+})
+```
 ## General Pagination Example
 
 In the following example, you'll see `Loading` on the initial query, but then as you move forward we'll use the next/previous buttons as a _fetching_ indicator while any non-cached query is performed. When you go back, the cached data will be served instantaneously.


maruhgar (commented):
Syntax error in code snippet corrected

maruhgar (commented):
Second attempt fixing syntax error

maruhgar (commented):
Fix syntax error in code snippet

maruhgar (commented):
Include Post interface to fix compilation error

phryneas:
@Shrugsy can you please take a look if this okay at this position in the flow of the general document?

Shrugsy:
> @Shrugsy can you please take a look if this okay at this position in the flow of the general document?

I think it's probably better off as a `h3` within the `Pagination Recipes` subheader, rather than as a sibling of `Pagination Recipes`, as it does come across as a recipe to me.

I've got a few other thoughts I'll leave as code comments shortly

docs/rtk-query/usage/pagination.mdx
```diff
@@ -76,6 +76,43 @@ const PostList = () => {
 };
 ```
 
+## Pagination and Caching
+
+If you have caching enabled for your paginated queries, it maybe a good idea to invalidate the `LIST` cache as well when you do a `delete` mutation.
```
Shrugsy, line 81:
```suggestion
It is a common use-case to utilize tag invalidation to perform
[automated re-fetching](./automated-refetching.mdx) with RTK Query.

A potential pitfall when combining this with pagination is that your paginated query may only
provide a _partial_ list at any given time, and hence not `provide` tags for entity IDs that
fall on pages which aren't currently shown. If a specific entity is deleted that falls on an
earlier page, the paginated query will not be providing a tag for that specific ID, and will
not be invalidated to trigger re-fetching data. As a result, items on the current page that
should shift one item up will not have done so, and the total count of items and/or pages
may be incorrect.

A strategy to overcome this is to ensure that the `delete` mutation always `invalidates` the
paginated query, even if the deleted item is not _currently_ provided on that page. We can
leverage the concept of
[advanced invalidation with abstract tag ids](./automated-refetching.mdx#advanced-invalidation-with-abstract-tag-ids)
to do this by `providing` a `'Posts'` tag with the `'PARTIAL-LIST'` ID in our paginated query,
and `invalidating` that corresponding tag for any mutation that should affect it.
```

This adds more detail about the specific scenario we're trying to address, and specifically refers to the tag invalidation usage page to provide context about the topic.

It also describes it using a 'PARTIAL-LIST' tag ID rather than just 'LIST' in order to assist in the distinction.

docs/rtk-query/usage/pagination.mdx
```diff
@@ -76,6 +76,43 @@ const PostList = () => {
...
+        }      
+      },      
+      // Invalidates queries that subscribe to this Post `id` as well as the `LIST`.
+      invalidatesTags: (result, error, id) => [
+        { type: "Posts", id },
+        { type: "Posts", id: "LIST" },
+      ],
+    }),
+  }),
+})
```
Shrugsy, line 114:
```suggestion
// file: api.ts
import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react'
interface Post {
  id: number
  name: string
}
interface ListResponse<T> {
  page: number
  per_page: number
  total: number
  total_pages: number
  data: T[]
}

export const postApi = createApi({
  reducerPath: 'postsApi',
  baseQuery: fetchBaseQuery({ baseUrl: '/' }),
  tagTypes: ['Posts'],
  endpoints: (build) => ({
    listPosts: build.query<ListResponse<Post>, number | void>({
      query: (page = 1) => `posts?page=${page}`,
      // highlight-start
      providesTags: (result, error, page) =>
        result
          ? [
              // Provides a tag for each post in the current page,
              // as well as the 'PARTIAL-LIST' tag.
              ...result.data.map(({ id }) => ({ type: 'Posts' as const, id })),
              { type: 'Posts', id: 'PARTIAL-LIST' },
            ]
          : [{ type: 'Posts', id: 'PARTIAL-LIST' }],
      // highlight-end
    }),
    deletePost: build.mutation<{ success: boolean; id: number }, number>({
      query(id) {
        return {
          url: `post/${id}`,
          method: 'DELETE',
        }
      },
      // Invalidates the tag for this Post `id`, as well as the `PARTIAL-LIST` tag,
      // causing the `listPosts` query to re-fetch if a component is subscribed to the query.
      // highlight-start
      invalidatesTags: (result, error, id) => [
        { type: 'Posts', id },
        { type: 'Posts', id: 'PARTIAL-LIST' },
      ],
      // highlight-end
    }),
  }),
})
```

This changes a few things:
- Shows the types explicitly
- Includes the `listPosts` paginated query and the tags it provides (no point invalidating tags when there is no other query providing them)
- Uses the `PARTIAL-LIST` tag for clarity
- Removes the "// Or from '@reduxjs/toolkit/query/react'" comment, which is already the same import, and not really necessary to include for this section of the docs
- Makes the tag providing/invalidating more obvious

Shrugsy (changes requested):
@maruhgar Could you please look over these changes and add them if you agree? IMO the changes should help clarify a lot about the topic

maruhgar (commented):
Applying code review suggestions

Shrugsy (approved):

Shrugsy:
LGTM

PR #1449, opened 2021-08-28
fix: `createAsyncThunk` union return values fall back to allowing only single member

### LINKED ISSUE #1156: Typescript: createAsyncThunk type infer not working correctly in 1.6.0 with multiple enum
In 1.5.1 work fine, in 1.6.0 not.
If explicitly typed, everything ok.
```
Argument of type '() => Promise<Example>' is not assignable to parameter of type 'AsyncThunkPayloadCreator<Example.A, void, {}>'.
  Type 'Promise<Example>' is not assignable to type 'MaybePromise<Example.A | RejectWithValue<unknown, unknown>>'.
    Type 'Promise<Example>' is not assignable to type 'Promise<Example.A | RejectWithValue<unknown, unknown>>'.
      Type 'Example' is not assignable to type 'Example.A | RejectWithValue<unknown, unknown>'.
        Type 'Example.B' is not assignable to type 'Example.A | RejectWithValue<unknown, unknown>'.ts(2345)
```

```
export enum Example{
  A = 0,
  B = 1
}

export const bug = createAsyncThunk("foo/bar", async () => {
  if (Math.random() > 0.5) {
    return Example.A;
  } else {
    return Example.B;
  }
});
```

phryneas:
This should fix #1156

### DIFF
--- a/packages/toolkit/src/createAsyncThunk.ts
+++ b/packages/toolkit/src/createAsyncThunk.ts
@@ -177,7 +177,7 @@ type GetSerializedErrorType<ThunkApiConfig> = ThunkApiConfig extends {
   ? GetSerializedErrorType
   : SerializedError
 
-type MaybePromise<T> = T | Promise<T>
+type MaybePromise<T> = T | Promise<T> | (T extends any ? Promise<T> : never)
 
 /**
  * A type describing the return value of the `payloadCreator` argument to `createAsyncThunk`.
--- a/packages/toolkit/src/tests/createAsyncThunk.typetest.ts
+++ b/packages/toolkit/src/tests/createAsyncThunk.typetest.ts
@@ -137,6 +137,16 @@ const anyAction = { type: 'foo' } as AnyAction
   expectType<RejectValue>(unwrapResult(returned))
 })()
 
+/**
+ * regression #1156: union return values fall back to allowing only single member
+ */
+;(async () => {
+  const fn = createAsyncThunk('session/isAdmin', async () => {
+    const response: boolean = false
+    return response
+  })
+})()
+
 {
   interface Item {
     name: string


packages/toolkit/src/createAsyncThunk.ts
```diff
@@ -177,7 +177,7 @@ type GetSerializedErrorType<ThunkApiConfig> = ThunkApiConfig extends {
   ? GetSerializedErrorType
   : SerializedError
 
-type MaybePromise<T> = T | Promise<T>
+type MaybePromise<T> = T | Promise<T> | (T extends any ? Promise<T> : never)
```
Shrugsy, line 180: ❓ 

For my own learning, do you know *why* it works with this change, but not otherwise? I'm guessing it's something to do with `IsUnknown`, but my brain is turning to jelly trying to understand this

phryneas: https://artsy.github.io/blog/2018/11/21/conditional-types-in-typescript/ - look for this paragraph:

> All we did is wrap the meat of our logic in a flimsy tortilla of inevitability, since the outer condition <code>A extends any</code> will, of course, always be true

In this case, assume

```ts
type Union = 'A' | 'B'

type X = Promise<Union> // will be Promise<'A' | 'B'>
type Y = Union extends any ? Promise<Union> : never // will be Promise<'A'> | Promise<'B'>

```

Shrugsy: aha, so a case of this then: https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types

I'll probably have to encounter this a couple more times before it sinks in to me that it occurs with `boolean` too (which I guess is 'just' a union of `true | false`.

Thanks for the explanation & link, I'll go over that article in full to brush up my knowledge

phryneas: Yeah, that is happens with `boolean` was completely new to me as well.

bfricka: I just have one question about this now: Isn't `T extends any ? Promise<T> : never` essentially the only type now? `T` will always extend `any` so this will always be `Promise<T>` and not `T`, right? I'm probably missing something.

Shrugsy: @bfricka it would have to look like this to fit the behaviour you describe:

```ts
type MaybePromise<T> = T extends any ? Promise<T> : T
```

As it is, it's still a union of three possibilities. It might make more sense if you format it a bit differently:
```ts
type MaybePromise<T> = 
  | T
  | Promise<T>
  | (T extends any ? Promise<T> : never)
```

So the fact that `T extends any` is only relevant for the third scenario. The other unions of `T` and `Promise<T>` don't care whether `T extends any` is true or not

Issue #1451, opened 2021-08-29
fetchBaseQuery does not work with node-fetch due to usage of response.clone()


GU5TAF:
When fetchBaseQuery is used in a node environment where fetch has been poly/ponyfilled requests fail to complete due the usage of `response.clone()`. This is a known issue with [node-fetch#custom-highwatermark](https://github.com/node-fetch/node-fetch#custom-highwatermark) and is solved by resolving the cloned response in parallel.

[Minimal Repro](https://codesandbox.io/s/fetchbasequery-broken-with-node-fetch-irklu?file=/src/services/pokemon.ts)

GU5TAF:
Closing due to issue being addressed in https://github.com/reduxjs/redux-toolkit/pull/1277

PR #1452, opened 2021-08-29
fix: resolve cloned response in parallel in fetchBaseQuery

GU5TAF:
Fixes https://github.com/reduxjs/redux-toolkit/issues/1451

### DIFF
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -238,7 +238,10 @@ export function fetchBaseQuery({
 
     let resultData
     try {
-      resultData = await handleResponse(response, responseHandler)
+      // note: node-fetch will hang when using response.clone() unless we resolve the cloned response in parallel
+      // https://github.com/node-fetch/node-fetch#custom-highwatermark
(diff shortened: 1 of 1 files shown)


phryneas:
Thanks for the PR! This has already been solved over in #1277 (with https://github.com/reduxjs/redux-toolkit/pull/1277/commits/be2dbb7300b258f6db5ce6c26aa3e7e9137145fc ), which also adds some other changes to `fetchBaseQuery`, so this change would probably result in a merge error and mostly be removed after merge of that PR.

So I'm gonna close this one, I hope that doesn't discourage you from future PRs.

Sorry you had to go down this rabbit hole. I know from personal experience how frustrating that bug is.

Until we release the next version, you can use the CodeSandbox build of this PR - it will stay around. Also, if you are in a SSR environment, feedback in #1277 would be great :)

GU5TAF:
Ah, I skimmed that PR but missed that it had been addressed 🤦‍♂️ Happy to hear it's being fixed, it was indeed a frustrating evening of wondering why fetch wasn't working 😅

phryneas:
Yeah, that cost me about two days I believe :sweat_smile:

PR #1453, opened 2021-08-30
fix lint 'consistent-type-imports' rules

ryota-murakami:
Fixed lint [consistent-type-imports](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-imports.md#prefer) rule based on prefer type setting at [.eslintrc.js](https://github.com/reduxjs/redux-toolkit/blob/master/.eslintrc.js#L19) :)

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -465,7 +465,7 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
   api,
   moduleOptions: { batch, useDispatch, useSelector, useStore },
 }: {
-  api: Api<any, Definitions, any, any, CoreModule>
+  api: Api<any, Definitions, any, any>
   moduleOptions: Required<ReactHooksModuleOptions>
 }) {
   return { buildQueryHooks, buildMutationHook, usePrefetch }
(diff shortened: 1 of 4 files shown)


packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -465,7 +465,7 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
   api,
   moduleOptions: { batch, useDispatch, useSelector, useStore },
 }: {
-  api: Api<any, Definitions, any, any, CoreModule>
+  api: Api<any, Definitions, any, any>
```
phryneas, line 468: this seems to be a type change. was that intended?

ryota-murakami: Ah, this change coming from WebStorm IDE support that says

"_Explicit type argument any, Definitions, any, any, CoreModule can be replaced with any, Definitions, any, any_"
Screen Shot 2021-09-04 at 23 03 49

Because type `API`'s 4th generics argument has set default type `CoreModule`.

Screen Shot 2021-09-04 at 23 04 33

Would you prefer bring back original explicit way?  
Personally original is more friendly for external contributors.

Issue #1447, opened 2021-08-26
`setOne/setMany` missing from `createEntityAdapter` documentation


jjspace:
I recently stumbled across issue #954 when looking for something else. I saw that it was closed but I didn't even know those methods existed because they're not shown in the CRUD functions section of `createEntityAdapter` (see below). It would be great if these could be added and described there so people don't have to look through the change log to notice they were added.

https://redux-toolkit.js.org/api/createEntityAdapter#crud-functions

![20210826_132340](https://user-images.githubusercontent.com/8007967/131008213-3ca136d9-b0de-49c6-a61c-4be4591dfe68.png)

markerikson:
I'd have to check the implementation, but I'm pretty sure the `set` functions completely replace any existing data, in a couple different ways.  So, `setOne/setMany` will guarantee that the items you just provided end up in the state exactly as-is, while leaving other items intact. `setAll` completely removes all existing data and replaces everything with exactly what you provided.

Agreed that we could use some better documentation here.  If you've got some time to submit PRs to improve this, I'd appreciate it!

jjspace:
@markerikson I opened up a PR with some updated docs. Please take a look when you get a chance!

Issue #954, opened 2021-03-28
Add `setOne/setMany` CRUD APIs to createEntityAdapter


markerikson:
We can currently add new items or merge into existing items, but there's no built-in way to _replace_ items.

NgRx actually added a `setOne` API shortly after we ported the adapter logic over;

https://github.com/ngrx/platform/issues/2369

We might as well port `setOne`, and add `setMany` also.

HHK1:
Waouh I *just* stumbled upon that issue, I love it when the exact feature request you wished for is already opened in the issues. 

Can I implement this or is someone else working on it already ?

markerikson:
@HHK1 No one's working on it yet - go for it!

Note that the entity adapter logic is split into two files, `unsorted_state_adapter.ts` and `sorted_state_adapter.ts`.  Most of the adapter CRUD functions have split implementations on either side, but a couple of them (the `remove*` functions) are only implemented in the `unsorted` file and then imported and reused in the `sorted` file.

I imagine that we would probably want to have `setOne` and `setMany` share a common implementation internally.  For example, I think one of the `add*` implementations really just has `addMany()`, and `addOne` is implemented as `addMany([entity], state)` (ie, turning the single item into a one-item array).  We probably want to do something similar here.

HHK1:
>(ie, turning the single item into a one-item array)

Actually I went the other way around, just like addMany calls addOne in a for loop, do you see any issues with it ?

```
function setOneMutably(entity: T, state: R): void {
    const key = selectIdValue(entity, selectId)
    if (!(key in state.entities)) {
      state.ids.push(key)
    }
    state.entities[key] = entity
  }

  function setManyMutably(
    newEntities: T[] | Record<EntityId, T>,
    state: R
  ): void {
    newEntities = ensureEntitiesArray(newEntities)
    for (const entity of newEntities) {
      setOneMutably(entity, state)
    }
  }
```

I'm done with the unsorted, adding the same to the sorted adapter right now.

mr-new-0509:
Hello, dear.
I can't find setOne in entityAdapter. Please let me know the reason. In the doc, it is clearly described setOne is existed.
But I can't find it and the error is occured.
Please help me.
Thanks.

markerikson:
@mr-new-0509 what version of RTK are you using? It was only added in 1.6.0.

mr-new-0509:
I am using 1.5.1.
Thanks a lot.
[...]

PR #1460, opened 2021-08-31
Docs: add `setOne` and `setMany`, describe `add` vs `set` vs `upsert`

jjspace:
- Adds `setOne` and `setMany` CRUD functions to `createEntityAdapter` documentation
- Adds an info block to expand on the differences between the `add*`, `set*` and `upsert*` functions. I took the behavior from my understanding of [these](https://github.com/reduxjs/redux-toolkit/blob/2ff1fa4364ce0dfa82dbbd3187aa0c63e57d5582/packages/toolkit/src/entities/sorted_state_adapter.ts#L26) functions

Fixes: https://github.com/reduxjs/redux-toolkit/issues/1447

Also note that these functions are only available in RTK v1.6 and up as that's when they were [added](https://github.com/reduxjs/redux-toolkit/issues/954). I'm not sure if there's any convention for pointing out things like this. Personally the project we use this on is still on v1.5 so we can't use this until we update and it would be nice to know that looking at the documentation

### DIFF
--- a/docs/api/createEntityAdapter.mdx
+++ b/docs/api/createEntityAdapter.mdx
@@ -203,9 +203,11 @@ export interface EntityAdapter<T> extends EntityStateAdapter<T> {
 
 The primary content of an entity adapter is a set of generated reducer functions for adding, updating, and removing entity instances from an entity state object:
 
-- `addOne`: accepts a single entity, and adds it.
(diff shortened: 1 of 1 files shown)


markerikson:
Looks good!

Yeah, I've seen various conventions for trying to point out new content. Maybe we could come up with a little badge, kinda like we do in the Redux Style Guide, but with text like "New in 1.6".

Issue #1456, opened 2021-08-30
1.6 breaks code thunk return type when rejectWithValue is used


bfricka:
Prior to 1.6, returning `rejectWithValue` (evaluates to `RejectValue<unknown>`, if it's not explicitly typed), would not change the return type of the `AsyncThunk`. For example:

```ts
const fetchFoo = createAsyncThunk('foo', async (_: void, { dispatch, rejectWithValue }) => {
  try {
    return await dispatch(fetch<Foo>())
  } catch (e) {
    return rejectWithValue(e)
  }
})

// Then
try {
  const foo = await dispatch(fetchFoo()).then(unwrapResult) // 1.5.x: Foo, 1.6.x: unknown
} catch (e) {
  // Do something with the actual rejected value
   console.log(e.body?.code)
}
```
This issue persists even after applying the https://github.com/reduxjs/redux-toolkit/pull/1449 fix.

I'm not sure exactly what's going on here, because I honestly don't understand the magic behind the `AsyncThunkPayloadCreatorReturnValue`, neither in 1.5 nor 1.6. It looks like `unwrapResult` and its associated types didn't change at all.

phryneas:
Could you open a PR with a failing type test for this? That would make it easier to fix it.

markerikson:
Available in the 1.7 betas.

PR #1461, opened 2021-08-31
Add test for #1456

bfricka:
- Test `createAsyncThunk` + `rejectWithValue` without generics. This test works
    in 1.5.1 but not in 1.6.0 - 1.6.1 as described in #1456

### DIFF
--- a/packages/toolkit/src/tests/createAsyncThunk.typetest.ts
+++ b/packages/toolkit/src/tests/createAsyncThunk.typetest.ts
@@ -7,7 +7,10 @@ import type { AxiosError } from 'axios'
 import apiRequest from 'axios'
 import type { IsAny, IsUnknown } from '@internal/tsHelpers'
 import { expectType } from './helpers'
-import { AsyncThunkPayloadCreator } from '@internal/createAsyncThunk'
+import type {
+  AsyncThunkFulfilledActionCreator,
+  AsyncThunkRejectedActionCreator,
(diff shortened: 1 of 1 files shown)


phryneas:
Curiously this only seems to happen if a payload creator returns a result asynchronously, but returns `rejectWithValue` synchronously. Fascinating stuff like that even happens out in the wild, given we also have the `condition` option ^^

I'm gonna take a look.

phryneas:
I could not find a way to make this inferrable any more. 
With `rejectWithMeta` adding another complexity level on this, I think the TypeScript compiler will just not work with inferring this.

Given that this only happens in the edge case "no reject type defined, return value async and rejection value sync" and can be circumvented with  either of
* a relatively simple type annotation of `<boolean, void>`
* by making the whole function `async` 
* by returning the rejection as a resolved promise as well

I would keep this open and add a "help wanted" label.

If someone finds a fix for it, I'd be happy to accept it, but at this point I'm not sure it can be solved and if it is, it would probably take a significant amount of work to do, while we have a lot other priorities going on right now.

If you find other, more common, cases I'm happy to repriotize tyhis.

phryneas:
Seems that this is solved in the current 1.7 beta. Can you please verify that?

nickpelone:
I had this same issue and can verify that `1.7.0-beta.0` makes this infer correctly again. Thank you!

PR #1476, opened 2021-09-03
add `reset` method to useMutation hook

### LINKED ISSUE #1455: Feature Request: Callback to Reset Mutation Status
I would like for mutation hooks to return a callback to reset mutation status back to their initial state so that the isSuccess, isError, etc. flags are all false.

The rational is for cases like modals whose contents change in response to mutation status change where the modal may need to be re-opened after closing. It's possible to get this behavior by restructuring your component hierarchy so that the hook's component gets unmounted, or by abstracting access to state, but the first isn't always possible and the second is messier than simply resetting the mutation status as a counter to firing the mutation's trigger function.

Example:

```tsx
import React from 'react';
import {useAddToCartMutation} from './api';
import Modal from './Modal';

interface Props {
  isOpen?: boolean;
  onClose?: () => void; // Consumer will use this to set isOpen to false
}

const AddToCartModal = (props: Props) => {
  const [addToCart, addToCartMutation] = useAddToCartMutation();

  const handleAddToCart = () => { addToCart(); }
  
  const handleClose = () => { setIsOpen(false); props.onClose(); }

  useEffect(() => {
   if (props.isOpen) {
     addToCartMutation.reset(); // <--- HERE
   }
  }, [props.isOpen]);

  return <Modal open={props.isOpen}>
    {addToCartMutation.isFetching ? <div>Adding to cart...</div> :
     addToCartMutation.isError ? <><div>An error occurred</div><div><button onClick={handleClose}>Close</div> :
     addToCartMutation.isSuccess ? <><div>Thank you!</div><div><button onClick={handleClose}>Close</div> :
     <button onClick={handleAddToCart}>Add to Cart</button>
  </Modal>
};
```

phryneas:
This would resolve #1455.

Also, it eliminiates all `useRef` usages in the `useMutation` hook, which is probably not a bad thing.


docs/rtk-query/api/created-api/hooks.mdx
```diff
@@ -360,7 +360,10 @@ selectFromResult: () => ({})
 
 - **Returns**: A tuple containing:
   - `trigger`: A function that triggers an update to the data based on the provided argument. The trigger function returns a promise with the properties shown above that may be used to handle the behavior of the promise
-  - `mutationState`: A query status object containing the current loading state and metadata about the request, or the values returned by the `selectFromResult` option where applicable
+  - `mutationState`: A query status object containing the current loading state and metadata about the request, or the values returned by the `selectFromResult` option where applicable.
+    Additionally, this object will contain
+    - a `reset` method to reset the hook back to it's original state and remove the current result from the cache
+    - an `originalArgs` property that contain the argument passed to the last call of the `trigger` function.
```
Shrugsy, line 366:
```suggestion
    - an `originalArgs` property that contains the argument passed to the last call of the `trigger` function.
```

docs/rtk-query/api/created-api/hooks.mdx
```diff
@@ -360,7 +360,10 @@ selectFromResult: () => ({})
 
 - **Returns**: A tuple containing:
   - `trigger`: A function that triggers an update to the data based on the provided argument. The trigger function returns a promise with the properties shown above that may be used to handle the behavior of the promise
-  - `mutationState`: A query status object containing the current loading state and metadata about the request, or the values returned by the `selectFromResult` option where applicable
+  - `mutationState`: A query status object containing the current loading state and metadata about the request, or the values returned by the `selectFromResult` option where applicable.
+    Additionally, this object will contain
+    - a `reset` method to reset the hook back to it's original state and remove the current result from the cache
```
Shrugsy, line 365: ❓ 

Based on the description, I would have assumed that `originalArgs` gets reset as well. Should that be the case?

Otherwise, is it worth pointing out that it doesn't get reset?

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -717,47 +722,37 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
       )
 
       const currentState = useSelector(mutationSelector, shallowEqual)
-      const originalArgs = promiseRef.current?.arg.originalArgs
+      const originalArgs = promise?.arg.originalArgs
       const finalState = useMemo(
         () => ({
           ...currentState,
           originalArgs,
+          reset: promise?.unsubscribe,
```
Shrugsy, line 753:
```suggestion
          reset: promise?.unsubscribe ?? (() => undefined),
```

As it is, this results in lying to the types: Before any mutation is triggered, `reset` will be undefined rather than `() => void`, and hence result in a `TypeError` if attempted to be called. It's probably better to let it do a no-op call in that situation

phryneas: Very good catch, I gotta fix that one!

Shrugsy: ❓ 

Just a bit of a lingering thought:

Given that the `reset` method is essentially an alias to the latest `promise.unsubscribe` property, it may come across as having counter-intuitive behaviour from the user's perspective.

i.e. a user will not intuitively expect that they *need* to be concerned about closures over `reset`, and that they *have to* access the latest one to get the desired 'resetting' behaviour.

For example, the below will *never* work from the user's perspective, as `result.reset` will be for the old, already-unsubscribed promise at that point, and won't unsubscribe from the fresh trigger.  The behaviour will appear as though `reset()` was never called at all.
```
  const [trigger, result] = useEditFooMutation();

  async function mutateAndResetAfter1s() {
    trigger("mutateAndResetAfter1s");
    await new Promise((r) => setTimeout(r, 1000));
    result.reset();
  }
```

An option to overcome this could be to use a stable 'reset' function that itself refers to the latest `promise.unsubscribe` via a ref, but I'd like to hear your thoughts on this

phryneas: I get where you're coming from. In that case it might be sensible to move it out of the `result` though to make clear you are not resetting "that result". (and here I had hoped I had gotten rid of all refs)

```
const [trigger, result, {reset}] = useEditFooMutation();
```

Shrugsy: Yeah, I can see it going either way really. What I'm anticipating is that if it's left as-is, even with `reset` being attached to the `result` object, people will overlook that as an 'implementation quirk', as their goal is to 'reset the hook', rather than 'reset a specific result'.

I'm honestly not sure which of the two options I prefer personally. It would be nice to avoid additional elements in the returned tuple, as well as avoiding the ref usage. And the hypothetical above might not ever come up in a real situation anyway.

For the particular situation above, we could also consider having another `reset` attached to the promise returned by `trigger()` itself, and go all-in on 'reset is for a particular result, not the hook in general'

Shrugsy: > For the particular situation above, we could also consider having another reset attached to the promise returned by trigger() itself, and go all-in on 'reset is for a particular result, not the hook in general'

Actually would `triggerReturn.unsubscribe()` already handle that?

phryneas: Yes it would.
We could also rename this `reset` function to `unsubscribe` and just stay with one naming

Shrugsy: Yeah I like the sound of that, it should prevent any confusion

phryneas: Hmm. On second thought, that could also add more confusion in combination with #1477.

My idea is there that you can have multiple components using the same mutation result by manually specifying a cache key:

```
const [trigger, result] = useEditFooMutation({ fixedCacheKey: "fooBar" });
```

Now, `result.unsubscribe()` would be a weird way of saying "reset that mutation result for all of those components".

Maybe instead I should rename `unsubscribe` to `reset` for mutations in general?

Shrugsy: Ah, that's a good point.

> Maybe instead I should rename unsubscribe to reset for mutations in general?

Yeah, at the moment that sounds like the most consistent outcome.

phryneas: Then let's think about wording here for a minute. 

* ~~unsubscribe~~
* reset
* removeFromCache
* remove
* ...more ideas?

Shrugsy: The only others I can currently come up with:  
- restore
- restoreCache

But I prefer just `reset` out of all of those options.

markerikson: `nukeItAll`?  `terminateWithExtremePrejudice`? :)

taejs:
Is there no need to update query results using same tags?

phryneas:
@taejs after just clearing a mutation result? That does not mean the mutation has run again or anything was sent to the server, it just means the component does not want to display the result anymore

LhonRafaat:
when is this gonna be available?

phryneas:
@LhonRafaat next minor release. Could still be a month or two, given what else we want in there. In the meantime you can either use the CodeSandbox build linked above or do something like


```js

const lastMutation = useRef(undefined)
const [trigger] = useMyMutation()


// when triggering:
lastMutation.current = trigger(arg);


// to reset
lastMuation.current.unsubscribe()
```

Shrugsy:
@phryneas the spelling & wording threads are still open, but besides that, LGTM!

docs/rtk-query/api/created-api/hooks.mdx
phryneas: With the current implementation they will get reset.

PR #1477, opened 2021-09-03
allow for "shared component results" using the `useMutation` hook

phryneas:
This is still a bit WIP, but *should* already work


packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -781,35 +782,60 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+          setPromise(undefined)
+        } else if (fixedCacheKey) {
+          dispatch(
+            api.internalActions.removeMutationResult({
+              requestId,
+              fixedCacheKey,
+            })
+          )
+        }
+      }, [dispatch, fixedCacheKey, promise, requestId])
```
Shrugsy, line 838: 💁

I think this logic makes it so that when using a fixed cache key, the component that calls `reset` needs to call it twice before it actually removes the result for that key

e.g.

- Components A & B with a shared mutation key
- Component A triggers the mutation
- Component A stores the promise in state
- Components A & B both receive the mutation result
- Component A calls `reset`
- The promise state in component A is set to undefined, but the mutation result for the cache key is left alone
- Components A & B both still receive the mutation result
- Component A calls `reset` again
- The mutation result for the cache key is now removed

phryneas: Good catch. My thought was: Setting `promise` to `undefined` should trigger the `useEffect` cleanup further up, which also triggers a `reset`.

But unfortunately, that skips in that case:
```js
      useEffect(
        () => () => {
          if (!promise?.arg.fixedCacheKey) {
            promise?.reset()
          }
        },
        [promise]
      )
```

Could you open a PR against the integration branch and maybe also add a test for this to prevent regressions?

Shrugsy: yup, no problem

phryneas: Thank you!

PR #1496, opened 2021-09-10
feat(createAsyncThunk): async condition

### LINKED ISSUE #1494: Async `condition` in `createAsyncThunk`
I need my thunk to detect that another instance of the same operation is already running and to fulfill the promise (without dispatching any actions) only after that other instance is done. The code that dispatches the thunk doesn't care who started the operation, it just needs it to complete. As I have more than one such place in my code, I'd prefer to encapsulate this logic into the thunk, which would be possible if `options.condition` could return a promise. That looks like a simple change. I can give a shot at a PR.

thorn0:
closes #1494

### DIFF
--- a/packages/toolkit/src/createAsyncThunk.ts
+++ b/packages/toolkit/src/createAsyncThunk.ts
@@ -553,12 +553,15 @@ If you want to use the AbortController to react to \`abort\` events, please cons
       const promise = (async function () {
         let finalAction: ReturnType<typeof fulfilled | typeof rejected>
         try {
-          const conditionResult = options?.condition?.(arg, { getState, extra })
-          if (
-            conditionResult === false ||
-            (conditionResult instanceof Promise &&
-              (await conditionResult.catch(() => false)) === false)
-          ) {
+          let conditionResult = options?.condition?.(arg, { getState, extra })
+          if (conditionResult instanceof Promise) {
+            try {
+              conditionResult = await conditionResult
+            } catch {
+              conditionResult = false
+            }
+          }
+          if (conditionResult === false) {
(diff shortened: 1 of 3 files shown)


packages/toolkit/src/createAsyncThunk.ts
```diff
@@ -553,11 +553,15 @@ If you want to use the AbortController to react to \`abort\` events, please cons
       const promise = (async function () {
         let finalAction: ReturnType<typeof fulfilled | typeof rejected>
         try {
-          if (
-            options &&
-            options.condition &&
-            options.condition(arg, { getState, extra }) === false
-          ) {
+          let conditionResult = options?.condition?.(arg, { getState, extra })
+          if (conditionResult instanceof Promise) {
```
thorn0, line 557: Is `instanceof` good enough or do we want to support 3rd-party thenables?

thorn0: Also what should happen when an error is thrown in a synchronous `condition`?

phryneas: Hmm. `"then" in conditionResult" might be better, you're right.

As for an error in `condition`, so far that was just thrown as it was and I'm pretty okay. Might even make sense to remove the try..catch block above to mimick that.

thorn0: > As for an error in condition, so far that was just thrown as it was and I'm pretty okay. Might even make sense to remove the try..catch block above to mimick that.

Okay, let's do it that way. Bur I somewhat dislike the fact that in this case `rejected` is dispatched without corresponding `pending`.

phryneas (approved):

PR #1497, opened 2021-09-10
Use `builder` in example rather than `build`

mungojam:
The text above says it is `builder` as do the examples elsewhere

### DIFF
--- a/docs/rtk-query/usage/mutations.mdx
+++ b/docs/rtk-query/usage/mutations.mdx
@@ -40,8 +40,8 @@ const api = createApi({
     baseUrl: '/',
   }),
   tagTypes: ['Post'],
-  endpoints: (build) => ({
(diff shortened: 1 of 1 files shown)


phryneas:
We use `builder` in `extraReducers` of `createSlice`, but `build` in `createApi` for `endpoints`.

So it should probably be fixed the other way round - in the text.

mungojam:
> We use `builder` in `extraReducers` of `createSlice`, but `build` in `createApi` for `endpoints`.
> 
> So it should probably be fixed the other way round - in the text.

It's `builder` [here in the quickstart](https://redux-toolkit.js.org/tutorials/rtk-query#create-an-api-service) for creating the endpoints. So my code had `builder` as I started with that quick-start.

phryneas:
Yeah, then we probably need to fix it over there.

Issue #1339, opened 2021-07-22
RTKQ: Data stays as old value and loading immediately jumps back to false when fetching fresh data from different endpoint + query param combination


agusterodin:
Hey, I am experiencing an issue when using the useQuery auto-generated hook. If I change query parameters, the data remains the same (I expect it to be undefined so I can show loading placeholder) and loading jumps from "true" immediately back to "false" despite the data for that given combination not being loaded yet. I have keepUnusedDataFor set to 0 btw (not sure if relevant or not).

The docs claim: 

> isLoading refers to a query being in flight for the first time for the given endpoint + query param combination. No data will be available at this time.

This is really hard to reproduce in a Codepen since it relies on the server having some sort of delay. I wasn't aware of this bug until I used an endpoint that is extremely slow (4+ sec response time).

Hopefully this attached video, code snippet, and console log demonstrate the issue effectively. Please let me know if there are any other details I can provide!!

Notice that despite the argument changing from "monthly" to "weekly" it is still providing the "monthly" data instead of data being undefined while still loading.

**Component**
```
import { useSelector } from 'react-redux'
import MetricGraphs from './MetricGraphs'
import MetricStats from './MetricStats'
import { getOverviewMetricTimeframe } from 'state/campaigns/overview/overview'
import { useGetAggregateCampaignMetricsQuery } from 'api/campaign'
import MetricTimeframeSelector from './MetricTimeframeSelector'

export default function Metrics() {
  const currentlySelectedTimeframe = useSelector(getOverviewMetricTimeframe)
  const { data, isLoading, originalArgs } = useGetAggregateCampaignMetricsQuery(currentlySelectedTimeframe)
  console.table({ currentlySelectedTimeframe, originalArgs, isLoading, data })

  return (
    <>
      <div className="relative">
        <MetricTimeframeSelector />
      </div>
      <div className="relative grid grid-cols-1 gap-5 sm:grid-cols-2 lg:grid-cols-4 mt-5">
        <MetricStats metricResults={data} />
        <MetricGraphs metricResults={data} />
      </div>
    </>
  )
}
```

**Recreation In Browser**
https://user-images.githubusercontent.com/10248395/126681813-9656f46e-46a8-4a2e-b316-9229291699b5.mov

aaronwolmutt:
Have you tried using isFetching for this? 

See: https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#usequery
```
[...]

PR #1500, opened 2021-09-12
add `currentData` property to hook results.

phryneas:
This would be one possible solution for #1339.

Feedback very welcome.

### TODO:
- [ ] Tests
- [ ] Docs

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -307,6 +307,12 @@ export type UseQueryStateResult<
 type UseQueryStateBaseResult<D extends QueryDefinition<any, any, any, any>> =
   QuerySubState<D> & {
     /**
+     * Where `data` tries to hold data as much as possible, also re-using
+     * data from the last arguments passed into the hook, this property
+     * will always contain the received data from the query, for the current query arguments.
+     */
+    currentData?: ResultTypeFrom<D>
+    /**
      * Query has not started yet.
      */
     isUninitialized: false
@@ -342,10 +348,20 @@ type UseQueryStateDefaultResult<D extends QueryDefinition<any, any, any, any>> =
         | { isLoading: true; isFetching: boolean; data: undefined }
         | ({
             isSuccess: true
-            isFetching: boolean
+            isFetching: true
             error: undefined
           } & Required<
             Pick<UseQueryStateBaseResult<D>, 'data' | 'fulfilledTimeStamp'>
+          >)
+        | ({
+            isSuccess: true
+            isFetching: false
+            error: undefined
+          } & Required<
+            Pick<
+              UseQueryStateBaseResult<D>,
+              'data' | 'fulfilledTimeStamp' | 'currentData'
+            >
           >)
         | ({ isError: true } & Required<
             Pick<UseQueryStateBaseResult<D>, 'error'>
@@ -420,6 +436,7 @@ const queryStatePreSelector = (
   return {
     ...currentState,
     data,
+    currentData: currentState.data,
     isFetching,
     isLoading,
     isSuccess,
--- a/packages/toolkit/src/query/tests/unionTypes.test.ts
+++ b/packages/toolkit/src/query/tests/unionTypes.test.ts
@@ -115,6 +115,20 @@ describe.skip('TS only tests', () => {
       expectExactType(false as false)(result.isError)
     }
 
+    expectExactType('' as string | undefined)(result.currentData)
+    // @ts-expect-error
+    expectExactType('' as string)(result.currentData)
+
+    if (result.isSuccess) {
+      if (!result.isFetching) {
+        expectExactType('' as string)(result.currentData)
+      } else {
+        expectExactType('' as string | undefined)(result.currentData)
+        // @ts-expect-error
+        expectExactType('' as string)(result.currentData)
+      }
+    }
+
     // @ts-expect-error
     expectType<never>(result)
     // is always one of those four


packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -420,6 +436,7 @@ const queryStatePreSelector = (
   return {
     ...currentState,
     data,
+    currentData: currentState.data,
```
HansBrende, line 439: Correct me if I am wrong, but wouldn't this be basically the equivalent of doing:
```typescript
const currentData = isFetching ? data : undefined;
```
in our own code? I would think you'd need to examine the `args` associated with `lastResult` to see if they are shallowly equal to the args of `currentState`, something like:

```
currentData = currentState.data ?? (lastResult?.arg === currentState.arg ? lastResult?.data : undefined)
```
if the comment 
> this property will always contain the received data from the query, for the *current query arguments*

is to be true. (i.e., distinguishing cases of polling from cases of arg changes).

But, correct me if I am wrong here. I am not too familiar with the internal workings of this codebase.

agusterodin:
I wish I had something more insightful to add (based on my limited knowledge of the inner workings of how this library works) but for what it is worth: 

currentData behaves exactly as advertised in the comment that shows up in intellisense. It solves the issue I was personally having. Just tried it out in my code and can confirm.

In general I really like this solution compared to other approaches like adding isLoadingFirstTimeForArg or having a keepPreviousData option in the hook (like React-Query does).

packages/toolkit/src/query/react/buildHooks.ts
phryneas: `currentState` is always for the currently selected endpoint and `lastResult` is the last result the hook had cached. So `lastResult` can also be from another `arg`, but `currentState` never can.

HansBrende: @phryneas Right... but my question is: this solution doesn't seem to distinguish between polling and arg changes. 
The behavior I'd expect for `currentData` is that if the args haven't changed (e.g., because of polling), then it would use the last result for those args (which might be cached in lastResult in that case).

HansBrende:
@agusterodin out of curiosity, did you also try it out for polling situations? I would expect that `currentData` doesn't get reset back to undefined while a polling operation is taking place (since the args haven't changed).

phryneas:
> @agusterodin out of curiosity, did you also try it out for polling situations? I would expect that `currentData` doesn't get reset back to undefined while a polling operation is taking place (since the args haven't changed).

It doesn't get reset, because also the cache entry does not disappear. But you now have the direct distinction: 
`isSuccess && !isFetching` => fresh data in `currentData`. `currentData` available otherwise. Might just be from an older request for these args.

packages/toolkit/src/query/react/buildHooks.ts
phryneas: Data once present for the current args will not disappear from the cache while polling, and this essentially gives you data for the current args directly from the cache. 
Once data is set in the cache, it will always be held, even while loading or an error occurs later. 
See the slice definition here: https://github.com/reduxjs/redux-toolkit/blob/261410a91eb095a7611550a464117d3993d14b44/packages/toolkit/src/query/core/buildSlice.ts#L121-L169

HansBrende:
@phryneas great! If `currentData` _doesn't_ get reset back to undefined if the args _don't_ change, then that's exactly what I'm looking for.

One other question: I'd expect that `currentData` gets set to undefined whenever the `skipToken` is passed in as the current args... will that be the case in this implementation?

agusterodin:
Just went back and checked, can confirm that currentData remains stable even when polling is enabled (which is desirable behavior) 👍

phryneas:
> One other question: I'd expect that `currentData` gets set to undefined whenever the `skipToken` is passed in as the current args... will that be the case in this implementation?

From the back of my head, yes - but please validate.

msutkowski (approved):
This makes sense to me - @Shrugsy perhaps we can add meaningful examples to the docs based on the use cases from https://github.com/reduxjs/redux-toolkit/issues/1339#issuecomment-917588375 once/if this lands?

HansBrende:
@phryneas I've just now installed and tested it: `currentData` _does_ maintain the expected behavior with `skipToken`. Nice work! 🎉 

Only other recommendation I can think of: I expect most devs will end up migrating from `data` to `currentData` as soon as they discover the potential integrity problems with `data` (which are not immediately obvious until you realize that your UI is behaving weirdly and have to dig down to the root cause). In fact, I can see `currentData` as becoming sort of the default recommended approach to take, whereas "data" would then only be used in edge cases such as paginated queries, etc. where you know that you really _do_ want that behavior. That being the case, it might be prudent to name it something slightly more succinct and "default-sounding", such as, simply: `result`.

All-in-all, great work on this! It definitely fixes the UI bugs I've been having. Hope it can be released soon :)

phryneas:
> Only other recommendation I can think of: I expect most devs will end up migrating from `data` to `currentData` as soon as they discover the potential integrity problems with `data` (which are not immediately obvious until you realize that your UI is behaving weirdly and have to dig down to the root cause). In fact, I can see `currentData` as becoming sort of the default recommended approach to take, whereas "data" would then only be used in edge cases such as paginated queries, etc. where you know that you really _do_ want that behavior. That being the case, it might be prudent to name it something slightly more succinct and "default-sounding", such as, simply: `result`.

Honestly, I think it is niche and will probably also stay niche - the current behaviour is not by accident, but very much intended. I can see where it makes sense to use `currentData` in polling situations, but for most simpler applications I think `data` more closely models what people actually want - at least until we get suspense for data fetching and then we'll probably have to ship a second set of hooks or something and that whole question becomes irrelevant.

HansBrende:
@phryneas sounds good; I disagree with the "niche" part, but whatever you name it, I will use it! Again, nice work on this. Will be looking forward to the next release 😄

Shrugsy:
> This makes sense to me - @Shrugsy perhaps we can add meaningful examples to the docs based on the use cases from [#1339 (comment)](https://github.com/reduxjs/redux-toolkit/issues/1339#issuecomment-917588375) once/if this lands?

yep, can do

saschametz:
The `currentData` addition is exactly what we need for our migration from react-query 👍

In our case we need the `currentData` property most of the time and the `data` property only for some parts of our application. The only thing that was confusing at first was that the `isLoading` property is not set to `false` for subsequent queries. To get the desired loading behavior, we simply use `currentData === undefined`.

---

For example, we want to use `currentData` for a paginated query and display a loading indicator if the current data has not yet been fetched:

Loading the first page:

1. `data = undefined`, `currentData = undefined` => `isLoading = true`, `isFetching = true`
2. `data = ["foobar 1"]`, `currentData = ["foobar 1"]` => `isLoading = false`, `isFetching = false`

Loading the second page:

3. `data = ["foobar 1"]`, `currentData = undefined` => `isLoading = false`, `isFetching = true`
4. `data = ["foobar 2"]`, `currentData = ["foobar 2"]` => `isLoading = false`, `isFetching = false`

Note: The other option `isFetching` wouldn't solve this problem, because we don't want to display a loading indicator when there is already data in the cache.

react-query doesn't has this "problem" that `isLoading` is `false` in step 3 because they only have one `data` prop. Whether the prop `data` contains previous data or not depends on an additional option [`keepPreviousData: true`](https://react-query.tanstack.com/guides/paginated-queries#better-paginated-queries-with-keeppreviousdata).

---

**Example:**
https://codesandbox.io/s/examples-query-react-basic-forked-wiezj

Example 2.):
While the first loading indicator is displayed, the loading indicator for the second page is missing due to the described behavior.

PR #1340, opened 2021-07-22
Adjust `isLoading`/`isFetching` explanation

phryneas:
Triggered by https://github.com/reduxjs/redux-toolkit/issues/1339

@Shrugsy can you give this a read-over if I'm making sense?

PR #1520, opened 2021-09-17
fix "isLoading briefly flips back to `true`" #1519

### LINKED ISSUE #1519: isLoading briefly flips back to `true`
> Also:  is only for the first fetch of a hook. If you are changing query args, it will never go back to true again. I'm not entirely sure what behaviour you would want there.

@phryneas hi, you say  will never go back to true. But when I test the pagination example，when I set page,  will become true  first, and then become false. Is this expected?

demo link:
https://codesandbox.io/s/admiring-wiles-6rm5t?file=/src/features/posts/PostsManager.tsx

Screen Shot 2021-09-17 at 14 18 59

_Originally posted by @zousandian in https://github.com/reduxjs/redux-toolkit/issues/1339#issuecomment-921533014_

phryneas:
This fixes #1519.

While working on it, I noticed one edge case: if the query result is `undefined`, this flipping will still happen due to underlying assumptions we have.

It might make sense to have the cache convert `undefined` to `null`, but since that change would go further, I am not including it in this one. It still fixes an important bug as is.

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -406,16 +406,17 @@ const queryStatePreSelector = (
   lastResult: UseQueryStateDefaultResult<any>
 ): UseQueryStateDefaultResult<any> => {
   // data is the last known good request result we have tracked - or if none has been tracked yet the last good result for the current args
-  const data =
-    (currentState.isSuccess ? currentState.data : lastResult?.data) ??
-    currentState.data
+  let data = currentState.isSuccess ? currentState.data : lastResult?.data
+  if (data === undefined) data = currentState.data
+
+  const hasData = data !== undefined
 
   // isFetching = true any time a request is in flight
   const isFetching = currentState.isLoading
   // isLoading = true only when loading while no data is present yet (initial load with no data in the cache)
-  const isLoading = !data && isFetching
+  const isLoading = !hasData && isFetching
   // isSuccess = true when data is present
-  const isSuccess = currentState.isSuccess || (isFetching && !!data)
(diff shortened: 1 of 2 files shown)


phryneas:
It seems we had even tests that tested for the old behaviour, that `isLoading` briefly jumped back to `true` on `refetch`. 
@Shrugsy do you know if we have documented something like that anywhere? I'm pretty sure that was never the intended use.

Shrugsy:
> It seems we had even tests that tested for the old behaviour, that isLoading briefly jumped back to true on refetch.
@Shrugsy do you know if we have documented something like that anywhere? I'm pretty sure that was never the intended use.

@phryneas AFAIK the closest thing to documenting that was the old wording that was fixed under #1340. But for this specific case, with a brief glance over the docs with they keywords `isLoading` and `true` independently, I can't find anything documenting that behaviour, so it should be all good on that end.

Shrugsy (approved):
Behaviour looks more consistent, and new test logic looks solid. LGTM!

phryneas:
> Edit: looking at it closer, I think the old wording before #1340 did describe the behaviour prior to the changes on this branch. i.e. it considered 'isLoading' to be tied to the query + args combo having a cache entry

Just to be clear for anyone reading this in the future: that wasn't really how it was actually behaving - it just turned `isLoading` on for the brief moment of `isUninitialized` and then disabled it again for the actual `isLoading` part. So this is really just fixing an overlooked timing inconsistency.

msutkowski (approved):
:trophy:. I can't think of any thing else to handle here as your new test simplifies the intent of the test that was updated :tada:

Issue #1526, opened 2021-09-19
RTK Query - Infinite loop when arg is nested object


lukascivil:
#### The problem
![image](https://user-images.githubusercontent.com/7409802/133936668-e03e9595-0b28-48d4-8021-84061a6151b2.png)

The problem **infinite loop with multiple calls** occurs when filter is an object inside another object, which assembles arg. 

```tsx
  const [filter, setFilter] = useState({email: "cafe@gmail.com"})
```

If filter is a direct value, as string or number, or the object with the same key, the problem does not occur, example:

```tsx
  const [filter, setFilter] = useState("cafe@gmail.com")
```

The "useGetUsersQuery" (example below) hook will make multiple calls as if it were in loop.

I don't know if this is expected behavior with wrong use or a possible bug, however I would like to understand the hook behavior based on **prev Arg** and **next Arg**, how does Hook handle this?

#### Full example with infinite loop

```tsx
export const TabLocalList: FC = () => {
  const [filter, setFilter] = useState({ email: '' })
  const currentSort = { field: 'created_at', order: 'ASC' }
  const { isFetching, data, refetch } = useGetUsersQuery({
    filter,
    pagination: { page: 1, perPage: 50 },
    sort: { field: 'created_at', order: 'ASC' }
  })

  return (
    <LocalList
      isLoading={isFetching}
      data={data?.data}
      onRefresh={refetch}
      onSubmit={values => {
        setFilter(values as any) // {created_at_start: "..."}
      }}
      sort={currentSort}
      filters={[<TextInput source="created_at_start" validate={required()} />]}
    >
      <Datagrid>
        <TextField source="id" label="Id" />
        <TextField source="email" label="E-mail" />
      </Datagrid>
    </LocalList>
  )
}
```

#### I did some examples and got different results
```tsx
      onSubmit={values => {
        // Works ok
        // setFilter({ email: values?.email } as any)

        // Works ok (If values were {email: "..."}) , same key with diferent value
        // setFilter(values as any)

        // Works ok
        // setFilter({ email: '55' } as any) // OK

        // Crash with infinite loop, add a new prop
        // setFilter({ email: '55', created_at_start: '' } as any)

        // Crash with infinite loop, remove email and use new key for diferent filter
        // setFilter({ created_at_start: '' } as any)
      }}
```

From the above results, I believe the internal check to see if the args have changed may be causing this issue.

lukascivil:
[...]

PR #1531, opened 2021-09-20
[RFC] 🥅 Improve feedback for infinite query loops

Shrugsy:
- Throw an error when query args are detected as changing every
render too frequently, after factoring shallow stabilisation attempts
- Expand docs explanation about stable query arg requirements

The topic of infinite re-renders with unstable query args comes up every now and then (e.g. #1526). This aims to provide more explicit, immediate feedback about the cause, and point to the relevant part of the docs as a solution.

A few things to consider:
- Is this safe from false positives?
  > I think it should be safe. The criteria to throw an error is "value changes 10 times in 10 consecutive renders within 1000ms, *after* attempting stabilization". I don't think that criteria would ever intersect with intentional behaviour.
- Should it actually be throwing an error, or only logging an error message?
  > I think throwing an error is better here. If it's meeting that criteria, I think it's a safe bet to assume that it will either freeze your browser tab, continue looping indefinitely, or throw react's internal infinite loop error ("Uncaught Error: Too many re-renders. React limits the number of renders to prevent an infinite loop."). I believe that throwing our error earlier is more useful than just logging alongside any of those three options, and has less chance of being missed.
- Is the additional runtime code overhead actually worth having for what is a relatively niche case?
  > tbh I'm not sure about this

### DIFF
--- a/docs/rtk-query/usage/queries.mdx
+++ b/docs/rtk-query/usage/queries.mdx
@@ -123,7 +123,29 @@ The `queryArg` param will be passed through to the underlying `query` callback t
 
 :::caution
 
-The `queryArg` param is handed to a `useEffect` dependency array internally. RTK Query tries to keep the argument stable by performing a `shallowEquals` diff on the value, however if you pass a deeper nested argument, you will need to keep the param stable yourself, e.g. with `useMemo`.
+The `queryArg` param is handed to a `useEffect` dependency array internally.
+RTK Query tries to keep the argument stable by performing a `shallowEquals` diff on the value,
+however if you pass a deeper nested argument, you will need to keep the param stable yourself,
+e.g. with `useMemo`.
+
+```diff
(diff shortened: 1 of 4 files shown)


Shrugsy:
CSB examples are here to demo the new behaviour. For the setup: a component is added which uses a nested argument for a query which always returns an error. The component will be rendered once you click the 'Toggle error query' button.

## v1.6.1 behaviour example
Note: The tab might freeze or you might observe it looping indefinitely in the console.  
https://codesandbox.io/s/examples-query-react-basic-with-nested-arg-error-forked-08g8n?file=/src/App.tsx

## Branch example
Rather than freezing or looping indefinitely, a meaningful error message is instead shown.
https://codesandbox.io/s/examples-query-react-basic-with-nested-arg-error-kxl7b?file=/src/App.tsx

Shrugsy:
Closed due to #1533

PR #1541, opened 2021-09-22
fix(useLazyQuery): added docs for preferCache option

akashshyamdev:
Hey,

This PR is based on the discussion we had [here] (https://github.com/reduxjs/redux-toolkit/discussions/1279). I've updated the documentation, but still can't find the type which returns the `trigger` function. I saw a definition of a `trigger()` method that matched your description but it was part of the `useLazyQuerySubscription` and not `useLazyQuery`.

Looking forward to your reply! Thanks.


phryneas:
This should turn up [here](https://deploy-preview-1541--redux-starter-kit-docs.netlify.app/rtk-query/api/created-api/hooks#uselazyquery) but I fear that our build caching is somehow interacting with that. 
@markerikson it might be necessary that we disable that build caching again :(

As for the type of the trigger method, that would be this line I think:
https://github.com/reduxjs/redux-toolkit/blob/28e201c9fdbbe476b183609c2e02d46649758a24/packages/toolkit/src/query/react/buildHooks.ts#L188

PR #1552, opened 2021-09-27
Update createSlice.mdx

tylercaceres:
counter.actions.decrement returns an action creator.
However, counter.actions.decrement.type returns "counter/decrement" as stated.


markerikson:
I _think_ that the current example _should_ work as-is.  RTK action creators override the `.toString()` implementation on the action creator function, and I'm pretty sure that the template literal usage here should try to stringify the value.  Can you double-check the behavior?

tylercaceres:
> I _think_ that the current example _should_ work as-is. RTK action creators override the `.toString()` implementation on the action creator function, and I'm pretty sure that the template literal usage here should try to stringify the value. Can you double-check the behavior?

Ah you are right, sorry. I was not using string literals in my example.

PR #1570, opened 2021-10-02
fix mistake in docs "automated-refetching.mdx"

3b7ameed:


markerikson:
I _think_ this may be correct as-is, actually.

By using a `queryFn`, and having it return null, it looks like it has the net effect of simply invalidating some tags with no other effect.

phryneas:
Yes, that is intended like this. It is a mutation that is not initiating a trip to the server, by using queryFn to bypass baseQuery. 

Am 3. Oktober 2021 01:36:20 MESZ schrieb Mark Erikson ***@***.***>:
>I _think_ this may be correct as-is, actually.
>
>By using a `queryFn`, and having it return null, it looks like it has
>the net effect of simply invalidating some tags with no other effect.

-- 
Sent from my Android phone with K-9 Mail. Please excuse my brevity.

3b7ameed:
Ah sorry guys I didn't know that

markerikson:
No worries, we appreciate the effort!

Issue #1469, opened 2021-09-02
[Query] Options for multi value parameters?


colemars:
Hello 👋 
      
Currently, `fetchBaseQuery` utilizes `UrlSearchParams` for preparing its query params. `UrlSearchParams` will parse multi value params (e.g. `param=[1,2,3]`) as a comma separated list, `...&param=1,2,3`.

However,  it's not uncommon when working with various 3rd parties to encounter different requirements. 

Most notably, AWS API Gateway [looks for repeating keys](https://aws.amazon.com/blogs/compute/support-for-multi-value-parameters-in-amazon-api-gateway/) (`...&param=1&param=2&param=3`).

Would exposing some sort of config here to support other multi value parameter strategies be valid?

as a minimal, hasty example:

```diff
diff --git a/node_modules/@reduxjs/toolkit/src/query/fetchBaseQuery.ts b/node_modules/@reduxjs/toolkit/src/query/fetchBaseQuery.ts
index 49605fc..b9222a6 100644
--- a/node_modules/@reduxjs/toolkit/src/query/fetchBaseQuery.ts
+++ b/node_modules/@reduxjs/toolkit/src/query/fetchBaseQuery.ts
@@ -22,6 +22,7 @@ type CustomRequestInit = Override<
 export interface FetchArgs extends CustomRequestInit {
   url: string
   params?: Record<string, any>
+  paramsOpts?: { arrayFormat: 'default' | 'repeat' }
   body?: any
   responseHandler?: ResponseHandler
   validateStatus?: (response: Response, body: any) => boolean
@@ -182,6 +183,7 @@ export function fetchBaseQuery({
       headers = new Headers({}),
       body = undefined,
       params = undefined,
+      paramsOpts = undefined,
       responseHandler = 'json' as const,
       validateStatus = defaultValidateStatus,
       ...rest
@@ -216,7 +218,22 @@ export function fetchBaseQuery({
 
     if (params) {
       const divider = ~url.indexOf('?') ? '&' : '?'
-      const query = new URLSearchParams(stripUndefined(params))
+      let query = new URLSearchParams(stripUndefined(params))
+
+      if ('arrayFormat' in paramsOpts) {
+        switch (paramsOpts.arrayFormat) {
+          case 'repeat':
+            let temp = new URLSearchParams();
+            Object.entries(params).forEach(([k, v]) => {
+              if (Array.isArray(v)) {
+                v.forEach(o => temp.append(k, o))
[...]

PR #1594, opened 2021-10-08
Support a custom paramsSerializer on fetchBaseQuery

msutkowski:
I missed this when we originally built `fetchBaseQuery`, but ran into this yesterday at work. A lot of ruby/elixir/whatever APIs natively handle a bracket params format, so in order to support this, we'd need to allow for a custom paramsSerializer as proposed.


msutkowski:
> seems very similar in spirit to the proposal in this issue: #1469

I didn't see that thread - I'll drop a comment there.

Shrugsy:
LGTM! Maybe this should go into the `v1.7.0-integration` branch though?

Shrugsy (approved):
LGTM! Maybe this should go into the v1.7.0-integration branch though?

phryneas (approved):

PR #1616, opened 2021-10-16
🐛 Fix bug with `useMutation` shared results

Shrugsy:
- Fix a scenario where the component calling `reset` needed to call it twice when using shared component results

Addresses the scenario described here: https://github.com/reduxjs/redux-toolkit/pull/1477#discussion_r730278289

A regression test is included.

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -827,7 +827,8 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
       const reset = useCallback(() => {
         if (promise) {
           setPromise(undefined)
-        } else if (fixedCacheKey) {
+        }
+        if (fixedCacheKey) {
           dispatch(
             api.internalActions.removeMutationResult({
               requestId,
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -825,17 +825,19 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
       const originalArgs =
         fixedCacheKey == null ? promise?.arg.originalArgs : undefined
       const reset = useCallback(() => {
-        if (promise) {
-          setPromise(undefined)
-        }
-        if (fixedCacheKey) {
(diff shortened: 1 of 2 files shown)


Shrugsy:
To test the demos below, try triggering a mutation, then resetting from the same component (i.e. click trigger, then click the reset button next to it)

## Before
https://codesandbox.io/s/examples-query-react-basic-forked-185pk?file=/src/App.tsx

`reset` needs to be clicked twice to reset if resetting from the component that called the trigger

## After
https://codesandbox.io/s/examples-query-react-basic-forked-i9vlk?file=/src/App.tsx

`reset` correctly resets immediately when resetting from the component that called the trigger

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -827,7 +827,8 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
       const reset = useCallback(() => {
         if (promise) {
           setPromise(undefined)
-        } else if (fixedCacheKey) {
+        }
+        if (fixedCacheKey) {
           dispatch(
             api.internalActions.removeMutationResult({
               requestId,
```
phryneas, line 834: Hmm. Do we want to wrap the whole thing in a `batch` now? :thinking:

Shrugsy: In it's current state without `batch`, calling `reset()` only seems to result in one render for each component using the hook, including the one that called the original `trigger`. Is there some other benefit that `batch` will provide?

Regardless of the above, I'm not against batching it anyway as a "might help, and can't hurt" scenario

phryneas: It depends how you call it. If you call it in an effect or click handler, React will automatically batch and only rerender once. If you do it in something async, for example after a timeout, it would lead to two renders

Shrugsy: ah, for some reason I assumed click handlers were exempt from the default batching. It's definitely worth batching this, I'll change that

phryneas:
LGTM :)

Issue #1658, opened 2021-10-27
UpdateQueryData function callback isn't called


maisterr:
I'm trying to implement an optimistic update function by manually update the cache, here is my code which has some differences from the example in the documentation. First console log inside onQueryStarted works and gets the data, but the callback inside updateQueryData does not work and console.log(patchCollection) return object with patches zero-length array

What is wrong with this and what do I need to pass in the args argument of the updateQueryData function it this case?

```
export const postsApi = createApi({
    reducerPath: 'posts',
    baseQuery: fetchBaseQuery({ baseUrl: BASE_URL }),
    endpoints: (build) => ({
        getPosts: build.query({
            query: () => '/posts',
        }),
        addPost: build.mutation({
            query(body) {
                return {
                    url: `/posts`,
                    method: 'POST',
                    body,
                }
            },
            onQueryStarted(post, { dispatch, queryFulfilled }) {
                console.log('works')
                const patchCollection = dispatch(
                    postsApi.util.updateQueryData('getPosts', undefined, (cachedPosts) => {
                        console.log("won't work")
                        cachedPosts.push(post)
                    }),
                )
                console.log(patchCollection)
                queryFulfilled.catch(patchCollection.undo)
            },
        }),
    }),
})
```

tkharuk:
struggled as well.
In my case, I expect any argument in query builder, but mistakenly passed one arg to hook, thus `updateQueryData` never triggered.

I've read `a cache key, used to determine which cached dataset needs to be updated` multiple times, but it didn't help me much :(

phryneas:
@Shrugsy could you maybe take a look at the docs there? This seems to confuse a lot of people.

Huespal:
I have a doubt related.
How can one know the exact params used in the previous query while on the query trying to apply the updateQueryData?
So, for example, a GET /something exists and is performed with params { number: 1 }.
User continues to navigate and 'number' parameter changes from 1 to 3, to 25, etc.
Lately a POST /something is called with an optimistic update (which uses updateQueryData) over the GET. 
How do I know here, in the PUT, that page is still 1 and not 3, or 25?
And this example is with only one parameter, so it can be a lot more complex. 
I think it will be a lot easier if the params or arguments where skipped to allow the update on data only with the 'getSomething' endpoint name. Now, it complicates things a lot.

Thanks

phryneas:
[...]

PR #1679, opened 2021-11-02
📝 Clarify `updateQueryData` usage

Shrugsy:
Hopefully clarifies a bit more about how to use `updateQueryData` to prevent confusion such as in #1658


phryneas:
This is great, thank you!

PR #1680, opened 2021-11-02
Move RTK-Query OpenAPI Codegen into Monorepo, prepare release 1.0

phryneas:
This pretty much completely overhauls the way the Codegen is used, as suggested in https://github.com/rtk-incubator/rtk-query-codegen/issues/63

Could some people please give this a try and report back?

The new README is here: https://github.com/reduxjs/redux-toolkit/blob/rtk-query-codegen-openapi/packages/rtk-query-codegen-openapi/README.md

Please note that right now you will have to use the CI build, so call
```bash
npx @rtk-query/codegen-openapi@https://pkg.csb.dev/reduxjs/redux-toolkit/commit/4b8f6b4f/@rtk-query/codegen-openapi/_pkg.tgz <configfile>
```
Where `4b8f6b4f` should reference the latest commit.


markerikson:
Side note: we're going to have to figure out a better monorepo release system. I've been using https://github.com/sindresorhus/np , but it wants to tag and push every time you do a release. That doesn't work for monorepos with multiple packages.

phryneas:
Gotta figure out why the CI is acting weirdly tomorrow. That's it for today.

packages/rtk-query-codegen-openapi/test/generateEndpoints.test.ts
```diff
@@ -0,0 +1,179 @@
...
+describe('yaml parsing', () => {
+  it.skip('should parse a yaml schema from a URL', async () => {
+    // TODO
+    const result = await generateEndpoints({
+      apiFile: './tmp/emptyApi.ts',
+      schemaFile: `https://petstore3.swagger.io/api/v3/openapi.yaml`,
+      hooks: true,
+    });
+    expect(result).toMatchSnapshot();
+  });
```
phryneas, line 141: @msutkowski could you look into this one? That really seems to be broken, but I have no idea why - you added that back then. 

It works with `.json` o_O

grumpyTofu:
@phryneas
I tested and here is what I am running into:

config file:

```ts
import { ConfigFile } from "@rtk-query/codegen-openapi");

const {
  parsed: { OPENAPI_DOCS },
} = require("dotenv").config({ path: "./.env.local" });

const config = {
  schemaFile: OPENAPI_DOCS,
  apiFile: "./src/app/emptyApi.ts",
  //apiImport: "emptyApi",
  outputFile: "./src/app/rtk-api.ts",
  //exportName: "emptyApi",
  hooks: true,
};

export default config;
```

---

it might be due to how I am running this? However, it doesn't seem to like my config file.
```bash
$ node ./node_modules/@rtk-query/codegen-openapi/lib/bin/cli.js openapi.dev.config.ts
(node:27794) UnhandledPromiseRejectionWarning: /Users/austinfelix/apps/finsense-ui/openapi.dev.config.ts:10
export default config;
```

However, when I change my config to the following and run it with ts-node it works fine:

```ts
const { generateEndpoints } = require("@rtk-query/codegen-openapi");

const {
  parsed: { OPENAPI_DOCS },
} = require("dotenv").config({ path: "./.env.local" });

const config = {
  schemaFile: OPENAPI_DOCS,
  apiFile: "./src/app/emptyApi.ts",
  //apiImport: "emptyApi",
  outputFile: "./src/app/rtk-api.ts",
  //exportName: "emptyApi",
  hooks: true,
};

generateEndpoints(config);
```
---

I also noticed the export name is also behaving inconsistently. 

```ts
const injectedRtkApi = api.injectEndpoints({
  endpoints: (build) => ({
    // endpoints
  }),
  overrideExisting: false,
});
export { injectedRtkApi as api };

 // types

export const {
 // queries & mutations
} = enhancedApi;

```

- regardless of the exportName property, the api variable name is always injectedRtkApi. It needs to take on the name of the exportName property or enhancedApi by default to match the query and mutation exports at the end of the file

Also, this line always stays the same regardless:
```ts
export { injectedRtkApi as api };
```
would it be better to just export from the declaration instead? that way the export always matches the exportName property from config and the final file looks like:

```ts
export const enhancedApi = api.injectEndpoints({
  endpoints: (build) => ({
    // endpoints
  }),
  overrideExisting: false,
});

 // types

export const {
 // queries & mutations
} = enhancedApi;
```

---

Finally, one small thing that I noticed is that the name of the library is missing an "e" in the package.json

```json
  "name": "@rtk-query/codgen-openapi"
```

phryneas:
> it might be due to how I am running this? However, it doesn't seem to like my config file.
> 
> ```shell
> $ node ./node_modules/@rtk-query/codegen-openapi/lib/bin/cli.js openapi.dev.config.ts
> (node:27794) UnhandledPromiseRejectionWarning: /Users/austinfelix/apps/finsense-ui/openapi.dev.config.ts:10
> export default config;
> ```

This is very weird. But I have to admit that I only tested it with `esbuild-runner` installed, not with `ts-node` installed, so I should maybe test that too.

This is really all the error message you get?

grumpyTofu:
@phryneas 
No, there is a bunch of extra stacktrace after. Here is the whole thing:

```bash
$ node ./node_modules/@rtk-query/codegen-openapi/lib/bin/cli.js openapi.dev.config.ts
(node:33479) UnhandledPromiseRejectionWarning: /Users/austinfelix/apps/finsense-ui/openapi.dev.config.ts:10
export default config;
^^^^^^

SyntaxError: Unexpected token 'export'
    at wrapSafe (internal/modules/cjs/loader.js:1001:16)
    at Module._compile (internal/modules/cjs/loader.js:1049:27)
    at Module.m._compile (/Users/austinfelix/apps/finsense-ui/node_modules/ts-node/src/index.ts:1365:23)
    at Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Object.require.extensions.<computed> [as .ts] (/Users/austinfelix/apps/finsense-ui/node_modules/ts-node/src/index.ts:1368:12)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:14)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:92:18)
    at run (/Users/austinfelix/apps/finsense-ui/node_modules/@rtk-query/codegen-openapi/src/bin/cli.ts:45:26)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:33479) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:33479) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

---

I did switch to esr and it works perfectly with the original config file. However, it may just be a local development problem. I think most people will use npx as described in the docs once its released. You might see some create a script and run it with ts-node if they want to do something really custom.

phryneas:
At this point I really have no idea any more why those test don't work - they run perfectly fine locally.

If anyone wants to experiment with them, I'd be really grateful.

Lokaltog:
@phryneas proposed fix: #1703

packages/rtk-query-codegen-openapi/src/bin/cli.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (require.resolve('esbuild') && require.resolve('esbuild-runner')) {
+    require('esbuild-runner/register');
+  }
+  ts = true;
+} catch {}
+
+try {
+  if (!ts) {
+    if (require.resolve('typescript') && require.resolve('ts-node')) {
+      require('ts-node/register/transpile-only');
```
phryneas, line 18: > @phryneas No, there is a bunch of extra stacktrace after. Here is the whole thing:
> 
> ```shell
> $ node ./node_modules/@rtk-query/codegen-openapi/lib/bin/cli.js openapi.dev.config.ts
> (node:33479) UnhandledPromiseRejectionWarning: /Users/austinfelix/apps/finsense-ui/openapi.dev.config.ts:10
> export default config;
> ^^^^^^
> 
> SyntaxError: Unexpected token 'export'
>     at wrapSafe (internal/modules/cjs/loader.js:1001:16)
>     at Module._compile (internal/modules/cjs/loader.js:1049:27)
>     at Module.m._compile (/Users/austinfelix/apps/finsense-ui/node_modules/ts-node/src/index.ts:1365:23)
>     at Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
>     at Object.require.extensions.<computed> [as .ts] (/Users/austinfelix/apps/finsense-ui/node_modules/ts-node/src/index.ts:1368:12)
>     at Module.load (internal/modules/cjs/loader.js:950:32)
>     at Function.Module._load (internal/modules/cjs/loader.js:790:14)
>     at Module.require (internal/modules/cjs/loader.js:974:19)
>     at require (internal/modules/cjs/helpers.js:92:18)
>     at run (/Users/austinfelix/apps/finsense-ui/node_modules/@rtk-query/codegen-openapi/src/bin/cli.ts:45:26)
> (Use `node --trace-warnings ...` to show where the warning was created)
> (node:33479) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
> (node:33479) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
> ```
> 
> I did switch to esr and it works perfectly with the original config file. However, it may just be a local development problem. I think most people will use npx as described in the docs once its released. You might see some create a script and run it with ts-node if they want to do something really custom.


So I believe we will have to add some config here so it uses a lower transpilation target/commonJS output. I'll have to experiment with that in the next few days.

Lokaltog:
@phryneas Happy to see that all checks are passing now!

Are there any remaining issues preventing this PR from being merged and published to npm? I have a couple of projects that I'd like to use this code generator for, and I have time to contribute to getting this PR ready and merged. I've been testing the code generator and it works well for a couple of OpenAPI 3.0 contracts I've tested with.

I've also been experimenting a bit with tweaking the output format from the code generator, so I'm quite familiar with the code base on both this project and oazapfts. Please let me know if there's anything I can help with.

phryneas:
@Lokaltog Things still open until this goes out:

* [ ] my [latest comment](https://github.com/reduxjs/redux-toolkit/pull/1680#discussion_r744322993) regarding ts-node and .ts config files (this is up for grabs, otherwise I'll do it some time this week)
* [ ] waiting for @lindapaiste to move https://github.com/rtk-incubator/rtk-query-codegen/pull/82 over here
* [ ] @msutkowski wanted to change the filter to also accept a callback function
* [ ] checking for new oazapfts releases & potentially bumping the dependency

And waiting for some more feedback - but at this point I think we're getting to the point of having enough of that :)

On the oazapfts side, I think [this issue with the generation of enums](https://github.com/rtk-incubator/rtk-query-codegen/issues/37#issuecomment-954472165) could need solving, but I'm not sure if there is maybe already an open PR over there - please check that before getting to it :)

Lokaltog:
Great! I've resolved conflicts in https://github.com/rtk-incubator/rtk-query-codegen/pull/82 and submitted #1712 with @lindapaiste's changes. I'll look into the other issues later today.

phryneas:
Also, one last thing for the list:

* [ ] extend README and extend Documentation on the Homepage (or at least link the new README)

Lokaltog:
@grumpyTofu @phryneas I've attempted to reproduce the ts-node issue but on my setup both ts-node and esr works fine. I've installed the latest CI package and tested it stock, and I've tested by commenting out the esr import in cli.js (forcing it to use ts-node) but it works fine. TS is already configured to output CJS modules so I'm not sure what's wrong here.

@grumpyTofu could you provide a repo with a clean, minimal config that reproduces this issue?

lindapaiste:
I chatted with @msutkowski yesterday about a `filterEndpoints` callback function and I thought that was a good idea.  Is that something that we want _instead of_ my PR with the `excludeEndpoint` option, or _in addition_ to it?  I got the impression from Matt that we didn't want to go overboard with options.

Lokaltog:
I'm thinking that it's not necessary with both options as a `filterEndpoints` callback allows for the same filtering as `excludeEndpoints`, only with inverted checks. A callback is also more flexible so I'm voting for a single option for filtering.

grumpyTofu:
@Lokaltog Unfortunately, I don't have the repo that I am working with on github. However, the config is posted in the comments above. I am also just running it with a yarn script that runs the cli.js file using node and passes the config file as the argument.

phryneas:
@grumpyTofu could you please test the version I just pushed? I think your `import` problem should be solved with that.

packages/rtk-query-codegen-openapi/test/generateEndpoints.test.ts
phryneas: ping @msutkowski

grumpyTofu:
@phryneas Was this related to something in my PR? Or are you talking about the broken import Issue #1753? It looks like the broken import issue was opened by @ChuckJonas. However, I can help with the testing efforts. 

I can try to replicate the original issue so that I can test if that is the case, but it might be easier if @ChuckJonas tests this to see if it solves the original issue.

ChuckJonas:
@phryneas I was just running the [code from the example in the docs ](https://redux-toolkit.js.org/rtk-query/usage/code-generation)when I got the broken import issue.

phryneas:
@grumpyTofu sorry, "export" :sweat_smile: 
I was talking about this one:

```
$ node ./node_modules/@rtk-query/codegen-openapi/lib/bin/cli.js openapi.dev.config.ts
(node:33479) UnhandledPromiseRejectionWarning: /Users/austinfelix/apps/finsense-ui/openapi.dev.config.ts:10
export default config;
^^^^^^

SyntaxError: Unexpected token 'export'
    at wrapSafe (internal/modules/cjs/loader.js:1001:16)
```

lindapaiste:
Can we got PR #1738 merged in?  This PR is needed in order to support queries with method `POST` (which I am personally very excited about!).

We are now allowing the user to specify whether an endpoint is a `query` or a `mutation` rather than inferring based on the `method`.  Therefore we can no longer assume that all queries will have method `GET`.

lindapaiste:
> I am also just running it with a yarn script that runs the cli.js file using node and passes the config file as the argument.

I have encountered the same sort of issue as @grumpyTofu.  I would consider it "user error" because I'm pretty sure that it's caused by trying to run the script in an environment that doesn't have access to TS.

The error is not from _inside_ the rtk-query script.  It never gets far enough to execute the codegen.  The error is that node cannot open a `.ts` config file.  If I switch the config to a `.js` file with `module.exports = ` syntax then it works.  There is no problem executing the codegen.   If it's not the package itself that's causing the problem then I don't think that the package can fix it?

---
**Detailed explanation of errors**

My setup: calling `npm run-script generate` -> calls `node generate.js` -> executes the following file:
```
const {execSync} = require('child_process');

const DIR = "./src/API";

// download and merge all openapi.json files
execSync(`npx openapi-merge-cli --config ${DIR}/openapi-merge.json`);
// create TS types and RTK query API
execSync(`npx @rtk-query/codegen-openapi ${DIR}/codegen-config.ts`);
```

I have `ts-node` installed globally but I don't think `execSync` can use it.  So I get the error:
```
Encountered a TypeScript configfile, but neither esbuild-runner nor ts-node are installed.
node:child_process:903
    throw err;
    ^

Error: Command failed: npx @rtk-query/codegen-openapi ./src/API/codegen-config.ts
Encountered a TypeScript configfile, but neither esbuild-runner nor ts-node are installed.
```

So I changed the `.ts` to a `.js` and removed the `ConfigFile` type.  Then I get the error regarding export syntax:
```
Error: Command failed: npx @rtk-query/codegen-openapi ./src/API/codegen-config.js
C:\Users\Linda Paiste\PhpstormProjects\monaco-stable\src\API\codegen-config.js:18
export default config;
^^^^^^

SyntaxError: Unexpected token 'export'
```

So then I changed my config file to use `module.exports = config;` and that worked fine.  Which confirms that the codegen itself can run in the node `execSync` environment.

---
**Avoiding Errors**


The problem does not happen when I call the commands directly from the script.  Instead using the script to execute a file that calls the commands, I can cut out the middle step and put this in my `package.json`:
```
  "scripts": {
    "generate": "npx openapi-merge-cli --config ./src/API/openapi-merge.json npx @rtk-query/codegen-openapi ./src/API/codegen-config.ts",
  },
```

grumpyTofu:
@lindapaiste Yes, there is definitely an issue with the script implementation. It's not really related to the use of Javascript vs Typescript but more of an ES module thing. There are several fixes such as changing to a .mjs or .mts file extension which makes the script into an ES module OR change the export to ```module.exports = config;``` as you have pointed out. Perhaps the best solution though was to just use esbuild-runner as recommended by @phryneas and then you can write your package.json script as you normally would and it will solve this issue for you. Based on his recent commits, this is what was internally implemented so the cli will perform this behind the scenes. Personally, I like that method and have been using that instead for more than just this use case. Thanks to @phryneas for teaching me something new ;)

@phryneas to follow up on testing, everything looks great and the example config file from the docs should work fine now!

phryneas:
This is now released as `@rtk-query/codegen-openapi@1.0.0-alpha.1`. If no bug reports come in anytime soon, it will be the final 1.0.0

PR #1684, opened 2021-11-03
feat: generic type for `BaseQueryApi`

JulienKode:
## Description 

This pull request is to see if it's needed or not, or if it's useful or not to have generic type for BaseQueryAPI which can be different based on the extra argument that we inject at store level https://github.com/reduxjs/redux-toolkit/issues/1685 

I didn't test this example, this might not work


JulienKode:
I close the PR as it's not required by a lot of people https://github.com/reduxjs/redux-toolkit/issues/1685#issuecomment-961379658

PR #1686, opened 2021-11-03
allow for circular references by building reducer lazily on first reducer call

phryneas:
This would solve most long-standing problem of circular references between slices by calling the builder callback lazily when the reducer is called first instead of immediately on reducer creation.

### DIFF
--- a/packages/toolkit/src/createSlice.ts
+++ b/packages/toolkit/src/createSlice.ts
@@ -1,4 +1,5 @@
-import type { Reducer } from 'redux'
+import type { AnyAction, Reducer } from 'redux'
+import { createNextState } from '.'
 import type {
   ActionCreatorWithoutPayload,
   PayloadAction,
@@ -7,7 +8,11 @@ import type {
   _ActionCreatorWithPreparedPayload,
 } from './createAction'
 import { createAction } from './createAction'
-import type { CaseReducer, CaseReducers } from './createReducer'
+import type {
+  CaseReducer,
+  CaseReducers,
+  ReducerWithInitialState,
+} from './createReducer'
 import { createReducer, NotFunction } from './createReducer'
 import type { ActionReducerMapBuilder } from './mapBuilders'
 import { executeReducerBuilderCallback } from './mapBuilders'
@@ -253,19 +258,16 @@ export function createSlice<
 >(
   options: CreateSliceOptions<State, CaseReducers, Name>
 ): Slice<State, CaseReducers, Name> {
-  const { name, initialState } = options
+  const { name } = options
   if (!name) {
     throw new Error('`name` is a required option for createSlice')
   }
+  const initialState =
+    typeof options.initialState == 'function'
+      ? options.initialState
+      : createNextState(options.initialState, () => {})
+
   const reducers = options.reducers || {}
-  const [
-    extraReducers = {},
-    actionMatchers = [],
-    defaultCaseReducer = undefined,
-  ] =
-    typeof options.extraReducers === 'function'
-      ? executeReducerBuilderCallback(options.extraReducers)
-      : [options.extraReducers]
 
   const reducerNames = Object.keys(reducers)
 
@@ -294,19 +296,36 @@ export function createSlice<
       : createAction(type)
   })
 
-  const finalCaseReducers = { ...extraReducers, ...sliceCaseReducersByType }
-  const reducer = createReducer(
-    initialState,
-    finalCaseReducers as any,
-    actionMatchers,
-    defaultCaseReducer
-  )
+  function buildReducer() {
+    const [
+      extraReducers = {},
+      actionMatchers = [],
+      defaultCaseReducer = undefined,
+    ] =
+      typeof options.extraReducers === 'function'
+        ? executeReducerBuilderCallback(options.extraReducers)
+        : [options.extraReducers]
+
+    const finalCaseReducers = { ...extraReducers, ...sliceCaseReducersByType }
+    return createReducer(
+      initialState,
+      finalCaseReducers as any,
+      actionMatchers,
+      defaultCaseReducer
+    )
+  }
+
+  let _reducer: ReducerWithInitialState<State>
 
   return {
     name,
-    reducer,
+    reducer(state, action) {
+      return (_reducer ??= buildReducer())(state, action)
+    },
     actions: actionCreators as any,
     caseReducers: sliceCaseReducersByName as any,
-    getInitialState: reducer.getInitialState,
+    getInitialState() {
+      return (_reducer ??= buildReducer()).getInitialState()
+    },
   }
 }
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/tests/createSlice.test.ts
```diff
@@ -191,7 +191,8 @@ describe('createSlice', () => {
                 .addCase('increment', (state) => state + 1)
                 .addCase('increment', (state) => state + 1),
           })
-        ).toThrowErrorMatchingInlineSnapshot(
+          slice.reducer(undefined, { type: 'unrelated' })
```
phryneas, line 194: This is the only downside to this I could find. As the builder is now called lazily on first reducer call, `createSlice` will not immediately toggle these errors.
But they will be triggered immediately on adding this to a store, so I'm pretty sure that would be okay.

phryneas:
Tests are failing because apparently the integration branch tests are just currently failing. Tests for the PR run fine locally.

markerikson:
@phryneas Per #1687, the build failures appear to be due to the `version: "1.7.0-beta.0"` string in `package.json` causing Yarn v2 to fail to install the local RTK package.

Yarn v3 seems to fix that issue, but the docs aren't building clean atm.

As a workaround, just change the version string back to `"1.6.2"` on this branch and it _should_ build okay.

packages/toolkit/src/createSlice.ts
```diff
@@ -253,19 +258,16 @@ export function createSlice<
 >(
   options: CreateSliceOptions<State, CaseReducers, Name>
 ): Slice<State, CaseReducers, Name> {
-  const { name, initialState } = options
+  const { name } = options
   if (!name) {
     throw new Error('`name` is a required option for createSlice')
   }
+  const initialState =
```
markerikson, line 265: :hand: Doesn't this interfere with the lazy init logic I just added to `createReducer`?  We should be able to just pass this through.

phryneas: Since `createReducer` will be called later than it was before, I wanted to keep the freezing behaviour as soon as possible. It should not really interfere with it, the "work" is just being done twice.

phryneas:
Grr. TS 3.9 does not have `??=` yet.

phryneas:
And it builds green. W00p w00p.

markerikson:
🚢 🚢 :shipit:

MatthewWid:
Hi there. Is there a plan to add similar functionality for `createReducer` itself too?

After refactoring a module in my project from `createReducer` back to `createSlice` this update certainly fixed the circular dependency issue (amazing!) but it would be nice if `createReducer` itself had the same behaviour.

markerikson:
@MatthewWid That shouldn't really be a concern in most cases. The circular dependency issue only shows up with `createSlice` because you may have `sliceA` wanting to listen to an action defined in `sliceB` and vice-versa.  Since `createReducer` does not define any new action types, that scenario shouldn't happen.

Do you have an actual example of a `createReducer` use that is causing a circular dependency?

MatthewWid:
@markerikson The original code is proprietary so I cannot post it exactly, unfortunately, but I have the taken time to create a repository with a reproduction of what I am talking about:

https://github.com/MatthewWid/rtk-createreducer-circular-dependencies

Running instructions are in `README.md`, I hope it's easy to follow! For reference, the original project follows the project structure from [Bulletproof React](https://github.com/alan2207/bulletproof-react) which is why it is organised the way it is.

See `src/index.ts` - it simply runs a series of actions and prints the store state after each.

`master` demonstrates the issue, showing that there is a circular dependency stemming from an action created by `createAsyncThunk` used in a separate reducer that is undefined and throws:

```
  builderCallback(builder)
  ^
TypeError: Cannot read property 'fulfilled' of undefined
    at /src/features/users/slices/user.reducer.ts:13:24
    at executeReducerBuilderCallback (/node_modules/.pnpm/@reduxjs+toolkit@1.7.0-rc.0/node_modules/@reduxjs/toolkit/src/mapBuilders.ts:194:3)
    at createReducer (/node_modules/.pnpm/@reduxjs+toolkit@1.7.0-rc.0/node_modules/@reduxjs/toolkit/src/createReducer.ts:220:9)
    at Object.<anonymous> (/src/features/users/slices/user.reducer.ts:6:42)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Module.m._compile (/node_modules/.pnpm/ts-node@10.4.0_d1ba71cee4528a43e185cc3d6fa907dc/node_modules/ts-node/src/index.ts:1371:23)
    at Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Object.require.extensions.<computed> [as .ts] (/node_modules/.pnpm/ts-node@10.4.0_d1ba71cee4528a43e185cc3d6fa907dc/node_modules/ts-node/src/index.ts:1374:12)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
 ELIFECYCLE  Command failed with exit code 1.
```

This occurs with the `createPost` action in the `user.reducer.ts` file.

If you then checkout the `slice-fix` branch it implements the exact same functionality, but simply refactored to use slices instead (and is on the `@next` version of RTK). You can see that it works just fine after. Hooray for the update!

Ideally, the first form on the `master` branch would work by lazy-loading the initialization for `createReducer` too.

Thanks so much for your responsiveness and work on RTK!

markerikson:
@MatthewWid yeah, the problem here is you've managed to create a real circular dependency based on your own folder structure:

- `posts.reducer` imports `deleteUser`
- `users.reducer` imports `createPost`

and this is made worse by the fact that you are using `feature/myFeature/index.ts` files to re-export everything together, and then having those reducer files import from that other feature's index files.

This can be avoided entirely given the current file structure if you just have the reducers import directly from those separate action files instead, ie, `import { createPost } from 'features/posts/post.actions'`.

We might be able to update `createReducer` with a similar implementation what we did in `createSlice`, but the bigger issue here is really how you've structured your logic and imports in the first place.

MatthewWid:
@markerikson That's understandable. As I say it's structured from [Bulletproof React](https://github.com/alan2207/bulletproof-react/blob/master/docs/project-structure.md) (that I believe is loosely similar to the Redux docs feature-folder recommendations?):

> Everything from a feature should be exported from the index.ts file which behaves as the public API of the feature.

Otherwise you're necessarily relying on the internal structure of one feature from another feature so they are not as decoupled as they could be.

Regardless, you're right in that the circular dependency is created by myself here, though the fact that this update seems to fix this exact issue only for `createSlice` and not `createReducer` when the former is supposed to be convenient way to write the latter (among other things) seems unexpected to me.

Thanks for considering, anyway!

markerikson:
The Redux docs do say "use feature folders", but personally I've always disliked the idea of "re-export _everything_ from an `index.ts` file".

Modifying `createReducer` _should_ be feasible, it's just never been brought up as the cause of a circular dependency problem before that I can think of because typically you _aren't_ generating action types in that same file the way `createSlice` does.

phryneas:
Tbh, that bulletproof react recommendation might make sense for externally exported stuff, but that would still not mean that within that module you should import it from the index. That's how you get circularities - the index should only be used from the outside.

PR #1687, opened 2021-11-04
[TESTING ONLY] Figure out CI issues with 1.7 integration branch

markerikson:


markerikson:
For posterity, here's the initial symptoms I was investigating:

- As of commit 5e36d3a on the 1.7 branch, everything built fine
- I pushed 6c0dafb, which _only_ contained a `version: "1.7.0-beta.0"` string change, and suddenly the type tests started failing
- Lenz's PR in #1686, based off that commit, was also failing

After investigation, my conclusions are this is _specifically_ an issue with:

- our CI test approach of running `yarn pack` to generate a local build, and then install it back into the workspace
- because Yarn v2 doesn't like the version string with characters in it
- and then doesn't actually install/extract the package correctly
- causing the tests to be importing an _old_ actual published version of RTK instead

Some additional evidence from the logs in https://twitter.com/thblckjkr/status/1456112118097489920 :

> every build that succeeds, in the "Install deps -> resolution step" section has something along the lines of
> 
> "@reduxjs/toolkit@workspace:packages/toolkit [HASH]"
> 
> And the ones that fail don't have it.

Updating from Yarn 2.4 to Yarn 3.1 _appears_ to correctly install the local RTK package, and thus the type tests pass.

markerikson:
Everything's clean. Closing to create a new branch against `master`.

merceyz:
This is an interesting issue so I looked into it to figure out what changes in Yarn fixed this, here are my findings.

When the version of `@reduxjs/toolkit` is `1.6.0` this is the layout on disk
```
/node_modules/@reduxjs/toolkit@workspace:packages/toolkit
/packages/toolkit/node_modules/@reduxjs/toolkit@file:./package.tgz
```

When the version is `1.7.0-beta.0` it becomes
```
/node_modules/@reduxjs/toolkit@npm:1.6.2
```

When updating to Yarn v3 it becomes
```
/node_modules/@reduxjs/toolkit@npm:1.6.2
/packages/toolkit/node_modules/@reduxjs/toolkit@file:./package.tgz
```

It fetching the older version from the registry is because some of your workspaces are depending on `@reduxjs/toolkit@^1.6.0` which doesn't match `1.7.0-beta.0` so that is valid behaviour.

However `packages/toolkit/node_modules/@reduxjs/toolkit@file:./package.tgz` not showing up appears to have been a hoisting bug that was fixed in https://github.com/yarnpkg/berry/pull/2481

markerikson:
Gotcha, thanks for looking into it and confirming the research!

(if nothing else this gave us an excuse to bump to Yarn 3, so.... winning? :) )

markerikson:
So actually, my other question is: this _really_ only broke when the version string in the locally-published package has a `-` in the name, like `1.6.2-b` .  What caused _that_ part? :)

merceyz:
The way https://github.com/npm/node-semver works a range has to explicitly allow pre-releases, in other words `1.6.2-b` doesn't satisfy `^1.6.0`

https://github.com/npm/node-semver/tree/cb1ca1d5480a6c07c12ac31ba5f2071ed530c4ed#prerelease-tags

PR #1688, opened 2021-11-04
Update Yarn from 2.4 to 3.1

markerikson:
This PR:

- Updates Yarn to v3.1

This appears to resolve the typetest failures we were seeing in the 1.7 integration branch, which were apparently happening because Yarn v2 doesn't want to install a local package tarball correctly if the version name has a `-` in it, like `1.7.0-beta.0`.  As a result, the typetests get run against an older actual install (like the real 1.6.2).  Yarn v3 does correctly install the local package.

Research from that:

https://github.com/reduxjs/redux-toolkit/pull/1687#issuecomment-960447941

If the docs builds fail here, they should probably succeed after clearing the cache and rebuilding.


phryneas:
Absolutely in favor of this!

FWIW, the size build step is probably just failing because the previous build step was failing - nothing to worry about there.

markerikson:
Okay, if you're cool with this, I'll go ahead and merge to `master`, and you can pull it over to `1.7-integration`.

PR #1693, opened 2021-11-05
bug fixes PR 1680

grumpyTofu:
#1680

### DIFF
--- a/packages/rtk-query-codegen-openapi/package.json
+++ b/packages/rtk-query-codegen-openapi/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@rtk-query/codgen-openapi",
+  "name": "@rtk-query/codegen-openapi",
   "version": "1.0.0-alpha.1",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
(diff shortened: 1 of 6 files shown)


phryneas:
Thanks for all the testing!

As for the variable name problem:

I'd prefer to keep the behaviour as it is, but fixing the bug with the rename-on-export.

When we keep the names predictable within the file, we can't run into weird naming conflicts.

That way the import and export could for example have the same name without any problem.

So, the idea would be

```js

import { importName as api } from './importFile'

const injectedRtkApi = api.injectEndpoints({ ... })

export { injectedRtkApi as exportName }

export { useMyQuery } = injectedRtkApi

```

packages/rtk-query-codegen-openapi/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@rtk-query/codgen-openapi",
```
phryneas, line 2: Good catch! That would have been weird :rofl:

grumpyTofu:
@phryneas Good call, I didn't even think about someone specifying them as the exact same name in their config, but now that you mention it, I could see that happening.

phryneas:
thanks a lot!

PR #1699, opened 2021-11-06
fetchBaseQuery: return `null`on empty body for JSON. Add DevWarnings.

phryneas:
Just something I wanted to do for a while.

### DIFF
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -54,7 +54,7 @@ const handleResponse = async (
 
   if (responseHandler === 'json') {
     const text = await response.text()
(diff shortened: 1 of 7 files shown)


packages/toolkit/src/query/fetchBaseQuery.ts
```diff
@@ -54,7 +54,7 @@ const handleResponse = async (
 
   if (responseHandler === 'json') {
     const text = await response.text()
-    return text.length ? JSON.parse(text) : undefined
+    return text.length ? JSON.parse(text) : null
```
phryneas, line 57: Note that this is a almost-breaking change, but a very minor one.

It is necessary as we cannot serialize `undefined` when transferring data from server to client in SSR scenarios.

PR #1703, opened 2021-11-07
fix: use correct working directory for CLI test

Lokaltog:
I had some time to investigate this issue, and I think I found the root cause (and a fix). Test actions run successfully both locally and on actions: https://github.com/Lokaltog/redux-toolkit/runs/4130398633

    GitHub actions override the working directory in tests, which causes
    `path.resolve` to point to nonexisting files. There are some mentions of
    similar issues on SO.
    
    This fix resolves paths from `process.env.PWD` which points to the
    process working directory which is set correctly both on local runs and
    GitHub actions.

Update: The pipeline is still failing, but the issue appears unrelated to this PR. I'm unable to reproduce the build issue locally, and from the logs it appears related to the GraphQL generator.


phryneas:
Thank you so much! 
I really had no idea what was causing this and was getting desperate.

PR #1706, opened 2021-11-07
Update action listener middleware to cover additional use cases

markerikson:
This PR updates the action listener middleware based on the changes proposed in https://github.com/reduxjs/redux-toolkit/discussions/1648#discussioncomment-1599805 :

- Removes the `type: "module"` line from the listener middleware `package.json` file to avoid unwanted ESM handling
- Sets up Jest+TS testing
- Copies the test file from #547 
- Adds additional tests for the `matcher` and `actionPredicate` listener overloads
- Rewrites the internals to track all listeners in a `Map<idString, ListenerEntry>`
- Removes the `stopPropagation` method
- Renames the `When` values to `'beforeReducer'` and `'afterReducer'`, and adds an additional `'both'` value
- Wraps listeners in a `try/catch` to handle errors
- Hacks in runtime support for an `extra` value ala thunks, but without good typing
- Adds a `getOriginalState` method that captures the state before the action was dispatched
- Tries and fails to get thunk-typed dispatching working
- Adds a `condition` method to enable long-running async workflows, based on https://docs.temporal.io/docs/typescript/workflows/#condition

For any review, please focus on actual code per intent - let's save the bikeshedding for the discussion thread after I get an updated alpha release out :)

### DIFF
--- a/.github/workflows/listenerTests.yml
+++ b/.github/workflows/listenerTests.yml
@@ -1,0 +1,31 @@
+name: CI
+on: [push, pull_request]
+
+jobs:
+  build:
+    name: Test Listener Middleware on Node ${{ matrix.node }}
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: ['14.x']
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
+      - name: Use node ${{ matrix.node }}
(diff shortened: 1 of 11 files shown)


markerikson:
Ignoring the docs failure - another "can't find `formik` while building the docs" thing that is normally solved by rebuilding docs without cache.

PR #1710, opened 2021-11-09
Update usePrefetchImmediately snippet

msutkowski:
I opted to show using `useAppDispatch` here, even though it would be fine to `useDispatch`. I figured being that we're showing types, it's probably worthwhile to encourage it.

I thought about including that if you were to provide `options` that you'd want to memo them, but also feel like that goes beyond this scope.

### DIFF
--- a/docs/rtk-query/usage/prefetching.mdx
+++ b/docs/rtk-query/usage/prefetching.mdx
@@ -89,9 +89,9 @@ export function usePrefetchImmediately<T extends EndpointNames>(
   arg: Parameters<typeof api.endpoints[T]['initiate']>[0],
   options: PrefetchOptions = {}
 ) {
-  const dispatch = useDispatch()
+  const dispatch = useAppDispatch()
   useEffect(() => {
-    dispatch(api.util.prefetch(endpoint, arg, options))
+    dispatch(api.util.prefetch(endpoint, arg as any, options))
   }, [])
 }
 


docs/rtk-query/usage/prefetching.mdx
```diff
@@ -89,10 +89,10 @@ export function usePrefetchImmediately<T extends EndpointNames>(
   arg: Parameters<typeof api.endpoints[T]['initiate']>[0],
   options: PrefetchOptions = {}
 ) {
-  const dispatch = useDispatch()
+  const dispatch = useAppDispatch()
   useEffect(() => {
-    dispatch(api.util.prefetch(endpoint, arg, options))
-  }, [])
+    dispatch(api.util.prefetch(endpoint, arg as any, options))
+  }, [arg, dispatch, endpoint, options])
```
phryneas, line 95: If we add a dependency array here, we will need to keep `arg` and `options` stable. 
Either we add a note to the docs or we export the tools we have in the `query/react` package.

@markerikson what are your thoughts on exposing `useShallowStableValue` and maybe even `useStableQueryArgs` (the latter would need changes to be usable outside of the hooks internals)?

msutkowski: Yea, I had that in the description, but wasn't sure if we need to talk about keeping those things stable as a part of this work or not. 100% agree with exposing `useShallowStableValue` :)

msutkowski: @markerikson Any thoughts here? I can always just revert this back to a dep array of `[]` just so we can get the `arg as any` in.

markerikson: too many other things going on, no thoughts here :)

phryneas: keep it with the `[]` for now and we'll think of the rest for the next release when we have resources?

PR #1711, opened 2021-11-09
Only set originalArgs if they're not undefined

### LINKED ISSUE #1709: RTK Query - Error serializing undefined in getServerSideProps
as @phryneas asked, I reproduced the same error in [this repo](https://github.com/elshanx/rtkq-beta-ssr-error) with pokemon api

the problem occurs when dispatching a thunk in getServerSideProps without an argument like this `store.dispatch(getPokemon.initiate())`

but `store.dispatch(getPokemon.initiate(null))` works just fine

msutkowski:
Fixes #1709. 

I believe we can just skip setting this property in the substate when it's `undefined`. I double-checked all of the invalidation behavior and this appears to be a non-issue. 

For a user that uses `originalArgs`, there is no difference in the check they'd have to use for this anyways?

https://codesandbox.io/s/rtkq-beta-ssr-repro-7q819 - CSB running this build

### DIFF
--- a/packages/toolkit/src/query/core/buildSlice.ts
+++ b/packages/toolkit/src/query/core/buildSlice.ts
@@ -37,7 +37,6 @@ import {
   copyWithStructuralSharing,
 } from '../utils'
 import type { ApiContext } from '../apiTypes'
-import { defaultMemoize } from 'reselect'
 
 function updateQuerySubstateIfExists(
   state: QueryState<any>,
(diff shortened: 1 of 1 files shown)


packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -37,7 +37,6 @@ import {
   copyWithStructuralSharing,
 } from '../utils'
 import type { ApiContext } from '../apiTypes'
-import { defaultMemoize } from 'reselect'
```
msutkowski, line 40: This was a :zombie: import.

phryneas:
I think this should do just fine. I'll give it another look in the evening just to be double sure.

packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -145,7 +144,9 @@ export function buildSlice({
           updateQuerySubstateIfExists(draft, arg.queryCacheKey, (substate) => {
             substate.status = QueryStatus.pending
             substate.requestId = meta.requestId
-            substate.originalArgs = arg.originalArgs
+            if (typeof arg.originalArgs !== 'undefined') {
```
phryneas, line 147:
```suggestion
            if (arg.originalArgs !== undefined) {
```

PR #1713, opened 2021-11-09
match endpoints with a function, exclude endpoints

lindapaiste:
Following up with the discussion [here](https://github.com/reduxjs/redux-toolkit/pull/1680#issuecomment-963892589), I am moving [my PR](https://github.com/rtk-incubator/rtk-query-codegen/pull/82) from the rtk-query-codegen repo to this monorepo and also adding the ability to filter endpoints using a function.  The function is called with the operation name and the operation definition -- are these the desired arguments?  Should it be the operation definition only?

I am still [unclear](https://github.com/reduxjs/redux-toolkit/pull/1680#issuecomment-964423073) on whether we actually *want* the `excludeEndpoint` option.

TODO: update documentation

### DIFF
--- a/packages/rtk-query-codegen-openapi/src/types.ts
+++ b/packages/rtk-query-codegen-openapi/src/types.ts
@@ -55,9 +55,9 @@ export interface CommonOptions {
 
 export type TextMatcher = string | RegExp | (string | RegExp)[];
 
-export type EndpointMatchFunction = (operationName: string, operationDefinition: OperationDefinition) => boolean;
+export type EndpointMatcherFunction = (operationName: string, operationDefinition: OperationDefinition) => boolean;
 
-export type EndpointMatcher = TextMatcher | EndpointMatchFunction;
+export type EndpointMatcher = TextMatcher | EndpointMatcherFunction;
 
 export interface OutputFileOptions extends Partial<CommonOptions> {
   outputFile: string;
--- a/packages/rtk-query-codegen-openapi/src/types.ts
+++ b/packages/rtk-query-codegen-openapi/src/types.ts
@@ -53,14 +53,21 @@ export interface CommonOptions {
   hooks?: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


phryneas:
Operation name and operation definition sound fine to me.

I think at this point we can really go without the `excludeEndpoint` option, as it can be replicated quite easily using `filter(endpointName){ return !(/pattern/.test(endpointName)) }`

lindapaiste:
> I think at this point we can really go without the `excludeEndpoint` option, as it can be replicated quite easily using `filter(endpointName){ return !(/pattern/.test(endpointName)) }`

Sounds good to me, I'll modify this PR to remove the option.

packages/rtk-query-codegen-openapi/src/types.ts
```diff
@@ -53,14 +53,20 @@ export interface CommonOptions {
   hooks?: boolean;
 }
 
+export type TextMatcher = string | RegExp | (string | RegExp)[];
+
+export type EndpointMatchFunction = (operationName: string, operationDefinition: OperationDefinition) => boolean;
+
+export type EndpointMatcher = TextMatcher | EndpointMatchFunction;
```
msutkowski, line 60:
```suggestion
export type TextMatcher = string | RegExp | (string | RegExp)[];

export type EndpointMatcherFunction = (operationName: string, operationDefinition: OperationDefinition) => boolean;

export type EndpointMatcher = TextMatcher | EndpointMatcherFunction;
```

Just to keep our naming consistent - we typically refer to these as matcher functions throughout the code and docs.

msutkowski (commented):
Just quick naming suggestions, but otherwise looks great. Thanks so much for doing this!

msutkowski (approved):
Looks good to me. @phryneas if you wanna double-check as well? :bow:

phryneas:
Looks good to me, let's get this in! :)

PR #1714, opened 2021-11-10
Codegen: can customize body property

lindapaiste:
This PR adds a `bodyArgName` option to the openapi codegen which can be a `string` (ie. `'body'`) or a `function`.

I'm open to any feedback/changes.  Mostly I just want some way to override the generated name for the `'POST'` body property so that I don't have query args like `{ someReallyLongAndAnnoyingName: body }`.

If this code looks good, I would probably move the `type QueryArgDefinition` from `generate.ts` to `types.ts` because I've introduced a circular import which I do not like (It's not a *problem* because it's just types, but it's not ideal).

### DIFF
--- a/packages/rtk-query-codegen-openapi/src/generate.ts
+++ b/packages/rtk-query-codegen-openapi/src/generate.ts
@@ -15,7 +15,7 @@ import {
 } from 'oazapfts/lib/codegen/tscodegen';
 import type { OpenAPIV3 } from 'openapi-types';
 import { generateReactHooks } from './generators/react-hooks';
-import type { EndpointOverrides, GenerationOptions, OperationDefinition } from './types';
+import type { EndpointMatcher, EndpointOverrides, GenerationOptions, OperationDefinition, TextMatcher } from './types';
 import { capitalize, getOperationDefinitions, getV3Doc, isQuery as testIsQuery, removeUndefined } from './utils';
 import type { ObjectPropertyDefinitions } from './codegen';
 import { generateCreateApiCall, generateEndpointDefinition, generateImportNode } from './codegen';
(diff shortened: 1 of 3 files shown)


msutkowski:
This also looks fine to me, but let's wait to rebase it after #1713 gets merged.

phryneas:
Hmm. Before we merge this: is there not a way for the schema to just specify a name for the body? :thinking:

lindapaiste:
> Hmm. Before we merge this: is there not a way for the schema to just specify a name for the body? 🤔

You mean in the OpenAPI itself?  I'm sure there is a way, but this feature would be useful if the consumer of the API does not have the ability to modify the actual API definitions.

phryneas:
It might make sense to pause a bit and give this some research and general thought.

* are there tools out there that could be used to pre-process an OpenAPI schema to rename things?
* are there situations, where such a tool could not do a rename without massively restructuring the whole openapi? (What about body type specifications that are not a Ref, but inline? Can we define a name there at all?)

Because if we add something for this, sooner or later we will end up adding more options to rename endpoints and arguments.
And at that point, it will probably impact maintainability a lot.

So if there were such a tool and if it were possible in all cases, I would prefer just using & propagating that.

If there are cases where that is just not feasible though, we'll probably have to add this, but then it might make sense to not have a "rename the body arg" option, but a "rename all args due to this mapping object" option or something like that from the start.

phryneas:
I'll add this to the "codegen 1.1" milestone so we can still discuss this without blocking a potentially soon 1.0 release

lindapaiste:
@msutkowski @phryneas I've learned that there is an option `x-codegen-request-body-name` but it doesn't seem to do anything.

https://openapi-generator.tech/docs/swagger-codegen-migration/#body-parameter-name

This probably needs to be handled in oazapfts rather than here. I'll see what I can do.

PR #1723, opened 2021-11-12
Overload resolution possible solution

josh-degraw:
Some of the tests still fail, and I didn't figure out how to do everything you were wanting to do here, but as far as type-checking goes I think I was able to figure out at least a halfway decent approach for at encapsulating the overloads.

### DIFF
--- a/packages/action-listener-middleware/src/globals.ts
+++ b/packages/action-listener-middleware/src/globals.ts
@@ -1,3 +1,0 @@
-import type { Dispatch, AnyAction } from '@reduxjs/toolkit'
-
-export interface AddListenerGlobals {}
--- a/packages/action-listener-middleware/src/globals.ts
+++ b/packages/action-listener-middleware/src/globals.ts
@@ -1,0 +1,3 @@
+import type { Dispatch, AnyAction } from '@reduxjs/toolkit'
+
+export interface AddListenerGlobals {}
--- a/packages/action-listener-middleware/src/index.ts
+++ b/packages/action-listener-middleware/src/index.ts
@@ -392,20 +392,6 @@ export const addListenerAction = createAction(
     }
   }
 ) as TypedAddListenerAction<unknown>
-// as unknown as BaseActionCreator<
-//   ListenerEntry<unknown, Dispatch<AnyAction>>,
-//   'actionListenerMiddleware/add'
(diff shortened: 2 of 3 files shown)


markerikson:
Yeah, this looks _very_ promising!

image

Correct inference of the action based on the action creator, and correct state type carried through.  Let me play with this some more.

markerikson:
The immediate issue I think I see, reading through this without running it yet, is that there's now a mismatch between `createListenerEntry` and `AddListenerOverloads`.  `cLE` is expecting a union of options objects. `ALO` is expecting alternate overloads.  The _types_ for the overloads definitely look like they work right, but the casting of all that dummy code I have strewn all over is broken because the actual behaviors/args are completely different.

Let me see if I can pull these pieces together and make them agree with each other.

markerikson:
I've added some type tests to clarify intended behavior.  I also fixed a type bug in `TypedAddListenerAction` (`D` was getting passed as a state arg instead of `S`) and added a `TypedAddListener` type.

Things are looking good so far!  The extraction of those `TypedAdd*` types is exactly what I was thinking of, particularly being able to configure the return type.

Biggest thing I'm seeing atm type-wise is that the `currentState, previousState` args in listener predicates aren't picking up state and are coming through as `any`, and atm I can't figure out where that `any` is coming from.  There's also the mismatch of `createListenerEntry` args I mentioned above.

I'm going to try backing up to the original start of the types changes and also play with @phryneas 's ideas in https://github.com/reduxjs/redux-toolkit/compare/master...phryneas:ts-fix-listenerMw?expand=1 a bit.

josh-degraw:
I managed to get the state types inferred correctly (overload ordering lol). I'll keep playing with the other bits

markerikson:
FYI, I think I've got something working locally, based on a combination of your earlier commits and Lenz's PR branch. I have _one_ type test that is not cooperating that I will probably need your help with. Will push up a PR momentarily.

markerikson:
@josh-degraw PR is up - see https://github.com/reduxjs/redux-toolkit/pull/1730 .

markerikson:
I'm gonna close this now that #1730 is merged in. Thank you for taking the time to put this together and file the PR - it was a huge help!

PR #1726, opened 2021-11-12
remove typeof undefined checks where not necessary 🐃🪒

phryneas:
This is just a bit of yak shaving, but I noticed we are doing the `typeof x === 'undefined'` check in quite some places where we would not need to. I changed it to `x === undefined` (or `x == null` if there was an `x === null` check going on in parallel anyways).

I kept the behaviour for non-local variables like `window`, `navigator`, `process` and `AbortController`.

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -58,8 +58,8 @@ import type { BaseQueryFn } from '../baseQueryTypes'
 // Copy-pasted from React-Redux
 export const useIsomorphicLayoutEffect =
   typeof window !== 'undefined' &&
-  typeof window.document !== 'undefined' &&
-  typeof window.document.createElement !== 'undefined'
+  window.document &&
+  window.document.createElement
     ? useLayoutEffect
     : useEffect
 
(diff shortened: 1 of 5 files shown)


packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -55,8 +55,8 @@ import { useShallowStableValue } from './useShallowStableValue'
 // Copy-pasted from React-Redux
 export const useIsomorphicLayoutEffect =
   typeof window !== 'undefined' &&
-  typeof window.document !== 'undefined' &&
-  typeof window.document.createElement !== 'undefined'
+  window.document &&
+  window.document.createElement
     ? useLayoutEffect
     : useEffect
```
msutkowski, line 61: Can this be even shorter? It looks like there are [many implementations](https://github.com/Andarist/use-isomorphic-layout-effect/blob/master/src/index.ts) that are just:

`typeof document !== 'undefined' ? useLayoutEffect : useEffect`

msutkowski (commented):
Overall looks good to me, just a quick question about the ismorphic layout effect check.

packages/toolkit/src/query/react/buildHooks.ts
markerikson: Hmm. I would normally say "those other implementations are probably wrong", but Andarist normally knows what he's doing.

I'm going to say leave this alone - it's there, we know it works.

PR #1728, opened 2021-11-12
Fix broken link

vmkevv:
Change `invalidatesTags` to `invalidatestags` because the first one is not working in the docs.


markerikson:
Thanks!

PR #1730, opened 2021-11-12
Rewrite action listener middleware types for better DX

markerikson:
This PR:

- Adds a bunch of type tests for desired action listener middleware behavior
- Splits out normal and type imports
- Removes an unneeded "options" generic
- Combines changes from @josh-degraw in https://github.com/reduxjs/redux-toolkit/pull/1723 and @phryneas in https://github.com/phryneas/redux-toolkit/tree/ts-fix-listenerMw to come up with the desired API and TypeScript DX behavior we want

The major changes in API/DX are:

- Both `middleware.addListener` and `addListenerAction` now take the exact same params - an `options` object containing the `listener` callback and `when` option, with five possible variations in how the middleware decides what actions will trigger the listener callback:
  - a plain action type string
  - an RTK action creator
  - an RTK "matcher" function
  - a "listener predicate" like `(action, currentState, originalState) => boolean`
  - a "listener predicate" which is also a type guard for the action
- The common overloads are now extracted as a reusable type that accepts the right return value
- It now exports `TypedAddListener<State>` and `TypedAddListenerAction<State>` types that can be used to create "pre-typed" versions of `middleware.addListener` and `addListenerAction`
- There's a `createListenerEntry` function that does the real work of looking at the options and assembling a listener entry
- `createActionListenerMiddleware()` defaults the state to `unknown`
- The middleware now includes `addListenerAction` as one of its attached fields, which means that if you do `createActionListenerMiddleware<MyState>()`, then `middleware.addListener/addListenerAction` are pre-typed.

The net results are that it should be fairly straightforward for users to ensure proper typing for their listener callbacks with the right state and action types, similar to how we define "pre-typed" hooks.

The _one_ thing that appears to be not working atm is that a `ListenerPredicate` of `(action, currentState, previousState) : action is PayloadAction<number> => {}` does not appear to be inferring the type of `action` in the listener callback (see test file line 813).  You can see the failure in the test job run.

### DIFF
--- a/packages/action-listener-middleware/src/index.ts
+++ b/packages/action-listener-middleware/src/index.ts
@@ -73,17 +73,12 @@ export type WithMiddlewareType<T extends Middleware<any, any, any>> = {
 export type MiddlewarePhase = 'beforeReducer' | 'afterReducer'
 
 export type When = MiddlewarePhase | 'both' | undefined
-type WhenFromOptions<O extends ActionListenerOptions> =
-  O extends ActionListenerOptions ? O['when'] : never
 
 /**
  * @alpha
  */
-export interface ActionListenerMiddlewareAPI<
-  S,
-  D extends Dispatch<AnyAction>,
-  O extends ActionListenerOptions
-> extends MiddlewareAPI<D, S> {
+export interface ActionListenerMiddlewareAPI<S, D extends Dispatch<AnyAction>>
+  extends MiddlewareAPI<D, S> {
   getOriginalState: () => S
   unsubscribe(): void
   condition: ConditionFunction<AnyAction, S>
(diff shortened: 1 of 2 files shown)


packages/action-listener-middleware/src/index.ts
```diff
@@ -124,18 +146,165 @@ export interface CreateListenerMiddlewareOptions<ExtraArgument = unknown> {
...
+
+  <LP extends AnyActionListenerPredicate<S>>(
+    options: {
+      actionCreator?: never
+      type?: never
+      matcher?: never
+      predicate: LP
+      listener: ActionListener<AnyAction, S, D>
+    } & ActionListenerOptions
+  ): Return
```
josh-degraw, line 199: I'll have to check on my computer to verify, but making this overload to be the first one on top should fix the state inference problem.

When I did that on my branch, I then saw a different error which I fixed using the  `ListenerPredicateGuardedActionType` I put in there.

markerikson: What I'm seeing is an action inference problem (which might be what you meant).  I just tried moving this one up, and didn't see a difference.  I did already grab `LPGuardedActionType` from your branch over to here.

markerikson:
I will poke at this a bit further, but this is a big enough improvement I'm inclined to just comment out the one failing typetest and ship this as 0.3.  (assuming the package changes work as intended as well.)

markerikson:
I'm sorta figuring out why this is failing.  The `ListenerPredicate<Action, State.` type doesn't seem match unless you explicitly declare the argument types in your predicate?

No types declared:

image

With explicit types:

image

I suppose the question is if there's any way to get the predicate to kick in even if there's no explicit args declared...?

markerikson:
Going to merge this as-is. I've got some other functionality I want to work on now that the types are basically in place.

Issue #1725, opened 2021-11-12
[RTK Query] Clearing whole API for new user session


ayushsharma82:
Hi,

I'm using RTK Query in my React Native app and I would like to know if there is any way we can clear the whole API for new user session? What I'm trying to achieve is to reset all query hooks and mark them uninitialized. 

Please let me know if someone has found a way to do this. I'll mark this as a feature request for RTK query ( Feature: a reset function ). 

Thanks for the amazing library.

Grapedge:
Is this what you need? [resetApiState](https://redux-toolkit.js.org/rtk-query/api/created-api/cache-management-utils#resetapistate)

phryneas:
`resetApiState` alone will not reset the hooks, as the hooks are written to always keep the "last known state" in `data`. They have the `currentData` property in the `1.7` beta though, which would reset with that.

ayushsharma82:
@phryneas  Thanks for the clarity on this topic! So should I try the 1.7@beta-1 release?

markerikson:
@ayushsharma82 yes, please do! we'd appreciate any feedback on the new functionality overall.

ayushsharma82:
One Question: Do we also mark them uninitialized with v1.7?

phryneas:
Nope, not at this moment.

ayushsharma82:
So I've tested v1.7.0@beta-1 with my app which is near to production. Everything seems to run smoothly and resetApiState seems to reset all data in the queries.

ayushsharma82:
A function to reset all internal hooks like `isUnitialized`, `isSuccess` etc. would be a lot helpful as my placeholder components depend on it. Right now, data seems to be re-fetched after dispatching `resetApiState` but the internal hooks ( `is Uninitialized` etc ) break the loading state of the application.

phryneas:
Yeah, I agree that using `resetApiState` should also reset the hooks. I'll see if I can find a quick fix. Would you be willing to do some testing with that?

ayushsharma82:
Sure, I'll test your changes. Thanks for the effort.

ayushsharma82:
Fixed with #1735

bryantobing12:
is there a way to use `resetApiState` but only for certain endpoint, not the whole API?

phryneas:
No, you can only rerun the query or stop using it.

noahfaro:
[...]

PR #1735, opened 2021-11-13
 fix: api.util.resetApiState should reset useQuery hooks

phryneas:
This should provide a fix for #1725.

@ayushsharma82 could you please try this? In a few minutes, the CodeSandbox bot should comment on this issue and one of the links there should lead to installation instructions for this build.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,6 +18,12 @@ module.exports = {
       'error',
       { prefer: 'type-imports', disallowTypeAnnotations: false },
     ],
+    'react-hooks/exhaustive-deps': [
+      'warn',
+      {
+        additionalHooks: '(usePossiblyImmediateEffect)',
+      },
+    ],
   },
   overrides: [
     // {
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -55,6 +55,7 @@ export type QueryActionCreatorResult<
   unsubscribe(): void
   refetch(): void
   updateSubscriptionOptions(options: SubscriptionOptions): void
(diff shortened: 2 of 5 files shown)


ayushsharma82:
@phryneas Tried this fix with my app and I can confirm that hooks are getting reset by looking at the placeholder components.

I'll also generate a small example to confirm that all hooks are getting reset or not.

ayushsharma82:
@phryneas  I've made this small [demo](https://codesandbox.io/s/reduxjs-redux-toolkit-forked-qu9n6?file=/package.json:265-273) to track this issue. It does reset all the hooks but data is never loaded again if the query is present during the dispatch of `resetApiState`. Let me know if this is an intended behaviour.

Demo uses c7d0efeb

phryneas:
Hmm you're right, that is not really optimal behaviour. I'll have to look into that tomorrow, but that could be tricky.

phryneas:
@ayushsharma82 I think this should do it, but I'll have to reread that tomorrow. Your test seems to pass now - can you experiment a bit more with that please?

ayushsharma82:
Sure, I'll add some more steps to that demo.

ayushsharma82:
Works! ( https://codesandbox.io/s/reduxjs-redux-toolkit-forked-qu9n6?file=/src/App.tsx ) Added a 'hide' button to simulate query being removed and added again before/after dispatching `resetApiState`.

dreyks:
Is this statement in docs still relevant or should it be removed?
https://github.com/reduxjs/redux-toolkit/blob/4afef4af288d1352975175de2f6dce98ce3759e4/docs/rtk-query/api/created-api/api-slice-utils.mdx#L371

PR #1737, opened 2021-11-13
chore(action-listener-middleware): add counter listener example

FaberVitale:
### Description

Adds a counter example that uses action-listener-middleware to periodically (setInterval) update values and to change theme color scheme.

### Why

- showcases action-listener-middleware features.
- adds a handly guinea pig app to test action-listener changes.

demo: https://githubbox.com/FaberVitale/redux-toolkit/tree/action-listener-middlware/add-counter-example/examples/action-listener/counter


FaberVitale:
@markerikson  I think this example,  inspired by the [saga counter example](https://github.com/redux-saga/redux-saga/tree/master/examples/counter), is mature enough to be merged.

demo: https://codesandbox.io/s/examples-action-listener-counter-82ext

I'm still not sure what's the best way to load listeners:
I've settled for this approach https://github.com/reduxjs/redux-toolkit/pull/1737/files#diff-e619b8395679d4c6ccf667624031f549b2ca764dd58948a146ab40d5354e0c3eR63

It is similar to the way I used to initialize sagas.

Feel free to suggest a better version.

---

### Sidenotes

We should either switch to CRA5 or migrate all the examples to Vite.

markerikson:
Yeah, atm I think the main possible approaches for setting up initial listeners are:

- export listeners or a setup function from a slice file, import those into app setup logic, add to middleware in app setup
- `export middleware` from `middleware.ts`, import middleware into slices, add in slices
- adding at runtime

Exporting listener entries from a slice directly doesn't entirely work right now, I think, because we don't export `createListenerEntry`.  Part of me wants to treat that as an internal implementation detail, but I could see how it might be useful.

Agreed that the examples probably would benefit from converting to Vite, but that's low priority right now.

Lemme get back to you on this example PR later in the holiday break.

FaberVitale:
### createListenerEntry

> Exporting listener entries from a slice directly doesn't entirely work right now, I think, because we don't export `createListenerEntry`. Part of me wants to treat that as an internal implementation detail, but I could see how it might be useful.

Looking at the code it seems that `createListenerEntry` is exported and I'd leave it as implementation detail.

---

### createSlice integration

> export middleware from middleware.ts, import middleware into slices, add in slices

Integrate listeners in `createSlice` would not be a bad idea.

I've sketched out an hypothetical integration

```ts
// counter.slice.ts
  name: 'counter',
  initialState: { value: 0 },
  reducers: {
    increment(state) {
      state.value += 1
    },
    decrement(state) {
      state.value -= 1
    },
  },
  listeners: {
    increment(action, listenerApi) {  // getState here has type `{ [counterSlice.name]: ReturnType<typeof counterSlice.reducer> }`
      console.log('new state', listenerApi.getState());
    }
  },
  extraReducers: {},
  extraListeners: {
    // object that maps strings to ListenerEntryOptions
  },
})

// ---------

// MyComp.component.ts

function MyComp() {
  useEffect(() => {
    counterSlice.addListeners(actionListenerMiddleware);

    return () => {  
      counterSlice.removeListeners(actionListenerMiddleware);
    }
  }, [])
}
```

---

### Simplify addListener config object

I'm wondering if we can find a way to simplify the config object passed to addListeners:
I'd like to drop `predicate` and leave just `actionCreator` and match.
Hopefully it does not create issues too many issues with TS types.

markerikson:
Hmm.  Adding a `listeners` field to `createSlice` is an interesting idea, but we should hold off on that for a while until we see the middleware getting some real adoption _and_ people say that they'd like a shorter way to define listeners.

We _need_ to keep the `predicate` option specifically because it's the only one that lets you check _state_ changes rather than just matching against specific actions:

https://github.com/reduxjs/redux-toolkit/discussions/1648#discussioncomment-1710420

Technically the `matcher` option is the one that's more redundant, because all `matcher` functions are actually `predicates` in terms of signature and behavior. But there may be enough of a difference in the types and action matching that they should be left as separate options.

markerikson:
Awright, let's go ahead and get this in. Thank you!

PR #1738, opened 2021-11-14
Codegen: don't drop the method on a query unless it is 'GET'

lindapaiste:
Using the `endpointOverrides` setting it is now possible to have a `query` with `method: 'POST'`.  If the query `method` is anything other than `'GET'`, it should be included in the generated API.

Perhaps the `isQuery` check is not useful at all anymore and it should just be `verb.toUpperCase() === 'GET'` instead of `isQuery && verb.toUpperCase() === 'GET'`?


phryneas:
Let's get this in, sorry for the delay.

PR #1748, opened 2021-11-16
fix(docs): fix the docs' example error.

LukerSpringtree:
action type is Object, doesn't have  'startsWith' and 'endsWith';
It will make an error like the picture below~
image


LukerSpringtree:
@phryneas 
hi~ How can I pass all checks~

phryneas:
I don't know, the code here looks good and the error is in another file. @markerikson could you take a look please?

LukerSpringtree:
@markerikson

PR #1751, opened 2021-11-16
export `RetryOptions` interface from `retry.ts`

colemars:
fixes re-exporting a typed `retry` wrapped `baseQuery` when `tsConfig.declaration = true`. 

Specifically resolves error:

> Exported variable '{retry_wrapped_basequery}' has or is using name 'StaggerOptions' from external module "*/node_modules/@reduxjs/toolkit/dist/query/retry" but cannot be named.

### DIFF
--- a/packages/toolkit/src/query/retry.ts
+++ b/packages/toolkit/src/query/retry.ts
@@ -23,7 +23,7 @@ async function defaultBackoff(attempt: number = 0, maxRetries: number = 5) {
   )
 }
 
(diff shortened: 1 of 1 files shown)


msutkowski (approved):
Good with me. @phryneas or @markerikson should this point against master or v1.7 integration?

phryneas:
Generally fine with me, but I'd suggest we rename it to `RetryOptions` before that.

PR #1763, opened 2021-11-23
fix issue where the global RTK object got overwritten in the UMD files

### LINKED ISSUE #1761: React umd files under dist overwrites each other
Hi!

I had some problem with importing the umd version of RTK and RTK Query at the same time from a cdn.
Both files overwrites the global RTK object with this line in the bundled file: 
`(global = typeof globalThis !== 'undefined' ? globalThis : global || self, factory(global.RTK = {}));`

This results in that the last file that is loaded will overwrite the first file.

Affected files:
* @reduxjs/toolkit/dist/rtk-query.umd.js
* @reduxjs/toolkit/dist/query/rtk-query-react.umd.js

The issue can be seen here: https://codepen.io/Antignote/pen/gOxJJGB

Is there any solution to this or is it a bug?

Antignote:
fix #1761 where the global variable RTK gets overwritten by the other entries in the UMD build

### DIFF
--- a/packages/toolkit/scripts/build.ts
+++ b/packages/toolkit/scripts/build.ts
@@ -97,18 +97,21 @@ const entryPoints: EntryPointOptions[] = [
     folder: '',
     entryPoint: 'src/index.ts',
     extractionConfig: 'api-extractor.json',
+    globalName: 'RTK',
   },
   {
     prefix: 'rtk-query',
     folder: 'query',
     entryPoint: 'src/query/index.ts',
     extractionConfig: 'api-extractor.query.json',
+    globalName: 'RTK.QUERY',
   },
   {
     prefix: 'rtk-query-react',
     folder: 'query/react',
     entryPoint: 'src/query/react/index.ts',
     extractionConfig: 'api-extractor.query-react.json',
+    globalName: 'RTK.QUERY.REACT',
   },
 ]
 
@@ -250,10 +253,7 @@ async function bundle(options: BuildOptions & EntryPointOptions) {
 /**
  * since esbuild doesn't support umd, we use rollup to convert esm to umd
  */
-async function buildUMD(outputPath: string, prefix: string) {
-  // All RTK UMD files share the same global variable name, regardless
-  const globalName = 'RTK'
-
+async function buildUMD(outputPath: string, prefix: string, globalName: string) {
   for (let umdExtension of ['umd', 'umd.min']) {
     const input = path.join(outputPath, `${prefix}.${umdExtension}.js`)
     const instance = await rollup.rollup({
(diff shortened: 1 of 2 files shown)


packages/toolkit/scripts/build.ts
```diff
@@ -97,18 +97,21 @@ const entryPoints: EntryPointOptions[] = [
...
     entryPoint: 'src/index.ts',
     extractionConfig: 'api-extractor.json',
+    globalName: 'RTK',
   },
   {
     prefix: 'rtk-query',
     folder: 'query',
     entryPoint: 'src/query/index.ts',
     extractionConfig: 'api-extractor.query.json',
+    globalName: 'RTK.QUERY',
```
markerikson, line 107: :hand: Does this actually work? I think it would be safer/better to go with a `CONSTANT_CASE` name instead

packages/toolkit/scripts/build.ts
```diff
@@ -97,18 +97,21 @@ const entryPoints: EntryPointOptions[] = [
...
   },
   {
     prefix: 'rtk-query-react',
     folder: 'query/react',
     entryPoint: 'src/query/react/index.ts',
     extractionConfig: 'api-extractor.query-react.json',
+    globalName: 'RTK.QUERY.REACT',
```
markerikson, line 114: :hand: Let's use the same name for both RTKQ entries.  Users should only be loading one or the other anyway, and I think it's better to keep the same name either way.

markerikson (changes requested):

packages/toolkit/scripts/build.ts
Antignote: Yes this works, it will generate this line in the UMD file: `(global = typeof globalThis !== 'undefined' ? globalThis : global || self, factory((global.RTK = global.RTK || {}, global.RTK.QUERY = {})));`

Shall I keep it and just make the same change for RTKQ-React or do you want me to rename it as a new global object like `RTKQ`?

packages/toolkit/scripts/build.ts
Antignote: Yes, I'll fix this in the next commit and when we have settled where the entry will be placed in the global namespace.

packages/toolkit/scripts/build.ts
markerikson: Let's make both of these be `window.RTKQ`.

Antignote: I've pushed new code to the PR

packages/toolkit/scripts/build.ts
Antignote: this is fixed in the latest commit

markerikson:
Thanks!

PR #1785, opened 2021-11-30
Feat/add take to action listener middleware

FaberVitale:
Inspired by: https://github.com/reduxjs/redux-toolkit/discussions/1648#discussioncomment-1723650


markerikson:
Nice! Not how I would have written it myself, but this looks like it oughta work. I'll merge this, and then I want to go ahead and split the code up into separate files.

PR #1786, opened 2021-12-01
Override unwrap behavior for buildInitiateQuery, update tests

msutkowski:
This is an attempt at solving the issue where if you execute multiple lazy queries  and unwrap them, anything after the first would throw with a `ConditionError` as requests 2+ would be `condition: false` due to a request being in flight.

### DIFF
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -279,31 +279,26 @@ Features like automatic cache collection, automatic refetching etc. will not be 
 
         const { requestId, abort } = thunkResult
 
-        const aggregatePromise = Promise.all([
-          runningQueries[queryCacheKey],
-          thunkResult,
-        ]).then(() =>
-          (api.endpoints[endpointName] as ApiEndpointQuery<any, any>).select(
-            arg
-          )(getState())
-        )
-
         const statePromise: QueryActionCreatorResult<any> = Object.assign(
-          aggregatePromise,
+          Promise.all([runningQueries[queryCacheKey], thunkResult]).then(() =>
+            (api.endpoints[endpointName] as ApiEndpointQuery<any, any>).select(
+              arg
+            )(getState())
+          ),
           {
             arg,
             requestId,
             subscriptionOptions,
             queryCacheKey,
             abort,
-            unwrap() {
-              return aggregatePromise.then<any>((result) => {
-                if (result.isError) {
-                  throw result.error
-                }
+            unwrap: async function () {
+              const result = await statePromise
 
-                return result.data
-              })
+              if (result.isError) {
+                throw result.error
+              }
+
+              return result.data
             },
             refetch() {
               dispatch(
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -276,20 +276,35 @@ Features like automatic cache collection, automatic refetching etc. will not be 
         })
         const thunkResult = dispatch(thunk)
         middlewareWarning(getState)
-        const { requestId, abort, unwrap } = thunkResult
+
+        const { requestId, abort } = thunkResult
+
+        const aggregatePromise = Promise.all([
+          runningQueries[queryCacheKey],
+          thunkResult,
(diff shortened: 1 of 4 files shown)


packages/toolkit/src/query/core/buildInitiate.ts
```diff
@@ -276,20 +276,35 @@ Features like automatic cache collection, automatic refetching etc. will not be
...
-            unwrap,
+            unwrap() {
+              return aggregatePromise.then<any>((result) => {
+                if (result.isError) {
+                  throw result.error
+                }
+
+                return result.data
+              })
+            },
```
msutkowski, line 307: The only thing that I think that could be interesting _(or confusing?)_ here is what happens regarding `unwrap` and `abort`. If abort is called, the initial request will be aborted, but a consumer will still be able to unwrap this. At the same time, I wouldn't want the abort to ignore the aggregate result, if that makes sense. I don't think there is any real issue with this implementation, but it _could be_ confusing. I don't have any better ideas or solutions for this though.

packages/toolkit/src/query/core/buildInitiate.ts
```diff
@@ -276,20 +276,35 @@ Features like automatic cache collection, automatic refetching etc. will not be
...
+          )(getState())
+        )
+
         const statePromise: QueryActionCreatorResult<any> = Object.assign(
-          Promise.all([runningQueries[queryCacheKey], thunkResult]).then(() =>
-            (api.endpoints[endpointName] as ApiEndpointQuery<any, any>).select(
-              arg
-            )(getState())
-          ),
+          aggregatePromise,
```
phryneas, line 292: It should work just fine without these changes.

packages/toolkit/src/query/core/buildInitiate.ts
phryneas: `.abort` of the originally running promise should set the query into an error state, so I think this would result in an error here, which seems reasonable.

You bring up another problem though: `.abort()` on this would do nothing if it just piggy-backs on another promise that started earlier. But otoh, I guess that's okay, too - you would not want to abort something else by accident I guess?

phryneas: Can we make this function `async`? The .then usage here irritates me for some reason :sweat_smile:

packages/toolkit/src/query/tests/buildHooks.test.tsx
```diff
@@ -915,6 +922,111 @@ describe('hooks tests', () => {
       expect(screen.queryByText(/An error has occurred/i)).toBeNull()
       expect(screen.queryByText('Request was aborted')).toBeNull()
     })
+
+    test('useLazyQuery does not throw on ConditionError and instead returns the aggregate error', async () => {
```
phryneas, line 926:
```suggestion
    test('unwrapping the useLazyQuery trigger result does not throw on ConditionError and instead returns the aggregate error', async () => {
```

phryneas (approved):

packages/toolkit/src/query/core/buildInitiate.ts
```diff
@@ -289,7 +291,15 @@ Features like automatic cache collection, automatic refetching etc. will not be
             subscriptionOptions,
             queryCacheKey,
             abort,
-            unwrap,
+            unwrap: async function () {
```
phryneas, line 294:
```suggestion
            async unwrap() {
```

packages/toolkit/src/query/tests/buildHooks.test.tsx
```diff
@@ -915,6 +922,111 @@ describe('hooks tests', () => {
...
+      }
+
+      render(<User />, { wrapper: storeRef.wrapper })
+
+      const fetchButton = screen.getByRole('button', { name: 'Fetch User' })
+      fireEvent.click(fetchButton)
+      fireEvent.click(fetchButton) // This technically dispatches a ConditionError, but we don't want to see that here. We want the real result to resolve and ignore the error.
+
+      await waitFor(() => {
+        const dataResult = screen.getByTestId('error')?.textContent
```
SaulMoro, line 1014: should be
```ts
const dataResult = screen.getByTestId('result')?.textContent
```

PR #1792, opened 2021-12-02
Add a "jobs" system to the listener middleware

markerikson:
This is very experimental stuff atm - I'm just pushing so the code is visible and I get some CI runs.

The idea here is to give the listener middleware additional capabilities that better match the feature set of sagas: cancelation of previous tasks for a a given listener, and the ability to launch child tasks.  This would enable a large variety of use cases, like implementing the equivalent of `takeLatest()`, `fork()`, etc.

For reference, I asked about how people use those APIs here: https://twitter.com/acemarke/status/1465017211265994757

I've copied in the source code from https://github.com/ethossoftworks/job-ts , which is a very neat little lib that has a `Job` class that already implements parent/child hierarchical tasking and cancelation.

The first step was wrapping the call to the listener function in a new `Job` instance.  This gives us a `JobHandle` that we can pass in as part of the `listenerApi`.  Additionally, the `Job` library uses https://github.com/ethossoftworks/outcome-ts as a functional-`Option`-style handler for success/error results.  I used that to consolidate the error handling for listeners, although I did drop the "sync/async" aspect to the error reporting.

Conceptually, I then want the `condition/take` functions to throw `JobCancelationError`s if the current listener is canceled, and that's where my brain halted for the night.  I was starting to try to restructure `createTakePattern`, but am realizing that A) the parent job doesn't give me a notice when it's canceled, and B) the call to `addListener()` wouldn't ever give me access to the nested `listenerApi.job` because the point is that cancelation happens before its listener ever runs.

So, a couple thoughts:

- I could try adding an `AbortController` to the `Job` class so that you could listen for that
- I could try rewriting `take/condition` to be based around the `Job.pause()` function instead

The latter approach is appealing, because there's already a `Job.delay()` function that has a timeout, it just doesn't return anything. It would be trivial to do the same thing but with a return value.

I really feel like I'm onto something here, but it's late and the pieces just aren't _quite_ coming together in my head.

### DIFF
--- a/packages/action-listener-middleware/src/index.ts
+++ b/packages/action-listener-middleware/src/index.ts
@@ -114,8 +114,12 @@ export const createListenerEntry: TypedCreateListenerEntry<unknown> = (
     predicate = options.actionCreator.match
   } else if ('matcher' in options) {
     predicate = options.matcher
-  } else {
+  } else if ('predicate' in options) {
     predicate = options.predicate
+  } else {
+    throw new Error(
+      'Creating a listener requires one of the known fields for matching against actions'
+    )
   }
 
   const id = nanoid()
--- a/packages/action-listener-middleware/src/index.ts
+++ b/packages/action-listener-middleware/src/index.ts
@@ -30,6 +30,15 @@ import type {
   ListenerErrorInfo,
 } from './types'
 
+import {
+  Job,
+  SupervisorJob,
+  JobHandle,
+  JobCancellationReason,
+  JobCancellationException,
+} from './job'
+import { Outcome } from './outcome'
+
 export type {
   ActionListener,
   ActionListenerMiddleware,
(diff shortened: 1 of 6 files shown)


FaberVitale:
https://github.com/reduxjs/redux-toolkit/pull/1799

markerikson:
Update: As of merging #1799 , we now have `take` and `condition` functions that appear to correctly re-throw `JobCancellationException` errors when the containing listener function gets canceled.  That should enable the equivalent of `takeLatest`, by calling `listenerApi.cancelPrevious()` at the start of a listener.  (This does assume that the previous instances are paused waiting for `take/condition` or the `listenerApi.job.*` methods that pause).

It's late and I want to do some more polishing before I publish a 0.4.0 (including updating the README), but this is looking _very_ encouraging so far.

markerikson:
Just added a test file demonstrating equivalence with multiple Redux-Saga operators:

```ts
 PASS  src/tests/effectScenarios.test.ts
  Saga-style Effects Scenarios
    √ throttle (51 ms)
    √ debounce / takeLatest (60 ms)
    √ takeEvery (24 ms)
    √ takeLeading (106 ms)
    √ fork + join (22 ms)
    √ fork + cancel (33 ms)
    √ canceled (69 ms)
```

I think that's sufficient evidence that this branch is worth merging and releasing.

PR #1799, opened 2021-12-02
feature/middleware-jobs -> implement take using jobs

FaberVitale:
I have played a bit with an implementation of take that uses the Job API that passes that skipped test.

I'm not sure that I have covered all the edge cases but it seems solid so far.


markerikson:
I just reworked this to correctly handle throwing exceptions in `take` when the parent job is canceled (I think), and filled out the "cancel previous jobs" test. Looking good so far - thank you!

PR #1800, opened 2021-12-03
Fix maxRetries documentation

SebastienGllmt:
The parameters `maxRetries` is not actually a maximum number of retries. It's just a maximum value on the exponential term.

Ideally the variable name would be updated because it's a very misleading name, but at least we should update the comment

### DIFF
--- a/packages/toolkit/src/query/retry.ts
+++ b/packages/toolkit/src/query/retry.ts
@@ -12,7 +12,7 @@ import { HandledError } from './HandledError'
  * 5. 9600ms * random(0.4, 1.4)
  *
  * @param attempt - Current attempt
- * @param maxRetries - Maximum number of retries
+ * @param maxRetries - Maximum value of the exponential
  */
 async function defaultBackoff(attempt: number = 0, maxRetries: number = 5) {
   const attempts = Math.min(attempt, maxRetries)


phryneas:
I'm not sure what you are implying here - with `maxRetries: 5`, the query will run a maximum 1 + 5 times and then fail.

SebastienGllmt:
What the variable does in the context of `retryWithBackoff` (where it acts as a max # of retries) is different from what it does in the context of the backoff function (where it is simply a clamp on the exponential used) which is why I only changed the comment in the backoff function documentation.

However, I do think that more generally probably a new optional parameter called `clamp` should be added to `retry` and, if it isn't set, the clamp use for the backoff will be `maxRetries`

phryneas:
Hm, I agree that something like a `maximumStagger` or something like that might be useful - but I'm kinda on the edge about the change suggested here.

As the code stands here, `maxRetries` never plays any role as it will never be called with `attempt > maxRetries` in the first place.

Also, the function itself is not public. We are only using it's `@remark` docblock for the documentation and that is all it is externally. 
And right now it is more in as a blueprint for copy-pasting that function. And another function would inherit the signature and could internally be doing something completely different, so in that case the current comment would be correct, while mentioning an "exponential" would not play any role in most copy-pasted functions.

PR #1801, opened 2021-12-03
Export StaggerOptions type

SebastienGllmt:
If you want to make a function that only accepts a base query that has retry functionality supported, you would do this by having the function take in a `BaseQueryFn` that has `StaggerOptions` as part of its `DefinitionExtraOptions`, but this can't be done because the `StaggerOptions` isn't exported


phryneas:
This is already done in https://github.com/reduxjs/redux-toolkit/pull/1751, so it should already be available in [1.7.0-rc.0](https://github.com/reduxjs/redux-toolkit/releases/tag/v1.7.0-rc.0).

But thanks for the PR, really appreciated :)

PR #1804, opened 2021-12-03
update naming references in docs

connected-tcaceres:
the actual api references are 'providesTags' and 'invalidatesTags' per docs:
https://redux-toolkit.js.org/rtk-query/api/createApi#providestags
https://redux-toolkit.js.org/rtk-query/api/createApi#invalidatestags


phryneas:
Good catch, thanks!

PR #1807, opened 2021-12-05
Add listener middleware Job class test file and initial usage test

markerikson:
This PR:

- Updates `Job.pause()` to include the existing `_cancelPromise` so that it bails out early if cancelation happens before the actual source promise resolves
- Copies in the `job.test.ts` file from the original `job` library (and updates the assertions to work with Jest)
- Adds the first of hopefully several usage scenario examples: a controllable long-polling loop based off of one I wrote years ago with sagas ( https://gist.github.com/markerikson/5203e71a69fa9dff203c9e27c3d84154 )

### DIFF
--- a/packages/action-listener-middleware/README.md
+++ b/packages/action-listener-middleware/README.md
@@ -59,7 +59,7 @@ The Redux community has settled around three primary side effects libraries over
 
 - Thunks use basic functions passed to `dispatch`. They let users run arbitrary logic, including dispatching actions and getting state. These are mostly used for basic AJAX requests and logic that needs to read from state before dispatching actions
 - Sagas use generator functions and a custom set of "effects" APIs, which are then executed by a middleware. Sagas let users write powerful async logic and workflows that can respond to any dispatched action, including "background thread"-type behavior like infinite loops and cancelation.
-- Observables use RxJS obesrvable operators. Observables form pipelines that do arbitrary processing similar to sagas, but with a more functional API style.
+- Observables use RxJS observable operators. Observables form pipelines that do arbitrary processing similar to sagas, but with a more functional API style.
 
 All three of those have strengths and weaknesses:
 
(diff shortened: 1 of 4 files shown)


FaberVitale:
I read your tweet about dropping jobs and using AbortController instead.

I think it is doable moreover I think we could play a bit with async generators:
complex logic can be achieved using them.

---

I believe we should drop `when` too:
redux-saga and redux-observable run after the reducers for good reasons.
I cannot think of any use case that when fulfills especially because we already provide `getOriginalState`.

We are just increasing the cognitive load for no apparent benefit.
I have personally worked with redux-saga and redux-observable in production apps and I find the "when" option more confusing than helpful.

I'm going to write a RFC in the other thread and a PR to drop this option.
what do you think?

markerikson:
Hmm. I may not be fully understanding what you're suggesting about "async generators". I would prefer that this middleware _not_ use generator functions, specifically because that's basically just reimplementing `redux-saga` at that point.  (Obviously dealing with cancelation would be a whole lot easier if we _were _using generators, but the point is this is supposed to be a simpler alternative.)  Are you saying that using async generators _inside_ the listener could maybe act as a replacement for "jobs"?  If so, what would that look like?

I think Lenz's concerns with jobs atm are a combination of bundle size and API size.  Right now, `dist/esm/index.js` is about 6K minified.  Half of that is the `Job` class, and about 2.5K is the middleware itself.  I think we could maybe shave off 1-2K if we dropped `Job` and used a simpler implementation based around `AbortController` for the `take/condition/cancelPrevious` behavior, but I'm not sure what the final savings would be.

I agree that there probably aren't as many reasons to run the listeners before the dispatched action.  It would be a good idea to go review #237 and #547 to see why we decided to add that in the first place.  I _think_ part of it was because Lenz originally had that `listenerApi.preventDefault()` function that could be used to cancel the action from reaching the reducers, but we removed that on the grounds that it was also confusing.

If you can do some research on why we added `when` and some brainstorming as to whether there are still any use cases where we might want to run the listeners `'beforeReducer'`, that would be very helpful.

Also please feel free to try out something with `AbortControllers` that drops the `Job` class. I'm not saying we'll merge it right away :)  but it would be good to compare implementations.

I did see an interesting implementation at https://gist.github.com/andrewcourtice/ef1b8f14935b409cfe94901558ba5594 that looked like it might be a good starting point for something like that.

PR #1818, opened 2021-12-10
Fix incorrect RTKQ endpoint definition types for correct selector typings

### LINKED ISSUE #1817: [v1.7.0-rc.0] Typing queries causes type issues with selectors
I upgraded my app to v1.7.0-rc.0 and had selectors which were previously working gain type issues. It looks like selectors which are created based on the endpoints' selectors type state as never, which causes selectors composed from them to error in typescript due to lack of overlapping types for state.

Here is a sandbox which reproduces the issue: https://codesandbox.io/s/rtk-query-selector-typing-issue-lzs3r?file=/src/api.ts

markerikson:
This PR:

- Fixes a bug in the `ReducerPath` type that was extracting the wrong generic from `EndpointDefinitions`, causing a cascade that resulted in typed Reselect selectors getting a `state: {}` parameter
- Fixes what _appeared_ to be a bug in `optimisticUpdates.test.ts` and additionally works around some odd behavior with mocking `baseQuery()` that was resulting in extra console errors

Fixes #1817 .

As best as I can tell, the incorrect `ReducerPath` type was there since at least 1.6.0. I think we always had a bug here, but the use of Reselect 4.0 with its looser types must have masked it.  When we updated to Reselect 4.1.x with improved type inference, we uncovered this bug.

### DIFF
--- a/packages/toolkit/src/query/endpointDefinitions.ts
+++ b/packages/toolkit/src/query/endpointDefinitions.ts
@@ -464,8 +464,9 @@ export type QueryArgFrom<D extends BaseEndpointDefinition<any, any, any>> =
 export type ResultTypeFrom<D extends BaseEndpointDefinition<any, any, any>> =
   D extends BaseEndpointDefinition<any, any, infer RT> ? RT : unknown
 
-export type ReducerPathFrom<D extends EndpointDefinition<any, any, any, any>> =
-  D extends EndpointDefinition<any, any, any, infer RP> ? RP : unknown
+export type ReducerPathFrom<
+  D extends EndpointDefinition<any, any, any, any, any>
+> = D extends EndpointDefinition<any, any, any, any, infer RP> ? RP : unknown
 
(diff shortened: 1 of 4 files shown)


packages/toolkit/src/query/endpointDefinitions.ts
```diff
@@ -464,8 +464,9 @@ export type QueryArgFrom<D extends BaseEndpointDefinition<any, any, any>> =
 export type ResultTypeFrom<D extends BaseEndpointDefinition<any, any, any>> =
   D extends BaseEndpointDefinition<any, any, infer RT> ? RT : unknown
 
-export type ReducerPathFrom<D extends EndpointDefinition<any, any, any, any>> =
-  D extends EndpointDefinition<any, any, any, infer RP> ? RP : unknown
+export type ReducerPathFrom<
```
phryneas, line 467: Daaamn, good find!

PR #1823, opened 2021-12-11
feat: implement saga-style effects using abortController

FaberVitale:
This PR adds an implementation of action-listener-middleware that drops Job in favor of [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController):

This change reduces the bundle size compared to the [current version](https://github.com/reduxjs/redux-toolkit/pull/1807#issuecomment-988195750).

### Build

```bash
$ yarn workspace @rtk-incubator/action-listener-middleware run build
Build "actionListenerMiddleware" to dist/esm:
       1703 B: index.modern.js.gz
       1519 B: index.modern.js.br
Build "actionListenerMiddleware" to dist/module:
      2.33 kB: index.js.gz
      2.09 kB: index.js.br
Build "actionListenerMiddleware" to dist/cjs:
      2.32 kB: index.js.gz
      2.09 kB: index.js.br
```

### Tests

```bash
$ yarn workspace @rtk-incubator/action-listener-middleware run test
 PASS  src/tests/effectScenarios.test.ts
 PASS  src/tests/listenerMiddleware.test.ts

Test Suites: 2 passed, 2 total
Tests:       45 passed, 45 total
Snapshots:   0 total
Time:        4.499 s, estimated 5 s
Ran all test suites.
```

### Bonus

I've originally experimented with a `Task` object but that implementation wasn't compatible with the requirements.
I've left  [code + tests in a gist](https://gist.github.com/FaberVitale/0ea2adf79ac8854042808137308a1441) if there's someone interested.


markerikson:
Oooo, fascinating! I'm having a "sit on the couch and play games" day today, but I'll try to look at this tomorrow!

packages/action-listener-middleware/src/index.ts
```diff
@@ -65,73 +67,109 @@ function assertFunction(
...
+        try {
+          assertActive(signal)
+        } catch (err) {
+          reject(err)
+        }
+        resolve()
+      }, timeoutMs)
+    })
+  }
+}
```
phryneas, line 91:
```suggestion
async function assertActive(signal: AbortSignal, reason?: string) {
  if (signal.aborted) {
    throw new TaskAbortError(reason)
  }
}

function createDelay(signal: AbortSignal) {
  return async function delay(timeoutMs: number): Promise<void> {
    await assertActive(signal)
    await new Promise((resolve) =>resolve, timeoutMs)
    await assertActive(signal)
  }
}
```

Might be worth exploring to write `assertActive` as an `async` function - at least `createDelay` would become significantly shorter. Not sure about the more complex functions though.

phryneas: Otoh... it wouldn't even need to be async for that change in `createDelay` 🤔

phryneas: I'm microoptimizing again.. 😅  this is looking very good. I'm not really awake enough to perfectly understand everything yet (I'm not on top with the full PR). but what I understand goes into a great direction 👍

FaberVitale: > I'm microoptimizing again.. 😅 this is looking very good. I'm not really awake enough to perfectly understand everything yet (I'm not on top with the full PR). but what I understand goes into a great direction 👍

Done 👍

FaberVitale:
Aligned PR with master.

I've skipped the test defined in usesCases.test because `fork` does currently not support a pause-like functionality and
I'm not convinced it should.

Potential changes to evaluate:

- encapsulate abortController in forkedTask.
- pass a `taskExecutorAPI` object as the first argument of the fork executor: 
  it could contain an AbortSignal and other utils.

markerikson:
Skimming the PR, a few thoughts:

- Looks like the `index.modern.js` ESM file dropped from 6K to 4K min, which is good
- I see a few places we could shave a few more bytes, although that's a final optimization step once we're actually settled on the API
- Is it worth still keeping that `Outcome` class around? Looks like the only use now is `Outcome.try()`, and it's getting exposed via `forkedTask.promise`. The listed use cases in the `@ethossoftware/outcome` repo are valid, but if we're not going to make much use of it we probably ought to just drop it
- I _did_ think that `pause` was useful, actually - the ability to wait for any promise in a cancelation-aware way seems important. Can we add that back in?
- Can we add some more tests around the forking behavior/API?

FaberVitale:
Changes

* forked task executor now receives an API object: `{ pause, delay, signal }`.
* added listener.fork tests.
* re-enabled 'Long polling loop' test
* removed Outcome; it has been replaced by an ADT.
* micro optimizations to keep the size at 4.0K.
* added `pause` to listenerApi.
* encapsulated abortController in `fork()` output.
* `listenerApi.delay` now rejects as soon as the listener is cancelled.

packages/action-listener-middleware/src/index.ts
```diff
@@ -334,6 +351,64 @@ export function createActionListenerMiddleware<
...
+          currentPhase,
+          extra,
+          signal: internalTaskController.signal,
+          fork,
+          unsubscribe: entry.unsubscribe,
+          subscribe: () => {
+            listenerMap.set(entry.id, entry)
+          },
+          cancelPrevious: () => {
+            entry.pendingSet.forEach((controller, _, set) => {
```
markerikson, line 390: :hand: I actually have had a concern with `cancelPrevious()` for a while now.

As implemented, it's really not "cancel _previous_", it's "cancel _all other instances regardless of which came first".  Which is a valid thing to do, but it's not semantically the same :)

It almost makes me wonder if we should have separate `cancelPrevious()` and `cancelOthers()` functions?  Or maybe rename this?

markerikson (approved):

markerikson:
This looks great overall!  I'm going to merge it, make a couple more tweaks as a new PR, and then put out 0.5.0.

PR #1836, opened 2021-12-16
bugfxing on useState

yanCode:
Note:  this only a bug-fixing on examples. 

In `mutation` example,  This bug can be reproduced by enable the edit mode of a  post, then shift the post in the post list, you can find the name of the post in edit mode is not synced at all.

the reason is,  it uses the value from **props** as the initial value of `setState` hook, however,  the  value of **state** cannot be  re-initialized if the **props** changes.  For details,  please refer to https://stackoverflow.com/questions/54865764/react-usestate-does-not-reload-state-from-props

### DIFF
--- a/examples/query/react/mutations/src/features/posts/PostDetail.tsx
+++ b/examples/query/react/mutations/src/features/posts/PostDetail.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useState, useEffect } from 'react'
 import { useHistory, useParams } from 'react-router-dom'
 import {
   useDeletePostMutation,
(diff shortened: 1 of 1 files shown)


examples/query/react/mutations/src/features/posts/PostDetail.tsx
```diff
@@ -29,7 +29,10 @@ const EditablePostName = ({
   onCancel: () => void
   isLoading?: boolean
 }) => {
-  const [name, setName] = useState(initialName)
```
phryneas, line 32: even with adding the `useEffect`, this line is still required I think?

yanCode: @phryneas , yes, that line is required, that's my bad.

Shrugsy:
I would argue that this is better handled by tying the `EditablePostName` component to the distinct post, rather than hacking a `useEffect` on like that.

```diff
  <EditablePostName
+ key={post.id}
```

The usage of `useEffect` obscures the intent of the design a lot. Constantly re-syncing `'initialName'` to `'name'` leaves one wondering why the state would be there in the first place rather than the parent, whereas tying the component to a specific post retains the original spirit without obscurity

phryneas:
Closing due to lack of activity.

PR #1860, opened 2021-12-22
Fix #1853 - `updateMany` on sorted entity adapter

### LINKED ISSUE #1853: Using updateMany in adapter with custom sortComparer ignores several updates with same id
When I use updateMany in an adapter with a custom sortComparer, I only see effects of the first update when sending several with the same id. See reproducer [here](https://codesandbox.io/s/exehh?file=/src/App.tsx).

jakeboone02:
Fixes #1853.

@markerikson stated that the preferred outcome would be to combine the sorted and unsorted logic into a single implementation, but I haven't done that yet.

So, feel free to merge this PR as is and I can combine the sorted/unsorted logic later, or let me know if you just want to wait until I do the combining in this PR. In either case I probably won't get to that until after the holidays.

### DIFF
--- a/packages/toolkit/src/entities/sorted_state_adapter.ts
+++ b/packages/toolkit/src/entities/sorted_state_adapter.ts
@@ -71,32 +71,40 @@ export function createSortedStateAdapter<T>(
     return updateManyMutably([update], state)
   }
 
-  // eslint-disable-next-line @typescript-eslint/prefer-readonly-parameter-types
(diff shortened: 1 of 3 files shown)


markerikson:
Yeah, this is a lower-priority fix given that there are workarounds in userland (ie, combine all your own updates first).  No rush - I'd like to take this chance to consolidate this code.

PR #1863, opened 2021-12-24
Fix named export used as default export in example

siwalikm:
Problem: In the rtk-query tutorial ([link](https://redux-toolkit.js.org/tutorials/rtk-query#wrap-your-application-with-the-provider)) in store.ts, we have a names export called store. In our entry file, we are importing it as default export. Probably a typo.

Fix: Updated the import statement.


markerikson:
Thanks!

PR #1871, opened 2021-12-29
📝 Add RTKQ typescript error handling section

Shrugsy:
As per title - provides explicit guidance & examples on error handling with RTKQ + typescript (although more specifically; error type narrowing)

Note that I *didn't* show usage of inline handling *without* unwrap, mainly just to try not to make it appear too overwhelming

### DIFF
--- a/docs/rtk-query/usage-with-typescript.mdx
+++ b/docs/rtk-query/usage-with-typescript.mdx
@@ -497,3 +497,184 @@ function MaybePost({ id }: { id?: number }) {
   return <div>...</div>
 }
 ```
+
+## Type safe error handling
+
+When an error is gracefully provided from a [`base query`](./api/createApi.mdx#baseQuery), RTK query will provide the error
+directly. If an unexpected error is thrown by user code rather than a graceful handled error,
+that error will be transformed into a `SerializedError` shape. Users should ensure that they are checking which
+kind of error they are dealing with prior to attempting to access it's properties. This can be done in a type safe manner either
+by using a type guard, e.g. by checking for [discriminated properties](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing),
+or using a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).
+
+When using [`fetchBaseQuery`](./api/fetchBaseQuery.mdx), as your base query,
+errors will be of type `FetchBaseQueryError | SerializedError`. The specific shapes of those types can be seen below.
+
+```ts title="FetchBaseQueryError type"
+export type FetchBaseQueryError =
+  | {
+      /**
+       * * `number`:
+       *   HTTP status code
+       */
+      status: number
+      data: unknown
+    }
+  | {
+      /**
+       * * `"FETCH_ERROR"`:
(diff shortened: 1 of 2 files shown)


docs/rtk-query/usage-with-typescript.mdx
```diff
@@ -497,3 +497,184 @@ function MaybePost({ id }: { id?: number }) {
   return <div>...</div>
 }
 ```
+
+## Type safe error handling
+
+When an error is gracefully provided from a [`base query`](./api/createApi.mdx#baseQuery), RTK query will provide the error
+directly. If an unexpected error is thrown by user code rather than a graceful handled error,
```
msutkowski, line 504:
```suggestion
directly. If an unexpected error is thrown by user code rather than a handled error,
```

docs/rtk-query/usage-with-typescript.mdx
```diff
@@ -497,3 +497,184 @@ function MaybePost({ id }: { id?: number }) {
...
+that error will be transformed into a `SerializedError` shape. Users should ensure that they are checking which
+kind of error they are dealing with prior to attempting to access it's properties. This can be done in a type safe manner either
```
msutkowski, line 506:
```suggestion
that error will be transformed into a `SerializedError` shape. Users should make sure that they are checking which kind of error they are dealing with before attempting to access its properties. This can be done in a type safe manner either
```

msutkowski (commented):
Overall content looks good to me! I think this is explicit enough to show that folks can make & compose their own guards for their specific API error responses 🎉

docs/rtk-query/usage-with-typescript.mdx
```diff
@@ -497,3 +497,184 @@ function MaybePost({ id }: { id?: number }) {
...
+       **/
+      status: 'PARSING_ERROR'
+      originalStatus: number
+      data: string
+      error: string
+    }
+  | {
+      /**
+       * * `"CUSTOM_ERROR"`:
+       *   A custom error type that you can return from your `fetchFn` where another error might not make sense.
```
phryneas, line 547:
```suggestion
       *   A custom error type that you can return from your `queryFn` where another error might not make sense.
```
Uh-oh. I think I messed that up :sweat_smile: 
Could you please also update that in the comment it came from?

docs/rtk-query/usage-with-typescript.mdx
```diff
@@ -497,3 +497,184 @@ function MaybePost({ id }: { id?: number }) {
...
+      return (
+        <div>
+          <div>An error has occurred:</div>
+          <div>{errMsg}</div>
+        </div>
+      )
+    }
+
+    // you can access all properties of `SerializedError` here
+    return <div>{error.message}</div>
```
phryneas, line 593:
```suggestion
    else {
        // you can access all properties of `SerializedError` here
        return <div>{error.message}</div>
    }
```
technically not necessary, but I'd add it for clarity - people might skim over the `return` above.

phryneas (changes requested):
Minor change requests, but overall this is a great addition :blush: :+1:

docs/rtk-query/usage-with-typescript.mdx
Shrugsy: done & done

phryneas:
Let's get this in :)

Thanks a lot!

Issue #1861, opened 2021-12-22
A case where `action` is not typed correctly when using addMatcher 


srmagura:
This is not a serious issue, but it can cause type safety issues if you are not paying attention. Feel free to close if you think it is trivial.

Minimal repro: https://github.com/srmagura/rtk-add-matcher-repro

Code:

```ts
import { createAction, createReducer } from "@reduxjs/toolkit";

interface SpecialAction {
  // Uncomment the next line, and then `action` will be typed correctly
  // type: string;
  payload: number;
}

function isSpecialAction(action: any): action is SpecialAction {
  return action.type.startsWith("special/");
}

const myReducer = createReducer(0, (builder) => {
  builder.addMatcher(isSpecialAction, (state, action) => {
    // Hover over `action` to see that its type is `AnyAction`, not `SpecialAction`
    return action.payload;
  });
});
```

phryneas:
Hmm yeah, because the type of the type guard does not contain a `type` field the overload does not work. We should probably widen the requirement there a bit to all type guards, not just specific ones. 
Good catch!

I'll mark this as help wanted & good first issue since I myself won't be able to get to this before the middle of january - maybe someone can already take over :)

markerikson:
Should be out in 1.7.2, I think.

PR #1895, opened 2022-01-05
Fix `addMatcher` typings

crcarrick:
References #1861

Previously, the `action` argument passed down to the `reducer` function of `addMatcher(matcher, reducer)` would be incorrectly typed as `AnyAction` when the type predicate being tested by the `matcher` function didn't include a `type` property.

This PR attempts to resolve the issue:

- Widens the type argument of `ActionMatcher` from `A extends AnyAction` to `A`.
- Maintains the previous behavior of defaulting the type of the `action` argument to `AnyAction` when there is no type predicate.
- Intersects the `A` type argument with `AnyAction` to satisfy the second type argument to `CaseReducer` and offer better completion inside the reducer function.
- Adds a type test to the mapBuilders.typetest.ts file that verifies that the `action` argument is now correctly typed.

> "If you changed external-facing types, make sure to also build the project locally and include the updated API report file etc/redux-toolkit.api.md in your pull request."

I was unable to accomplish this. I ran the monorepo `build` script and the @reduxjs/toolkit `build` scripts but this file was never updated. All of the `build:*` scripts in the @reduxjs/toolkit workspace's package.json file include the `--skipExtraction` flag which seems like it would prevent the extraction tool from running. `addMatcher()` and `ActionMatcher` _are_ referenced in the etc/redux-toolkit.api.md file, so I think the file should be updated with this PR. If somebody could point me in the right direction on how to get the tooling to do that, I would be grateful.

Thanks for taking the time to look at this and let me know what you all think!

### DIFF
--- a/packages/toolkit/src/createReducer.ts
+++ b/packages/toolkit/src/createReducer.ts
@@ -15,8 +15,11 @@ import type { NoInfer } from './tsHelpers'
  */
 export type Actions<T extends keyof any = string> = Record<T, Action>
 
-export interface ActionMatcher<A> {
-  (action: any): action is A
+/**
+ * @deprecated use `TypeGuard` instead
+ */
+export interface ActionMatcher<A extends AnyAction> {
+  (action: AnyAction): action is A
 }
 
 export type ActionMatcherDescription<S, A extends AnyAction> = {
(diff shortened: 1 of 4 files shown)


phryneas:
Tbh., I'm not even sure if the extraction step still works or if the CI actually uses it.

I've used this PR to add some minor refactoring (introducing a new `TypeGuard` type for this) and I've added a few more type assertions in the tests. Generally, this looks good for me.

phryneas:

PR #1902, opened 2022-01-09
fix(alm): prevent memory leaks caused by getOriginalState

FaberVitale:
This PR prevents memory leaks caused by `getOriginalState` and adds the following BREAKING CHANGE:
`getOriginalState` now throws error if it is called asynchronously.

See relevant discussion: https://github.com/reduxjs/redux-toolkit/discussions/1648#discussioncomment-1932820

### Build log

```bash
Build "actionListenerMiddleware" to dist/esm:
       1592 B: index.modern.js.gz
       1431 B: index.modern.js.br
Build "actionListenerMiddleware" to dist/module:
      2.25 kB: index.js.gz
      2.01 kB: index.js.br
Build "actionListenerMiddleware" to dist/cjs:
      2.24 kB: index.js.gz
         2 kB: index.js.br
```

### Bundle size

```bash
stat -f%z packages/action-listener-middleware/dist/esm/index.modern.js

3572
```

### Test log

```bash
﻿﻿PASS src/tests/listenerMiddleware.test.ts
PASS src/tests/effectScenarios.test.ts
PASS src/tests/fork.test.ts
PASS src/tests/useCases.test.ts

Test Suites: 4 passed, 4 total
Tests:       62 passed, 62 total
Snapshots:   0 total
Time:        1.93 s, estimated 2 s
Ran all test suites.
```

### DIFF
--- a/packages/action-listener-middleware/README.md
+++ b/packages/action-listener-middleware/README.md
@@ -210,8 +210,7 @@ The `listenerApi` object is the second argument to each listener callback. It co
 
 - `dispatch: Dispatch`: the standard `store.dispatch` method
 - `getState: () => State`: the standard `store.getState` method
-- `getOriginalState: () => State`: returns the store state as it existed when the action was originally dispatched, _before_ the reducers ran.
-  This method can only be called synchronously, it throws error otherwise.
+- `getOriginalState: () => State`: returns the store state as it existed when the action was originally dispatched, _before_ the reducers ran. (**Note**: this method can only be called synchronously, during the initial dispatch call stack, to avoid memory leaks. Calling it asynchronously will throw an error.)
 
 `dispatch` and `getState` are exactly the same as in a thunk. `getOriginalState` can be used to compare the original state before the listener was started.
 
--- a/packages/action-listener-middleware/README.md
+++ b/packages/action-listener-middleware/README.md
@@ -210,7 +210,8 @@ The `listenerApi` object is the second argument to each listener callback. It co
 
 - `dispatch: Dispatch`: the standard `store.dispatch` method
 - `getState: () => State`: the standard `store.getState` method
-- `getOriginalState: () => State`: returns the store state as it existed when the action was originally dispatched, _before_ the reducers ran
+- `getOriginalState: () => State`: returns the store state as it existed when the action was originally dispatched, _before_ the reducers ran.
+  This method can only be called synchronously, it throws error otherwise.
 
 `dispatch` and `getState` are exactly the same as in a thunk. `getOriginalState` can be used to compare the original state before the listener was started.
 
--- a/packages/action-listener-middleware/src/index.ts
+++ b/packages/action-listener-middleware/src/index.ts
@@ -31,7 +31,7 @@ import type {
   ForkedTaskExecutor,
   ForkedTask,
 } from './types'
-import { assertFunction, catchRejection, INTERNAL_NIL_TOKEN } from './utils'
+import { assertFunction, catchRejection } from './utils'
 import { TaskAbortError } from './exceptions'
 import {
   runTask,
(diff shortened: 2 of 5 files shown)


packages/action-listener-middleware/src/index.ts
```diff
@@ -432,33 +432,51 @@ export function createActionListenerMiddleware<
...
 
     // Need to get this state _before_ the reducer processes the action
-    const originalState = api.getState()
-    const getOriginalState = () => originalState
+    let originalState: S | typeof INTERNAL_NIL_TOKEN = api.getState()
+
+    // `getOriginalState` can only be called synchronously.
+    // @see https://github.com/reduxjs/redux-toolkit/discussions/1648#discussioncomment-1932820
+    const getOriginalState = (): S => {
+      if (originalState === INTERNAL_NIL_TOKEN) {
```
phryneas, line 440: `undefined` should really be enough here. No reducer is ever allowed to return `undefined`, so `getState()` cannot be `undefined` either.

FaberVitale: Maybe I'm missing something but I can't find an explicit invariant here: https://github.com/reduxjs/redux/blob/master/src/createStore.ts#L238

Where I can find it?

markerikson: Strictly speaking, I don't think we have such an invariant for the _root_ state.

`combineReducers` does have that check for _slice_ reducers, though:

https://github.com/reduxjs/redux/blob/793947f9a33d841ce91eb15dcaf3884d8aeb43b2/src/combineReducers.ts#L70-L75

FaberVitale: I'm aware and I'm sorry that's a really nitpicky  but you technically can use reducer: `() => undefined` with redux-toolkit.

DEMO: https://codesandbox.io/s/redux-store-reducer-undefined-mtner?file=/index.html

This demo shows that it is possible to end up in a `getOriginalState() === undefined` even when using  `configureStore` and `createSlice`.

phryneas: Huh, wasn't aware that that is combineReducers only. Then this seems necessary, yeah. Out of interest: does it go a few bytes down if the token is in the same file?

markerikson: Heh, this is actually a very similar issue to what I ran into with Reselect 4.1.x :)  Users occasionally wanted to have selectors that _do_ return `undefined`, so I had to use a similar `NOT_FOUND_TOKEN` approach internally.

FaberVitale: > Huh, wasn't aware that that is combineReducers only. Then this seems necessary, yeah. Out of interest: does it go a few bytes down if the token is in the same file?

It is slightly smaller `1589` vs `1592`.

Odd, as far as I can see `index.modern.js` does not preserve the original `esm` modules.

phryneas: It will create a new scope for the new file, a new variable to contain exports and inject that into the "module scope" of the other file. So it's a bit more code. But yeah, those are common to all modules so it makes sense that it gzips down to only 3 more bytes.

packages/action-listener-middleware/README.md
```diff
@@ -210,7 +210,8 @@ The `listenerApi` object is the second argument to each listener callback. It co
 
 - `dispatch: Dispatch`: the standard `store.dispatch` method
 - `getState: () => State`: the standard `store.getState` method
-- `getOriginalState: () => State`: returns the store state as it existed when the action was originally dispatched, _before_ the reducers ran
+- `getOriginalState: () => State`: returns the store state as it existed when the action was originally dispatched, _before_ the reducers ran.
+  This method can only be called synchronously, it throws error otherwise.
```
markerikson, line 214:
```suggestion
- `getOriginalState: () => State`: returns the store state as it existed when the action was originally dispatched, _before_ the reducers ran. (**Note**: this method can only be called synchronously, during the initial dispatch call stack, to avoid memory leaks. Calling it asynchronously will throw an error.)
```

PR #1903, opened 2022-01-09
Remove duplicate documentation block

ryanauj:
This looks like the same block of documentation as [lines 52-60](https://github.com/reduxjs/redux-toolkit/blob/master/docs/rtk-query/overview.md?plain=1#L52-L60). Would it make sense to only include the first instance?

### DIFF
--- a/docs/rtk-query/overview.md
+++ b/docs/rtk-query/overview.md
@@ -83,16 +83,6 @@ The functionality included in RTK Query quickly pays for the added bundle size, 
 
 ### Create an API Slice
 
-RTK Query is included within the installation of the core Redux Toolkit package. It is available via either of the two entry points below:
-
-```ts
-import { createApi } from '@reduxjs/toolkit/query'
-
-/* React-specific entry point that automatically generates
-   hooks corresponding to the defined endpoints */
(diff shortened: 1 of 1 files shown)


markerikson:
Eh... honestly if anything, I'd flip it. Remove the "These two entry points" + the code block in the "What's Included" section.

Although... that section is also making the point that it's part of RTK but a separate entry point.

Hmm. @phryneas , @msutkowski : any thoughts on the content here?

msutkowski:
> Eh... honestly if anything, I'd flip it. Remove the "These two entry points" + the code block in the "What's Included" section.
> 
> Although... that section is also trying to get across that it's part of RTK but a separate entry point, so having the import illustration is potentially useful.
> 
> Hmm. @phryneas , @msutkowski : any thoughts on the content here?

Agreed on removing the block from 'What's Included' if we change this at all. But...I don't think any change is really needed here as I'd rather things be explicit and redundant 😅 when it's not way overboard.

phryneas:
We're gonna keep this in for now - better have an example too many than not enough.

Thanks a lot for the PR!

Issue #1904, opened 2022-01-10
Cache invalidate based on headers


sriteja777:
### For RTK Query

Currently the cache key takes only the query args into consideration for deciding validation of cache.
We have an endpoint `getPosts` and in the headers of this GET request we are sending the JWT token of the user. The response is different based on whether the token is present or not.  We use `prepareHeaders` to set the token in requests instead of sending it as arg in every request.

    baseQuery: fetchBaseQuery({
      baseUrl: apiConfig.baseUrl,
      prepareHeaders: (headers, { getState }) => {
        // By default, if we have a token in the store, let's use that for authenticated requests
        const token = (getState() as RootState).user.token
        if (token) {
          headers.set(Constants.AUTH_HEADER, token)
        }  
        return headers
      },

 But with current setup the `getPosts` will not refetch data from server if the token in Header changes. Is there any way / workaround to achieve this?
 - One solution is to send the token as query arg, but token needs to be sent every where the posts are required. Is there any way to minimize this?

phryneas:
Hmm, if you can derive that from the response, you can add some kind of `"UNAUTHENTICATED"` tag to queries that were not authenticated.
At the moment we do not have `meta` available in `providesTags/invalidatesTag` at the moment, but it would probably be consistent with `transformResponse` to add it to these two functions too. Then you could look in there and look at the actual request & response including headers.

bever1337:
I just submitted a PR to pass `meta` arg to result description functions. This is my first redux commit, so the changes lean more towards "perfectly serviceable" than "wow, such beautiful code". Please let me know what adjustments I can make.

phryneas:
The `meta` argument has been released in https://github.com/reduxjs/redux-toolkit/releases/tag/v1.7.2

This should be possible now.

PR #1910, opened 2022-01-12
feat(meta): Passes meta to result description functions [#1904]

bever1337:
The internal and public interfaces for result descriptions are getting a little spread out, but I didn't see an easy solution to simplify the result description callback interface without a breaking change. This seems perfectly functional and passes tests. I didn't see any existing testing for `calculateProvidedBy`, so I didn't try to take that on.

After feedback, I will be glad to update the corresponding docs.

References #1904

### DIFF
--- a/packages/toolkit/src/query/core/buildThunks.ts
+++ b/packages/toolkit/src/query/core/buildThunks.ts
@@ -513,6 +513,7 @@ export function calculateProvidedByThunk(
     isFulfilled(action) ? action.payload : undefined,
     isRejectedWithValue(action) ? action.payload : undefined,
     action.meta.arg.originalArgs,
+    action.meta,
     assertTagType
   )
 }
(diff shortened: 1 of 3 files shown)


markerikson:
Seems reasonable. Thanks!

bever1337:
Thanks @markerikson . I'll put together a second PR for docs changes in the next couple of days.

packages/toolkit/src/query/core/buildThunks.ts
```diff
@@ -513,6 +513,7 @@ export function calculateProvidedByThunk(
     isFulfilled(action) ? action.payload : undefined,
     isRejectedWithValue(action) ? action.payload : undefined,
     action.meta.arg.originalArgs,
+    action.meta,
```
phryneas, line 516: Sorry for chiming in later, but I'm like 80% sure that this is the "wrong meta". 

I'm pretty sure this would need to be `action.meta.baseQueryMeta`

phryneas:

PR #1926, opened 2022-01-15
Pass baseQueryMeta into calculateProvidedBy

msutkowski:
Patches the work done in https://github.com/reduxjs/redux-toolkit/pull/1910 by ensuring that we're exposing `baseQueryMeta` aka `Request` and `Response` objects... which matches the `transformResponse` behavior for consistency. I debated on doing a full use case example here, but it seemed like it wouldn't really provide any real value and would be better served in usage docs.

Also note: I didn't update the docs for this as @bever1337 is working on those.

### DIFF
--- a/packages/toolkit/src/query/core/buildThunks.ts
+++ b/packages/toolkit/src/query/core/buildThunks.ts
@@ -5,9 +5,8 @@ import type {
   BaseQueryError,
   QueryReturnValue,
 } from '../baseQueryTypes'
-import { BaseQueryArg } from '../baseQueryTypes'
 import type { RootState, QueryKeys, QuerySubstateIdentifier } from './apiState'
-import { QueryStatus, CombinedState } from './apiState'
+import { QueryStatus } from './apiState'
 import type { StartQueryActionCreatorOptions } from './buildInitiate'
 import type {
   AssertTagTypes,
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/query/core/buildThunks.ts
```diff
@@ -500,20 +500,30 @@ In the case of an unhandled error, no tags will be "provided" or "invalidated".`
...
+
   return calculateProvidedBy(
     endpointDefinitions[action.meta.arg.endpointName][type],
     isFulfilled(action) ? action.payload : undefined,
     isRejectedWithValue(action) ? action.payload : undefined,
     action.meta.arg.originalArgs,
-    action.meta,
+    isAnyOf(isRejected, isFulfilledQuery, isFulfilledMutation)(action)
+      ? action.meta.baseQueryMeta
+      : undefined,
```
msutkowski, line 526: I don't know if all of this ceremony is really necessary, but it's probably better than `(action.meta as any).baseQueryMeta` or something like that. Very open to ideas here @phryneas if you have one :)

bever1337:
Hey @msutkowski 👋 I just got around to submitting a PR for a docs update, let me know if it's up to snuff.

https://github.com/reduxjs/redux-toolkit/pull/1934

Would it be easier for your workflow if we merged my branch to yours instead of master? I haven't looked too closely, so I'm not sure what overlap there is. Thanks!

packages/toolkit/src/query/core/buildThunks.ts
phryneas: I think that can be done with less changes - I pushed my suggestion.

msutkowski: Thanks! I need to get a 'ways to narrow types' checklist solidified in my brain.

phryneas:

PR #1928, opened 2022-01-16
fix: clearPolls should clearTimout for dead subscriptions

### LINKED ISSUE #1691: polling polls one additional time after unsubscribe
Full Disclosure: I'm using [ngrx-rtk-query](https://github.com/SaulMoro/ngrx-rtk-query) with Angular. It might be their problem.

**Environment**
- @reduxjs/toolkit: 1.6.2
- ngrx-rtk-query: 2.3.0
- node: 14.16.1
- npm: 6.14.2
- OS: Windows 10
- Browser: Edge (Chromium)

**Summary**
I have an endpoint that takes paging params, e.g., `{pageNumber: number, pageSize: number}`. I refetch the data every 30 seconds to ensure the latest data is visible to the user.

rtk-query will send the last request of any query that didn't send prior to the arg change.

For example, if a user is on page 1, then goes to page 2, the following requests will be sent in this order:

```
/api/reservedwords?pageNumber=1&pageSize=10
/api/reservedwords?pageNumber=2&pageSize=10
/api/reservedwords?pageNumber=1&pageSize=10 // should not be sent
/api/reservedwords?pageNumber=2&pageSize=10
```

Here's the relevant code

```typescript
// reserved-words.api.ts
    baseQuery: fetchBaseQuery({ baseUrl: '/api' }),
    endpoints: (build) => ({
        getReservedWords: build.query<PagedResource<ReservedWordResource>, ReservedWordsQueryParams | void>({
            query: (params: ReservedWordsQueryParams) => ({
                url: 'reservedwords',
                params
            }),
            providesTags: [cacheTags.reservedWords]
        })
    })
});

export const { useGetReservedWordsQuery } = api;

// reserved-words-api.facade.ts
...
@Injectable({
    providedIn: 'root'
})
export class ReservedWordsApiFacade {
    public getReservedWords$(params: Observable<ReservedWordsQueryParams>): ReturnType<typeof useGetReservedWordsQuery> {
        return useGetReservedWordsQuery(params, { pollingInterval: 30000 });
    }
}

// reserved-words.component.ts
...
export class ReservedWordsComponent
    public query$ = new BehaviorSubject<ReservedWordsQueryParams>({ pageNumber: 1, pageSize: 10 });
    ...

    constructor(private reservedWordsApi: ReservedWordsApiFacade) {}

    public ngOnInit(): void {
        const sub = this.reservedWordsApi
            .getReservedWords$(this.query$)
            .pipe(tap((x) => console.log('from component', x.originalArgs)))  // Works fine here. Only logs when expected
            .subscribe((res) => {
                // do stuff with res
            });

        this.subscriptions.push(sub);
    }
...

---

Update: Remove Facade Code to make more readable.

### PR DESCRIPTION
**Pull request by djbreen7**
Fixes #1691

### DIFF
--- a/packages/toolkit/src/query/core/buildMiddleware/polling.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/polling.ts
@@ -45,6 +45,11 @@ export const build: SubMiddlewareBuilder = ({
           clearPolls()
         }
 
+        if (api.internalActions.unsubscribeQueryResult.match(action)) {
+          const { queryCacheKey } = action.payload
+          clearPoll(queryCacheKey)
+        }
+
         return result
       }
 
@@ -91,8 +96,6 @@ export const build: SubMiddlewareBuilder = ({
       const querySubState = state.queries[queryCacheKey]
       const subscriptions = state.subscriptions[queryCacheKey]
 
-      clearExpiredPolls(state.subscriptions)
-
       if (
         !querySubState ||
         querySubState.status === QueryStatus.uninitialized
@@ -119,25 +122,15 @@ export const build: SubMiddlewareBuilder = ({
     }
 
     function clearPolls() {
-      for (const [key, poll] of Object.entries(currentPolls)) {
-        if (poll?.timeout) clearTimeout(poll.timeout)
-        delete currentPolls[key]
+      for (const [key] of Object.entries(currentPolls)) {
+        clearPoll(key)
       }
     }
 
-    function clearExpiredPolls(subscribers: SubscriptionState) {
-      for (const [key, poll] of Object.entries(currentPolls)) {
-        const sub = subscribers[key]
-
-        if (sub) {
-          const isActiveSubscription = Object.entries(sub).length
-
-          if (poll?.timeout && !isActiveSubscription) {
-            clearTimeout(poll.timeout)
-            delete currentPolls[key]
-          }
-        }
-      }
+    function clearPoll(key: string) {
+      const poll = currentPolls[key];
+      if (poll?.timeout) clearTimeout(poll.timeout)
+      delete currentPolls[key]
     }
   }
 
--- a/packages/toolkit/src/query/core/buildMiddleware/polling.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/polling.ts
@@ -1,4 +1,4 @@
-import type { QuerySubstateIdentifier, Subscribers, SubscriptionState } from '../apiState'
+import type { QuerySubstateIdentifier, Subscribers } from '../apiState'
 import { QueryStatus } from '../apiState'
 import type {
   QueryStateMeta,
--- a/packages/toolkit/src/query/core/buildMiddleware/polling.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/polling.ts
@@ -90,6 +90,13 @@ export const build: SubMiddlewareBuilder = ({
       const state = api.getState()[reducerPath]
       const querySubState = state.queries[queryCacheKey]
       const subscriptions = state.subscriptions[queryCacheKey]
+      const { queryCacheKey: omitted, ...oldSubscriptions } = state.subscriptions
+
+      const hasDeadSubscriptions = Object.entries(oldSubscriptions).some(x => !Object.entries(x[1] ?? {}).length)
+      
+      if (hasDeadSubscriptions) {
+        clearPolls();
+      }
 
       if (
         !querySubState ||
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by djbreen7** [comment]
Looks like checks may just be failing, regardless of PR content. I noticed this on the last few PRs.

Regarding this PR:
It works for me, but it might not be "THE" solution. Maybe calling `clearPolls` is overkill and I should call `clearTimeout` on dead `currentPolls` associated with dead subscriptions instead?

**Comment by msutkowski** [comment]
> Looks like checks may just be failing, regardless of PR content. I noticed this on the last few PRs.
> 
> Regarding this PR: It works for me, but it might not be "THE" solution. Maybe calling `clearPolls` is overkill and I should call `clearTimeout` on dead `currentPolls` associated with dead subscriptions instead?

I wouldn't clear all of the polling intervals here. If we can be more granular as to remove _only_ the dead ones that are related to a given subscription change. 

I think the big thing we're not accounting here for is:

setup: timer set for `15000`

1. start poll for args `{ page: 1, search: 'thing' }`
2. change query inputs and now look for `{ page: 2, search: 'thing' }`.
3. 10s later page back to `{ page: 1, search: 'thing' }`

In this case, I'd imagine we'd leave the timers intact until the cache entry actually drops _(meaning that when switching back to the args in 1, it'd poll in 5 seconds, instead of restarting at 15)_. So... I'm not sure if doing this on arg change is accurate? 

So to summarize my view: I think it's okay to let polls run even if a query isn't currently mounted, and then only remove its timer in accordance with `keepUnusedDataFor` settings and the actual `onCacheEntryRemoved`. Does that make sense? Or am I missing something? :sweat_smile: 


Sooooo, maybe a solution could be as simple as this:

```ts
        if (api.internalActions.removeQueryResult.match(action)) {
          delete currentPolls[action.payload.queryCacheKey]
        }
```

phryneas:
@msutkowski removeQueryResult would definitely be too late - polling is always bound to a component being mounted.

`clearPolls` on the other hand is definitely too much, since it removes all polls for all cache keys.

msutkowski:
> @msutkowski removeQueryResult would definitely be too late - polling is always bound to a component being mounted.
> 
> `clearPolls` on the other hand is definitely too much, since it removes all polls for all cache keys.

@djbreen7 Can you give this a shot?

```ts
if (api.internalActions.unsubscribeQueryResult.match(action)) {
          const { queryCacheKey } = action.payload
          const existingPoll = currentPolls[queryCacheKey]
          if (existingPoll?.timeout) {
            clearTimeout(existingPoll.timeout)
            delete currentPolls[queryCacheKey]
          }
        }
```

I believe this works like you'd expect :)

phryneas:
@msutkowski seems like a good direction, but that poll should only be cleared if there are no other active polls for the same endpoint left

djbreen7:
@msutkowski Shoot! I didn't see the conversation before my latest commit. 

What you're proposing makes sense to me. I just tested it and it works for my use case. I'm happy to change directions to go with your idea once we work out the kinks that @phryneas mentioned.

> ... that poll should only be cleared if there are no other active polls for the same endpoint left

Can you provide a little more clarification for me? My understanding is that the problem is _because_ polls remain active for the same endpoint when they shouldn't. 

In my example in #1691 , the problem is that that a new poll is added when pagination changes without removing old ones. In other words, if I change the query from `{pageNumber: 1}`, to `{pageNumber: 2}`, I'll have two active polls even though nothing is currently subscribed to the poll with the query `{pageNumber: 1}`.

phryneas:
Sorry, I meant endpoint/argument combination.

You can have the same query mounted with multiple different polls:

```
// component A:
useMyQuery(null, { pollInterval: 5 })


// component B:
useMyQuery(null, { pollInterval: 3 })
```

now when component A unmounts, but component B is not unmounted, the polling should not be interrupted, since there is still a subsciber waiting for this to poll.

phryneas:
@msutkowski you made a very interactive example of all that polling stuff. do you still have that codesandbox?

msutkowski:
@djbreen7 I can't easily patch into this PR, but I opened one and tagged this. Are you comfortable adding tests for this, or would you like me to tackle that? I realized we don't have any... and I have free time today :)

djbreen7:
@msutkowski 

> @djbreen7 I can't easily patch into this PR, but I opened one and tagged this. Are you comfortable adding tests for this, or would you like me to tackle that? I realized we don't have any... and I have free time today :)

You'll definitely be faster than I will. This is my first foray into the redux-toolkit source code. I'm around today too, so I'm happy to help out wherever I can. We could go as far as pairing on a LiveShare if there's a slack group or discord you can DM me a link.

msutkowski:
> @msutkowski
> 
> > @djbreen7 I can't easily patch into this PR, but I opened one and tagged this. Are you comfortable adding tests for this, or would you like me to tackle that? I realized we don't have any... and I have free time today :)
> 
> You'll definitely be faster than I will. This is my first foray into the redux-toolkit source code. I'm around today too, so I'm happy to help out wherever I can. We could go as far as pairing on a LiveShare if there's a slack group or discord you can DM me a link.

You can always find Lenz, Mark and myself in the Reactiflux discord in #redux. My username is msutkowski#0001.

djbreen7:
Closing in favor of #1933

PR #1932, opened 2022-01-17
Add a dev/watch mode script

msutkowski:
This PR adds a simple `watch` mode that we can use while actively developing against the examples... or whatever else people use watch modes for outside of `jest` 😄 . It isn't perfect, but better than what we've got.

What happens when running `yarn dev` is that it'll go through one complete build, and from then on out, it'll only do an incremental build of the affected package. `umd` bundles **_will not be rebuilt_** as there isn't a straightforward way to know what work is left to do in the `onRebuild` callback, and I don't think it's worth the effort to support dev mode for `umd` either. If someone _really_ wanted that, we could implement a simple polling system and run the rollup umd portion... but, probably not worth it :)


markerikson:
@msutkowski this still something you want to add?

msutkowski:
> @msutkowski this still something you want to add?

We can close this 🎉

Issue #1283, opened 2021-07-11
[RED-10] `prefetch` creates a suscription that can not be unsubscribed


phryneas:
Currently, `prefetch` creates a subscription that cannot be unsubscribed.

Ideally, `prefetch` should either
* subscribe and directly unsubscribe, to just keep the cache data around until cache collection
* subscribe and return a means to unsubscribe

It could also just take an option to switch between those two behaviours.

<sub>[RED-10](https://linear.app/redux/issue/RED-10/prefetch-creates-a-suscription-that-can-not-be-unsubscribed)</sub>

bever1337:
Are there ideas for how `usePrefetch` could return `unsubscribe` without a breaking change? _Technically_ you can assign a function as a property of another function, but I don't think that's a good solution for a public API.

bever1337:
In PR #1938 I toyed around with prefetch resolving the query void OR the result, exposing `unwrap`, `unsubscribe`, etc. That was pretty cool I thought. 5 seconds later, I realized this is a convoluted path to get identical behavior as manually initiating a query. After chatting in reactiflux about this feature, it might make more sense to just always unsubscribe after prefetch. No configuration options are necessary, instead RTKQ just always cleans up the orphaned subscriptions.

bever1337:
The next approach will be to provide a config option like `cleanup`, where false is the default (and current) behavior. Unsubscribing by default is a breaking change, so the last commit needs to be stepped back. There's still no need to expose `unsubscribe` in my opinion.

vladamx:
Bumping this since it can really lead to unpredictable behavior in the app. In general, subscription management is not straightforward.

markerikson:
@vladamx could you clarify what you mean by "subscription management is not straightforward" ?

vladamx:
@markerikson I mainly meant that hanging or global subscriptions are not easy to find just by statically reading the code. Those subscriptions cause data to never update regardless of keepUnusedFor. Devtools help to spot their count but does not pinpoint where are those subs.

dorsharonfuse:
@markerikson This literally just introduce a bug in our app where the cache wouldn't clean up because of that dangling subscription.
This needs to be addressed..

markerikson:
@dorsharonfuse honestly, we're pretty with a lot of other things right now. We'll happily accept PRs that fix an issue like this, but this is not a priority for us maintainers to work on atm.

markerikson:
[...]

Issue #1282, opened 2021-07-11
Proper handling of non-subscribing `initiate` calls.


phryneas:
Right now, a call to `endpoint.initiate(..., { subscribe: false })` has the following behaviour:

* dispatches pending, does *not create a cache entry*
* dispatches fulfilled, updates cache entry *if one exists*
* returns a promise of the result of `selector(...)(getState())`

While this is great for just updating an existing state value without actually subscribing (we use it for `refetch`), it will return an uninitialized state if there never was a cache entry before. That part, I would consider a bug.

We could have it return the fulfilled action itself (#1277 would guarantee that we always have one of those, even in "skipped, because pending" scenarios), but that would change the general return value.

We could also just change the signature of `{ subscribe: false }` to always return `void`. I would consider that valid, too.

In both of these cases, we should probably change LazyQuery in #1280 to always subscribe to state. Since that was never properly exposed and a bug at the given time, that could still be a "bug fix" and not a "breaking change".

bever1337:
I found this ticket because I was trying to find any documentation on what `{ subscribe: false }` actually does. I was looking for a solution that solved "execute and resolve this endpoint once, automatically clean up any query subscriptions." 

> We could also just change the signature of { subscribe: false } to always return void. I would consider that valid, too.

I think this change would mean that `{ subscribe: false }` could _not_ be used to create a "once" behavior from RTKQ endpoints. Since that is a feature I personally want, I am glad to make the PR if you can please clarify the alternative solution you'd like to see.

> While this is great for just updating an existing state value without actually subscribing (we use it for refetch), it will return an uninitialized state if there never was a cache entry before. That part, I would consider a bug.

This tracks. The necessary change is not to initialize cache when there are no subscriptions.

> We could have it return the fulfilled action itself (#1277 would guarantee that we always have one of those, even in "skipped, because pending" scenarios), but that would change the general return value.

This is where I am confused. Can you please explain what the type change is?

I don't want to unfairly promote the changes I want, so please let me know if the preferred solution is to return void. I am still glad to contribute. Thanks for your help.

PR #1938, opened 2022-01-18
feat(query): Adds subscribe option to prefetch, automatically unsubscribes if not present [#1283]

bever1337:


packages/toolkit/src/query/core/buildThunks.ts
```diff
@@ -430,6 +433,17 @@ In the case of an unhandled error, no tags will be "provided" or "invalidated".`
...
+  const hasSubscribe = (options: any): options is { subscribe: boolean } =>
+    'subscribe' in options
+  const handlePrefetchSubscription = (
+    result: QueryActionCreatorResult<any>,
+    subscribe: boolean
+  ): void | QueryActionCreatorResult<any> => {
+    if (subscribe) {
+      return result
+    }
+    result.unwrap().finally(result.unsubscribe)
```
bever1337, line 445: Should this catch and swallow the error?

packages/toolkit/src/query/core/buildThunks.ts
```diff
@@ -430,6 +433,8 @@ In the case of an unhandled error, no tags will be "provided" or "invalidated".`
   const hasMaxAge = (
     options: any
   ): options is { ifOlderThan: false | number } => 'ifOlderThan' in options
+  const handlePrefetchSubscription = (result: QueryActionCreatorResult<any>) =>
+    result.unwrap().finally(result.unsubscribe)
```
bever1337, line 437: I opted to `unwrap` because that seems closer aligned to the lifecycle the end-user is expecting. My hope is it also sidesteps worrying about `keepUnusedDataFor` being set to 1 second or other cache edge-cases I can't think of.

phryneas:
I would keep the "automatic unsubscribe" out of this PR in general and keep the scope of this to only adding a `subscribe` yes/no option. 

The rest should be done in a separate PR handling #1282 that would automatically subscribe *all* queries but automatically unsubscribe them after the thunk has resolved if `subscribe: false`.

PR #1944, opened 2022-01-21
Fix link and improved referring location

geoalexidis:
- Fixed a broken link
- Refer to Response page instead of Body (the better referring location based on the context of this section)


msutkowski (approved):

msutkowski:
Thanks @geoalexidis!

Issue #1782, opened 2021-11-29
RTQK: Automatic re-auth - multiple requests


cmtmc:
Hi guys and thanks for the toolkit.

"Automatic re-authorization by extending fetchBaseQuery" from [customizing queries](https://redux-toolkit.js.org/rtk-query/usage/customizing-queries) is pretty nice. But suppose we run several queries at once - all of them return 401, first reauth disables old refresh token on server - and all rest queries fail to reauth. What is the recommended way for such case?

Using axios, we can use its response interceptor - setInterval for all other queries after first (which does re-auth) and re-call them when re-auth is completed ([here](https://stackoverflow.com/questions/60941326/how-to-handle-jwt-refresh-token-on-client-side-with-multiple-requests-api-call) and [here](https://stackoverflow.com/questions/60749813/how-to-avoid-multiple-token-refresh-requests-when-making-simultaneous-api-reques)). Do I need to make Axios baseQuery​ (as setInterval needs promises to work with async)- or it is a typical thing with an easy solution?

Thanks!

phryneas:
You can always add a [Mutex](https://blog.mayflower.de/6369-javascript-mutex-synchronizing-async-operations.html)

andrew-moiseyev:
Thanks a lot, I've used [async-mutex](https://github.com/DirtyHairy/async-mutex) - and it seems to work fine now.

andrew-moiseyev:
Thanks, using [async-mutex](https://github.com/DirtyHairy/async-mutex) :

```
const mutex = new Mutex();

const baseQueryWithReauth = async (args, api, extraOptions) => {
  let result = await baseQuery(args, api, extraOptions);

  if (result.error && result.error.status === 401) {
    if (!mutex.isLocked()) {
      const release = await mutex.acquire();

      try {
        const refreshResult = await baseQuery(
          {
            url: 'auth/refresh/',
            method: 'POST',
            body: { getToken(), getRefreshToken() },
          },
          api,
          extraOptions,
        );

        if (refreshResult.data) {
          api.dispatch(tokenUpdated(refreshResult.data));

          // retry the initial query
          result = await baseQuery(args, api, extraOptions);
        } else {
          api.dispatch(logout());
        }
      } finally {
        release();
      }
    } else {
      await mutex.waitForUnlock();
      result = await baseQuery(args, api, extraOptions);
    }
  }

  return result;
};

const baseApi = createApi({
  baseQuery: baseQueryWithReauth,
[...]

PR #1947, opened 2022-01-21
Feature/docs re authorization mutex example

DeniKucevic:
Adding example for using mutex to prevent multiple queries when using re-authorization as found here :

https://github.com/reduxjs/redux-toolkit/issues/1782#issuecomment-985985120

### DIFF
--- a/docs/rtk-query/usage/customizing-queries.mdx
+++ b/docs/rtk-query/usage/customizing-queries.mdx
@@ -392,7 +392,7 @@ const baseQueryWithReauth: BaseQueryFn<
 
 #### Preventing multiple unauthorized errors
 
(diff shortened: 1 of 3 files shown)


docs/rtk-query/usage/customizing-queries.mdx
```diff
@@ -390,6 +390,80 @@ const baseQueryWithReauth: BaseQueryFn<
 }
 ```
 
+#### Preventing multiple unauthorized errors
+
+Implementing [`async-mutex`](https://github.com/DirtyHairy/async-mutex) to prevent multiple calls to '/refreshToken' when multiple calls fail with [`401 Unauthorized`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) errors.
```
phryneas, line 395:
```suggestion
Using [`async-mutex`](https://github.com/DirtyHairy/async-mutex) to prevent multiple calls to '/refreshToken' when multiple calls fail with [`401 Unauthorized`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) errors.
```

phryneas:
Thanks! Sorry I missed this one.

DeniKucevic:
No worries, as you can see we can get lost in the work 😄

Issue #1896, opened 2022-01-06
[RTKQ]: copyWithStructuralSharing peformance issue


zoubingwu:
I was testing with mock data, some request returns an array of thousands of items, and it freezes for like 10+ second so I found this:

![image](https://user-images.githubusercontent.com/18644268/148364667-a988a8b3-aae1-4b4c-af02-88c7282aca4c.png)

I know with that amount of data it should be done with paginations or other techniques but this happens in real world. 

The `JSON.parse` is fairly quick, and most of the time it was executing `copyWithStructuralSharing`, so just wondering what it does and is it possible to make it better?

phryneas:
Already 3 seconds for JSON.parse? 
Tbh, I would not have expected anyone to handle these amounts of data. 

At the moment, there is no way to skip it, but I could see us adding a `structuralSharing: false` option to endpoints.

phryneas:
Hey @zoubingwu. @msutkowski has implemented an option for this over in https://github.com/reduxjs/redux-toolkit/pull/1954 - could you please try out the CI build and report back if it solves your problem?

zoubingwu:
@phryneas Yep, just tried `https://pkg.csb.dev/reduxjs/redux-toolkit/commit/738faa66/@reduxjs/toolkit` this build and indeed it worked as expected. Thanks!

phryneas:
Released in https://github.com/reduxjs/redux-toolkit/releases/tag/v1.7.2

PR #1954, opened 2022-01-24
RTKQ: configurable `structuralSharing` on endpoints/queries/createApi

msutkowski:
Addresses #1896.

For the example in that issue, there isn't much that can be done outside of just disabling the structural sharing optimization. This PR enables that by adding a flag to `createApi`, accepting a flag on an individual query endpoint, and allowing a hook/initiate to specify the desired value.

### DIFF
--- a/packages/toolkit/src/query/apiTypes.ts
+++ b/packages/toolkit/src/query/apiTypes.ts
@@ -46,6 +46,7 @@ export type Module<Name extends ModuleName> = {
       | 'refetchOnFocus'
       | 'refetchOnReconnect'
       | 'tagTypes'
+      | 'structuralSharing'
     >,
     context: ApiContext<Definitions>
   ): {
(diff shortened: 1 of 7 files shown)


msutkowski:
@phryneas Do you think it's a better idea to consider doing this at the endpoint level as well instead of an all or nothing approach?

phryneas:
Wanted to comment on that one - yes, I'd prefer this as a non-global escape hatch.

msutkowski:
> Wanted to comment on that one - yes, I'd prefer this as a non-global escape hatch.

Alright, updated so that the order of precedence is `call site (hook/initiate) flag > endpoint flag > global flag`. I was going to remove the global flag, but I do like the flexibility of being able to opt-out, and opt-in very easily.

phryneas (approved):

phryneas:

PR #1957, opened 2022-01-24
Add no retry endpoint list options.

kahirokunn:
It would be useful to have a whitelist for `retry` as well.


phryneas:
Instead of having a string array here, I would suggest adding something like `{ skipRetry: boolean }` to the `AdditionalDefinitionExtraOptions`

kahirokunn:
thx. I will try it.

kahirokunn:
I close this PR because I realized that I can do the same thing by setting the `extraOptions: { maxRetries: 0 }`.

PR #1972, opened 2022-01-28
rtk-query-codegen-openap: Fixed relative `apiFile` path when generating on Windows

andrejpavlovic:
### Summary

Running the generator on Windows where the `outputFile` and `apiFile` are not in the same folder, the generated import `api` import statement would use backslash instead of forward slashes due to operating system differences.

### Issue
Sample configuration:
```typescript
const config: ConfigFile = {
  schemaFile: './openapi-definition.yaml',
  apiFile: './api.ts',
  outputFile: 'generated/index.ts',
}
```

Before fix:
```typescript
import { api } from '..\\api'
```

After fix:
```typescript
import { api } from '../api'
```

### Solution
Normalize `apiFile` path to use forward slashes after relative path computation.


phryneas:
Seems reasonable, thank you!

PR #1974, opened 2022-01-29
Refetch should not happen if no active subscribers

### LINKED ISSUE #1530: RTK Query - refetchOnFocus refetches unmounted queries
We're using the [`refetchOnFocus`](https://redux-toolkit.js.org/rtk-query/api/createApi#refetchonfocus) option and set [`keepUnusedDataFor`](https://redux-toolkit.js.org/rtk-query/api/createApi#keepunuseddatafor) to 5 minutes.

The problem is that when the user navigates through the application, this creates a lot of API requests at once when the page is refocused.

This question already seems to have been up for discussion in the old repo:
- https://github.com/rtk-incubator/rtk-query/pull/90#discussion_r540368910
- https://github.com/rtk-incubator/rtk-query/issues/99
---
It would be very useful to change the default behaviour of this (refetch mounted queries only on refocus) or have an option to disable the fetching of unmounted queries.

Changing the default behaviour would be the preferable way IMHO. We don't have the problem of stale data for unmounted queries because we're using the [`refetchOnMountOrArgChange`](https://redux-toolkit.js.org/rtk-query/api/createApi#refetchonmountorargchange) option.

AlexanderArvidsson:
Fixes #1530

All logic that checks if there are active subscriptions in the code does a falsey check as well as a keys length check on the subscription sub state, except for the code that refetches valid queries. 
[Here is an example from the handleUnsubscribe function.](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts#L98)
[Here is another from invalidateTags.](https://github.com/reduxjs/redux-toolkit/blob/d4e95caa26ea1c76a1ecead8f09b1897b4444d40/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts#L75)
This PR includes the same check to resolve this issue. See #1530 for more information.
Attempting a refetch for data that has no subscribers should remove it from the cache, similar to `invalidateTags`, because the data would become stale relative to other cache data which would be refetched on focus / reconnect.

### DIFF
--- a/packages/toolkit/src/query/core/buildMiddleware/windowEventHandling.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/windowEventHandling.ts
@@ -37,6 +37,7 @@ export const build: SubMiddlewareBuilder = ({
 
         if (
           !subscriptionSubState ||
+          Object.keys(subscriptionSubState).length === 0 ||
           !querySubState ||
           querySubState.status === QueryStatus.uninitialized
         )
--- a/packages/toolkit/src/query/core/buildMiddleware/windowEventHandling.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/windowEventHandling.ts
@@ -39,25 +39,25 @@ export const build: SubMiddlewareBuilder = ({
         const querySubState = queries[queryCacheKey]
         const subscriptionSubState = subscriptions[queryCacheKey]
 
-        if (!subscriptionSubState || !querySubState) return
+        if (!subscriptionSubState || !querySubState) continue
 
-        if (Object.keys(subscriptionSubState).length === 0) {
-          api.dispatch(
-            removeQueryResult({
-              queryCacheKey: queryCacheKey as QueryCacheKey,
-            })
-          )
-        } else if (querySubState.status !== QueryStatus.uninitialized) {
-          const shouldRefetch =
-            Object.values(subscriptionSubState).some(
-              (sub) => sub[type] === true
-            ) ||
-            (Object.values(subscriptionSubState).every(
-              (sub) => sub[type] === undefined
-            ) &&
-              state.config[type])
+        const shouldRefetch =
+          Object.values(subscriptionSubState).some(
+            (sub) => sub[type] === true
+          ) ||
+          (Object.values(subscriptionSubState).every(
+            (sub) => sub[type] === undefined
+          ) &&
+            state.config[type])
 
-          if (shouldRefetch) {
+        if (shouldRefetch) {
+          if (Object.keys(subscriptionSubState).length === 0) {
+            api.dispatch(
+              removeQueryResult({
+                queryCacheKey: queryCacheKey as QueryCacheKey,
+              })
+            )
+          } else if (querySubState.status !== QueryStatus.uninitialized) {
             api.dispatch(refetchQuery(querySubState, queryCacheKey))
           }
         }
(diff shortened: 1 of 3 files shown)


phryneas:
Ha, timing ^^

What I said in the other thread pretty much applies here - "not fetching" in that case is not really an option, as that would leave stale data in the cache. In this case, the data would need to be removed from the cache, like it is in the case of cache invalidation. 
Could you please adjust the PR accordingly?

AlexanderArvidsson:
I will adjust the PR tomorrow with the fixes, as per our discussion in the issue.

AlexanderArvidsson:
@phryneas I changed the logic now and wrote a test for it. I considered placing the logic in `cleanup.test.tsx` but since `refetchValidQueries` is only run when refetching on focus or reconnect, I added it there.

Is it necessary to also include the same test for reconnect? They run the same function, so if the normal tests for reconnect passes, and my new test passes, then it should be guaranteed to pass on reconnect too.

packages/toolkit/src/query/core/buildMiddleware/windowEventHandling.ts
```diff
@@ -35,24 +39,27 @@ export const build: SubMiddlewareBuilder = ({
...
+              (sub) => sub[type] === true
+            ) ||
+            (Object.values(subscriptionSubState).every(
+              (sub) => sub[type] === undefined
+            ) &&
+              state.config[type])
+
+          if (shouldRefetch) {
+            api.dispatch(refetchQuery(querySubState, queryCacheKey))
+          }
```
AlexanderArvidsson, line 62: Should `removeQueryResult` be guarded by `shouldRefetch` as well? What do you think @phryneas ?

packages/toolkit/src/query/core/buildMiddleware/windowEventHandling.ts
```diff
@@ -35,24 +39,27 @@ export const build: SubMiddlewareBuilder = ({
...
-            (sub) => sub[type] === true
-          ) ||
-          (Object.values(subscriptionSubState).every(
-            (sub) => sub[type] === undefined
-          ) &&
-            state.config[type])
-
-        if (shouldRefetch) {
-          api.dispatch(refetchQuery(querySubState, queryCacheKey))
+        if (!subscriptionSubState || !querySubState) return
```
phryneas, line 42: On re-reading this... shouldn't this `return` always have been a `continue`?

packages/toolkit/src/query/core/buildMiddleware/windowEventHandling.ts
phryneas: Yes, moving it into that guard might be a good way. That way we remove as little as necessary from the cache.

packages/toolkit/src/query/core/buildMiddleware/windowEventHandling.ts
AlexanderArvidsson: Very good catch, I'm surprised this hasn't caused more severe issues. I'll adjust it soon.

phryneas:
I'l try tol take a look at it this evening or tomorrow.
Afaik, @markerikson wanted to do a patch release soon so we should be able to get this in.

phryneas:
Looks good to me. Let's get this in.

phryneas (approved):

phryneas:

Issue #1976, opened 2022-01-30
createSelector: Argument of type 'DefaultRootState' is not assignable to parameter of type


akichim21:
When I raise the version of @reduxjs/toolkit from 1.6.2 to 1.7.1, I get the following error.

```
npx tsc

error TS2345: Argument of type 'DefaultRootState' is not assignable to parameter of type '{ readonly [$CombinedState]?: undefined; hoge: hogeEntityState; fuga: fugaEntityState; ... 20 more ...; suge: sugeEntityState; }'.
  Type 'DefaultRootState' is missing the following properties from type '{ readonly [$CombinedState]?: undefined; hoge: hogeEntityState; fuga: fugaEntityState; ... 20 more ...; suge: sugeEntityState; }': hoge, fuga, and 19 more.
```

```
import { useSelector } from "react-redux";
import { hogeSelector } from "./hogeSlice.ts";

  // error argument state
  const hoge = useSelector((state) =>
    hogeSelector(state, fuga.id)
  );
```

hogeSlice.ts
```
import {
createSelector,
createEntityAdapter,
EntityState,
createSlice,
} from "@reduxjs/toolkit";
import { RootState } from "store";
import { HYDRATE } from "next-redux-wrapper";

const adapter = createEntityAdapter<Hoge>();

const initialState: EntityState<Hoge> = adapter.getInitialState({});

/** slice **/
export const hogeSlice = createSlice({
  name: "hoge",
  initialState,
  reducers: {},
  extraReducers: (builder) => {
    builder
      .addCase(HYDRATE, (state, action: any) => {
        return { ...state, ...action.payload.hoge };
      });
});

export const {
  selectAll: hogesSelector
} = adapter.getSelectors((state: RootState) => state.hoge);

export const hogeSelector = createSelector(
  [
    hogesSelector,
    (state, fugaId: number) => fugaId,
  ],
  (entities, fugaId) => {
    return xxxx;
  }
);
```

store.ts
```
import { useDispatch } from "react-redux";
import logger from "redux-logger";
import { createWrapper } from "next-redux-wrapper";
import {
  combineReducers,
  configureStore,
  Middleware,
  // getDefaultMiddleware,
} from "@reduxjs/toolkit";

import { reducers } from "./reducers";

const rootReducer = combineReducers({
  ...reducers,
});

const createStore = () => {
  return configureStore({
    reducer: rootReducer,
    middleware: (getDefaultMiddleware) =>
[...]

Issue #1914, opened 2022-01-12
RTKQ: Selector type errors


dhenneke:
I want to add a RTK Query API to an existing store. I also want to use the selectors rather than the hooks, so I can use the values in other selectors. However, I'm getting typing errors.

See this Codesandbox for an example: https://codesandbox.io/s/redux-toolkit-selector-typing-forked-pylgv

---

This is the error message:
```
Argument of type '{ testSlice: {}; testApi: CombinedState<{ getTest: QueryDefinition<void, BaseQueryFn<any, unknown, unknown, {}, {}>, never, string[], string>; }, never, "testApi">; }' is not assignable to parameter of type 'RootState<{ getTest: QueryDefinition<void, BaseQueryFn<any, unknown, unknown, {}, {}>, never, string[], string>; }, never, string>'.
  Property 'testSlice' is incompatible with index signature.
    Type '{}' is missing the following properties from type 'CombinedState<{ getTest: QueryDefinition<void, BaseQueryFn<any, unknown, unknown, {}, {}>, never, string[], string>; }, never, string>': queries, mutations, provided, subscriptions, configts(2345)
```

What is odd is that the selector expects a type `RootState<{ getTest: QueryDefinition<...>; }, never, string>`, but I would expect it to be `RootState<{ getTest: QueryDefinition<...>; }, never, "testApi">`. Somewhere in `testApi.endpoints.getTest.select()`, the name of the `ReducerPath` gets lost and is replaced with `string`.

In this case, the selector tries to enforce the type of `store.testApi` to `store.testSlice` which won't fit. When I remove the `testSlice`, it works because there is no conflicting property.

phryneas:
That should be fixed by #1977. Could you please try the CI build available at https://ci.codesandbox.io/status/reduxjs/redux-toolkit/pr/1977 (see "local install instructions") and verify if it fixes the problem for you?

dhenneke:
@phryneas Yes this solves the problem in my project. Thanks for the fix 👍

phryneas:
Released in https://github.com/reduxjs/redux-toolkit/releases/tag/v1.7.2

PR #1977, opened 2022-01-30
fix `reducerPath` for query definitions

### LINKED ISSUE #1879: RTK Query endpoint data selectors expecting an incorrect RootState type
Say you have a redux store setup using redux toolkit. your store type will look something like this

```
type Store = EnhancedStore<{ sliceA: SliceA }>;
```

and when you integrate the autogenerated RTK Query reducer into an existing redux store ([link to docs](https://redux-toolkit.js.org/rtk-query/api/created-api/redux-integration)) your store type will look something like 

```
type Store = EnhancedStore<{ sliceA: SliceA, api: CombinedState<...> }>;
```

which will fail type checking when passing in the state from `store.getState` into a [RTK Query endpoint data selector](https://redux-toolkit.js.org/rtk-query/usage/usage-without-react-hooks#accessing-cached-data--request-status) which is expecting a type of:

```
// apiState.d.ts
export declare type RootState<Definitions extends EndpointDefinitions, TagTypes extends string, ReducerPath extends string> = {
    [P in ReducerPath]: CombinedState<Definitions, TagTypes, P>;
};
```

since every reducer path in the store is not of the `CombinedState<...>` type

phryneas:
Fixes #1879 #1914 #1976 

I really don't know how I could have missed that.

### DIFF
--- a/packages/toolkit/src/query/endpointDefinitions.ts
+++ b/packages/toolkit/src/query/endpointDefinitions.ts
@@ -390,10 +390,10 @@ export type EndpointBuilder<
    */
   query<ResultType, QueryArg>(
     definition: OmitFromUnion<
-      QueryDefinition<QueryArg, BaseQuery, TagTypes, ResultType>,
+      QueryDefinition<QueryArg, BaseQuery, TagTypes, ResultType, ReducerPath>,
       'type'
     >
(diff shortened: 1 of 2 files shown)


markerikson:
Hah, nice!

msutkowski (approved):
Confirmed working by authors of the linked issues as well 👍

PR #1984, opened 2022-02-01
Add `ignoreActions` flag to serializable state middleware

### LINKED ISSUE #1967: More control over serializableCheck option in middleware?
I'm using redux-toolkit with redux-persist. In order to avoid serialization errors, middleware is configured to ignore redux-persist actions:

```
 serializableCheck: {
 		ignoredActions: [FLUSH, REHYDRATE, PAUSE, PERSIST, PURGE, REGISTER]
},
```

I don't see a point of checking if actions are serializable if I am not using time-travel or other features. If action has a payload that is not serializable, it doesn't mean I am going to put that data as-is in the store. For example, my app works with dates and I would like to use dates in actions and store them in the state as strings. Or I might be using Set or Map to send data in action, but reducer will store that data as serializable data in the store.

Is there any option to skip checking **only for actions**? All of them? But leave this checking for the state.

I expected something like this, but this is not working:

```
 serializableCheck: {
 		ignoredActions: ['*']
},
```

msutkowski:
Adds an `ignoreActions` flag to `serializableCheck` to balance out the `ignoreState` flag. Also re-ran api-extractor.

Fixes #1967

### DIFF
--- a/docs/api/serializabilityMiddleware.mdx
+++ b/docs/api/serializabilityMiddleware.mdx
@@ -57,9 +57,14 @@ interface SerializableStateInvariantMiddlewareOptions {
   warnAfter?: number
 
   /**
-   * Opt out of checking state, but continue checking actions
+   * Opt out of checking state. When set to `true`, other state-related params will be ignored.
    */
   ignoreState?: boolean
+
+  /**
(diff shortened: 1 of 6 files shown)


markerikson (approved):

phryneas (approved):

phryneas:

PR #1988, opened 2022-02-02
docs – it's → its (two spots); slight wording tweak

bryndyment:
- `it's` always expands to `it is` … meaning `re-fetch it's data` can expand to `re-fetch it is data` which is obviously incorrect
- `we recommend to use` → `we recommend using` (grammar tweak)


markerikson:
Thanks!

PR #2000, opened 2022-02-05
Reorganize serializable check conditions

### LINKED ISSUE #1996: Why `ignoreActions` and `ignoredActions` overrides `ignoreState`?
I want to check the state but ignore the data that is passed to the action.. But i can`t do that.
Сonfiguration
```
{
    ignoreActions: true,
    ignoreState: false,
}
```
is identical
```
{
    ignoreActions: true,
    ignoreState: true,
}
```
Why not make these two checks independent?

msutkowski:
Fixes #1996 

- `state` and `action` checks behave independently, allowing a user to `ignoreState` and still check actions, or `ignoreActions` and continue to check state


markerikson:
I can't say I'm thrilled about all the `!ignoreActions` checks :) But they were there already, I think.

OK, let's go with it.

PR #2001, opened 2022-02-05
Rewrite MiddlewareArray and gDM for better Dispatch inference

markerikson:
This is an attempt to rewrite the type inference for the `middleware` arg in `configureStore`, to better capture cases where a custom middleware overrides the return type of `dispatch`.

The motivation for this is specifically an issue with the action listener middleware:

```ts
const unsubscribe = store.dispatch(addListenerAction({actionCreator, listener});

// ERROR: _should_ be a void callback, but is typed as `{type: 'alm/add', payload: listenerEntry}`
unsubscribe()
```

Initial investigation shows that part of the issue is with the middleware itself, due to adding fields like `addListener` to the returned function.  Additionally, use of `TypedAddListenerAction` seems to help.

However, separate to that, I can reproduce the problem with a dummy middleware:

```ts
const dummyMiddleware: Middleware<
        {
          (action: Action<'actionListenerMiddleware/add'>): Unsubscribe
        },
        CounterState
      > = (storeApi) => (next) => (action) => {}

      const store1 = configureStore({
        // PROBLEM:
        middleware: (gDM) => gDM().prepend(dummyMiddleware),

        // WORKS
        //middleware: [dummyMiddleware, thunk]
      })

      const unsubscribe = store1.dispatch(
        addListenerAction({
          actionCreator: testAction1,
          listener: () => {},
        })
      )
```

Given that, it seems that something is wonky with the way `configureStore` infers the extensions to `dispatch` from the middleware types in conjunction with `getDefaultMiddleware`.

While the commit is a mess, the overall approach here is:

- Rewrite the `MiddlewareArray` type so that it tracks a specific _tuple_ of middleware types, rather than "an array that could be any one of these"
- Use some of the same tuple iteration and intersection techniques we used on Reselect to extract the actual dispatch extension types and intersect them
- Get a more exact final type for `dispatch` using that intersected type

~~What I have here _mostly_ works, except for the "null extraArg in ThunkAction" case.  This is because the `ThunkMiddlewareFor` type previously returned a union of two `ThunkMiddleware` types with different values for `ExtraArgument`, one `undefined` and one `null`.  This was okay when the middleware types were all getting unioned, but now that I've converted them to be a tuple, that doesn't work.~~

~~I'd prefer to drop the `null` case if possible.  The comment says "this is back-compat, since it was shown in the docs", but I don't know where/when the docs showed that.  Unfortunately this seems to be something that is actually used in the wild:~~

~~https://sourcegraph.com/search?q=context:global+lang:typescript+ThunkAction%3C.%2B%2C.%2B%2C.*null.*%3E&patternType=regexp~~

~~Granted, several of the hits I looked at are still using the core `createStore` instead of RTK, but that's still some potential breakage I'd like to avoid if possible.~~

### Update

Lenz and I have worked through the original implementation issues.  Current status:

- We've gotten seemingly correct inference working for middleware, and I've cleaned up the code
- Given that the `extraArg: null` pattern is not something we _want_ people to use (`unknown` would be correct today), and that there's only a couple of examples using it with RTK in the wild, we're comfortable treating that as a bug that needs to be fixed rather than a breaking change and just dropping support for that case
- The `MiddlewareArray.typetests.ts` file began failing in this PR. It appears to have been somehow related to the way we do a "package + install" step in our CI builds, while that file also tried to do an import of some local types. The workaround was to rewrite it to actually call `configureStore` for real instead of faking it at the type level.
- The types changes here now fail in TS 4.0 and earlier.  Given that Reselect now requires TS 4.1 anyway, and 4.6 is right around the corner, it's time to drop 3.9 and 4.0 from our support matrix
- We plan to release the middleware in RTK 1.8, but there's a chance we might need to do a 1.7.x patch release before then.  So, I've created a new `v1.8.0-integration` branch, and will merge these changes into there
- The middleware itself will need additional API changes, per [#1648 (comment)](https://github.com/reduxjs/redux-toolkit/discussions/1648?sort=new#discussioncomment-2118259) , and I'll do those as a separate branch + PR

### DIFF
--- a/packages/toolkit/src/tsHelpers.ts
+++ b/packages/toolkit/src/tsHelpers.ts
@@ -1,4 +1,5 @@
-import type { Middleware } from 'redux'
+import type { Middleware, Dispatch } from 'redux'
+import type { MiddlewareArray } from './utils'
 
 /**
  * return True if T is `any`, otherwise return False
@@ -65,20 +66,30 @@ export type IsUnknownOrNonInferrable<T, True, False> = AtLeastTS35<
   IsEmptyObj<T, True, IsUnknown<T, True, False>>
 >
 
-/**
- * Combines all dispatch signatures of all middlewares in the array `M` into
- * one intersected dispatch signature.
- */
-export type DispatchForMiddlewares<M> = M extends ReadonlyArray<any>
-  ? UnionToIntersection<
-      M[number] extends infer MiddlewareValues
-        ? MiddlewareValues extends Middleware<infer DispatchExt, any, any>
-          ? DispatchExt extends Function
-            ? IsAny<DispatchExt, never, DispatchExt>
-            : never
-          : never
-        : never
+// Appears to have a convenient side effect of ignoring `never` even if that's not what you specified
+export type ExcludeFromTuple<T, E, Acc extends unknown[] = []> = T extends [
+  infer Head,
+  ...infer Tail
+]
+  ? ExcludeFromTuple<Tail, E, [...Acc, ...([Head] extends [E] ? [] : [Head])]>
+  : Acc
+
+type ExtractDispatchFromMiddlewareTuple<
+  MiddlewareTuple extends any[],
+  Acc extends {}
+> = MiddlewareTuple extends [infer Head, ...infer Tail]
+  ? ExtractDispatchFromMiddlewareTuple<
+      Tail,
+      Acc & (Head extends Middleware<infer D, any> ? IsAny<D, {}, D> : {})
     >
+  : Acc
+
+export type ExtractDispatchExtensions<M> = M extends MiddlewareArray<
+  infer MiddlewareTuple
+>
+  ? ExtractDispatchFromMiddlewareTuple<MiddlewareTuple, {}>
+  : M extends Middleware[]
+  ? ExtractDispatchFromMiddlewareTuple<[...M], {}>
   : never
 
 /**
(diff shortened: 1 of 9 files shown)


packages/toolkit/src/tsHelpers.ts
```diff
@@ -81,6 +82,112 @@ export type DispatchForMiddlewares<M> = M extends ReadonlyArray<any>
...
+
+// Appears to have a convenient side effect of ignoring `never` even if that's not what you specified
+export type ExcludeFromTuple<T extends readonly any[], E> = T extends [
+  infer F,
+  ...infer R
+]
+  ? [F] extends [E]
+    ? ExcludeFromTuple<R, E>
+    : [F, ...ExcludeFromTuple<R, E>]
+  : []
```
phryneas, line 131:
```suggestion
export type ExcludeFromTuple<T, E, Acc extends unknown[] = []> =
    T extends [infer Head, ...infer Tail] ? ExcludeFromTuple<
        Tail,
        E,
        [...Acc, ...([Head] extends [E] ? [] : [Head])]
    >
    : Acc;
```

markerikson:
Weird - I force-pushed the branch, and I can see the new commit in https://github.com/reduxjs/redux-toolkit/commits/feature/gdm-array-types , but the PR isn't updating properly.

monim67:
I can't seem to find any issue related to this, is this bug fixed? If yes from which version? I am also facing the unsubscribe type as action bug, is there any work-around?

phryneas:
@monim67  This is a Pull Request that was merged into 1.8.0 one and a half years ago.

Issue #1940, opened 2022-01-19
Discriminator property for polymorphic types seems to be ignored.


apstechlead:
As the title suggest I can't get the generator to work with polymorphic types.

I attached a bare-minimum reproduction schema for demonstration.
The schema has an `Animal` base type and 2 types deriving from it, `Dog` and `Cat`. The `Animal` type has a `discriminator` property called `type` to distinguish between `Dog` and `Cat` types.

I then defined an endpoint which returns an array of `oneOf -> Cat or Dog`.

When I run this RTK generator the resulting types are as follows:

```typescript
export type AnimalType = "All" | "Cats" | "Dogs";
export type Animal = {
  type: AnimalType;
  id?: number;
  name?: string | null;
};
export type Dog = Animal;
export type Cat = Animal;
```

As you can see the result is very wrong. Both `Cat` and `Dog` are defined to equal `Animal` when instead they should extend it and append their own properties.

I used the exact same schema (at the end of this post) with other typescript generators such as `typescript-fetch` and the generated types are as expected.

I removed unnecessary comments for readability:
```typescript
export interface Animal {
    type: AnimalType;
    id?: number;
    name?: string | null;
}
export interface Dog extends Animal {
    dogUniqueProp?: string | null;
}
export interface Cat extends Animal {
    catUniqueProp?: string | null;
}
```

The deserialisation of the `Animal` type also takes this into account: 
```typescript
export function AnimalFromJSONTyped(json: any, ignoreDiscriminator: boolean): Animal {
    if ((json === undefined) || (json === null)) {
        return json;
    }
    if (!ignoreDiscriminator) {
        if (json['type'] === 'Cat') {
            return CatFromJSONTyped(json, true);
        }
        if (json['type'] === 'Dog') {
            return DogFromJSONTyped(json, true);
        }
    }
    return {
        
        'type': AnimalTypeFromJSON(json['type']),
        'id': !exists(json, 'id') ? undefined : json['id'],
        'name': !exists(json, 'name') ? undefined : json['name'],
    };
}
```

Swagger schema used in both examples:

```json
{
  "openapi": "3.0.1",
  "info": {
    "title": "Animals API",
    "version": "1.0.0"
  },
  "servers": [
    {
      "url": "https://localhost:8000"
    }
  ],
  "paths": {
    "/api/v1/animals": {
      "get": {
        "parameters": [
          {
            "name": "type",
            "in": "query",
            "schema": {
[...]

PR #2004, opened 2022-02-05
rtk-query-codegen-openapi: use patched oazapfts for intersection type gen

### LINKED ISSUE #2002: Inheritance in openApi code generation incorrect.
[Swagger Inheritance & Polymorphism](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/) is not generating correctly from the OpenApi definition. this is directly related to #1940. But i am recording this as a seperate issue because it references a different case but is most probably a related error in the code.

I have striped down the situation as to not clutter the issue.
When a openApi schema has an `Entity` base type and a subtype called `Project`, and is defined simply as `Project -> allOf -> Entity` with no descriminators or anything just a simple inheritance the `Project` entity is not correctly generated.

RTK Generator produces the following type output:
```typescript
export type Entity= {
  id?: string | null;
  name?: string | null;
  description?: string | null;
  status?: string | null;
};
export type Project = Entity
```
This is incorrect and seems to have 2 mistakes:

1. all additional properties on a subtype are lost due to the `Project = Entity`.
2. the missing properties are caused because the code gen creates `type`s not `interface`s which cannot be extended.

msutkowski:
~~I opened a PR over in oazapfts here: https://github.com/cellular/oazapfts/pull/207. I'd prefer to not be blocked on them merging that PR or a similar one that is referenced in the linked PR. So this would allow us to patch this behavior for our users in the short term.~~

## Update
This now uses a patched version of oazapfts (@rtk-query/oazapfts-patched) located at https://github.com/rtk-incubator/oazapfts

- Adds test for intersection types

Fixes #2002 #1940

### DIFF
--- a/packages/rtk-query-codegen-openapi/test/__snapshots__/generateEndpoints.test.ts.snap
+++ b/packages/rtk-query-codegen-openapi/test/__snapshots__/generateEndpoints.test.ts.snap
@@ -445,6 +445,40 @@ export type PatchApiV1ListByItemIdApiArg = {
 
 `;
 
+exports[`tests from issues issue #2002: should be able to generate proper intersection types 1`] = `
+import { api } from './tmp/emptyApi';
+const injectedRtkApi = api.injectEndpoints({
+  endpoints: (build) => ({
+    getApiV1Animals: build.query<GetApiV1AnimalsApiResponse, GetApiV1AnimalsApiArg>({
+      query: (queryArg) => ({
+        url: \`/api/v1/animals\`,
+        params: { type: queryArg['type'] },
+      }),
+    }),
+  }),
+  overrideExisting: false,
+});
+export { injectedRtkApi as enhancedApi };
+export type GetApiV1AnimalsApiResponse = /** status 200 Success */ (Dog | Cat)[];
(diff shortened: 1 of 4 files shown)


packages/rtk-query-codegen-openapi/test/__snapshots__/generateEndpoints.test.ts.snap
```diff
@@ -445,6 +445,40 @@ export type PatchApiV1ListByItemIdApiArg = {
...
+};
+export type AnimalType = 'All' | 'Cats' | 'Dogs';
+export type Animal = {
+  type: AnimalType;
+  id?: number;
+  name?: string | null;
+};
+export type Dog = Animal & {
+  dogUniqueProp?: string | null;
+};
```
msutkowski, line 474: Currently, we don't generate this intersection and only do `export type Dog = Animal`.

phryneas:
Could you publish that as a scoped package to npm? I'm not comfortable adding a dependency directly from github.

phryneas:
ping - could you please take a look if this is still not upstream and publish that package?

Issue #1720, opened 2021-11-11
Add `upsertQueryData` functionality


chrisjagoda:
I have a use case where I would like to pass an additional parameter when calling an endpoint, where I do not want this additional parameter to be cached. The first parameter, args, would still be serialized as the cache key as normal.

Would it be possible to provide an additional field to pass some additional data when calling an endpoint via a hook/initiate? My use case is for calling a query endpoint with initiate.

Perhaps an additional optional parameter after options? Something like`extraArgs`, `noCacheArgs`, or `nonCacheableArgs`. I don't know if it makes sense have it on the options parameter, https://redux-toolkit.js.org/rtk-query/api/created-api/endpoints#initiate, but that would work too.

chrisjagoda:
Specifically, my use case is to cache the result from a create api call in a different cache. I can do this, but it comes at the cost of having to override serializeQueryArgs, and having to have some additional logic in the providesTags field for postById.

If I could pass an additional parameter in postsClient.endpoints.postById.initiate, e.g. the response from createPost, I could see that I have that data in the queryFn for getById, and if so, return it, caching it, or if not, calling the api and cache it as normal.

`
import { createApi } from "@reduxjs/toolkit/query/react";
import posts from "some-api";
import { getByIdProvidesTags, getByIdQueryFn, mapClientResponse, serializeGetByIdArgs } from "./postsHelper";

export const POSTS = "posts";
export const POST_BY_ID = "postById";

const { get: getPosts, getById: getPostById, create: createPost, update: updatePost } = posts;

const postsApi = createApi({
    reducerPath: "posts",
    tagTypes: [POSTS, POST_BY_ID],
    baseQuery: ({ args, method }) => method(...args).then(mapClientResponse),
    serializeQueryArgs: serializeGetByIdArgs(POST_BY_ID),
    endpoints: (builder) => ({
        posts: builder.query({
            query: (args) => ({ args: [args], method: getPosts }),
            providesTags: (result, error, args) => [{ type: POSTS, id: JSON.stringify(args) }],
        }),
        postById: builder.query({
            queryFn: getByIdQueryFn(getPostById),
            providesTags: getByIdProvidesTags(POST_BY_ID),
        }),
        createPost: builder.mutation({
            query: (args) => ({ args: [args], method: createPost }),
            invalidatesTags: (result, error) => (error ? [] : [{ type: POSTS }]),
            onQueryStarted: (_args, { dispatch, queryFulfilled }) =>
[...]

PR #2007, opened 2022-02-07
Add `upsertQueryData` functionality per issue #1720

lindapaiste:
Is this the desired functionality?
- Should a user be allowed to manually insert an error, or just data?
- Should the third argument of the action creator be a function which is called with the previous data, even though that data might be `undefined`?
- Are the returned immer patches correct?  I will need help writing the unit tests on these.

### DIFF
--- a/packages/toolkit/src/query/core/buildSlice.ts
+++ b/packages/toolkit/src/query/core/buildSlice.ts
@@ -1,4 +1,4 @@
-import type { AnyAction, PayloadAction } from '@reduxjs/toolkit'
+import type { PayloadAction } from '@reduxjs/toolkit'
 import {
   combineReducers,
   createAction,
@@ -22,7 +22,7 @@ import type {
   ConfigState,
 } from './apiState'
 import { QueryStatus } from './apiState'
-import type { MutationThunk, QueryThunk } from './buildThunks'
+import type { MutationThunk, QueryThunk, QueryThunkArg } from './buildThunks'
 import { calculateProvidedByThunk } from './buildThunks'
 import type {
   AssertTagTypes,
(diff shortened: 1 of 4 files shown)


markerikson:
Oh hey, _very_ cool!  Thanks for putting this together!

I'll defer to @phryneas , but at least at first glance this looks like basically what I imagine this feature is supposed to do (and I see you based on the implementation off the suggested Immer patches approach in #1720 ).

phryneas (commented):
The problem I see here conceptually is that this adds data, but will never trigger `onCacheEntryAdded`. And since that cache entry now exists, `onCacheEntryAdded` will also not be called in the future for this cache entry.

So, at this moment I think this might not be the right way of going about this.

And if we only trigger the thunk, overwriting the `baseQuery`/`queryFn` call, it might still not be complete, since `initiate` contains logic to hold the "currently running request" for concurrecy reasons - and this *will* take a few ticks to completely go through all steps, so that should really be held.

=> it might make sense to add an internal option `forceQueryFn`to `initiate` that passes it on to `queryThunk`. That way it would behave exactly the same in terms of lifecycle handling.

So the implementation of `upsertQueryData` would become something like

```js
const forceQueryFn = Symbol();
const upsertQueryData = (endpointName, args, data) => (dispatch, getState) => {
  // get initiate for the right endpoint here
  dispatch(initiate(arg, { subscribe: false, forceRefetch: true, [forceQueryFn]: () => ({ data }) }))
}
```

The downside is that this would put all queries in a `loading` state, so there it could again switch between `updateQueryData` and `inititate` depending on context.

phryneas:
This is pretty much superseded by https://github.com/reduxjs/redux-toolkit/pull/2266, so I'm gonna close this one - but thanks a lot for the PR and the initiative!

PR #2009, opened 2022-02-08
📝 Link to essentials tutorial from RTKQ quick start

Shrugsy:
As per title - this hopes to remove some occurrences where people are unaware of the essentials tutorial when learning RTK Query.

I'm open to hear opinions on whether it is obvious enough, and whether it should be in an 'info' box or something instead

### DIFF
--- a/docs/rtk-query/overview.md
+++ b/docs/rtk-query/overview.md
@@ -22,6 +22,12 @@ description: 'RTK Query > Overview: a summary of the RTK Query data caching API 
 
 RTK Query is **an optional addon included in the Redux Toolkit package**, and its functionality is built on top of the other APIs in Redux Toolkit.
 
+:::info
+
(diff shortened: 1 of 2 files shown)


markerikson:
Yeah, let's make a few tweaks here:

- Go ahead and put this in an info box
- Have the link text say `the "Redux Essentials" tutorial`

Let's also put this in the "RTKQ Overview" page.  At a minimum we should link the Essentials tutorial in the "Further Information" section at the bottom.  I also feel like we should highlight the tutorials in an info box somewhere on the page, but I'm actually not sure where in that page it should go.  Maybe we should move "Further Information" up?  We could rename it to "Learn RTK Query", and put it _before_ the "Basic Usage" section.  I'm not 100% sold on that, but I do want the link to the tutorials to not get buried at the bottom of the page.  

Thoughts?

PR #2019, opened 2022-02-10
line 23, spelling mistake

Gichohi-Simon:
changed it from it should written to take a single argument, to it should be written to take a single argument.


markerikson:
Thanks!

PR #2020, opened 2022-02-10
fix: prevent duplicate ids when updating id of an element with the id of an existing element

### LINKED ISSUE #2018: Duplicate Ids when updating the id of an element with the id of an existing element
If you update the id of an entity with the id matching another element the entities are merged correctly but the Ids will be duplicated.

**How to reproduce**

1. Create state 
`     demoAdapter.addMany(state, [
        { id: "one", text: "one" },
        { id: "two", text: "two" },
        { id: "three", text: "three" },
        { id: "random", text: "random" }
      ]);`
2. Update state
`demoAdapter.updateOne(state, { id: "random", changes: { id: "three" } });`

**Expected:**
ids:
`["one","two","three"]`

**Actual:**
Ids:
`["one","two","three","three"]`

Also rewrote a counter example on code sandbox to show the issue:
https://codesandbox.io/s/wandering-https-775mm?file=/src/features/counter/counterSlice.js:517-590

Hope this is detailed enough, wrote it in a rush :). but otherwise let me know.

samatar26:
Fixes #2018

Think I'll probably have to update the sorted_state adapter too, but I'm unfamiliar with redux-toolkit and will need to take a closer look. Any pointers would be great 😛. Also let me know if it's too "expensive" to create a set from the array of ids and then changing it back to an array again.

### DIFF
--- a/packages/toolkit/src/entities/unsorted_state_adapter.ts
+++ b/packages/toolkit/src/entities/unsorted_state_adapter.ts
@@ -158,7 +158,9 @@ export function createUnsortedStateAdapter<T>(
         0
 
       if (didMutateIds) {
-        state.ids = state.ids.map((id) => newKeys[id] || id)
+        state.ids = Array.from(
+          new Set(state.ids.map((id) => newKeys[id] || id))
+        )
       }
     }
   }
--- a/packages/toolkit/src/entities/unsorted_state_adapter.ts
+++ b/packages/toolkit/src/entities/unsorted_state_adapter.ts
@@ -158,9 +158,7 @@ export function createUnsortedStateAdapter<T>(
         0
 
       if (didMutateIds) {
-        state.ids = Array.from(
-          new Set(state.ids.map((id) => newKeys[id] || id))
-        )
+        state.ids = Object.keys(state.entities)
       }
     }
   }
--- a/packages/toolkit/src/entities/tests/sorted_state_adapter.test.ts
+++ b/packages/toolkit/src/entities/tests/sorted_state_adapter.test.ts
@@ -1,6 +1,6 @@
-import type { EntityStateAdapter, EntityState } from '../models'
+import type { EntityAdapter, EntityState } from '../models'
 import { createEntityAdapter } from '../create_adapter'
-import { createAction } from '../../createAction'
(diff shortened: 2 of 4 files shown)


packages/toolkit/src/entities/unsorted_state_adapter.ts
```diff
@@ -158,7 +158,9 @@ export function createUnsortedStateAdapter<T>(
         0
 
       if (didMutateIds) {
-        state.ids = state.ids.map((id) => newKeys[id] || id)
+        state.ids = Array.from(
```
paolorossig, line 161: The stat.ids iteration is necessary? What do you think about:
`state.ids = Object.keys(state.entities)` ?

markerikson: I _think_ that could work, actually. Worth a shot?

paolorossig, line 161: I works well. Here's my suggestion:
```suggestion
        state.ids = Object.keys(state.entities)
```

packages/toolkit/src/entities/unsorted_state_adapter.ts
```diff
@@ -158,7 +158,9 @@ export function createUnsortedStateAdapter<T>(
...
+          new Set(state.ids.map((id) => newKeys[id] || id))
```
paolorossig, line 162:
```suggestion
```

packages/toolkit/src/entities/unsorted_state_adapter.ts
```diff
@@ -158,7 +158,9 @@ export function createUnsortedStateAdapter<T>(
...
+        )
```
paolorossig, line 163:
```suggestion
```

paolorossig (changes requested):

packages/toolkit/src/entities/unsorted_state_adapter.ts
samatar26: Updated and all the tests are passing too

markerikson:
Just rebased this and updated the tests.  Thanks!

Issue #2003, opened 2022-02-05
RTK 1.8 and 1.9 planning


markerikson:
> This was originally "1.8 planning", but 1.8 ended up just having the listener middleware.  So, keeping the issue around for both.

1.8:

- TS fixes for `configureStore` and `getDefaultMiddleware`
- Merge in the listener middleware

Initial notional items for 1.9:

- Drop support for TS < 4.1
- Actual string literal types for `createSlice`

What else would we want to get in?

Existing milestone: https://github.com/reduxjs/redux-toolkit/milestone/7

Looking through the issues, other candidates:

- Could go in a 1.8.x:
  - #1853 / #1860
- Probably 1.9:
  - #1720 
  - #1668 
  - #1139 
  - #975
  - #1059 

For the record, I think this probably ought to be the last feature release before we look at doing RTK 2.0 and dropping IE11 compat ( ref: #958 )

phryneas:
maybe a `cAT.pinTypes/typeFactory` thing?

markerikson:
Unrelated but as a note for later: we should link the RTKQ "Essentials" tutorial pages from the RTKQ "Overview" page:

https://www.reddit.com/r/webdev/comments/sc6eoc/redux_toolkit_query_is_no_one_using_it/

phryneas:
@lindapaiste that would be a *very* welcome contribution!

markerikson:
Throwing this here for lack of anywhere better:

https://www.reddit.com/r/reactjs/comments/q6lit4/thoughts_about_redux_reactquery_together/i0s3f27/

> One thing i found that it needs extra config is SSR, if you are planning on doing SSR with something like Next for example, then you need to setup next-redux-wrapper and split your state, and combining reducers with the apiSlice while also supporting a global HYDRATE action to sync up server state with client state was a bit of a pain, we had to set a reducer case for HYDRATE for each slice, which can get big.
> 
> RTK Query looks promising but SSR is certainly a bit more painful compared to the "initialData" approach of react-query, one thing i do like is that you can still use endpoints defined within your apiSlice on the server instead of custom fetch functions.
> 
[...]

PR #2024, opened 2022-02-12
v1.8.0 integration

markerikson:
This is our ongoing integration branch for changes that will be released as part of RTK 1.8.

See #2003 for related planning and discussion.


FaberVitale:
### Docs feedback

I've taken a look at the docs: https://deploy-preview-2024--redux-starter-kit-docs.netlify.app/api/createListenerMiddleware#overview and I'm quite impressed:

They're **really** good 💯  but I've noticed that there are things that could be improved.

#### Suggested changes

##### 1.  Expand `clearListeners` and `removeAllListeners` description

We have to specify that we're also cancelling active listeners.

---

##### 2.  Rename `removeAllListeners` to `clearAllListeners`

I think that the name `removeAllListeners` is problematic:

it implies a behavior similar to `removeListener` but this function will not only unsubscribe all entries but also **cancel** all running listeners.

When we `remove` listeners we're  just  unsubscribing entries.
When we `clear` listeners   we're  **removing** subscriptions and **cancelling** active listeners.

I believe that we have to rename `removeAllListeners` to `clearAllListeners`.

markerikson:
Okay, yeah, I think that makes sense.

Issue #2098, opened 2022-03-05
Add type parameter for Id in createEntityAdapter


Matt-Ord:
Currently `createEntityAdapter` provides no type checking for the id of an entity

This prevents type checking when using the selectors, and means that the return type of functions such as `getAllIds` returns `EntityId[]` rather than a type specific to the entity.

Introducing an (optional) type parameter for Id will solve all of these propblems, and also allow for the newtype pattern to ensure types between different adapters do not get mixed up
```typescript
type MyIdType = string & { readonly __tag: unique symbol }
```

markerikson:
Hmm. @phryneas , thoughts?

Matt-Ord:
Just thought I would drop another message to see what the status was with this - I'm happy to give more details of my use case if it would be helpful

phryneas:
I'll have time this friday. 
Generally, something like this might be useful, but there is a very fine line balancing "adding additional generic parameters in many places" and "is the feature worth it" - since additional generic parameters become part of a contract we cannot change any more and have to support pretty much indefinitely.

markerikson:
Available in the 2.0 betas.

PR #2099, opened 2022-03-05
Add type parameter for Id in createEntityAdapter

Matt-Ord:
Fixes issue #2098
Let me know if this change also needs any docs/ I've messed up the system for creating a PR


phryneas:
I've gone over this - from my side it's okay @markerikson . 
I've also roughly gone over https://github.com/reduxjs/redux-toolkit/pull/948 in the context of this - it doesn't seem like the types changes would overlap too much, probably a few types would get one new type argument from each PR, but most would be affected by only one of the two PRs.

phryneas (approved):

Matt-Ord:
Do I need to do anything to update the docs- I think currently they have still have the old types

packages/toolkit/src/entities/models.ts
```diff
@@ -24,7 +24,7 @@ export type Dictionary<T, Id extends EntityId> = Partial<Record<Id, T>>
 /**
  * @public
  */
-export type DictionaryNum<T> = Dictionary<TemplateStringsArray, number>
```
phryneas: I think that with this PR, this type could be pretty much completely removed - or only be kept for backwards compat, but not really referenced any more/

phryneas:
Yeah, docs PRs would be great where you see them necessary 👍

Matt-Ord:
I think I've updated the docs where it's needed - it should be ready to go

markerikson:
Hiya. I'm looking at this, and it seems like it makes some decidedly breaking changes to the types: it looks like we _have_ to pass `, Id` everywhere (including existing public types), and it also rewrites the `Dictionary` type.

Given that, I don't think we can ship this in a 1.x release, unless I'm missing something.

Matt-Ord:
I would agree this is probably a breaking change - although I think it will be 'worth it' when you come to a 2.0 release, since it's solved a real problem on a medium to large codebase I've been working on.

Ideally I would say swap the order of Id and T (and do the same on all types) but this would make it 'more breaking'  - is this the issue you have with the dictionary type?

Matt-Ord:
Just seen this pr was closed - is this not something you want to change anymore? I'm happy to put the work in to update it if we're coming up to a 2.0 release?

phryneas:
I think this was just an accident.

markerikson:
Uh. Yeah, _I_ don't have permission to reopen it either! I think the issue is it was targeted to the `v1.9` branch.

PR #2112, opened 2022-03-10
update multiple requests with queryFn example

v-lax:
PR is related to the following discussion - https://github.com/reduxjs/redux-toolkit/discussions/2110 . First time contributing and not sure if I missed anything here. 

This updates the typescript example on the docs but not sure if there is also another file that had to be updated for the javascript example. If there is please let me know and I will update accordingly!

### DIFF
--- a/docs/rtk-query/usage/customizing-queries.mdx
+++ b/docs/rtk-query/usage/customizing-queries.mdx
@@ -939,7 +939,7 @@ const api = createApi({
       async queryFn(_arg, _queryApi, _extraOptions, fetchWithBQ) {
         // get a random user
         const randomResult = await fetchWithBQ('users/random')
-        if (randomResult.error) throw randomResult.error
+        if (randomResult.error) return { error: randomResult.error as FetchBaseQueryError } 
         const user = randomResult.data as User
         const result = await fetchWithBQ(`user/${user.id}/posts`)
         return result.data


docs/rtk-query/usage/customizing-queries.mdx
```diff
@@ -939,7 +939,7 @@ const api = createApi({
       async queryFn(_arg, _queryApi, _extraOptions, fetchWithBQ) {
         // get a random user
         const randomResult = await fetchWithBQ('users/random')
-        if (randomResult.error) throw randomResult.error
+        if (randomResult.error) return { error: randomResult.error as FetchBaseQueryError }
```
phryneas, line 942: No need to JS examples is needed, it will be generated from this. I'm not sure that the type assertion here is required though - are you sure about that?

v-lax: Gotcha! And thats a good question. Im actually not sure. I added it in because if `await fetchWithBQ(`user/${user.id}/posts`)` results in an error, then we assert the type of the error with `FetchBaseQueryError` . 

I also found this in this docs - https://redux-toolkit.js.org/rtk-query/usage-with-typescript#error-result-example. So im not actually sure 😅

markerikson:
Thanks!

Issue #2055, opened 2022-02-21
RTQ: How to unsubscribe from a `useLazyQuery` 


nichita-pasecinic:
Hello 👋

In documentation https://redux-toolkit.js.org/rtk-query/usage/usage-without-react-hooks#removing-a-subscription
is specified a different way of subscription - un-subscription method for a `query`, but how to we do it with a hook ?

```
const [trigger, { data }] = useLazyEnpointQuery();
```
I cant pass a `skip` property to `useLazyQuery` that forces the unsubscription from the query, it can be used only with the `useQuery` hook. So how can I unsubscribe from a lazy query ?

Thanks for help!

nichita-pasecinic:
Hi,
Any updates on it? wouldn't be nice if the `useLazyEndpointQuery()` returned something like `unsubscribe` here :
```javascript
const [trigger, { data, unsubscribe }] = useLazyEnpointQuery();
```

phryneas:
At the moment, the only way to unsubscribe from a lazy query is to start a new one with a different argument or to unmount the component. 

It might be a nice to have to have an unsubscribe option, but it's not high in the priorities. PRs against `useLazyQuerySubscription` are welcome.

bleedingAyush:
I think the feature is added.

markerikson:
Out in https://github.com/reduxjs/redux-toolkit/releases/tag/v2.4.0 !

PR #2133, opened 2022-03-17
Add possibility to manually unsubscribe from useLazyQuery #2055

damianborowy:
Related to issue [#2055](https://github.com/reduxjs/redux-toolkit/issues/2055)

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -1,6 +1,6 @@
 import type { AnyAction, ThunkAction, ThunkDispatch } from '@reduxjs/toolkit'
 import { createSelector } from '@reduxjs/toolkit'
-import type { Selector } from '@reduxjs/toolkit'
+import type { Selector, Unsubscribe } from '@reduxjs/toolkit'
 import type { DependencyList } from 'react'
 import {
   useCallback,
@@ -194,7 +194,8 @@ export type UseLazyQuery<D extends QueryDefinition<any, any, any, any>> = <
 ) => [
   LazyQueryTrigger<D>,
   UseQueryStateResult<D, R>,
-  UseLazyQueryLastPromiseInfo<D>
+  UseLazyQueryLastPromiseInfo<D>,
+  Unsubscribe | undefined
 ]
 
 export type LazyQueryTrigger<D extends QueryDefinition<any, any, any, any>> = {
@@ -239,7 +240,7 @@ export type UseLazyQuerySubscription<
   D extends QueryDefinition<any, any, any, any>
 > = (
   options?: SubscriptionOptions
-) => readonly [LazyQueryTrigger<D>, QueryArgFrom<D> | UninitializedValue]
+) => readonly [LazyQueryTrigger<D>, QueryArgFrom<D> | UninitializedValue, Unsubscribe | undefined]
 
 export type QueryStateSelector<
   R extends Record<string, any>,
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -835,16 +836,16 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
           ...options,
           skip: arg === UNINITIALIZED_VALUE,
         })
 
         const info = useMemo(() => ({ lastArg: arg }), [arg])
         return useMemo(
-          () => [trigger, queryStateResults, info],
-          [trigger, queryStateResults, info]
+          () => [trigger, queryStateResults, info, unsubscribe],
+          [trigger, queryStateResults, info, unsubscribe]
```
phryneas, line 848: I would suggest to make `unsubscribe` part of the `info` object and maybe give it a better name. Open to suggestions.

damianborowy: That's some good advice, I missed the fact that `info` was already an object for some extra things so I can touch it without breaking anything 😄 not sure if `extras` is the best possible name though, waiting for further feedback!

phryneas:
We just did a grooming of PRs and discussed this: would you be okay to to a PR for a `reset` method instead that unsubscribed *and* set the query hook back to `isUninitialized`?

It feels inconsistent to us that after unsubscribing, the query state will still stay in the component for another 60 seconds - or even forever, if another component is still using it.

Or do you have a good use case that would only make sense with `unsubscribe`, but not with `reset`?

henkkasoft:
> `reset`

This is exactly what I am looking for I think.

I am using useLazyQuery with query params given by user. 
The API response is HTTP ERROR 422 when the user input is incorrectly formed.
UI will show modal of the error and then I want to recover without unmounting the component. So I would want get rid of the isError state and the user could modify the input and trigger the query again.

Is this possible somehow already? 
And if not, did I understood correctly that this PR (when unsubscribe  is extended as reset) will enable this kind of usage?

henkkasoft:
> We just did a grooming of PRs and discussed this: would you be okay to to a PR for a `reset` method instead that unsubscribed _and_ set the query hook back to `isUninitialized`?
> 
> It feels inconsistent to us that after unsubscribing, the query state will still stay in the component for another 60 seconds - or even forever, if another component is still using it.
> 
> Or do you have a good use case that would only make sense with `unsubscribe`, but not with `reset`?

This is exactly what I am looking for I think.

I am using useLazyQuery with query params given by user.
The API response is HTTP ERROR 422 when the user input is incorrectly formed.
UI will show modal of the error and then I want to recover without unmounting the component. So I would want get rid of the isError state and the user could modify the input and trigger the query again.

Is this possible somehow already?
And if not, did I understood correctly that this PR (when unsubscribe is extended as reset) will enable this kind of usage?

phryneas:
@henkkasoft it will not remove the value, just unsubscribe from it. It will be removed after 60 second if no other component uses it.

Maybe you are giving `isError` too much importance here?  
As it sounds "there is an error from the server" will 100% be true until the data is sent to the server without a 422 - so maybe you just need an additional local state to control that message?

stchristian:
Hi, I see that there was no activity on this PR for a while now. Is this still pending? I would be happy to take over this PR and implement the above mentioned `reset` method
@damianborowy @henkkasoft

markerikson:
@stchristian yeah, no one has looked at this in a year. Might be best to re-create it and file a new PR with the updates

stchristian:
@markerikson Before starting this feature, I would suggest adding a `skip` parameter instead of `reset` method, similar to `useQuery()` that would work the same way setting the hook to `uninitialized`. This way the query stays consistent with the existing interface.

henkkasoft:
I was able to prevent invalidation of the query in isError state by adding error check to providesTags function:
`return error ? [] : ['MyTag']`
Now error state queries are not invalidated when 'MyTag' is invalidated in some mutation's invalidatesTags function.

Not sure if this is still the best solution...

markerikson:
Replaced by #4689

Issue #2136, opened 2022-03-17
`sortComparer` not running after action that updates sort order


dwoznicki:
I have an entity type that I want to sort by a `priority` number, so I have a custom `sortComparer` function on my entity adapter. However, when I update the `priority` number using a slice action/reducer call, the entity IDs are not automatically re-sorted.

Here's a sample slice.
```typescript
type Thing = {
  id: string;
  value: string;
  priority: number;
};
const thingsAdapter = createEntityAdapter({
  sortComparer: (a: Thing, b: Thing) => a.priority - b.priority
});

export const thingsSlice = createSlice({
  name: "things",
  initialState: thingsAdapter.getInitialState(),
  reducers: {
    thingsSet: thingsAdapter.setAll,
    thingsReordered: (state, action) => {
      const ids: string[] = action.payload;
      for (let i = 0; i < ids.length; i++) {
        const id = ids[i];
        const thing = state.entities[id];
        thing.priority = i;
      }
    }
  }
});
```

When I call `thingsSet`, the IDs are properly ordered by priority, even if I pass in an array of things in a different order. When I call `thingsReordered` and pass in an array of IDs in a new order, the priority value gets updated, but the IDs are not re-sorted.

Here's a sandbox with a demo: https://codesandbox.io/s/entity-adapter-ordering-issue-6w5f0e?file=/src/components/ThingsList.tsx

markerikson:
@dwoznicki : yeah, you're manually updating the state. Sorting only kicks in if you're using one of the entity adapter update methods, like `thingsAdapter.updateAll`.

dwoznicki:
Got it. So it sounds like I can probably do something like
```javascript
thingsReordered: (state, action) => {
    const ids = action.payload;
    const updates = ids.map((id, i) => {
        return {
            id,
            changes: { priority: i },
        };
    });
    thingsAdapter.updateMany(updates);
}
```
I'm not sure if it's implied, but it may be worth explicitly mentioning in the [`sortComparer` document](https://redux-toolkit.js.org/api/createEntityAdapter#sortcomparer) that sorting only fires after calling one of the adapter CRUD functions.

Thanks!

PR #2137, opened 2022-03-17
Added a quick note about when sorting is triggered

dwoznicki:
I think this may be implied, but I did not realize that sorting only fires when an adapter CRUD function is called. Hence #2136


markerikson:
Thanks!

PR #2145, opened 2022-03-18
Elaborate on the documentation of preloadedState

dggsax:
In the Reactiflux discord, I asked the following question:

> "Is there a reason why the `PreloadedState` typing is not made such that it's recursively partial? For my testing, I would like to be able to set a value for part of a piece of state, such as controlling `showNavigationPanel` but allowing the `initialState` for the slice go ahead and define the theme:"
```ts
const initialState = {
  appState: {
    theme: "Dark",
    showNavigationPanel: true
  }
}

// preloaded state for test, I only care about controlling showNavigationPanel, I want theme to just be handled by the initialState for the slice
const preloadedState = {
  appState: {
    showNavigationPanel: false
  }
}
```

To which @markerikson replied "Your slice reducers will be given the pieces of state you provide at startup. Those can't be partial themselves.".  Which is a pretty obvious answer if you think about it.

With this PR, I hope to provide some elaboration on `preloadedState` in the documentation so that others who come to this conceptual question can achieve the same understanding. This is my first attempt at contributing to the RTK documentation, so please let me know if you have any feedback on how I can make the explanations/examples more concise.

### DIFF
--- a/docs/api/configureStore.mdx
+++ b/docs/api/configureStore.mdx
@@ -53,7 +53,7 @@ interface ConfigureStoreOptions<
    * function (either directly or indirectly by passing an object as `reducer`),
    * this must be an object with the same shape as the reducer map keys.
    */
-  preloadedState?: DeepPartial<S extends any ? S : S>
+  preloadedState?: PreloadedState<CombinedState<NoInfer<S>>>
 
   /**
    * The store enhancers to apply. See Redux's `createStore()`.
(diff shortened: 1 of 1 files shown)


docs/api/configureStore.mdx
```diff
@@ -114,6 +114,96 @@ If the DevTools are enabled by passing `true` or an object, then `configureStore
...
+  theme: 'Dark',
+  showNavigationPanel: false,
+}
+
+type RootState = {
+  appState: AppState
+}
+
+// Not correct, we don't provide a value for `theme`. You would get the following typescript error if you tried to assign preloadedState the `PreloadedState<RootState>` type: Property 'theme' is missing in type '{ showNavigationPanel: false; }' but required in type '{ theme: "Dark" | "Light"; showNavigationPanel: boolean; }'.
+const incorrectPreloadedState = {
```
dggsax, line 182: @phryneas is there a way that I can give this the type `PreloadedState<RootState>` and have the type error be displayed in the documentation without breaking the compilation of the MDX?

dggsax:
Publishing this PR to start getting some feedback on what I've written here.

docs/api/configureStore.mdx
```diff
@@ -114,6 +114,91 @@ If the DevTools are enabled by passing `true` or an object, then `configureStore
...
+export const initialAppState: AppState = {
+  theme: 'Dark',
+  showNavigationPanel: false,
+}
+
+type RootState = {
+  appState: AppState
+}
+
+// Not correct, we don't provide a value for `theme`. You would get the following typescript error if you tried to assign preloadedState the `PreloadedState<RootState>` type: Property 'theme' is missing in type '{ showNavigationPanel: false; }' but required in type '{ theme: "Dark" | "Light"; showNavigationPanel: boolean; }'.
```
markerikson, line 176: :hand: This line should be broken into multiple separate `//` comments, so that there's no horizontal scrolling needed

docs/api/configureStore.mdx
```diff
@@ -114,6 +114,91 @@ If the DevTools are enabled by passing `true` or an object, then `configureStore
 
 An optional initial state value to be passed to the Redux `createStore` function.
 
+This allows you to override part (or all) of the initial state values that you have defined for each of your reducers and slices. This is useful for things like hydrating the state from `localStorage` or for [writing tests](https://redux.js.org/usage/writing-tests) for your Redux applications.
+
+> **Note**: `preloadedState` is not recursively partial! For each part of your application's initial state that you decide to provide initial values for, you must define values for **all** properties of that part of state. They will be passed to your reducers as the initial state, and they themselves can't be partial.
+
+Let's discuss the above point some more and provide a few tips for how you fully leverage `preloadedState` in your application and tests. For example, let's stay we start with the following initial state:
```
markerikson, line 121: :hand: The info from here on down is useful, but I don't think it needs to be part of the main flow of the page.

Over in the Redux core docs, we have a `<DetailedExplanation>` component that toggles display of content, as soon at https://redux.js.org/style-guide/#structure-files-as-feature-folders-with-single-file-logic and other spots.

We don't have that component in this repo, but it should be straightforward to copy-paste it over and wrap that around all the content after the "Note" block.

markerikson (changes requested):

docs/api/configureStore.mdx
dggsax: Perfect! I'll look into adding that. Thank you

EskiMojo14:
closing this as outdated - Redux 5 now explicitly types PreloadedState per reducer, instead of using a "magic brand" that makes `combineReducers` state partial.

PreloadedState can be any value that is valid to pass to the reducer as state without expecting it to return it.

For example
```ts
// reducer can receive string or number state but will only ever return number
const mySpecialReducer: Reducer<number, Action, string> = (state, action) => {
  if (typeof state === "string") state = Number(state);
  switch (action.type) {
    case "increment":
      return state + 1;
    default:
      return state;
  }
}

const nextState = mySpecialReducer("1", { type: "increment2 }) // nextState = 2
```

Sorry this took so long to get looked at!

Issue #2186, opened 2022-03-31
Promises returned from dispatching manually initiated RTK queries doesn't handle in flight requests correctly


wesen:
The title is a bit long winded. 

There is a long discussion in reactiflux' redux channel, starting around [here](https://discord.com/channels/102860784329052160/103538784460615680/958459043825983559).

Here is an example to reproduce the problem: [bug-repro](https://github.com/wesen/redux-toolkit/tree/hack/2022-03-30/debug-promise-resolution-with-rtk-query/examples/query/react/bug-repro)

The problem is when multiple queries are `initiate()`d, `dispatch()`d, and we then waited on, in quick succession. 
Some promises will resolve "too" soon, and not wait on a previously initiated query that actually does a fetch. 
The way to see this in the example posted above is clicking the button, and seeing the promises return a `pending` store state.

The problem is that only a single runningQuery is kept in the runningQueries cache to a query key in `buildInitiate.ts`.

I am running into this problem (which many other people probably don't) because I am replacing a legacy jquery app with rtk-query calls, and a lot of backend communication is done with async jquery.ajax call chains. As I was replacing these in the app (with `forceRefetch: true` or more refined invalidation mechanism), I realized that my `await store.dispatch(api.endpoints.getCart.initiate())` promises were out of lockstep with the actual `fetch()` calls.

PR #2187, opened 2022-03-31
Fix manually initiate()d rtk-query promises

wesen:
This PR keeps track of `runningQueries` in `buildInitiate`
by using a `Record<queryCacheKey, Record<requestId, Promise>>` type for
`runningQueries` instead of tracking a single request per cache key.

That way, when multiple queries are actually in flight for the same cache key,
we now call `Promise.all` on all of them, and thus get the proper store 
result once a `forceRefetch` query finishes.

This is a bit problematic in that the call `getRunningOperationPromise` doesn't
really fit its name. Arguably, this was already not the case before, as it
was returning `runningQueries[cacheKey]`, yet the actual promises handed out
to callers were already `Promise.all(...)` variants.

I also couldn't figure out how to write a proper unit test (this was a bit too much
for me at once).

This addresses #2186

### DIFF
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -1,14 +1,15 @@
 import type {
   EndpointDefinitions,
+  QueryDefinition,
   MutationDefinition,
   QueryArgFrom,
-  QueryDefinition,
   ResultTypeFrom,
 } from '../endpointDefinitions'
 import { DefinitionType } from '../endpointDefinitions'
-import type { MutationThunk, QueryThunk } from './buildThunks'
-import type { AnyAction, SerializedError, ThunkAction } from '@reduxjs/toolkit'
-import type { RootState, SubscriptionOptions } from './apiState'
+import type { QueryThunk, MutationThunk } from './buildThunks'
+import type { AnyAction, ThunkAction, SerializedError } from '@reduxjs/toolkit'
+import type { SubscriptionOptions, RootState } from './apiState'
+import { QueryStatus } from './apiState'
 import type { InternalSerializeQueryArgs } from '../defaultSerializeQueryArgs'
 import type { Api, ApiContext } from '../apiTypes'
 import type { ApiEndpointQuery } from './module'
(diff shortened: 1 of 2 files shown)


phryneas:
Without actually having tested it, this is what a test could look like:

```js
import { createApi } from '../core'
import { fakeBaseQuery } from '../fakeBaseQuery'
import { setupApiStore } from './helpers'

let calls = 0
const api = createApi({
  baseQuery: fakeBaseQuery(),
  endpoints: (build) => ({
    increment: build.query<number, void>({
      async queryFn() {
        const data = calls++
        await Promise.resolve()
        return { data }
      },
    }),
  }),
})

const storeRef = setupApiStore(api)

test('multiple synchonrous initiate calls with pre-existing cache entry', async () => {
  const { store, api } = storeRef
  // seed the store
  const firstValue = await store.dispatch(api.endpoints.increment.initiate())

  expect(firstValue).toMatchObject({ data: 0, status: 'fulfilled' })

  // dispatch another increment
  const secondValuePromise = store.dispatch(api.endpoints.increment.initiate())
  // and one with a forced refresh
  const thirdValuePromise = store.dispatch(
    api.endpoints.increment.initiate(undefined, { forceRefetch: true })
  )
  // and another increment
  const fourthValuePromise = store.dispatch(api.endpoints.increment.initiate())

  const secondValue = await secondValuePromise
  const thirdValue = await thirdValuePromise
  const fourthValue = await fourthValuePromise

  // this would be perfect, but will probably not be possible unless we try to detect synchronously rejected calls and handle their return value synchronously, too
  expect(secondValue).toMatchObject({
    data: firstValue.data,
    status: 'fulfilled',
    requestId: firstValue.requestId,
  })
  // this is more relalistic
  expect(secondValue).toMatchObject({
    data: firstValue.data,
    status: 'pending',
    requestId: thirdValue.requestId,
  })

  expect(thirdValue).toMatchObject({ data: 1, status: 'fulfilled' })
  expect(thirdValue.requestId).not.toBe(firstValue.requestId)
  expect(fourthValue).toMatchObject({
    data: thirdValue.data,
    status: 'fulfilled',
    requestId: thirdValue.requestId,
  })
})
```

PR #2193, opened 2022-04-02
docs(rtk-query): update axios base query example

hgb123:
- add params for query parameters (https://github.com/hgb123/redux-toolkit/pull/new/doc-update)
- add error message in case of network error (https://github.com/axios/axios/pull/205)


phryneas:
Thanks!

PR #2197, opened 2022-04-03
Docs: Correct Redux Batch usage in configureStore example

EskiMojo14:
See #2196 -  Redux Batch requires being at the beginning and the end of the enhancers to work properly.

This updates the example in configureStore's API documentation to match this, using the callback function.

I used `existingEnhancersArray` as the name for the parameter as that was what @markerikson used in our discussion in Reactiflux. (happy to change if something else is desired, obviously)

Currently I haven't added a comment explaining it in the doc example - if one is desired I'd love some assistance on the wording for that.


markerikson:
Heh, if anything maybe we ought to pick a different example enhancer here?

The only real point of `enhancers` here is to demonstrate that the field exists and that you can pass an array.

EskiMojo14:
> Heh, if anything maybe we ought to pick a different example enhancer here?
> 
> The only real point of `enhancers` here is to demonstrate that the field exists and that you can pass an array.

haha, happy to make that change instead - got any in mind? i've not used any previously, just middlewares

phryneas:
But at the same time, could you please update that example over in https://redux.js.org/introduction/ecosystem#batching ? :laughing:

EskiMojo14:
> But at the same time, could you please update that example over in https://redux.js.org/introduction/ecosystem#batching ? 😆

neat, didn't know about that page - i'll check through the other enhancers there and see any of them can use the array form, and correct the redux batch example on that page

PR #2200, opened 2022-04-04
Generate .mjs file to allow importing from Node.js in ESM mode (#1960)

### LINKED ISSUE #1960: Cannot import redux-toolkit from a Node.js ESM module
I have a bit of an unusual setup where I use redux-toolkit also in the backend of a Node.js application. I am currently in the process of migrating my backend to [ESM modules](https://nodejs.org/api/esm.html), since some dependencies (in particular [node-fetch](https://github.com/node-fetch/node-fetch)) are starting to ship only ESM modules.

# Error description

When I try to import redux-toolkit in an mjs module using `import { createSlice } from '@reduxjs/toolkit';`, I am receiving the following error:
```
SyntaxError: Named export 'createSlice' not found. The requested module '@reduxjs/toolkit' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@reduxjs/toolkit';
const { createSlice } = pkg;
```

The workaround suggested in the error message does work for Node.js mjs files. The problem is that the code where I use redux-toolkit is shared by the backend (running on Node.js) and the frontend (compiled using webpack). With the workaround in place, webpack can now not compile the file anymore and gives the following error:
```
export 'default' (imported as 'toolkit') was not found in '@reduxjs/toolkit' (possible exports: MiddlewareArray, __DO_NOT_USE__ActionTypes, applyMiddleware, bindActionCreators, combineReducers, compose, configureStore, createAction, createAsyncThunk, createDraftSafeSelector, createEntityAdapter, createImmutableStateInvariantMiddleware, createNextState, createReducer, createSelector, createSerializableStateInvariantMiddleware, createSlice, createStore, current, findNonSerializableValue, freeze, getDefaultMiddleware, getType, isAllOf, isAnyOf, isAsyncThunkAction, isDraft, isFulfilled, isImmutableDefault, isPending, isPlain, isPlainObject, isRejected, isRejectedWithValue, miniSerializeError, nanoid, original, unwrapResult)
```

# Reason for the error

redux-toolkit is bundled in several different formats, among them `cjs` and `esm`. The bundles are referenced in `package.json` in the following way (`index.js` being a wrapper that includes the `cjs` bundle):
```json
  "main": "dist/index.js",
  "module": "dist/redux-toolkit.esm.js",
```

While the `module` property is probably supported by webpack and other bundlers, it does not seem to be supported by Node.js. Instead, Node.js uses the `exports` property to support different main files for different environments (see [here](https://nodejs.org/api/packages.html#nodejs-packagejson-field-definitions). Since that is not defined in this case, Node.js requires the file from the `main` property, which is a CommonJS bundle.

Even forcing Node.js to use the ESM bundle by doing `import { createSlice } from '@reduxjs/toolkit/dist/redux-toolkit.esm.js';` does not solve the problem. The problem is that Node.js interprets files as CommonJS unless they have a `.jsm` file extension or `"type": "module"` is defined in `package.json` (which then applies to all files in the package) (see [here](https://nodejs.org/api/packages.html#determining-module-system)).

Node.js does support importing CommonJS packages in most cases, but in the case of redux-toolkit for some reason it doesn’t work. I am not sure why, but none of my other dependencies had this problem.

# Possible solution

Setting `"type": "module"` is probably not an option, since that will break the CommonJS files.

The only solution that I can think of is to ship the ESM bundle as an `.mjs` file, either by renaming the current one or by creating a copy. The file can then be referenced in `package.json` like this:
```json
  "exports": {
    "import": "./dist/redux-toolkit.esm.mjs",
    "require": "./dist/index.js"
  },
```

This solution does not solve the problem completely, as redux-toolkit uses immer, which has a similar problem. I have reported that as immerjs/immer#901.

# Workaround

Use the import like this:
```javascript
import * as toolkitRaw from '@reduxjs/toolkit';
const { createSlice } = toolkitRaw.default ?? toolkitRaw;
```

or in Typescript:
```typescript
import * as toolkitRaw from '@reduxjs/toolkit';
const { createSlice } = ((toolkitRaw as any).default ?? toolkitRaw) as typeof toolkitRaw;
```

cdauth:
This pull request fixes #1960 while attempting to not introduce any breaking changes.

In addition to each existing `*.esm.js` bundle, an identical `*.esm.mjs` bundle is created. This makes sure that existing apps/libraries that for some reason import the existing bundle by its filename don't break.

An `exports` map is added to the main `package.json` of redux-toolkit. This map makes sure that Node.js will use the CommonJS bundle in CommonJS mode and the ESM bundle in ESM mode. To not break any projects that import redux-toolkit in an unconventional way, a wildcard export `"./*": "./*"` is added to allow importing any sub paths as before.

Items for `/query` and `/query/react` are also added to the exports map. So far these have been their own directories with their own `package.json`. However, according to my research, the current layout is not supported by Node.js in ESM mode, because modules are required to live in `node_modules/<package>` or in `node_modules/@<org>/<package>`. Importing a package from a sub directory such as `node_modules/@reduxjs/toolkit/query` will raise an `ERR_UNSUPPORTED_DIR_IMPORT` error in Node.js. Adding these items to the exports map fixes these imports.

Unless I made some mistake, this pull request should not introduce any breaking changes, so it can be published with the next patch or minor release.

For the next major release, I think the `*.esm.js` bundles can be removed. It would also be worth to consider relying exclusively on the export map and removing the `query` and `query/react` sub packages.

### DIFF
--- a/packages/toolkit/package.json
+++ b/packages/toolkit/package.json
@@ -24,6 +24,21 @@
   },
   "main": "dist/index.js",
   "module": "dist/redux-toolkit.esm.js",
+  "exports": {
+    ".": {
+      "import": "./dist/redux-toolkit.esm.mjs",
+      "require": "./dist/index.js"
+    },
+    "./query": {
+      "import": "./dist/query/rtk-query.esm.mjs",
+      "require": "./dist/query/index.js"
+    },
+    "./query/react": {
+      "import": "./dist/query/react/rtk-query-react.esm.mjs",
+      "require": "./dist/query/react/index.js"
+    },
+    "./*": "./*"
+  },
   "unpkg": "dist/redux-toolkit.umd.min.js",
   "types": "dist/index.d.ts",
   "devDependencies": {
--- a/packages/toolkit/scripts/build.ts
+++ b/packages/toolkit/scripts/build.ts
@@ -53,6 +53,14 @@ const buildTargets: BuildOptions[] = [
     minify: false,
     env: '',
   },
+  // ESM, embedded `process`, ES5 syntax: typical Webpack dev, mjs extension
+  {
+    format: 'esm',
+    name: 'esm',
+    minify: false,
+    env: '',
+    extension: 'mjs',
+  },
   // ESM, embedded `process`, ES2017 syntax: modern Webpack dev
   {
     format: 'esm',
@@ -134,10 +142,11 @@ async function bundle(options: BuildOptions & EntryPointOptions) {
     name,
     target = 'es2015',
     entryPoint,
+    extension = 'js',
   } = options
 
   const outputFolder = path.join('dist', folder)
-  const outputFilename = `${prefix}.${name}.js`
+  const outputFilename = `${prefix}.${name}.${extension}`
   const outputFilePath = path.join(outputFolder, outputFilename)
 
   const result = await build({
--- a/packages/toolkit/scripts/types.ts
+++ b/packages/toolkit/scripts/types.ts
@@ -12,6 +12,7 @@ export interface BuildOptions {
   minify: boolean
   env: 'development' | 'production' | ''
   target?: 'es2017'
+  extension?: 'js' | 'mjs'
 }
 
 export interface EntryPointOptions {


packages/toolkit/package.json
```diff
@@ -23,7 +23,22 @@
     "access": "public"
   },
   "main": "dist/index.js",
-  "module": "dist/redux-toolkit.esm.js",
+  "module": "dist/redux-toolkit.esm.mjs",
```
rschristian, line 26: This is a breaking change, which doesn't seem to be intended. Older tools that rely on `"module"` don't all support `.mjs` unfortunately.

This is unnecessary anyhow; `"module"` has always meant ESM, so you can continue to supply `.js` module without worry to it. All tools will be able to handle it correctly. It's only Node that requires `.[cm]js`

cdauth: So Node.js is not a tool? ;-)

So would an acceptable compromise be to reference the `.esm.js` module in the `module` field and the `.esm.mjs` module in the `exports` field?

rschristian: Eh, the runtime/build tool distinction can be a bit silly at times ;)

Yes, that's really the best way to go about it for now. Of course that can be changed in a major (as shipping the same file twice, just with a different extension is a bit silly), but that ensures no breakage. Nothing will consume `"exports"` without being aware of & supporting `.mjs`

cdauth: That's good to know. Are you aware of any overview somewhere of which tools support which field(s) in `package.json`?

rschristian: Webpack's [Package Exports docs](https://webpack.js.org/guides/package-exports/) are the best I am aware of. Huge wall of text, but covers a good amount of conditions, formats, and which popular tools consume what.

nevace:
Would it be possible to get this merged please?

markerikson:
Frankly, given what happened with Immer, I'm very hesitant to make _any_ changes to our build system for the foreseeable future :(  And I'm busy enough with other things that this is not at all on my priority list.

At a minimum, I'd need to see someone put together some example projects that use the CodeSandbox CI build as a dependency, and show using RTK in several different environments (Node ESM, Next, Vite, whatever) and all working successfully, before I'd start to feel potentially comfortable merging this.

rschristian:
> Frankly, given what happened with Immer, I'm very hesitant to make any changes to our build system for the foreseeable future

Immer's issue was solved in a day or so, with the correct change being slated for the next major; not really much of an issue, no? Ideally avoided for sure, but wasn't a disaster.

Totally understand unease, and wouldn't change this in anything but a semver major myself, but right now RTK is simply invalid w/ ESM in Node. This is a big part of the ecosystem and growing quickly, due to the actions of some maintainers who've gone ESM-only. 

> show using RTK in several different environments (Node ESM, Node CJS, Next, Vite, CRA) and all working successfully

Keep in mind, some of these tools break from the spec rather often.... Next in particular I've seen _a lot_ of issues in over the years. The implementation there for resolving modules seems to be far from static so would say that one should expect some amount of harmless noise from these.

Probably also worth mentioning which versions of these tools you want to maintain support for?

markerikson:
As maintainer, my number one concern is **don't break existing code**.

I am also sympathetic with "RTK doesn't work in X environment" and want to support that.

But given how fragile all this stuff is, _and_ that I don't have time to spend on this myself, I need you or other people in the community to do the legwork here before I'm willing to merge this.  I need to see working proof of this branch compiling and running okay in as many different environment setups as possible, ideally exercising all the different combinations of build outputs and module formats, so that I can see this isn't going to suddenly break things.

rschristian:
That's totally fair, and I hope my comment didn't come off as demanding or suggesting this needs to be done! Wasn't my intention. I'm not even a Redux user fwiw (though I have enjoyed reading about it from you over the years), I don't have any need to see this land myself.

I thought I'd just hop in as I've done a fair bit regarding build tools (preact-cli, wmr, microbundle, etc) and have spent way too many hours dealing with the Node exports setup in different environments.

I'd definitely +1 shelving for a 2.0, even if this as-is theoretically shouldn't break any modern tools. And while I know you never get as many beta users as you'd like, this is definitely the sort of thing that'd be good to test out that way. Might just need some quick patches afterwards is what I wanted to warn about, depending on the tools and if you want to support their (often incorrect) configs.

packages/toolkit/package.json
qnighy: How about adding the "module" branch to make it safer for Webpack? In my understanding, it corresponds with the top-level "module" field.

```javascript
    ".": {
      "module": "./dist/redux-toolkit.esm.js",
      "import": "./dist/redux-toolkit.esm.mjs",
      "require": "./dist/index.js"
    },
```

rschristian: Safer in what way? Webpack (as well as every other tool, minus any bugs) can handle `.mjs` within `"exports"` just fine. `"module"` is really only useful if you want to provide a different ESM build to some bundlers vs Node.

qnighy: > Safer in what way?

Webpack behaves slightly differently for `*.mjs` (javascript/esm) and `*.js` (javascript/auto):

- It no longer completes the `.js` extension.
- It no longer looks for the `__esModule` flag.
- It no longer polyfills `global` by default.
- It no longer allows non-ESM metadata like `module.hot`.

They are all subtle but I proposed this because you seem to be careful about breaking changes.

rschristian: That's not "safer" though, just allows usage of older (and looser) ESM-ish builds. Not much of a reason to do that if the package is going to output proper ESM anyways.

If you wanted to still output the ESM-ish builds, and do so via `"exports"` for some reason, that could be a valid solution. I just don't think that would solve any issue here, and would only be an extra entry and output file.

Happy to be wrong though if you have an example situation in mind.

markerikson:
Superseded by the work I did in #3095 .

Issue #1939, opened 2022-01-19
RTKQ: refetch() can't be a promise ?


vadeneev:
Could you help me to get why RTKQ useQuery's refetch is not a promise-like thing ? (what is motivation)
I mean - I can't await it e.t.c
same for lazyQuery and it's trigger

phryneas:
The lazyQuery trigger can be awaited in 1.7. 
As for refetch, I'm not sure from the back of my mind - historically that was problematic in some situations when a fetch was already running, now it should be technically possible but never came up I think.

phryneas:
@jpcody but you already have a `isFetching` property on the query hook result?

jpcody:
Yes. which mostly fits my needs, but not always.

```tsx
// Component A - listing all dogs
const { data, isFetching, isLoading, refetch } = useGetDogsQuery()
<PullToRefresh onRefresh={refetch} refreshing={isFetching && !isLodading} />

// Component B - showing a single dog
const fetchDog, { isFetching, isLoading } = useLazyGetDogByIdQuery()
const refetch = useCallback(() => fetchDog(id), [id])
<PullToRefresh onRefresh={refetch} refreshing={isFetching && !isLodading} />
```

Component A actually behaves as expected, but component B doesn't, which leads me to want an async refetch.

In component B, the lazy query only returns `isLoading` on the first component load, not any subsequent loads of different components that use the same query. So in my case (in React Native), if I load a component that lazy queries for dog 1, then a new component that lazy queries for dog 2, the second component's `isLoading` will always return false. So on subsequent loads, `isRefreshing` will always evaluate as true and the refreshing state will appear instead of the loading state.

That lazy query behavior is unexpected to me, and perhaps I'm doing something wrong on my end. But given that behavior, in my pull-to-refresh component, I'd like to simply `await` the value I pass in `onRefresh`, which could be either a refetch function or a lazy query.

jpcody:
Sorry, this is difficult to communicate about. What's unexpected to me is that when a component is re-mounted, and thus the query is re-issued with a different argument and different data is fetched, `isLoading` does not fire again. So after I lazily fetch 1 dog in a component in the above example, then fetching a different dog via re-mounting the same component always reports `isLoading` as false.

awdyson:
If nothing else, consistency in RTKQ's explicit calls would be nice.
The submit method in there is ingested by a smart RHF form that disables controls while that promise is alive.
[...]

PR #2212, opened 2022-04-07
return promise from query result & hook `refetch`

phryneas:
See #1939. I'll update tests later.


joel1st:
This would be useful when incorporating RTKQuery into an existing codebase with some existing imperative fetching logic. Looking forward to this being merged :)

kahirokunn:
Looking forward to this being merged :)

Juneezee:
We should also update the type of `refetch` in this page https://github.com/reduxjs/redux-toolkit/blob/c663ba24edbff6aed69a9effc2eda7738e219ffb/docs/rtk-query/api/created-api/hooks.mdx. `refetch` now returns a Promise since `v1.9.0`. The `refetch: () => void` is outdated.

Issue #2213, opened 2022-04-07
Examples Not Working


ahmetcaglayan:
https://redux-toolkit.js.org/rtk-query/usage/examples

React Hooks and Svelte example not working.

phryneas:
Both work fine for me. For the svelte example, you have to click the popup in the bottom right at the right moment when it comes up.

ahmetcaglayan:
<img width="1373" alt="Ekran Resmi 2022-04-07 15 07 43" src="https://user-images.githubusercontent.com/13018803/162185844-6e81e999-5393-45ec-a198-4e6114f5ac5b.png">

Is there a problem with my computer?

ahmetcaglayan:
"dependencies": {
    "@reduxjs/toolkit": "^1.5.0",
    "@rtk-incubator/rtk-query": "https://pkg.csb.dev/rtk-incubator/rtk-query/commit/e4ccd12a/@rtk-incubator/rtk-query",
    "msw": "0.21.3",
    "react": "17.0.0",
    "react-dom": "17.0.0",
    "react-redux": "7.2.2",
    "react-router-dom": "5.2.0",
    "react-scripts": "3.4.1"
  },
  
  @rtk-incubator/rtk-query I think there is an access problem. Can you update it with an updated link?

phryneas:
As I said, I just tested it. It's a very outdated version of RTK Query and we should probably update those examples, but when you click on "Open Sandbox" they will just work fine. I have not tested them embedded, since I'm on the road and don't have the screen size - but it should just be an embed of the same thing.

markerikson:
Yeah, we really _should_ update these to use the latest RTK.  I think there may also be some API changes that need to be applied ("entites" -> "tags", etc).  Unfortunately it's also kind of low priority, and it would be really helpful if someone could help us with that.

sidwebworks:
@markerikson I would like to help with this, I saw your tweet about this issue.

markerikson:
Thank you!

It _should_ basically be going through the `examples` folder in our repo and seeing which ones still refer to the RTK "incubator" package, and updating those to import from the actual RTK Query package instead and also fixing any broken options usages.

Longer-term, there may also be some CodeSandboxes in our RTK docs that are standalone or pointing to the old alpha repo.  Those ought to be updated to, but first step is fixing the ones that _are_ in the main RTK repo.

sidwebworks:
Got it, I'll get started right away.

[...]

PR #2219, opened 2022-04-08
Fix rtk-query react-hooks example

sidwebworks:
This pull request addresses #2213

- Removed all occurences of `@rtk-incubator/rtk-query`
- Fixed the RTK-Query `react-hooks` example and ported it to the repo at `examples/query/react/kitchen-sink`
- Updated `examples.mdx`


markerikson:
Let's get this in!

PR #2232, opened 2022-04-11
feat: Add graphql error customization in graphql-request-base-query

### LINKED ISSUE #2231: [Feature] Set custom error object in rtk-query/graphql-request-base-query
Proposal Feature:
`graphqlRequestBaseQuery` shoud  give the option to set the error response object, using the graphql response

Possible example:
```
  baseQuery: graphqlRequestBaseQuery({
    client,
    customErrors: ({ name, message, request, stack, response }) => {
      return {
        name,
        stack,
        message: response?.errors[0]?.extensions?.errorMessage,
        errorCode: response?.errors[0]?.extensions?.errorCode
      };
    }
  }),
  endpoints: () => ({}),
  refetchOnMountOrArgChange: true
});
```

In the example the **`customError`** callback provides the error returned from the `graphql-response` client

antoniodinovi:
resolve #2231


packages/rtk-query-graphql-request-base-query/src/GraphqlBaseQueryTypes.ts
```diff
@@ -12,14 +13,23 @@ export type PrepareHeaders = (
   api: Pick<BaseQueryApi, 'getState' | 'endpoint' | 'type' | 'forced' | 'extra'>
 ) => MaybePromise<Headers>
 
+export type ErrorResponse = {
+  message: string;
+  stack: string;
+  name: string;
+  [key: string]: unknown;
```
phryneas, line 20: Do you have a specific reason for the index signature? I'd rather leave that one out. We could also think about adding a generic error type, but I'd suggest adding that at a later date if the need really should come up.

antoniodinovi: Hi, I use the index signature to generalize the error response starting from the base error. And in my app, the first thing I do was add another field related to a specific requirement.

Here, for example, I add a property **errorCode**
```    
...
customErrors: ({ name, message, request, stack, response }) => {
      return {
        name,
        stack,
        message: response?.errors[0]?.extensions?.errorMessage,
        errorCode: response?.errors[0]?.extensions?.errorCode    // <---------- ADDITIONAL FIELD
      };
    }
```

But I agree with you, you can leave out the index field and on my side, i will try to find a simple and consistent way to generalize the response type

antoniodinovi: I pushed an update

```
export type ErrorResponse = {
  message: string;
  stack: string;
  name: string;
};

export type GraphqlRequestBaseQueryArgs<E = ErrorResponse> = (
  ...
  customErrors?: (args: ClientError) => E;
}
```


And the inizialitazion of the client becomes (passing the error response type): 
```

  baseQuery: graphqlRequestBaseQuery<{foo: number}>({
    ....
    customErrors: (args) => {
      return {foo: 1, };
    }
  }),
  ...
});
```

Or without passing error response type

```

  baseQuery: graphqlRequestBaseQuery({
    ....
    customErrors: (args) => {
      return {m:essage: "", stack: "", name: " };
    }
  }),
  ...
});
```

phryneas:
Yup, that looks good to me :)

Given that I don't really have a project to try this one, have you been using this (including the new types) & it works on all fronts as expected?

antoniodinovi:
I use the same implementation in one of my projects. 
I try to replicate it in a code sandbox https://codesandbox.io/s/brave-elgamal-jl1x23?file=/src/pages/index.tsx you can see
 ```{foo: "....."}``` on screen .
The error mapping is defined in lib/baseApi.ts.
```
    customErrors: ({ message, stack, name }) => {
      return { foo: message };
    }
```
`lib/base-request.ts` and `lib/GraphqlBaseQueryTypes.ts` are just copy and paste from my latest commit.

phryneas:
Can you try it with the package built by CI?

```sh
yarn add @rtk-query/graphql-request-base-query@https://pkg.csb.dev/reduxjs/redux-toolkit/commit/74b966f1/@rtk-query/graphql-request-base-query/_pkg.tgz
```

antoniodinovi:
> Can you try it with the package built by CI?
> 
> ```shell
> yarn add @rtk-query/graphql-request-base-query@https://pkg.csb.dev/reduxjs/redux-toolkit/commit/74b966f1/@rtk-query/graphql-request-base-query/_pkg.tgz
> ```

I  think there is a problem in the build with the import of `GraphqlBaseQueryTypes.ts`. The customError callback works as expected but the typing is broken (i think also in the latest release). I try to push an update

image

phryneas:
You can always find the latest Ci builds here: https://ci.codesandbox.io/status/reduxjs/redux-toolkit/pr/2232/

packages/rtk-query-graphql-request-base-query/src/GraphqlBaseQueryTypes.ts
```diff
@@ -27,7 +27,7 @@ export type GraphqlRequestBaseQueryArgs<E = ErrorResponse> = (
 ) & {
   requestHeaders?: RequestHeaders
   prepareHeaders?: PrepareHeaders,
-  customErrors?: (args: ClientError) => E;
+  customErrors?: (args: ClientError) =>  Partial<E | ErrorResponse>;
```
phryneas, line 30: Can you explain the reasoning behind this? :)

antoniodinovi: Yes sure, (in the meantime the build has failed with this update :) ). Anyway using just E (that shoud be initialized to ErrorResponse) in VS-Code the autocomplete for the base case doesn't works

in the image below return `foo` should not be possible i'm expecting `{message, stack, name}`

image

But if i define the return type it works, (i return `foo` instead of `bar`)

image

So i tried to add an explicit return type to enforce typings for the base case

antoniodinovi: In the latest commit, i removed the logic discussed before. In my local project it works.
The typization on base case is not so strong (1 image), but if i define a `customError` function i shoud be free to raise the error in the format that i need.
And if i want to enforce the typization i can just pass the type to `graphqlRequestBaseQuery<{myProps: any}>`.
In my opinion it can be fine. Anyway i'm ever open to other suggestions to improve the code ;)

packages/rtk-query-graphql-request-base-query/src/index.ts
```diff
@@ -3,17 +3,18 @@ import type { BaseQueryFn } from '@reduxjs/toolkit/query'
...
+  ErrorResponse,
   GraphqlRequestBaseQueryArgs,
   PrepareHeaders,
   RequestHeaders,
 } from './GraphqlBaseQueryTypes'
 
-export const graphqlRequestBaseQuery = (
-  options: GraphqlRequestBaseQueryArgs
+export const graphqlRequestBaseQuery = <T = ErrorResponse>(
+  options: GraphqlRequestBaseQueryArgs<T>
```
phryneas, line 13:
```suggestion
export const graphqlRequestBaseQuery = <E = ErrorResponse>(
  options: GraphqlRequestBaseQueryArgs<E>
```

packages/rtk-query-graphql-request-base-query/src/index.ts
```diff
@@ -3,17 +3,18 @@ import type { BaseQueryFn } from '@reduxjs/toolkit/query'
...
 ): BaseQueryFn<
   { document: string | DocumentNode; variables?: any },
   unknown,
-  Pick<ClientError, 'name' | 'message' | 'stack'>,
+  Pick<ClientError, 'name' | 'message' | 'stack'> | T,
```
phryneas, line 17:
```suggestion
  E,
```

phryneas (changes requested):
I would assume it would work like this - could you verify that please?

packages/rtk-query-graphql-request-base-query/src/index.ts
antoniodinovi: By the logical point of view shoud works, but i'm tryng and unfortunately typization doesn't works. 
This because  typescript infer `customizedError` to
```
E | {
    name: string;
    message: string;
    stack: string;
}
```
But using generics initialization our type is just `E` that can be `ErrorResponse`


To make it works I added this (is more or less the part you drop)

```suggestion
  ErrorResponse | E,
```

antoniodinovi: The build fails on pipeline, I need to do some other check.

antoniodinovi: In the last version i apply your suggestions, but i need to force `customizedErrors` to `E`. 
I think the main problem is that `E` can be initialized to `unknown` by `BaseQueryFn`  and this cause some conflict.

antoniodinovi:
Hi, some news on this review ?

phryneas:
Looks solid, let's get this in :)

phryneas:
Published as @rtk-query/graphql-request-base-query@2.1.0.

Issue #2175, opened 2022-03-30
RTK Query: retry condition by http status code


kahirokunn:
It would be useful if the RTK Query provided retry function had an option to not retry depending on the HTTP Status.

It would be useful to be able to set both a common setting and a per endpoint setting for this.

It is not limited to HTTP Status, but there are cases where you may not want to retry some API Responses.

There is an idea to allow the user to pass a function that API Response takes as input value and return a boolean as the return value.

kahirokunn:
@phryneas If this looks good, I'll implement it myself. 👀

markerikson:
Sounds potentially useful, sure!

phryneas:
@kahirokunn please go ahead :)

kahirokunn:
done 💪 
https://github.com/reduxjs/redux-toolkit/releases/tag/v1.9.0-alpha.1

PR #2239, opened 2022-04-14
retry condition by http error response

kahirokunn:
https://github.com/reduxjs/redux-toolkit/issues/2175

### DIFF
--- a/packages/toolkit/src/query/retry.ts
+++ b/packages/toolkit/src/query/retry.ts
@@ -34,8 +34,8 @@ export interface RetryOptions {
    */
   backoff?: (attempt: number, maxRetries: number) => Promise<void>
   /**
-   Callback to determine if a retry should be attempted. 
-   Return `true` for another retry and `false` to quit trying prematurely.
+   * Callback to determine if a retry should be attempted.
+   * Return `true` for another retry and `false` to quit trying prematurely.
    */
   shouldRetry?: (error: FetchBaseQueryError, args: BaseQueryArg<BaseQueryFn>, extraArgs: {
     attempt: number
--- a/packages/toolkit/src/query/retry.ts
+++ b/packages/toolkit/src/query/retry.ts
@@ -34,7 +34,8 @@ export interface RetryOptions {
    */
   backoff?: (attempt: number, maxRetries: number) => Promise<void>
   /**
-   * Function used to determine need retry error
+   Callback to determine if a retry should be attempted. 
+   Return `true` for another retry and `false` to quit trying prematurely.
    */
   shouldRetry?: (error: FetchBaseQueryError, args: BaseQueryArg<BaseQueryFn>, extraArgs: {
     attempt: number
--- a/packages/toolkit/src/query/retry.ts
+++ b/packages/toolkit/src/query/retry.ts
@@ -37,7 +37,7 @@ export interface RetryOptions {
    * Callback to determine if a retry should be attempted.
    * Return `true` for another retry and `false` to quit trying prematurely.
    */
-  shouldRetry?: (error: FetchBaseQueryError, args: BaseQueryArg<BaseQueryFn>, extraArgs: {
+  retryCondition?: (error: FetchBaseQueryError, args: BaseQueryArg<BaseQueryFn>, extraArgs: {
     attempt: number
     maxRetries: number
     baseQueryApi: BaseQueryApi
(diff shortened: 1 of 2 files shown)


kahirokunn:
@phryneas How about this for implementation?
One concern is that I was quite confused about the name of the option.
If you have a more appropriate name, I would appreciate suggestions.
Thx.

kahirokunn:
@markerikson @msutkowski Can anyone give me a review? 😢

phryneas:
Sorry I'm just coming back to this so late.

I think we could make this a bit more generic.

What do you think about a method `shouldRetry(error, args, baseQueryApi, extraOptions): boolean` ?

kahirokunn:
@phryneas Agree 👍 
How about now?

locothedev:
Hey I was looking for such a function 
@phryneas can we have some eyes on this nice MR?

rayyan224:
Nice Feature To have!
@phryneas if all is well lets merge 🥳

packages/toolkit/src/query/retry.ts
```diff
@@ -57,7 +68,12 @@ const retryWithBackoff: BaseQueryEnhancer<
     try {
       const result = await baseQuery(args, api, extraOptions)
       // baseQueries _should_ return an error property, so we should check for that and throw it to continue retrying
-      if (result.error) {
+      if (result.error && options.shouldRetry(result.error as FetchBaseQueryError, args, {
```
phryneas, line 71: If we are at `retry === options.maxRetries`, this will still be called. 

Yet, if the function returns `true`, no other retry will be made (see line 82). This is unexpected behavior for the user as they specifically signalled that they want to have a retry.

We have the following options:

1. if `retry === options.maxRetries`, do not call `shouldRetry` any more since it will not have any effect either way:
  this is the easiest solution
2. if `retry === options.maxRetries`, still call `shouldRetry`. If it returns `true`, somehow signal to the error handler that even though at that point, `retry > options.maxRetries` is `true`, another retry needs to be made.

Either option would need to be sufficiently documented. I am personally leaning towards 2., but it would complicate the code a bit.  
If you agree that 2. would be the preferable solution, maybe we could just have a default definition for `shouldRetry` to be `(_, __, {attempt, maxRetries}) => attempt < maxRetries` and swap out some logic around the retrying mechanism?

phryneas (changes requested):
I left some comments - this will still need a decision to be made and some changes.

Either way, could you please also add some test cases for this?

packages/toolkit/src/query/retry.ts
kahirokunn: I chose 2.
I've tried to implement it.
If the way it looks good, I will add a test at the end.

kahirokunn:
@phryneas Thx! If the way it looks good, I will add a test at the end.

phryneas:
It looks good! 

Could you please give it a small rename to `retryCondition` (we just discussed that we want to be a *little* more consistent 😅 ) and add some tests?

kahirokunn:
@phryneas Is this perfect?

jmberrueta:
I came here looking specifically for what was asked at https://github.com/reduxjs/redux-toolkit/issues/2175.

One comment though, would it make sense to have this still be agnostic of `FetchBaseQueryError`?
In one of our projects we're using a custom `baseQuery`, and one nice thing about the current retry implementation is it works no matter the baseQuery's error type.

Then, supper minor comment, it felt strange that, if provided, the `retryCondition` implementation should take care of considering the `maxRetries`. I was initially expecting `maxRetries` to be evaluated independently of whether a retry condition was provided or not.

phryneas:
@jmberrueta 
> One comment though, would it make sense to have this still be agnostic of FetchBaseQueryError?
In one of our projects we're using a custom baseQuery, and one nice thing about the current retry implementation is it works no matter the baseQuery's error type.

That's good input, yes, that should inherit the `Error` of the `baseQuery` it is used with, not `FetchBaseQueryError` by default

> Then (supper minor comment) it felt strange that, if provided, the retryCondition implementation should take care of considering the maxRetries. I was initially expecting maxRetries to be evaluated independently of whether a retry condition was provided or not.

I get the initial confusion, but this would be the best escape hatch we could get to allow for a dynamic number of retries based on received data.

jmberrueta:
> I get the initial confusion, but this would be the best escape hatch we could get to allow for a dynamic number of retries based on received data.

Yes, but that could still be achieved by not providing a `maxRetries` while having the `retryCondition` still check for a maximum number of retries on its own, right?
But I agree, that might be too much of an opinion. Either way, this looks like a great feature to have!

phryneas:
That is a fair argument - it might make sense to allow specifying either `maxRetries` **or** `condition`, but not both 🤔

markerikson:
Ok, I _think_ this is updated to match Lenz's requests.

@phryneas final check on this before I merge it?

kahirokunn:
Sorry for the slow work, I have been too busy with work for the past two months.
It is already late, but I was thinking of the following code.

https://github.com/kahirokunn/redux-toolkit/commit/acce9c332ac08486d24c131c430ce7fe3d84f367

The code you wrote instead this time is very helpful.

markerikson:
Let's go ahead and get this in. @kahirokunn , thanks for the work here!

locothedev:
@kahirokunn Awesome stuff man

Thanks! @markerikson

packages/toolkit/src/query/retry.ts
```diff
@@ -63,14 +112,26 @@ const retryWithBackoff: BaseQueryEnhancer<
...
           return e.value
         }
 
         // We don't know what this is, so we have to rethrow it
         throw e
       }
+
+      if (
+        e instanceof HandledError &&
+        !options.retryCondition(e.value as FetchBaseQueryError, args, {
```
phryneas, line 127: This should be `e.value.error`, not `e.value`.

phryneas: fixed in #2743

PR #2245, opened 2022-04-16
feat: add useSuspendAll hook & react/suspense example

FaberVitale:
### Description

Proof of concept of a suspense for data fetching implementation.
It is based on `useSuspendAll` a custom hook that parallelizes suspended queries and outputs
input arguments with remapped TS types.

### Demo

- site: https://rtkq-use-suspend-all-demo.netlify.app/
- source: https://github.com/FaberVitale/redux-toolkit/tree/feat/use-suspend-all/examples/query/react/suspense/src
- codesandbox: 

---

#### 1. Simple single query

```ts
  const [{  data, isLoading }] = useSuspendAll(
    useGetPokemonByNameQuery(name),
  )

 //  data is defined here, isLoading is false & isSkipped is boolean
```

#### 2. Skipped Query

```ts
 import { skipToken } from '@reduxjs/toolkit/query/react';
 
  const [{  data, isLoading }] = useSuspendAll(
    useGetPokemonByNameQuery(skipToken),
  )

 //  data here is Data | undefined, isLoading is boolean and isSkipped is true
```

#### 3. Parallel queries

```ts
  const [baseDataQuery, evolutionQuery] = useSuspendAll(
    useGetPokemonByNameQuery(base),
    useGetPokemonByNameQuery(evolution)
  )

 //  baseDataQuery.data is defined here
 //  evolutionQuery.dara is defined here
```

#### 4. Error Handling

```tsx
import * as React from 'react'
import { ErrorBoundary } from 'react-error-boundary'
import Comp from 'components/Comp';

function CompWithErrorBoundary(props) {
  return (
    <ErrorBoundary
      fallbackRender={({ resetErrorBoundary, error }) => (
        <div onClick={() => { 
         (error as any)?.retryQuery?.()
          resetErrorBoundary()
       }}>retry</div>)}>
      <React.Suspense
        fallback={<div>loading ...</div>}
      >
        <Comp {...props} />
      </React.Suspense>
    </ErrorBoundary>
  );
}
```

### Technical details

`useSuspendAll` expects objects that implement the `Suspendable` interface as arguments:

```ts
export interface Suspendable {
  /** 
   * Returns a promise if there's a pending request, 
   * throws error in case of unsuccessful request
   * or returns undefined
   */
  getSuspendablePromise(): Promise<unknown> | undefined
}
```

#### Changes to existing APIs.

Adds the following properties to `api.endpoints.useQuery` output:

- getSuspendablePromise
- isSkipped: `true` if query is skipped

#### New Apis

- useSuspendAll
- SuspenseQueryError

---

### Todo

- [ ] React 18 testing
- [ ] Figure out why `err instanceof SuspenseQueryError` doesn't work
- [ ] Add more tests
- [ ] Handle edge cases
- [ ] Should we suspend other rtk-query hooks (e.g. useLazyQuery)?

---

### Alternatives

- [react-query suspense mode](https://react-query.tanstack.com/guides/suspense)
- [useSwr  suspense mode](https://swr.vercel.app/docs/suspense)
- [jotail](https://jotai.org/docs/basics/async)

### Previous attempts

- https://github.com/reduxjs/redux-toolkit/pull/2149


packages/toolkit/src/query/react/exceptions.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export class SuspenseQueryError extends Error {
+  constructor(
+    public reason: unknown,
+    public endpointName: string,
+    public retryQuery: () => void
+  ) {
+    super(computeErrorMessage(reason, endpointName))
+    this.reason = reason
+    this.name = 'SuspenseQueryError'
```
phryneas, line 27:
```suggestion
    super(computeErrorMessage(reason, endpointName))
    this.reason = reason
    this.name = 'SuspenseQueryError'
    Object.setPrototypeOf(this, SuspenseQueryError.prototype);
```

see https://www.typescriptlang.org/docs/handbook/2/classes.html#inheriting-built-in-types

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -93,11 +99,15 @@ export interface MutationHooks<
...
+  R extends Record<string, any> = UseQueryStateDefaultResult<D>,
+  Arg extends QueryArgFrom<D> | SkipToken = QueryArgFrom<D> | SkipToken
 >(
   arg: QueryArgFrom<D> | SkipToken,
   options?: UseQuerySubscriptionOptions & UseQueryStateOptions<D, R>
-) => UseQueryStateResult<D, R> & ReturnType<UseQuerySubscription<D>>
+) => UseQueryStateResult<D, R> &
+  ReturnType<UseQuerySubscription<D>> &
+  Suspendable &
+  IdleState<Arg>
```
phryneas, line 110: I'd suggest splitting this into an overloaded signature, one with `arg: SkipToken` and one without

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -502,6 +512,87 @@ type GenericPrefetchThunk = (
...
+          pendingPromise = api.util.getRunningOperationPromise(
+            name as any,
+            args
+          )
+
+          if (!pendingPromise) {
+            throw new Error(
+              `[rtk-query][react]: invalid state error, expected getRunningOperationPromise(${name}, ${queryStateResults.requestId}) to be defined`
+            )
+          }
```
phryneas, line 581: Do you see this ever happening? Also, are we saving another promise than `prefetch` returns into the pending promises? Otherwise we could reuse that.

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -502,6 +512,87 @@ type GenericPrefetchThunk = (
...
+    // We do not suspend if a query is skipped:
+    // @see https://github.com/vercel/swr/pull/357#issuecomment-627089889
+    if (!isSkipped && typeof queryStateResults.data === 'undefined') {
+      if (queryStateResults.isLoading) {
+        let pendingPromise = api.util.getRunningOperationPromise(name, args)
+
+        if (!pendingPromise) {
+          prefetch(args, {
+            force: true,
+          })
```
phryneas, line 570:
```suggestion
          retry()
```
or am I missing something?

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -848,17 +939,46 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
         })
+
+        const stableArg = useStableQueryArgs(
+          options?.skip ? skipToken : arg,
+          serializeQueryArgs,
+          context.endpointDefinitions[name],
+          name
+        )
+
+        const prefetch = usePrefetch(name as any)
```
phryneas, line 956: using prefetch here might be problematic since that creates a value subscription that cannot be unsubscribed. if we use that here we have to look into that

packages/toolkit/src/query/react/exceptions.ts
```diff
@@ -0,0 +1,29 @@
...
+  let message = `rtk-query suspense error ${queryKey}: `
+
+  if (reason instanceof Error) {
+    message += reason
+  } else if (typeof reason === 'object' && reason !== null) {
+    ;[reason?.status, reason?.code, reason?.error].forEach((value) => {
+      if (value) {
+        message += ` ${value}`
+      }
+    })
```
phryneas, line 11: please no `.forEach` in RTK code ^^
just use a for..of loop :)

packages/toolkit/src/query/react/suspense-utils.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export function useSuspendAll<
+  G extends SuspendableResource<any>,
+  T extends SuspendableResource<any>[]
+>(
+  ...suspendables: readonly [G, ...T]
+): UseSuspendAllOutput<readonly [G, ...T]> {
+  if (!suspendables.length) {
+    throw new TypeError('useSuspendAll: requires one or more arguments')
+  }
```
phryneas, line 55: Why not just allow calling it with an empty array? :)

packages/toolkit/src/query/react/suspense-utils.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  let promises = suspendables
+    .map(getSuspendable)
+    .filter(isPromiseLike) as Promise<unknown>[]
+
+  if (promises.length) {
+    throw Promise.all(promises)
```
phryneas, line 62: That would create a new promise each time, but we need to keep it stable. Maybe just throw the first unresolved promise? (We would need to add an `isResolved` property to the promises in that case)

phryneas:
This looks like a great approach, but with `prefetch` you have hit a hornet's nest of an old TODO that needs to be solved better ;)

I have skimmed the source and left some comments, but I have spent all day recording egghead video lessons, so my brain is mush. Very likely that some comments don't make sense or I missed some things.

But please keep going! 👍

packages/toolkit/src/query/react/suspense-utils.ts
FaberVitale: Well, it doesn't do anything if don't pass at least an argument but I guess that being too restrictive does not help much.

phryneas: I wanted to add a `useQueries` hook at some point that allows to call one query multiple times with an array of arguments, returning an array of results. With a dynamic amount of array elements there, 0 could be possible sometimes :)

packages/toolkit/src/query/react/buildHooks.ts
FaberVitale: > Do you see this ever happening?

No , it's an invariant that was useful while iterating on this feature. I can remove it.

> Also, are we saving another promise than prefetch returns into the pending promises? Otherwise we could reuse that.

usePrefetch() returns a consumer:

```ts
const prefetch: (arg: any, options?: PrefetchOptions | undefined) => void
```

I do not think that I can extract a promise from that.

I've skimmed a bit the type of `api.util.prefetch` and it seems that has type:

```ts
<EndpointName extends QueryKeys<Definitions>>(
      endpointName: EndpointName,
      arg: any,
      options: PrefetchOptions
    ): ThunkAction<void, any, any, AnyAction>
 ```

How can I extract a promise from that thunk action, I've not found a test that unwraps a promise.

Can you please provide more details?

File: packages/toolkit/src/query/react/buildHooks.ts (hunk 5 above)
Comment (FaberVitale) [reply packages/toolkit/src/query/react/buildHooks.ts:956]: Screenshot 2022-04-18 at 20 35 18


You're right :/

 I just need a way to fetch once and then to store  the result.

File: packages/toolkit/src/query/react/buildHooks.ts (hunk 3 above)
Comment (phryneas) [reply packages/toolkit/src/query/react/buildHooks.ts:581]: We would need a `return dispatch`  here everywhere this has only a `dispatch` to get that promise reliably: https://github.com/reduxjs/redux-toolkit/blob/e7a98eba3fbef2137109dc68caff305f603f4d1d/packages/toolkit/src/query/core/buildThunks.ts#L453-L468 

Perfectly viable to just add that.

File: packages/toolkit/src/query/react/buildHooks.ts (hunk 5 above)
Comment (phryneas) [reply packages/toolkit/src/query/react/buildHooks.ts:956]: I think at this point we probably need to overwork `prefetch` so that it takes something like a "keepFor" timer to add a time-based temporary subscription (maybe with a fixed subscription id of "prefetch" so that multiple prefetches just "bump up the timer").

I hate to bother you to go deeper into this rabbit hole - but could you look into this?

File: packages/toolkit/src/query/react/buildHooks.ts (hunk 3 above)
Comment (FaberVitale) [reply packages/toolkit/src/query/react/buildHooks.ts:581]: It's a breaking change

Comment (phryneas) [reply packages/toolkit/src/query/react/buildHooks.ts:581]: returning something from a function that returned `undefined` before? I wouldn't really call that breaking

File: packages/toolkit/src/query/react/buildHooks.ts (hunk 2 above)
Comment (FaberVitale) [reply packages/toolkit/src/query/react/buildHooks.ts]: done!

File: packages/toolkit/src/query/react/suspense-utils.ts (hunk 8 above)
Comment (FaberVitale) [reply packages/toolkit/src/query/react/suspense-utils.ts:51]: > That would create a new promise each time, but we need to keep it stable. Maybe just throw the first unresolved promise? (We would need to add an `isResolved` property to the promises in that case)

I've not seen suspense for data fetching implementation that relies on `isResolved` or similar.

> That would create a new promise each time

Yes but it is not a problem because you are suspending the rendering until all the promises are resolved and 
`runningQueries` does not keep promises that are settled.

Moreover it is the same approach used by [joitai waitForAll](https://github.com/pmndrs/jotai/blob/main/src/utils/waitForAll.ts)

File: packages/toolkit/src/query/react/exceptions.ts (hunk 1 above)
Comment (FaberVitale) [reply packages/toolkit/src/query/react/exceptions.ts:29]: https://github.com/reduxjs/redux-toolkit/pull/2245/files#diff-8b83915d0409b42b46a5dfdc3c600a5210168bd1ca5165fa0598ae4f820a66acR31

File: packages/toolkit/src/query/react/suspense-utils.ts (hunk 8 above)
Comment (phryneas) [reply packages/toolkit/src/query/react/suspense-utils.ts:51]: I remember reading in the working group that for suspense it is important to always get the same promise back, but right now I can't find the quote for it.

Comment (markerikson) [reply packages/toolkit/src/query/react/suspense-utils.ts:51]: Yep, pretty sure I saw that too, but don't immediately have a pointer either.

File: packages/toolkit/src/query/react/buildHooks.ts (hunk 5 above)
Comment (FaberVitale) [reply packages/toolkit/src/query/react/buildHooks.ts:977]: Done https://github.com/reduxjs/redux-toolkit/pull/2245/commits/f9944185dce95c781315e1d86d56a011c68a1fa5 :) 
Please let me know what you think 
I'm still not sure about adding a return type to prefetch but I'll look into it.

File: packages/toolkit/src/query/react/buildHooks.ts (hunk 3 above)
Comment (FaberVitale) [reply packages/toolkit/src/query/react/buildHooks.ts]: prefetch thunk and usePrefetch callback now return an object.

```ts
interface PrefetchActionCreatorResult {
  /**
   * Returns a promise that **always** resolves to `undefined`
   * when there are no pending requests initiated by input thunk.
   */
  unwrap(): Promise<void>,
  /**
   * Cancels pending requests.
   */
  abort(): void
}
```

packages/toolkit/src/query/react/suspense-utils.ts
FaberVitale: > Yep, pretty sure I saw that too, but don't immediately have a pointer either.

I'll look into it, I've got a couple of ideas on how to tackle this issue but I'd like to migrate the examples to react 18 first.

packages/toolkit/src/query/core/buildThunks.ts
```diff
@@ -435,37 +448,56 @@ In the case of an unhandled error, no tags will be "provided" or "invalidated".`
...
         )
-      const latestStateValue = (
-        api.endpoints[endpointName] as ApiEndpointQuery<any, any>
-      ).select(arg)(getState())
+
+        return {
+          unwrap() {
+            return initiateOutput.unwrap().then(noop, noop)
+          },
+          abort: initiateOutput.abort,
```
Hypnosphi, line 470: what about `unsubscribe`? Why not just `return initiateOutput` as is?

packages/toolkit/src/query/core/buildThunks.ts
```diff
@@ -435,37 +448,56 @@ In the case of an unhandled error, no tags will be "provided" or "invalidated".`
...
+            { forceRefetch, prefetch: options || true }
+          )
         )
-      const latestStateValue = (
-        api.endpoints[endpointName] as ApiEndpointQuery<any, any>
-      ).select(arg)(getState())
+
+        return {
+          unwrap() {
+            return initiateOutput.unwrap().then(noop, noop)
```
Hypnosphi, line 468: looks like this way you remove information about both fulfilled and rejected values

Hypnosphi:
Also, `initialize` accepts `{subscribe?: boolean}` option so maybe `prefetch` should as well

packages/toolkit/src/query/core/buildThunks.ts
FaberVitale: initiateOutput returns QueryActionCreatorResult

```ts
type QueryActionCreatorResult<
  D extends QueryDefinition<any, any, any, any>
> = Promise<QueryResultSelectorResult<D>> & {
  arg: QueryArgFrom<D>
  requestId: string
  subscriptionOptions: SubscriptionOptions | undefined
  abort(): void
  unwrap(): Promise<ResultTypeFrom<D>>
  unsubscribe(): void
  refetch(): void
  updateSubscriptionOptions(options: SubscriptionOptions): void
  queryCacheKey: string
}
```

It returns methods that do not fit the prefetch model like `updateSubscriptionOptions` and can cause more harm than anything else if misused.
I just want to expose the informations strictly useful.

Moreover we can always expose later more properties of initiate output if necessary, speaking of which we should at least expose `arg` and `queryCacheKey`.

packages/toolkit/src/query/core/buildThunks.ts
FaberVitale: Yeah it makes sense to at least return the output of unwrap

FaberVitale:
> Also, `initialize` accepts `{subscribe?: boolean}` option so maybe `prefetch` should as well

I'm afraid I do not agree with this change.

The whole point of prefetch is to put data in the store because you anticipate that you are going to query later.
It wasn't part of the prefetch api previously and I do not believe that it should be added with these changes.

Moreover there's already `initiate(arg, { subscribe: false })` if the user doesn't want put fetched data in the store.

FaberVitale:
### Prefetch PR

I'd like to move the prefetch changes (see https://github.com/reduxjs/redux-toolkit/issues/1283) in another PR.
This one has become a bit too big and it is hard to follow, review and develop.
I plan to keep on developing this PR once the prefetch changes get merged.

Hypnosphi:
Sorry, I didn't realize `{subscribe: false}` doesn't put data in the store, I thought it only opts out of automatic refetching. Is there any mechanism to do the latter without the former?

markerikson:
FYI, the React team just put up a PR that looks to add what might eventually be "official" support for Suspense triggering: `experimental_use(promise)`. 

It still throws internally, but only after introspecting the promise:

https://github.com/facebook/react/pull/25084

the PR says throwing directly would eventually be deprecated

FaberVitale:
> [...]
> [facebook/react#25084](https://github.com/facebook/react/pull/25084)
> 
> the PR says throwing directly would eventually be deprecated

---

### Caching promises

>Another feature is that the promises do not need to be cached between attempts. Because we assume idempotent execution of components, React will track the promises that were used during the previous attempt and reuse the result. You shouldn't rely on this property, but during initial render it mostly just works. Updates are trickier, though, because if you used an uncached promise, we have no way of knowing whether the underlying data has changed, so we have to unwrap the promise every time. It will still work, but it's inefficient and can lead to unnecessary fallbacks if it happens during a discrete update.


A shallow equalilty check of the pending promises list should be enough to avoid issues during updates.

markerikson:
Kevin Ghadyani put up a vid with him talking to Tanner about how React Query does Suspense.

Haven't even watched it yet, but throwing it here for reference:

https://twitter.com/Sawtaytoes/status/1567606755214983170

PR #2246, opened 2022-04-16
Remove Unused Import

adamaveray:
Removes an unused import from `tsHelpers.ts` that causes a `tsc` warning when `--noUnusedLocals` is enabled:

```
node_modules/@reduxjs/toolkit/src/tsHelpers.ts:1:27 - error TS6196: 'Dispatch' is declared but never used.

1 import type { Middleware, Dispatch } from 'redux'
```


markerikson:
Thanks!

PR #2260, opened 2022-04-19
Update createSlice.mdx

ravshan-samandarov-busuu:
Do not duplicate keys, use `name` property.

### DIFF
--- a/docs/api/createSlice.mdx
+++ b/docs/api/createSlice.mdx
@@ -273,8 +273,8 @@ const user = createSlice({
 })
 
 const reducer = combineReducers({
-  counter: counter.reducer,
-  user: user.reducer,
+  [counter.name]: counter.reducer,
+  [user.name]: user.reducer,
 })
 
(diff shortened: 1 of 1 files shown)


markerikson:
Hiya.  Appreciate the PR, but tbh I'm not overly fond of using the `[someSlice.name]: ` approach myself, especially when the names are fairly clear.

ravshan-samandarov-busuu:
@markerikson this pattern is already used in Query API examples as `api.reducerPath`. It's about being fond of or not, I just wanted to add consistency. But, still, thank you for quick response.

phryneas:
In the RTK Query examples, it is necessary that the reducer is always mounted at the exact name of `reducerPath`, because otherwise the middleware will not work - so it *has* to be equal and that's why we are showing it that way.

On the other hand, it's totally fine to have a slice with `name: 'users'` and mount it as `auth: users.reducer` into your Redux store if you please so.

PR #2261, opened 2022-04-19
Fix link to explanation about non-serializable vals

dominikdosoudil:
Former link redirected to 404

### DIFF
--- a/docs/usage/usage-guide.md
+++ b/docs/usage/usage-guide.md
@@ -1026,7 +1026,7 @@ export const usersAdapter = createEntityAdapter({
 
 ## Working with Non-Serializable Data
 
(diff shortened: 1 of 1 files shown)


phryneas:
That seems to be an error with out documentation build system currently - that link has been working for years and *should* be the right link.

dominikdosoudil:
@phryneas oh, okay, in that case this PR should be probably closed, right?

phryneas:
Yeah, probably. Thank you a lot for bringing this to our attention though!

markerikson:
FYI this turned out to be a new Docusaurus _feature_ :)  After some debate we ended up going with the changed URLs.  There was another PR that made this same fix, and I just reopened and merged that one.  Thanks for the help!

PR #2262, opened 2022-04-19
Chore: Fix broken link

txndai:
## Work done
- Update & test that new link is the right one

### DIFF
--- a/docs/usage/usage-guide.md
+++ b/docs/usage/usage-guide.md
@@ -1026,7 +1026,7 @@ export const usersAdapter = createEntityAdapter({
 
 ## Working with Non-Serializable Data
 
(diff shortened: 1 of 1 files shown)


markerikson:
Hiya! Appreciate the PR, but the change in URLs is actually a bug / problem with an upgrade to the latest Docusaurus version.  We _want_ to keep the URLs the same :)

Given that I'm going to close this, but thank you for noticing and helping out!

markerikson:
Well, guess what :)  Turns out this was actually a new Docusaurus _feature_, and we ended up deciding to go with the URL change.  So, I'll merge this. 

Thank you!

txndai:
@markerikson gotcha!

PR #2266, opened 2022-04-20
implemented upsertQueryData functionality per #1720 #2007

barnabasJ:
I used the comments in PR #2007 to implement a proof of concept for the upsertQueryData functionality. 

I copied and adapted the optimisticUpdate test code to test this. 

I added some comments to the `does update non existing values` test.

* I'm not sure about the behaviour if there are currently no active subscribers.
* The undo might not work as expected.

Please let me know what you think of this

### DIFF
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -13,7 +13,7 @@ import { QueryStatus } from './apiState'
 import type { InternalSerializeQueryArgs } from '../defaultSerializeQueryArgs'
 import type { Api, ApiContext } from '../apiTypes'
 import type { ApiEndpointQuery } from './module'
-import type { BaseQueryError } from '../baseQueryTypes'
+import type { BaseQueryError, QueryReturnValue } from '../baseQueryTypes'
 import type { QueryResultSelectorResult } from './buildSelectors'
 
 declare module './module' {
@@ -34,10 +34,11 @@ declare module './module' {
   }
 }
 
-export interface StartQueryActionCreatorOptions {
+export interface StartQueryActionCreatorOptions  {
   subscribe?: boolean
   forceRefetch?: boolean | number
   subscriptionOptions?: SubscriptionOptions
+  forceQueryFn?: () => QueryReturnValue
 }
 
 type StartQueryActionCreator<
@@ -179,6 +180,7 @@ export type MutationActionCreatorResult<
   unsubscribe(): void
 }
 
+
 export function buildInitiate({
   serializeQueryArgs,
   queryThunk,
@@ -259,7 +261,7 @@ Features like automatic cache collection, automatic refetching etc. will not be 
     endpointDefinition: QueryDefinition<any, any, any, any>
   ) {
     const queryAction: StartQueryActionCreator<any> =
-      (arg, { subscribe = true, forceRefetch, subscriptionOptions } = {}) =>
+      (arg, { subscribe = true, forceRefetch, subscriptionOptions, forceQueryFn } = {}) =>
       (dispatch, getState) => {
         const queryCacheKey = serializeQueryArgs({
           queryArgs: arg,
@@ -270,6 +272,7 @@ Features like automatic cache collection, automatic refetching etc. will not be 
           type: 'query',
           subscribe,
           forceRefetch,
+          forceQueryFn,
           subscriptionOptions,
           endpointName,
           originalArgs: arg,
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -34,11 +34,13 @@ declare module './module' {
   }
 }
 
-export interface StartQueryActionCreatorOptions  {
+export const forceQueryFnSymbol = Symbol('forceQueryFn')
+
+export interface StartQueryActionCreatorOptions {
   subscribe?: boolean
   forceRefetch?: boolean | number
   subscriptionOptions?: SubscriptionOptions
-  forceQueryFn?: () => QueryReturnValue
+  [forceQueryFnSymbol]?: () => QueryReturnValue
 }
 
 type StartQueryActionCreator<
@@ -180,7 +182,6 @@ export type MutationActionCreatorResult<
   unsubscribe(): void
 }
 
-
 export function buildInitiate({
   serializeQueryArgs,
   queryThunk,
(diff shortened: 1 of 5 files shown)


packages/toolkit/src/query/core/buildInitiate.ts
```diff
@@ -258,7 +260,7 @@ Features like automatic cache collection, automatic refetching etc. will not be
     endpointDefinition: QueryDefinition<any, any, any, any>
   ) {
     const queryAction: StartQueryActionCreator<any> =
-      (arg, { subscribe = true, forceRefetch, subscriptionOptions } = {}) =>
+      (arg, { subscribe = true, forceRefetch, subscriptionOptions, forceQueryFn } = {}) =>
```
phryneas, line 263: This should be hidden from user space, so
```
// somewhere else
export const forceQueryFnSymbol = new Symbol
```
```suggestion
      (arg, { subscribe = true, forceRefetch, subscriptionOptions, [forceQueryFnSymbol]: forceQueryFn } = {}) =>
```

packages/toolkit/src/query/core/buildThunks.ts
```diff
@@ -255,6 +265,62 @@ export function buildThunks<
       return ret
     }
 
+  const upsertQueryData: UpsertQueryDataThunk<EndpointDefinitions, State> =
```
phryneas, line 268: I'm not sure if we need to go with the full "updateRecipe" here. I think it would also be fine to have the user say "this is the data, put it into the endpoint or overwrite everything that is there".
Or do you think it would be used a lot?

barnabasJ: It just felt the most natural to me. I don't necessarily know if the other query already has data or not, and this allows me to handle both cases easily. In cases where the data is only a primitive / simple value you can just return the new value from the recipe, but it is also possible to reuse data if the draft is already there. Otherwise I would need to manually check if there is already cached data or always specify both updateRecipe and upsertValue.

packages/toolkit/src/query/core/buildThunks.ts
```diff
@@ -255,6 +270,62 @@ export function buildThunks<
...
+          ret.patches.push({ op: 'replace', path: [], value })
+          ret.inversePatches.push({
+            op: 'replace',
+            path: [],
+            value: currentState.data,
+          })
+        }
+        dispatch(api.util.patchQueryData(endpointName, args, ret.patches))
+      } else {
+        ret.inversePatches.push({
```
phryneas, line 305: I'm not sure about this part tbh. - undoing an upsert would be what exactly? 
Keeping the cache entry but replacing the value with `undefined` seems wrong. Removing the cache entry could have even more implications.

=> I would suggest that this method is not in any way undoable. Just chuck a value in there and be done with it.

In the same sense, I'm not sure if we really need an "upsertRecipe". Tbh., I'd be perfectly fine with this method just "putting a value into the store, overriding what is there". 
To me it feels like it's not really the place for fine-grained, undoable updates.

I'm open to arguments though - do you think it is a good idea - do you have use cases?

barnabasJ: Changing the signature of `updateQueryData` is probably also not desirable because it would mean a breaking change. I think the insertMethod could be something that is used with pessimistic updates only, but I think a different name would be more appropriate in this case. Maybe something along the lines of `primeQueryCache` but that would only signal the use case of adding data to an empty cache. 

Maybe `primeQueryCache` should only do something if the query wasn't executed before. This way you could call `updateQueryData` in an optimistic fashion and if the use case is that data should also be entered if the cache is empty `primeQueryCache` can be called after the query has successfully completed. This way the user also wouldn't need to check the current query states themselves.

barnabasJ: Thanks for the feedback. 
I should be able to add the changes by the end of the week.

barnabasJ: Something came up on the weekend, will make the changes in the next couple of days

phryneas:
That's looking pretty good right now!

One scenario I'm concerned about: What happens/should happen when `upsertQueryData` is called for a cache entry that is currently "in flight"? 

From the back of my head: With this current solution, the request in flight would continue and `upsertQueryData` would not have any effect (due to the `condition` of `initiate`).

We could change `condition` and in that case, the value from `upsertQueryData` would make it into the cache, but the result from the query would probably never make it in the state. 

Now the question: I think the second behaviour feels like the intended behaviour - after all, the `upsert` call was made after, which indicates that we consider that that is the more "up to date" data. @barnabasJ do you agree? @markerikson do you have thoughts on that?

phryneas:
(also I'm not sure on the `emptyCache` name - maybe more along the lines of `cacheEntryFound`?)

phryneas:
Problem is that with the current underlying logic, a reducer will only ever react to the last started request for a cache key and ignore all requests that would possibly run otherwise.

https://github.com/reduxjs/redux-toolkit/blob/8c23cd33971ae5a62bc19764adca6f6be2ed98ec/packages/toolkit/src/query/core/buildSlice.ts#L158

So either we have this start a "pending" and then the according "fulfilled" will be accept (but the one currently running will be ignored), or we don't start this and it will not make it into the cache.

Or we revamp all that logic, of course.

markerikson:
Huh. I'm not sure this PR is actually running any tests properly.

I just did a rebase against `v1.9-integration`, and removed the "add/revert" commits in the middle.

Looks like it's trying to run actual tests now.

Also, I don't think the tests got updated to match the "whole values, not `updateRecipe` callbacks" behavior. working on that now.

markerikson:
Ok, think I've got the tests updated to match what the current implementation logic is actually doing.  

Need to think about the "request in flight case" still, but at least the tests should be accurate.

markerikson:
So right now it looks like we're actually hitting this check inside of `queryThunk` if you attempt to call `upsertQueryData` while a request is in flight:

```ts
      // Don't retry a request that's currently in-flight
      if (requestState?.status === 'pending') return false
```

Because of that, the `queryThunk` for `upsertQueryData` bails out entirely, right away.

I've modified the logic so that the `if (isForced()) return true` check runs first instead.  Given that, the `upsertQueryData` call at least runs.

I've also tried changing the reducer logic to start with:

```ts
        .addCase(queryThunk.fulfilled, (draft, { meta, payload }) => {
          updateQuerySubstateIfExists(
            draft,
            meta.arg.queryCacheKey,
            (substate) => {
              if (substate.requestId !== meta.requestId) {
                if (
                  substate.fulfilledTimeStamp &&
                  meta.fulfilledTimeStamp < substate.fulfilledTimeStamp
                ) {
                  return
                }
              }
```

So, we _should_ end up going with whatever request resolves last.

I don't think this is necessarily "right" or "best" behavior, but I at least know what it _does_ now.

Having said that, I'm still not sure what the _final_ behavior sure be here and would appreciate suggestions.

markerikson:
Okay. This runs, tests pass, it compiles clean.

Let's merge this in and put out `1.9.0-alpha.1` so people can try it out, and we can tweak the edge case behavior if needed.

blaine-foreflight:
Hey! So glad I found this! Exactly what I needed. Just wanted to see if there was any documentation coming out with the release of this feature? I looked [here](https://deploy-preview-2401--redux-starter-kit-docs.netlify.app/) and didn't see any results when I searched for `upsertQueryData`

markerikson:
@blaine-foreflight : it's here:

https://deploy-preview-2401--redux-starter-kit-docs.netlify.app/rtk-query/api/created-api/api-slice-utils#upsertquerydata

I don't think Algolia indexes previews

PR #2273, opened 2022-04-22
Update overview.mdx

flexbox:
just to align naming `builder` and `build` 

like in
- https://redux-toolkit.js.org/rtk-query/usage/queries
- https://redux-toolkit.js.org/rtk-query/usage/mutations


markerikson:
Thanks!

Issue #1937, opened 2022-01-18
RTK Query: UseQuery return type seems to be faulty


ulrikap:
Hello,

I stumbled upon this issue while implementing RTK Query at my workplace. It seems like the return type of UseQuery hooks in interfaces results in a type that is different from the result of the hook when invoking it.

**Example** 

Given a regular setup like this:

baseApi
```ts
export interface IBaseApiEndpoints extends EndpointDefinitions {
  [key: string]: TBaseApiEndpoint<unknown, string, unknown>;
}

export const baseApi = createApi({
  reducerPath: "api",
  baseQuery: fetchBaseQuery({
    baseUrl: getApiUrl(),
  }),
  tagTypes,
  endpoints: () => ({}),
  
});
```

domainSpecificApi
```ts
const domainSpecificApi = baseApi.injectEndpoints({
    endpoints: build => ({
        fetchMyData: build.query<TSpecificReturnType, IArgs>({
            queryfn: (args) => someApiCall(args),
            providesTags: (_, __, params) => [{type: 'SomeTestTag', id: params.id}]
        })
    })
})

export const {
   useFetchMyDataQuery
} = domainSpecificApi

```
given the above setup (hope this is understandable)
`ReturnType<typeof useFetchMyDataQuery>`
does not result in the same type as this object receives.
`const data = useFetchMyDataQuery(param)`.

The first type results in a object with a single key `refetch` while the last object has all the fields defined in https://redux-toolkit.js.org/rtk-query/usage/queries#frequently-used-query-hook-return-values

**Question**
What is the preferred way to define a query hook result in a interface? In my current setup the goal is to get the result of a RTK hook and pass it as a argument to another function, preferably typed.

**Versions**
`"typescript": "4.5.4"`
`"@reduxjs/toolkit": "1.6.2"`

**tsconfig**
```
{
  "compilerOptions": {
    "allowJs": false,
    "target": "ESNext",
    "moduleResolution": "node",
    "noEmit": true,
    "strict": false,
    "esModuleInterop": true,
    "noImplicitReturns": true,
    "noUnusedLocals": true,
    "module": "esnext",
    "jsx": "preserve",
    "resolveJsonModule": true,
    "allowSyntheticDefaultImports": true
  },
}
```

markerikson:
We did end up adding a `TypedUseQueryHookResult` type to RTKQ.

PR #2276, opened 2022-04-24
add types for typed hook results

phryneas:
This has come up quite a bit, as just calling `ReturnType` on the hooks does not work - after all, depending on how the hook is called, it could be any kind of result.  ( a lot in chat, but also for example in #1937).

This PR exports some types that can be used to manually declare these types.

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -52,6 +52,7 @@ import { useStableQueryArgs } from './useSerializedStableValue'
 import type { UninitializedValue } from './constants'
 import { UNINITIALIZED_VALUE } from './constants'
 import { useShallowStableValue } from './useShallowStableValue'
+import { BaseQueryFn } from '../baseQueryTypes'
 
 // Copy-pasted from React-Redux
 export const useIsomorphicLayoutEffect =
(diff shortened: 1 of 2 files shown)


u900x600:
Thanks to @phryneas I ended up here here after some talk on the chat. Our use case to have access to hook result types is fairly simple: We have queries which depend on each other and we expect to have a bunch of them coming up. To illustrate that a bit:

```
import { createApi } from '@reduxjs/toolkit/query/react';

  …
  endpoints: b => ({
    getA: b.query<{aid: string}, void>(…)
    getB: b.query<{bid: string}, { aid: string }>(…)
  })
});

export {
  useGetAQuery,
  useGetBQuery
} = api
```

in order to `getB` we need to `getA` in advance and pipe the result as argument into `getB`. The simple solution to this is to create a third hook, that does the chaining with the help of the `skip` or `skipToken`. Assuming that there are more endpoints getC ... getZ  plus mutations postB … postZ which all depend on the data provided by `getA` this process can easily get repetitive and errorprone.

To come around this a »Higher Order Query Hook« will come in handy. So we started doing that and with the help of the source and the discord channel, I was able to setup something along those lines:

```
import { QueryDefinition, MutationDefinition } from '@reduxjs/toolkit/query';
import { UseQuery, UseMutation } from '@reduxjs/toolkit/src/query/react/buildHooks';

type QDef<QueryArg, ResultType> = QueryDefinition<QueryArg, any, string, ResultType, string>;

type MDef<QueryArg, ResultType> = MutationDefinition<QueryArg, any, string, ResultType, string>;

type QueryHook<A, B>  = UseQuery<QDef<A, B>>;
                                        | UseMutation<MDef<A, B>>;

const withAData = <A extends object, B>(query: QueryHook<A & { aid: string }, B>): QueryHook<A, B> => (arg, option) => …
```
**But** the above doesn't work with Query or Mutation, since there Return Type doesn't match. So, for now we have it working for queries, but to make it work for mutations as well we would need some more control over the return type.

Additionally it is cumbersome to use the existing types to create a proper constraint on the input argument(s).

The ability to create properly types abstractions over those query hooks would allow the creation of many primitives over queries: merging data, getC if getA otherwise getB - it would basically be the foundation to unlock reactive query combinators.

PR #2277, opened 2022-04-24
ApiProvider: correctly register listeners, allow to disable listeners

phryneas:
As mentioned in #2003, might even be worth to put it into a 1.8 patch depending how long we want to wait until 1.9.

### DIFF
--- a/packages/toolkit/src/query/react/ApiProvider.tsx
+++ b/packages/toolkit/src/query/react/ApiProvider.tsx
@@ -1,5 +1,5 @@
 import { configureStore } from '@reduxjs/toolkit'
-import type { Context } from 'react'
+import { Context, useEffect } from 'react'
 import React from 'react'
 import type { ReactReduxContextValue } from 'react-redux'
 import { Provider } from 'react-redux'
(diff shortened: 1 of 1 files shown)


packages/toolkit/src/query/react/ApiProvider.tsx
```diff
@@ -45,7 +45,13 @@ export function ApiProvider<A extends Api<any, {}, any, any>>(props: {
...
   )
   // Adds the event listeners for online/offline/focus/etc
-  setupListeners(store.dispatch, props.setupListeners)
+  useEffect(
+    (): undefined | (() => void) =>
+      props.setupListeners === false
+        ? undefined
+        : setupListeners(store.dispatch, props.setupListeners),
+    [props.setupListeners]
+  )
```
msutkowski, line 54: Hmmm, do we need to rework `setupListeners` and `initialized` make sure we're cleaning up the subscriptions here if we're putting it in an effect?

markerikson: Actually, that's a good point. Especially with React 18 double-running effects in dev.

phryneas: `setupListeners` returns an `unsubscribe` function so this should be fine as it is.

markerikson:
Let's put this in `master` now.

PR #2289, opened 2022-04-27
Update code example in RTK-Query docs

### LINKED ISSUE #2288: RTK-Query "Usage with Typescript" error code example broken
There appears to be a copy/paste error in the code example for using the errors from `fetchBaseQuery`. The error was destructured as `error` but is referenced as `err`, which appears to have been copied from a code example down below.

thezanke:
Fixes #2288


markerikson:
Thanks!

PR #2308, opened 2022-05-04
Re-export cast utility functions from immer

Methuselah96:


markerikson:
Yo!  Any particular reason we need to re-export these specifically?

Methuselah96:
They're necessary to make reducers type-safe in the situations mentioned [here](https://immerjs.github.io/immer/typescript#cast-utilities) (most commonly if there's a `readonly` array in the state) and it would be nice to be able to use them without having to add a dependency on Immer.

Methuselah96:
I can't think of a situation where I would use `castImmutable` in relation to RTK, so I would be fine with removing that one and just re-exporting `castDraft`.

phryneas:
Mark, Matt and I just discussed this a bit - we feel like this just doesn't happen often enough to add it as a re-export, and if it actually happens, adding `immer` as a dependency might actually be a better choice than us adding it for everyone.

Methuselah96:
How much it's used is highly dependent on how often one uses read-only arrays. I understand that most people aren't that strict about it and so it doesn't come up that often for those people, but I'm not sure I understand the downside of including it for those of us who like the strictness of making their state types fully read-only. I have no other reason to add immer as a dependency, so it would be really nice to not have to add it as a dependency just for this.

Methuselah96:
I'd be happy to just cut it down to just re-exporting `castDraft`, but it seems like a pretty straightforward re-export. Is there something particular about this re-export that you're concerned about long-term?

Methuselah96:
@markerikson @phryneas Any more thoughts on what the concerns are for adding the export? Am I right in thinking that most people don't run into this because they're not using readonly arrays? I made a [TS playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgYwgOwGbAOYFcoCmAyjNAQDQqECGMxANsMgXAL5wZQQhwDkAAoQAmuAB4ArAM4B6UhHoBrYDF4AoUJFhwYATzAskABWo76EakICCyGMHRsOXHgOFipsiPKUr14aPCRcSWICegIbaAdObj4aGwBaV1E1DX9tPQM4ABUMoQBVYKJQ8NIoAAlPBSinWIJqBKS1dTQ6KAx6lgA1anpcAgBheTCbOzREVTg4aWlsjPjJagwCAC4JuBohdHodODRcEAAjAihJZfW6zbRt3f2jqABtAF1VNem4ADkIeF19ecWV14zDZbHZ7Q7HU43cEPZ6sF6oNCSeDANDKYA9Ei0FbnCwguDdXoDIYlUZPOAAXjgTwA3C8CKJNPAEUi4AA3Hp9Qb0Ya2dCSIiMZgUqh1OgCpgEAAUCDWaGoIGxvHZhK5PNGkl45DWKLRGJgWK1k1czBOZxlk0mFiEqpJ6DOkqRBrg9V5aDOxlM5isI3QAB4CZziT60AA+ACUFJD4wtFsddAAdGAggALSUu0aJkxmCxhl4xyZvADqEFw9CEcEYChYpDgR2dBzC2ggKB69FW+amMzjBETKclyGoSIAIlBFjA08HM56c7mY6xDWwtaxc6pmQFnUJrUHXQ5KcrA9zbYjxcx4+m+fC+fB90TD8HJAAlAgiE3Cm82+8nnvG46XxHwJEyGFBEsDwQgSDIaU1h-KAzTWSZ323dUzkQu9XUfZ9cBNBd51UZcXh+FgH08GBMToYUnxgfA0ByfRfUIiAMDgQDCHjbACFI-U6BDVcrzgIICEsMAwCKHloDOWjnwKEIxPKSpfWIr4yIIKNKQE0SSmgP8WSwE4YADW81XsNTgiEkTigiKAHS4lhySjbt41QozEXuAAGR4VzefgYEkeJ6X0BJji4KBsmTYBJGY5MSzLWsWGoMYgugeNVF0pEDI-V14zBO5JF7SRUwAVhXK0MtGaUoRys57gKygAGZKAAJkeNhcyAA) demonstrating the issue. If you make the array non-readonly you'll see that it makes it possible to mutate it when using the state somewhere else in the app. I run into the need to use `castDraft` very frequently and I see people on my team just using a non-readonly array just to get rid of the error, which leads to less type-safety. I think it would make sense to include `castDraft` to encourage readonly state types where possible, or just to throw a bone to those of us who care about it. :)

markerikson:
It's a combination of a few things:

- Anything we add to the public API lives there forever
- We _try_ to aim for the "80% common use case" scenario, and I'm not sure this qualifies
- I'm also trying to avoid turning RTK into "export * from Immer and Reselect", which is why we also don't re-export `createStructuredSelector`
- This is the first I've heard of `readonly` types being an issue, and thus far you're the only person who's brought this up :)

phryneas:
It's not about the complexity of the function - but what if `immer` goes from 8 to 9 and removes it because they now do this on a type level with some new pattern? We'd need a major release to go to `immer` 9. 

There are things that are central to RTK where we can't be picky, and then there is stuff like this where risks like that have to be weighed a bit more.

Also, it's not free - if the consumer's bundler doesn't tree-shake well, it adds another 10 bytes to the library - no matter if actually used or not.

Methuselah96:
> It's not about the complexity of the function - but what if `immer` goes from 8 to 9 and removes it because they now do this on a type level with some new pattern? We'd need a major release to go to `immer` 9.
>
> There are things that are central to RTK where we can't be picky, and then there is stuff like this where risks like that have to be weighed a bit more.

Seems like you would need a major release in that case whether you exported `castDraft` or not and it would be easy to stub `castDraft` if necessary. I only mentioned the complexity since oftentimes a more complex function would be more likely to have breaking changes in its functionality over time, but the existence of the function seems to be fairly stable as well.

> Honestly, you're the first person I've seen doing that. I get that it could be a "better practice", but with immer automatically freezing things anyways, it's just never been any concern to me personally. ReadonlyArrays feel just very bulky to use for me because the ecosystem doesn't support that in the slightest.

To each their own, I agree that the ecosystem does not support readonly arrays well.

Thanks for you responses, I appreciate your time.

phryneas:
We have gone from immer 2 or so to currently immer 9 without the need for a bump - because the features we use & expose of it are stable "enough" (not 100% unfortunately, but nothing really breaking). 

But the more we start pulling in external dependencies that we have no control of (afaik we control every other dependency), the flakier it gets. That's just why we have a bad feeling adding more - especially since this is essentially a type helper and those patterns do change over time.

PR #2329, opened 2022-05-15
Improve JSON detection

### LINKED ISSUE #2282: Improve fetchBaseQuery content json verification
Currently in the docs regarding the body it says:
` // fetchBaseQuery automatically adds 'content-type: application/json' to the Headers and calls 
 JSON.stringify(patch)`
However it doesn't state that it only "stringifies" if the `content-type` is strictly `application/json`, I use `application/vnd.api+json` for example and it was not very easy to figure what was wrong. 

My proposal is to check if the content-type contains the json extension, so it will cover all json cases, if no one sees any issues with this proposal I'd be glad to push a PR for it, otherwise we could at least update the docs to avoid any confusion.

ShaunDychko:
Resolves #2282

### DIFF
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -43,8 +43,7 @@ const isJsonContentType = (headers: Headers) => {
     return contentType.startsWith('application/') && contentType.endsWith('json');
   }
   return false;
-}
-                                                                                            }
+}                                                                                          }
 
 const handleResponse = async (
   response: Response,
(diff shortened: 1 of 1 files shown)


packages/toolkit/src/query/fetchBaseQuery.ts
```diff
@@ -37,8 +37,13 @@ const defaultFetchFn: typeof fetch = (...args) => fetch(...args)
...
   response.status >= 200 && response.status <= 299
 
-const isJsonContentType = (headers: Headers) =>
-  headers.get('content-type')?.trim()?.startsWith('application/json')
+const isJsonContentType = (headers: Headers) => {
+  const contentType = headers.get('content-type')?.trim();
+  if (typeof contentType === 'string') {
+    return contentType.startsWith('application/') && contentType.endsWith('json');
+  }
+  return false;
```
phryneas, line 45:
```suggestion
  const contentType = headers.get('content-type')?.trim();
  return contentType?.startsWith('application/') && contentType.endsWith('json');
```

phryneas (commented):
Still not 100% sure if we generally can do this, but this would be shorter.

phryneas:
This is already handled over in #2331

PR #2330, opened 2022-05-15
Check that body isJsonifiable before stringify

### LINKED ISSUE #2282: Improve fetchBaseQuery content json verification

ShaunDychko:
Resolves #2282

### DIFF
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -218,7 +218,7 @@ export function fetchBaseQuery({
       config.headers.set('content-type', 'application/json')
     }
 
-    if (body && isJsonContentType(config.headers)) {
+    if (isJsonifiable(body) && isJsonContentType(config.headers)) {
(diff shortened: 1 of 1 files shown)


msutkowski:
Alternative and more flexible solution: https://github.com/reduxjs/redux-toolkit/pull/2331

phryneas:
I don't think this actually has something to do with https://github.com/reduxjs/redux-toolkit/issues/2282 - but it's a very good catch!

Thank you for the PR!

PR #2331, opened 2022-05-15
Add `isJsonContentType` predicate to `fetchBaseQuery`

msutkowski:
The other thing that you'd typically do with this is set the content type in the `prepareHeaders`, unless a generator is doing that by default.


packages/toolkit/src/query/fetchBaseQuery.ts
```diff
@@ -123,6 +123,15 @@ export type FetchBaseQueryArgs = {
     init?: RequestInit | undefined
   ) => Promise<Response>
   paramsSerializer?: (params: Record<string, any>) => string
+  /**
+   * By default, we only check for 'application/json' as the content-type for json. If you need to support another format, you can pass
+   * in a predicate function for your given api to get the same automatic stringifying behavior
+   * @example
+   * ```ts
+   * const isJsonContentType = (headers: Headers) => ["application/vnd.api+json", "application/json"].includes(headers.get("content-type")?.trim());
```
ShaunDychko, line 131:
```suggestion
   * const isJsonContentType = (headers: Headers) => ["application/vnd.api+json", "application/json", "application/vnd.hal+json"].includes(headers.get("content-type")?.trim());
```

ShaunDychko: Document adding a media type not already included by default.

ShaunDychko (changes requested):

packages/toolkit/src/query/fetchBaseQuery.ts
```diff
@@ -123,6 +123,15 @@ export type FetchBaseQueryArgs = {
     init?: RequestInit | undefined
   ) => Promise<Response>
   paramsSerializer?: (params: Record<string, any>) => string
+  /**
+   * By default, we only check for 'application/json' as the content-type for json. If you need to support another format, you can pass
```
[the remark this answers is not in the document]
msutkowski: I'm not opposed to doing this, but I'd prefer to have @phryneas and @markerikson chime in. We originally based this on how axios/redaxios work (if no headers set and it looks like json, make it json and stringify). If you were to use axios or anything, you'd create an instance and set your default headers on each request, the same as you'd need to do with this proposal. I'm not so sure we should operate any differently, but I could be wrong!

PR #2343, opened 2022-05-22
Update slice.ts

dirheimerb:
Small change, for better typescript experience. For state, typescript needs a type. I added
changeColorScheme(state: { colorScheme: string; }
to the state under reducers for themeSlice

### DIFF
--- a/examples/action-listener/counter/src/services/theme/slice.ts
+++ b/examples/action-listener/counter/src/services/theme/slice.ts
@@ -12,7 +12,7 @@ export const themeSlice = createSlice({
     colorScheme: 'light',
   } as ThemeState,
   reducers: {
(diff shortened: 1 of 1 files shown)


dirheimerb:
Small change, for better typescript experience.  For state, typescript needs a type.  I added
changeColorScheme(state`: { colorScheme: string; }`  
to the state under reducers for themeSlice

markerikson:
Hiya. Appreciate the PR, but this should actually be unnecessary.  `createSlice` already infers the type of `state` in each case reducer from the `initialState` value that was provided.  In this case, `initialState` was typed as `ThemeState`, so TS already knows that `state` in this case reducer is also `ThemeState`.

PR #2362, opened 2022-05-27
Elaborate on refreshToken examples

taylorkline:
I was silly and it took me a bit to realize that, just as with mutations and queries, the first argument to `baseQuery` can be a `string` in short form but a more verbose object argument may be provided. I think a longer form example will help people to realize how a custom baseQuery can work.

### DIFF
--- a/docs/rtk-query/usage/customizing-queries.mdx
+++ b/docs/rtk-query/usage/customizing-queries.mdx
@@ -380,7 +380,7 @@ const baseQueryWithReauth: BaseQueryFn<
   let result = await baseQuery(args, api, extraOptions)
   if (result.error && result.error.status === 401) {
     // try to get a new token
-    const refreshResult = await baseQuery('/refreshToken', api, extraOptions)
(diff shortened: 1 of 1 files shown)


phryneas:
To be honest, I'd rather not show POSTing the refresh token, or people might start doing that - the most secure way here would be to have the refresh token set at login by the server as a `httpOnly` cookie that is never accessed from JavaScript.

I would be okay with `{ url: '... }` I think, that should not distract too much from the real point of the example. :thinking: 
What do you think @markerikson ?

taylorkline:
> 

I see your point in the web space, but in React Native, we store the refresh token in biometric-protected keychain services / SharedPreferences, and fetch it out of that encrypted storage for refreshes. Though I see what you mean since most of the RTK Query examples are geared towards web.

phryneas:
Of course there are good reasons to do it - but they you know you want that - I just don't want to steer web users the wrong direction :)

PR #2363, opened 2022-05-28
Add 'content-type' ResponseHandler

### LINKED ISSUE #2354: The default responseHandler should respect Content-Type headers
Howdy!

I believe the default responseHandler should be a _little_ more intelligent to handle the response based on the content type header.

For example, in the default Go http package, [you can see it is typical to return a response body of "404 page not found"](https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/net/http/server.go;l=2101), "Unauthorized," etc. with non-200 responses.

It seems very counterintuitive that one must define a custom `responseHandler` for each and every endpoint to avoid RTK Query failing with `PARSING_ERROR` in this rather common use-case.

taylorkline:
Please let me know:

- If this PR is welcome, I can edit the docs as well before merge.
- Would you be open to a follow-up PR to make `content-type` the default for `responseHandler`?

Closes #2354.

### DIFF
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -54,16 +54,12 @@ const handleResponse = async (
     responseHandler = isJsonContentType(response.headers) ? 'json' : 'text'
   }
 
-  if (responseHandler === 'text') {
-    return response.text()
-  }
-
   if (responseHandler === 'json') {
     const text = await response.text()
     return text.length ? JSON.parse(text) : null
   }
 
-  return responseHandler
+  return response.text()
 }
 
 export type FetchBaseQueryError =
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -46,6 +46,7 @@ const handleResponse = async (
   response: Response,
   responseHandler: ResponseHandler
 ) => {
+  let x: never
   if (typeof responseHandler === 'function') {
     return responseHandler(response)
   }
@@ -62,6 +63,9 @@ const handleResponse = async (
     const text = await response.text()
     return text.length ? JSON.parse(text) : null
   }
+
+  x = responseHandler
+  return x
 }
 
 export type FetchBaseQueryError =
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -46,7 +46,6 @@ const handleResponse = async (
   response: Response,
   responseHandler: ResponseHandler
 ) => {
-  let x: never
   if (typeof responseHandler === 'function') {
     return responseHandler(response)
   }
@@ -64,8 +63,7 @@ const handleResponse = async (
     return text.length ? JSON.parse(text) : null
   }
 
-  x = responseHandler
-  return x
+  return responseHandler
 }
 
 export type FetchBaseQueryError =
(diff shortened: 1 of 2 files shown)


taylorkline:
Should be good to go.

markerikson:
Hiya.  PRs are always welcome, but can you clarify what specific problem this PR is meant to solve?  Is it related to #2282 ?

phryneas:
While I rarely see apis with mixed response-types, this might be handy for some cases - although it might also make it harder to spot if you expect an api to return an object and suddenly it starts returning text. #

So I'd say "let's add that, but not make it default".

Thoughs? @markerikson @msutkowski

taylorkline:
f3517e0f3f60c4e160c4f495bc4934d9f10b8ee3 was an exhaustiveness check to make sure all response handler types were accounted for.

phryneas:
Hmm. If something like that would make sense here, it should probably be a `throw` with a useful error message.

taylorkline:
> Hmm. If something like that would make sense here, it should probably be a `throw` with a useful error message.

But that would result in a run-time exception, whereas an exhaustiveness check with TypeScript would be a compile-time exception. Preferable to catch it at compile-time, no?

phryneas:
If catching it at compile time means shipping a bunch of bytes that are never executed to every library consumer, then no, it's definitely not worth that - there would be no start and no end to it.

The only thing I would see here would be to guard against errors of non-TS-users accidentally passing something unsupported in. But since we don't have done that before, I don't think we'll start now. I will go and remove the `return` statement though, leaving that in was an oversight.

phryneas:
I guess we should have some fallthrough default here, since `undefined` is actually a problematic value in many code branches. 
I will move the `text` check down and make it the default without a check.

alexandernst:
I'm testing this and it's not working properly. 

My API is returning an array instead of an object (content type is `application/json, charset=utf-8`).

markerikson:
@alexandernst : define "not working properly"?

It would really help if you could file a new issue, and provide either a CodeSandbox or a Replay ( https://replay.io/record-bugs ) showing what's happening

alexandernst:
> While I rarely see apis with mixed response-types, 

You'll get (most probably) plain text responses from pretty much any API deployed behind a reverse proxy (what you'd usually use when deploying to production) when the API itself is unavailable for whatever reason and the reverse proxy itself it issuing the reply.

> although it might also make it harder to spot if you expect an api to return an object and suddenly it starts returning text.

I really don't think that it's RTK's job to deal in any way (nor protect in any way) the contract of your API. IMHO this is totally out of the scope of what RTK should do, which is (again, imho), parse the responses from the server in the most intelligent possible way (which is what this PR is all about).

'content-type' should be the default behavior.

alexandernst:
> @alexandernst : define "not working properly"?

I beg a pardon, my environment was acting weird. Re-testing this on a clean env yielded the expected behavior.

phryneas:
> @phryneas
> 
> > While I rarely see apis with mixed response-types,
> 
> You'll get (most probably) plain text responses from pretty much any API deployed behind a reverse proxy (what you'd usually use when deploying to production) when the API itself is unavailable for whatever reason and the reverse proxy itself is issuing the reply. This will cause the "mixed response-types" (assuming that the API itself is returning JSON).

If that proxy usually forwards JSON, I would still expect that proxy to respond with a JSON error in a form similar to `{ status: 500, message: "internal server error" }`.

> 
> > although it might also make it harder to spot if you expect an api to return an object and suddenly it starts returning text.
> 
> I really don't think that it's RTK's job to deal in any way (nor protect in any way) the contract of your API. IMHO this is totally out of the scope of what RTK should do, which is (again, imho), parse the responses from the server in the most intelligent possible way (which is what this PR is all about).
> 
> 'content-type' should be the default behavior.

We might revisit this with 2.0, but for now that would be a breaking change that we can't just add.

alexandernst:
> If that proxy usually forwards JSON, I would still expect that proxy to respond with a JSON error in a form similar to `{ status: 500, message: "internal server error" }`.

If your API is actually down/unreachable (think about service crash, network unreachability, time-gap deployment, etc...) you'd still get text response by the reverse proxy.

> We might revisit this with 2.0, but for now that would be a breaking change that we can't just add.

Fair enough. 👍

PR #2378, opened 2022-06-01
Check initial state is draftable before using immer to freeze it.

### LINKED ISSUE #2377: createReducer throws if initialState isn't draftable
Encountered while testing converting a legacy app (that uses ImmutableJS) to createSlice.

The actual reducer implementation makes a [check](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/createReducer.ts#L260) whether the value is draftable, and avoids passing it to `createNextState` if it's not - however the `getInitialState` implementation **[always](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/createReducer.ts#L227-L233)** passes the initial state to `createNextState`, whether it's draftable or not.

This means that it'll throw if the initial state is not possible for immer to draft, as seen in [this codesandbox](https://codesandbox.io/s/adoring-hellman-u39qd6?file=/src/index.ts).

A fix for this (incoming PR 😎) would be to only pass the initial state to `createNextState` if it's draftable.

EskiMojo14:
Fixes #2377.

### DIFF
--- a/packages/toolkit/src/createReducer.ts
+++ b/packages/toolkit/src/createReducer.ts
@@ -75,6 +75,10 @@ function isStateFunction<S>(x: unknown): x is () => S {
   return typeof x === 'function'
 }
 
+function freezeDraftable<T>(val: T) {
+  return isDraftable(val) ? createNextState(val, () => {}) : val
+}
+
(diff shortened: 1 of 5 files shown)


phryneas:
Very good catch & thanks for the fix :)

Issue #2382, opened 2022-06-03
Asserting condition rejection of thunks in tests


josh-degraw:
Howdy, I'm trying to test some thunks, and my experience testing them is not what I would expect based on the provided API, apologies if I'm just misunderstanding.

Basically I'm just trying to assert that the body of the thunk is not executed if the `condition` is false. 

[Repro sandbox](https://codesandbox.io/s/thunk-condition-tester-seho28?file=/src/index.test.js)

```ts
import { createAsyncThunk, isRejected } from "@reduxjs/toolkit";

test("simple thunk test", async () => {
  let wasCalled = false;
  const thunk = createAsyncThunk(
    "Sample",
    async () => {
      wasCalled = true;
      return "hello";
    },
    {
      condition: () => false
    }
  );
  const resp = await thunk()(dispatch, getState, null);
  expect(wasCalled).toBe(false); // Passes
  expect(isRejected(resp)).toBe(true); // Passes
  expect(resp.meta.condition).toBe(false); // Fails >:(
});
```

So I have here that the condition always returns false, yet what is returned by the thunk has the condition set to true. I would expect this to be false when what is returned by the condition function is false. Is this not a valid assumption?

It looks like it is correctly not executing the body of the function, but unless the `meta.condition` is false, it seems like I have to do something hacky like above to check whether the promise body was entered during testing, which I would love to avoid for simplicity's sake.

josh-degraw:
So upon further inspection it seems like the condition is showing as `true` if the rejection was caused by the condition being `false`, it's `false` if the rejection is not because of the condition, and it's `undefined` if the action was not rejected. Is that accurate? If so some clarity in the docs/code comments around that would be helpful. Since it has the same name I would have expected it to hold the resolved value of the condition function, hence my confusion here.

phryneas:
You are just interpreting it the wrong way round: It is *rejected* with the reason of a *condition*.

In the case of a rejection with a payload, `action.meta.rejectedWithValue` will be true, in the case of an abortion, `action.meta.aborted` will be `true`.

josh-degraw:
> You are just interpreting it the wrong way round: It is _rejected_ with the reason of a _condition_.
> 
> In the case of a rejection with a payload, `action.meta.rejectedWithValue` will be true, in the case of an abortion, `action.meta.aborted` will be `true`.

[...]

PR #2383, opened 2022-06-03
Add documentation section about testing cancellation behavior

josh-degraw:
In reference to #2382, this PR adds some clarity about how to test cancellation behavior to the docs.

### DIFF
--- a/docs/api/createAsyncThunk.mdx
+++ b/docs/api/createAsyncThunk.mdx
@@ -507,6 +507,36 @@ const fetchUserById = createAsyncThunk(
 )
 ```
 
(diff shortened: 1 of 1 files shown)


phryneas:
Could you move this a bit away from the "testing" use case and more into "checking if a rejection is of type X or Y"? This information is not only good for testing, but can also be used in reducers.

phryneas (changes requested):
setting this to "changes requested"

josh-degraw:
Okay, I updated some of the wording there. Would you want an additional example of consuming it in a reducer or do you think just the wording change is sufficient?

Issue #2334, opened 2022-05-17
missing `else` in `enhanceEndpoints` could cause problems with function callback


phryneas:
https://github.com/reduxjs/redux-toolkit/blob/8c23cd33971ae5a62bc19764adca6f6be2ed98ec/packages/toolkit/src/query/createApi.ts#L280

This needs to be a `else` or the `Object.assign` will assign the function to the endpoint.

Also we could need a few tests for `enhanceEndpoints` and a few examples in the docs since it seems we don't have any.

PR #2386, opened 2022-06-05
Add Missing Else to enhanceEndpoints Function

kinson:
# Overview

Addresses #2334 .

This pr adds a missing `else` to the `enhanceEndpoints` function to improve the clarity of the function. It also adds an example of using it to the documentation.

Regarding the original issue:

> This needs to be a else or the Object.assign will assign the function to the endpoint.

Based on my understanding of `Object.assign`, passing a function as an assign will return a copy of the original object with no modifications. I created a [codesandbox to demonstrate it](https://codesandbox.io/s/admiring-davinci-6ifsjd?file=/src/index.ts) (forgive me if there is a better template to use for such a simple example). As a result, the function won't affect the endpoint as the code currently is written. Let me know if I am misunderstanding the issue though, or if my premise is incorrect.

Here is the code snippet again for reference:

https://github.com/reduxjs/redux-toolkit/blob/64a30d83384d77bcbc59231fa32aa2f1acd67020/packages/toolkit/src/query/createApi.ts#L275-L285

Even if the lack of an `else` will not cause a bug, it still does not make much sense to call `Object.assign` when a function is passed, because it does not do anything. So, I still added the `else` in this pr for clarity and readability.

The issue asked for tests as well, however when I went to adds tests I realized I was mostly duplicating a test that was already in the test suite:

https://github.com/reduxjs/redux-toolkit/blob/64a30d83384d77bcbc59231fa32aa2f1acd67020/packages/toolkit/src/query/tests/createApi.test.ts#L457-L501

I am happy to add more tests 🤓 , but I wondering what test cases may remain for this. I removed my test change from this pr because I didn't want to add more weight to the codebase without adding any testing value.

~Lastly, is there a way for me to preview the docs changes in the format they are rendered? I wanted to provide a before/after screenshot but I am probably missing a very obvious set of steps to get it up and running!~ CI did it for me 😍 (and I figured out how to do it locally as well)

Thanks for creating this package, and taking the time to read this wall of text ✌🏻 😄

### DIFF
--- a/docs/rtk-query/api/created-api/code-splitting.mdx
+++ b/docs/rtk-query/api/created-api/code-splitting.mdx
@@ -61,9 +61,20 @@ Returns an updated and enhanced version of the API slice object, containing the 
 
 This is primarily useful for taking an API slice object that was code-generated from an API schema file like OpenAPI, and adding additional specific hand-written configuration for cache invalidation management on top of the generated endpoint definitions.
 
-For example, `enhanceEndpoints` can be used to modify caching behaviour by changing the values of `providesTags`, `invalidatesTags`, and `keepUnusedDataFor`:
+For example, `enhanceEndpoints` can be used to modify caching behavior by changing the values of `providesTags`, `invalidatesTags`, and `keepUnusedDataFor`:
 
 ```ts
+// file: api.ts noEmit
+import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react'
+
+export const api = createApi({
+  baseQuery: fetchBaseQuery({ baseUrl: '/' }),
+  endpoints: (builder) => ({}),
+})
+
+// file: enhanceEndpoints.ts
+import { api } from './api'
+
 const enhancedApi = api.enchanceEndpoints({
   addTagTypes: ['User'],
   endpoints: {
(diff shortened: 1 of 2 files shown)


docs/rtk-query/api/created-api/code-splitting.mdx
```diff
@@ -60,3 +60,24 @@ Any provided tag types or endpoint definitions will be merged into the existing
 Returns an updated and enhanced version of the API slice object, containing the combined endpoint definitions.
 
 This is primarily useful for taking an API slice object that was code-generated from an API schema file like OpenAPI, and adding additional specific hand-written configuration for cache invalidation management on top of the generated endpoint definitions.
+
+For example, `enhanceEndpoints` can be used to modify caching behaviour by changing the values of `providesTags`, `invalidatesTags`, and `keepUnusedDataFor`:
+
+```ts
+const enhancedApi = api.enchanceEndpoints({
```
kinson, line 67: It looks like CI may have failed due to this line:
https://app.netlify.com/sites/redux-starter-kit-docs/deploys/629d1fee7ab8150008c71b33#L301

I will take a closer look and fix this!

phryneas:
Thank you for looking deeper into this!

I think the test is enough - to be honest, I just forgot about the existence of that test :) 

I've tweaked the example a bit so it works without the `: any` and am going to get this in - it makes the code base a bit more sane *and* improves the docs :tada:

kinson:
Thanks for the updates and finding a way to remove the pesky `any`!

PR #2407, opened 2022-06-12
Codegen: create an option to "flatten" the queryArg

lindapaiste:
It has always been a pain point for me that the RTK codegen creates unnecessarily complex arguments, ie. `{ someIdProperty: string }` instead of `string`.

This PR adds an option `flattenArg` (does this name make sense?  It can be changed) to the codegen config which will clean up the `queryArg`.  If the `queryArg` would contain exactly one property, then make that property itself be the `queryArg`.   This works whether the argument is in the path, body, or params.

before:
```
getOrderById: build.query<GetOrderByIdApiResponse, GetOrderByIdApiArg>({
    query: (queryArg) => ({ url: `/store/order/${queryArg.orderId}` }),
})

export type GetOrderByIdApiResponse =
    /** status 200 successful operation */ Order;
export type GetOrderByIdApiArg = {
    /** ID of order that needs to be fetched */
    orderId: number;
};
```
```
const { data } = useGetOrderByIdQuery({ orderId: 99 });
```

after:
```
getOrderById: build.query<GetOrderByIdApiResponse, GetOrderByIdApiArg>({
    query: (queryArg) => ({ url: `/store/order/${queryArg}` }),
})

export type GetOrderByIdApiResponse =
    /** status 200 successful operation */ Order;
export type GetOrderByIdApiArg =
    /** ID of order that needs to be fetched */ number;
```
```
const { data } = useGetOrderByIdQuery(99);
```

### DIFF
--- a/packages/rtk-query-codegen-openapi/src/generate.ts
+++ b/packages/rtk-query-codegen-openapi/src/generate.ts
@@ -56,6 +56,19 @@ function operationMatches(pattern?: EndpointMatcher) {
   };
 }
 
+function withQueryComment<T extends ts.Node>(node: T, def: QueryArgDefinition, hasTrailingNewLine: boolean): T {
+  const comment = def.origin === 'param' ? def.param.description : def.body.description;
+  if (comment) {
+    return ts.addSyntheticLeadingComment(
+      node,
+      ts.SyntaxKind.MultiLineCommentTrivia,
+      `* ${comment} `,
+      hasTrailingNewLine
+    );
+  }
+  return node;
+}
+
 export function getOverrides(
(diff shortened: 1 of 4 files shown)


phryneas:
Heyho, I just wanted to give quick feedback: this looks like a great addition!

We are currently all pretty swamped with real life, so PRs are sitting around for quite a while - so it might really take a bit until we come back to it - and I want to give you a bit of a heads up on that.

But this looks like a really great feature and I'm very much in favor that we'll get that in once things start cooling down a bit :+1:

msutkowski (approved):
Thanks for this @lindapaiste!

Issue #2416, opened 2022-06-15
Update docs to use type-only imports


raulsanchez1024:
Currently, not all imported types in the docs use type-only imports. Updating the imported types in the docs will make the imports a bit more consistent.

raulsanchez1024:
Closing because PR #2417 was merged :)

PR #2417, opened 2022-06-15
Updates docs to use type-only imports

raulsanchez1024:
Related to https://github.com/reduxjs/redux-toolkit/issues/2416. This updates the docs to use type-only imports


phryneas:
This is a very welcome contribution, thank you!

Unfortunately there have been some updates to the `kitchen-sink` example in the meantime, could you please give those a go-over? :)

PR #2420, opened 2022-06-16
[FEAT][AsyncThunk] Pass abort reason of thunk action to AbortController #2395

### LINKED ISSUE #2395: Pass abort reason of thunk action to AbortController
Hi,

I noticed that the abort reason of a thunk action is not getting passed down to the `AbortController` in line 578:
https://github.com/reduxjs/redux-toolkit/blob/64a30d83384d77bcbc59231fa32aa2f1acd67020/packages/toolkit/src/createAsyncThunk.ts#L575-L580

It would be useful to get the abort reason directly in the thunk action, so the thunk can cleanup its states.

thx

FaberVitale:
Closes #2395

### Description

`asyncThunkApi.signal.reason` is set to the first value provided to `asyncThunkHandle.abort`.

#### Other changes

- fixes `signal.aborted` not being set to `true` in the `AbortController` shim if `abort` is called.

#### Implementation details

It uses a utility function `abortControllerWithReason`, already used within the listenerMiddleware,  that patches reason if it is not available.

### DIFF
--- a/packages/toolkit/src/createAsyncThunk.ts
+++ b/packages/toolkit/src/createAsyncThunk.ts
@@ -7,6 +7,8 @@ import { createAction } from './createAction'
 import type { ThunkDispatch } from 'redux-thunk'
 import type { FallbackIfUnknown, IsAny, IsUnknown } from './tsHelpers'
 import { nanoid } from './nanoid'
+import type { AbortSignalWithReason } from './function-utils'
+import { abortControllerWithReason } from './function-utils'
 
 // @ts-ignore we need the import of these types due to a bundling issue.
 type _Keep = PayloadAction | ActionCreatorWithPreparedPayload<any, unknown>
@@ -541,7 +543,7 @@ export function createAsyncThunk<
             reason: undefined,
             throwIfAborted() {},
           }
-          abort() {
+          abort(reason?: any) {
             if (process.env.NODE_ENV !== 'production') {
               if (!displayedWarning) {
                 displayedWarning = true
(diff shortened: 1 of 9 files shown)


phryneas:
To be honest, I'm not really sure if I like the thought of adding more "polyfill-like" code. Actually, I'd like to remove the `AbortController` polyfill in general with RTK 2.0 (#2448). 

So I guess we'd need a bit research first: 
* At that point, what platforms do have `AbortController`, but not `reason`? 
* As these platforms will need a Polyfill anyways, which of the existing polyfills do not have `reason`? 
* If all platforms that have it natively support `reason`, wouldn't it make more sense to just PR into the missing polyfills?

FaberVitale:
### Why should these changes be included in a minor release (1.9)?

1. fix `signal.aborted` not being set to true in the polyfill that causes inconsistent behavior between runtimes.
2. Add support to `signal.reason` reusing code that is already part of `listenerMiddleware` logic.

---

> what platforms do have AbortController, but not reason?

### `signal.reason` support | 2022-06-25

[`abortSignal.reason`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort) is available in:

### server runtimes

- node [`^16.14.0`](https://nodejs.org/dist/latest-v16.x/docs/api/globals.html#abortsignalreason)
- node [`^17.2.0`](https://nodejs.org/dist/latest-v17.x/docs/api/globals.html#abortsignalreason)
- node [`^18.0.0`](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#abortsignalreason)
- deno `^1.16`

### browsers

- [can i use](https://caniuse.com/mdn-api_abortsignal_reason) `82.51%`

Samsung internet is a relevant mobile browser (especially in Germany) that does not support `signal.reason`.


### `createAsyncThunk` changes for 2.0.0

I agree that we should remove if possible obsolete polyfills in `2.0`.

PR #2431, opened 2022-06-21
docs: Updated references for redux-devtools to the correct github repo

kashifshamaz21:
The docs currently point to https://github.com/zalmoxisus/redux-devtools-extension for the devtools extension. But that repo is no longer the actively maintained one, the correct repo is https://github.com/reduxjs/redux-devtools.
This PR updates the references for the same.


markerikson:
Thanks!

PR #2433, opened 2022-06-21
Update createAsyncThunk example to handle getState typings

Retsam:
Was helping someone on Discord with an issue related to `getState` typings, and noticed that the example here, despite being in Typescript, didn't include the necessary generic parameters to actually compile.


markerikson:
Thanks!

PR #2436, opened 2022-06-22
🚿 Renovate kitchen sink

Shrugsy:
Renovates the kitchen sink example.

The kitchen sink example has been seen to mislead users into believing it is expected to have multiple API definitions in a single app, in situations where they should only have one
(see https://discord.com/channels/102860784329052160/103696749012467712/989088448671543316 https://discord.com/channels/102860784329052160/103538784460615680/988439535560065064 for a couple examples in the last couple of days)

Besides being additional unnecessary work for the user to write like this, it will cause issues such as tag invalidation not working across separate APIs.

This MR makes the following changes to the kitchen sink example to address this:
- All endpoints in separate files are now injected into a single base API definition

Additionally, the following has been changed in the same kitchen sink example:
- Fix an issue where nested routes were not loading (i.e. when trying to view a single 'Post')
- Update the 'new post' and 'edit post' inputs to use a native `form` element (which allows submitting the new post/edited post with the 'enter' key)


markerikson:
Awesome!

PR #2446, opened 2022-06-23
fix: abort controller fallback

JulienKode:
# Description 

Inside `createAsyncThunk` currently RTK have a fallback in case `AbortController` is not declared. https://github.com/reduxjs/redux-toolkit/blob/e307245e8e78868353c746531331257ea69ccab2/packages/toolkit/src/createAsyncThunk.ts#L530

This seems not the case for the `listenerMiddleware` https://github.com/reduxjs/redux-toolkit/blob/3d7bd2246df35a668a25d60a757f1f0b6df62798/packages/toolkit/src/listenerMiddleware/index.ts#L89

## How to reproduce the issue ? 

- Create a next.js project with node 14
- use listener middleware on server side 
- see the reference error of `AbortController`

### DIFF
--- a/packages/toolkit/src/abortController.ts
+++ b/packages/toolkit/src/abortController.ts
@@ -1,0 +1,28 @@
+let displayedWarning = false
+export const AC =
+    typeof AbortController !== 'undefined'
+        ? AbortController
+        : class implements AbortController {
+            signal = {
+                aborted: false,
+                addEventListener() {},
+                dispatchEvent() {
+                    return false
+                },
+                onabort() {},
+                removeEventListener() {},
+                reason: undefined,
+                throwIfAborted() {},
+            }
+            abort() {
+                if (process.env.NODE_ENV !== 'production') {
+                    if (!displayedWarning) {
+                        displayedWarning = true
+                        console.info(
+                            `This platform does not implement AbortController. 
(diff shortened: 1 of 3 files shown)


phryneas:
Hi, thanks for the PR!

At this point I am thinking we should rather remove that polyfill: IE has dropped and there is no more desktop browser supporting it. In node environments, that polyfill can cause more harm than good, since some `fetch` polyfills only work with very specific `AbortController` polyfills.

So my suggestion would be that we keep it for the time being and remove that polyfill in the 2.0 major. @markerikson what are your thoughts?

phryneas:
I haven't thought about it in-depth, but given that you don't need a polyfill in the browser, node 16 ships with AbortController, and (in the case of older node versions) on the server installing a polyfill also doesn't consume any "over the wire data", I think it would be okay to say "we expect an AbortController implementation to be available".

JulienKode:
This makes sense, in the same way as fetch

JulienKode:
Should I close this PR or it's still useful for pre 2.0 version ?

phryneas:
Good question - could you close this, but open an issue in the other direction a la "remove abortcontroller" so we can label that as 2.0 and track it? :)

JulienKode:
In addition what node.js 14 and lower can do because polyfill for node.js seems to be with direct import https://github.com/mo/abortcontroller-polyfill#using-it-on-nodejs

PR #2451, opened 2022-06-24
Add ThunkMiddleware to the re-exported types from redux-thunk

orta:
:wave: hi folks, I have just updated to the latest redux toolkit and found myself reaching into some of my transitive dependencies in order to accurately describe the types inside my app.  

( The code is some shared code across many redux stores which needs to send dispatches to stores, and sets a generic shape constraint on the state's type )

### Old Code (1.5.1)

```ts
import type { EnhancedStore } from "@reduxjs/toolkit"

export const setupHostConfig = (config: GameConfig) => <State extends ObjWithSession, Props extends {}>(
  store: EnhancedStore<State>
) => { ... } 
```

### New Code (1.8.2)

`yarn add redux-thunk`

```ts
import type { AnyAction, EnhancedStore } from "@reduxjs/toolkit"
import type { ThunkMiddleware } from "redux-thunk"

export const setupHostConfig = (config: GameConfig) => <State extends ObjWithSession, Props extends {}>(
  store: EnhancedStore<State, AnyAction, [ThunkMiddleware<State, AnyAction, undefined>]>
) => { ... } 
```

Given that ThunkMiddleware is effectively vendored from an API perspective, and there's already some vendored type exports - I just added it to the list.

### After this PR

```ts
import type { AnyAction, EnhancedStore, ThunkMiddleware } from "@reduxjs/toolkit"

export const setupHostConfig = (config: GameConfig) => <State extends ObjWithSession, Props extends {}>(
  store: EnhancedStore<State, AnyAction, [ThunkMiddleware<State, AnyAction, undefined>]>
) => { ... } 
```

### DIFF
--- a/packages/toolkit/src/index.ts
+++ b/packages/toolkit/src/index.ts
@@ -16,7 +16,7 @@ export type {
   ParametricSelector,
 } from 'reselect'
 export { createDraftSafeSelector } from './createDraftSafeSelector'
-export type { ThunkAction, ThunkDispatch } from 'redux-thunk'
+export type { ThunkAction, ThunkDispatch, ThunkMiddleware } from 'redux-thunk'
 
 // We deliberately enable Immer's ES5 support, on the grounds that
 // we assume RTK will be used with React Native and other Proxy-less


phryneas:
Ah, for using it as an argument. I haven't seen that one before, only as return value (where we try to let our users just keep TS inference). In this case, it's perfectly valid.

I'll merge this into the 1.9 integration branch, but fair warning: that one is not gonna drop very soon since we all have a lot of RL going on right now - so until then you are probably off best using the CodeSandbox build from our CI. (install instructions see [here](https://ci.codesandbox.io/status/reduxjs/redux-toolkit/pr/2451))

phryneas:
Hmm, even though I just merged the current `master` into `[v1.9-integration](https://github.com/reduxjs/redux-toolkit/tree/v1.9-integration)`, something doesn't seem to add up. Could you please try a rebase on your side?

orta:
Fixed up the branch. Yeah, I'm in no rush, my code works fine and next time I do a series of dep updates I can just revert my user-land changes 👍🏻

PR #2452, opened 2022-06-24
Fix correct alias of useDispatch function

oreillyross:
The useDispatch function should be aliased by adding the AppDispatch as a generic type argument


markerikson:
Hmm. Appreciate the PR, but no, that's not necessary - the given alias should be correct.

Issue #2456, opened 2022-06-26
RTK Query Kitchen Sink example on website uses incorrect import name


KaylaKremer:
When viewing the Kitchen Sink example, there is an error of `postsApi` being undefined in `authSlice.ts`.

This can be fixed by changing the incorrect import of `postApi` to `postsApi`.

<img width="1534" alt="Screen Shot 2022-06-25 at 10 05 14 PM" src="https://user-images.githubusercontent.com/22095244/175796425-65ed2430-8d14-43cc-b1a7-84620981ad20.png">

markerikson:
Thanks, could you file a PR to fix this?

KaylaKremer:
@markerikson Just put up the PR for it!

markerikson:
Thank you!

PR #2457, opened 2022-06-26
fix: incorrect import name for postsApi in authSlice

KaylaKremer:
This PR fixes a typo error for an import name in the kitchen sink example as described in [this issue](https://github.com/reduxjs/redux-toolkit/issues/2456).


markerikson:
Thanks!

PR #2458, opened 2022-06-26
Update Docusaurus to 2.b21

markerikson:
Seeing a bunch of SSR build errors locally, so we'll see what happens here.


markerikson:
Per Sebastien Lorber in https://github.com/facebook/docusaurus/issues/7686:


> @markerikson you have a custom DocPage component:
> 
> ```
> docLayoutComponent: '../src/theme/DocPageWithBraveWarning',
> ```
> 
> Our new entrypoint now provides multiple providers so that all subcomponents can access docs/version data easily with hooks anywhere (ie no props needed anymore):
> 
> ```tsx
> <DocsVersionProvider version={versionMetadata}>
>   <DocsSidebarProvider name={sidebarName} items={sidebarItems}>
>     <DocPageLayout>{docElement}</DocPageLayout>
>   </DocsSidebarProvider>
> </DocsVersionProvider>
> ```
> 
> Instead of providing a custom config entry point for that page (that you now have to maintain yourself over time, ensuring it remains compatible with Docusaurus), you can try to swizzle components instead.
> 
> Looks like @msutkowski added a Brave warning to Docusaurus documents: [reduxjs/redux-toolkit@599d4b8#diff-eaeffea00bcc3667ab8ddf261520fc47589a396fb6eb14afd916f40b98e4d682](https://github.com/reduxjs/redux-toolkit/commit/599d4b8940f5c9db1a181188c485d46766566499#diff-eaeffea00bcc3667ab8ddf261520fc47589a396fb6eb14afd916f40b98e4d682)
> 
> Not 100% sure but it looks like we can summarize this customization as:
> 
> ```tsx
> <div>
>   <BraveWarning />
>   <MDXProvider components={MDXComponents}>{children}</MDXProvider>
> </div>
> ```
> 
> This is a lot of copied code for such small customization.
> 
> We have split the `DocPage` component into much smaller parts to make these kinds of customizations much easier: you would swizzle a much smaller component instead. That remains a burden on your side to maintain, but a much smaller one, less likely to break on upgrades.
> 
> For example, solutions you can try:
> ## Swizzle eject `@theme/DocItem/Content`:
> 
> `yarn swizzle @docusaurus/theme-classic DocItem/Content --eject --typescript --danger`
> 
> ```tsx
> export default function DocItemContent({children}: Props): JSX.Element {
>   const syntheticTitle = useSyntheticTitle();
>   return (
>     <div className={clsx(ThemeClassNames.docs.docMarkdown, 'markdown')}>
>       {syntheticTitle && (
>         <header>
>           <Heading as="h1">{syntheticTitle}</Heading>
>         </header>
>       )}
>       <BraveWarning />
>       <MDXContent>{children}</MDXContent>
>     </div>
>   );
> }
> ```
> 
> Note: If the warning should also be displayed in MDX blog posts and MDX pages, you can also eject `MDXContent`, this component directly wraps any MDX content before rendering it.
> ## Swizzle wrap `@theme/DocItem/Content`:
> 
> `yarn swizzle @docusaurus/theme-classic DocItem/Content --wrap --typescript --danger`
> 
> ```tsx
> export default function ContentWrapper(props: Props): JSX.Element {
>   return (
>     <Content {...props}>
>       <BraveWarning />
>       {props.children}
>     </Content>
>   );
> }
> ```
> 
> Both solutions works. We usually prefer to wrap as it is less likely to break on an upgrade.
> 
> Note these components are still marked as "unsafe" to swizzle (require the `--danger` CLI flag) because we are not ready to commit to SemVer retro compatibility on those.
> 
> In any case: it is still better than ejecting the `DocPage` component like you previously did: it is much larger/complex and we don't provide SemVer compatibility on it either 😅 Like we have seen here: it broke on an upgrade due to internal changes.

Shrugsy:
@markerikson FYI I've done the following:

- Replaced the ejected `DocPage` with a wrapped `DocPage/Layout/Main`. I couldn't seem to get `DocItem` or `DocItem/Content` to show anything, but I might have just been doing something wrong on my side. Using `DocPage/Layout/Main` seems to put the brave warning in a reasonable spot (see brave-warning), and by wrapping rather than ejecting, we avoid the core issue that you quoted from Sebastien above
- Updated the links in `docs/tutorials/rtk-query.mdx` to complete relative paths. Dunno why they even worked before, but docusaurus was definitely complaining about them now, and treating them as relative paths in the `docs/tutorials` directory

Issue #2459, opened 2022-06-27
Add a dev-mode warning to `createSlice` to error if `initialState` is undefined


markerikson:
I've seen some folks misspell `initialState`, and thus end up with _no_ initial state.  

This leads to `combineReducers` doing the "Reducer `x` returned `undefined` blah blah" error, but it's not as helpful as it could be.

We should throw a dev-mode error inside of `createSlice` itself if the initial state is undefined.

dannielss:
I would like to solve this.

dannielss:
My first PR!! #2461

markerikson:
That's merged, so I'll close this.

PR #2461, opened 2022-06-27
Throw new error when initial state is undefined - #2459

dannielss:
Add a new error when initial state is undefined on createSlice.ts.

Issue: #2459

### DIFF
--- a/packages/toolkit/src/createSlice.ts
+++ b/packages/toolkit/src/createSlice.ts
@@ -262,6 +262,16 @@ export function createSlice<
   if (!name) {
     throw new Error('`name` is a required option for createSlice')
   }
+
+  if (
+    typeof process !== 'undefined' &&
+    process.env.NODE_ENV === 'development'
+  ) {
+    if(options.initialState === undefined) {
+      throw new Error('initial state must be different of undefined')
+    }
+  }
+
   const initialState =
     typeof options.initialState == 'function'
       ? options.initialState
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/createSlice.ts
```diff
@@ -262,6 +262,16 @@ export function createSlice<
   if (!name) {
     throw new Error('`name` is a required option for createSlice')
   }
+
+  if (
+    typeof process !== 'undefined' &&
+    process.env.NODE_ENV === 'development'
+  ) {
+    if(options.initialState === undefined) {
+      throw new Error('initial state must be different of undefined')
```
phryneas, line 271: This should probably be a `console.error`, so that (apart from this message) the application behaves 100% the same in dev and prod - there will be a crash down the line either way, but it should probably not a different crash depending on the environment.

dannielss: You're right, let me change this

dannielss:
@markerikson it should be throw new error or console.error?

markerikson:
Heh, sorry - I had assumed "throw", Lenz said "print a console error", so we'll go with Lenz's opinion here :)

markerikson:
@dannielss : yeah, let's go with:

```js
console.error('You must provide an `initialState` value that is not `undefined`.  You may have misspelled `initialState`')
```

dannielss:
Updated to console.error

markerikson:
Thanks! At a minimum this will end up going out in 1.9.  If we end up doing a 1.8.x minor release before then, it'll be part of that first.

Issue #2379, opened 2022-06-01
Redundant trailing slash is added when using root endpoint


ygrishajev:
Steps to reproduce:
1) declare an API with a `baseUrl`
2) add an endpoint that should query root `baseUrl` with params
3) call the endpoint

Expected url: `/api/foo?bar=baz`
Actual url: `/api/foo/?bar=baz`

As a result that redundant trailing slash added after path is causing issues in some apps/environments.
E.g. Next.js app would do a redirect to an endpoint without trailing slash, which would fail in iOS Chrome.

**Example client API declaration**:
```typescript
export const api = createApi({
  reducerPath: 'api',
  baseQuery: fetchBaseQuery({ baseUrl: '/api/foo' }),
  endpoints: (builder) => ({
    findAll: builder.query({
      query: (params) => ({ url: '', params })
    })
  })
});
```

The issue is tracking down to [joinUrls](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/query/utils/joinUrls.ts#L6) utility function and **could be fixed like so**:
```typescript
function joinUrls(base, url) {
    if (!base) {
        return url;
    }
    if (!url) {
        return base;
    }
    if (isAbsoluteUrl(url)) {
        return url;
    }
    base = withoutTrailingSlash(base);
    url = withoutLeadingSlash(url);
    
    const delimiter = url.startsWith('?') ? '' : '/' // <- detect a proper delimiter
  
    return base + delimiter + url;
}
```

I'm open to do a PR if this is a valid solution.

markerikson:
Yeah, if you could file a PR that would be appreciated!

ygrishajev:
> Yeah, if you could file a PR that would be appreciated!

@markerikson Sure, done. Pls review

phryneas:
Unfortunately that one is not ready yet - I'll reopen it here so we don't forget about that :)

PR #2465, opened 2022-06-29
Ensures there is no slash in url before query params

ygrishajev:
Fixes reduxjs/redux-toolkit#2379

### DIFF
--- a/packages/toolkit/src/query/tests/utils.test.ts
+++ b/packages/toolkit/src/query/tests/utils.test.ts
@@ -83,6 +83,8 @@ describe('joinUrls', () => {
 
     expect(joinUrls('', '/banana')).toBe('/banana')
     expect(joinUrls('', 'banana')).toBe('banana')
+
+    expect(joinUrls('/api', '?foo=bar')).toBe('/api?foo=bar')
   })
 
   test('correctly joins variations of absolute urls', () => {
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/query/tests/utils.test.ts
```diff
@@ -99,6 +101,10 @@ describe('joinUrls', () => {
     expect(joinUrls('https://example.com/api/', '/banana/')).toBe(
       'https://example.com/api/banana/'
     )
+
+    expect(joinUrls('https://example.com/api/', '?foo=bar')).toBe(
+      'https://example.com/api?foo=bar'
+    )
```
phryneas, line 107: I'm too late to the party :(

I feel like this scenario is a bug. There are servers that expect the slash and will lead you to a completely different page without it - completely independently of a search param being present.

It's fine to *not add* the slash - but actively removing ti is dangerous.

phryneas:
Could you please open another PR that keeps the slash if it is user-specified and only *not adds* it? Sorry for the hassle :sweat_smile:

ygrishajev:
@phryneas I see, so you mean some user specified parameter on `createApi` function?

phryneas:
On `fetchBaseQuery` - if `baseUrl` ends on a `/`, that should always stay in - likewise if `query` returns `{ url: 'foo/', params }`, the slash should stay as well, since the user explicitly specified it. It should just not appear out of the blue.

markerikson:
I'm gonna revert this for now - can you file a new PR?

PR #2476, opened 2022-07-01
add verb to lazyQuery docs

phryneas:


phryneas:
See https://twitter.com/GWeaths/status/1542592251305623552 - we might want to tweak the wording here further to prevent more confusion

PR #2480, opened 2022-07-02
Update devtools options TS types based on RDT 3.0

### LINKED ISSUE #1698: Handle changes to Redux DevTools Extension 3.0 configuration
Per https://github.com/reduxjs/redux-devtools/releases/tag/v3.0.0-beta.1 , the upcoming DevTools Extension 3.0 has some breaking changes. The biggest one is the renaming of options `actionsBlacklist` and `actionsWhitelist` renamed to `actionsDenylist` and `actionsAllowlist`.

We previously inlined our DevTools Extension config setup, because (if I remember right) the DevTools connection package wasn't written in ESM and was causing bundling de-opts.  That's currently in [`devtoolsExtension.ts`](https://github.com/reduxjs/redux-toolkit/blob/5fa0d4a3a7b2106fab56b5b489f4e8609d68fe61/packages/toolkit/src/devtoolsExtension.ts).

Perhaps we could switch back to having this as an external dep now that the extension and package are being actively maintained again.

@Methuselah96 , thoughts?

markerikson:
This PR:

- Copy-pastes the latest Redux DevTools Extension TS options over into our codebase, per [this `extension/index.ts` version](https://github.com/reduxjs/redux-devtools/blob/db1fcd12c0312c240e90c10a6bf3a7363b70b987/packages/redux-devtools-extension/src/index.ts)

Fixes #1698

### DIFF
--- a/packages/toolkit/src/devtoolsExtension.ts
+++ b/packages/toolkit/src/devtoolsExtension.ts
@@ -29,27 +29,26 @@ export interface EnhancerOptions {
    */
   maxAge?: number
   /**
-   * See detailed documentation at https://github.com/reduxjs/redux-devtools/blob/%40redux-devtools/extension%403.2.1/extension/docs/API/Arguments.md#serialize
+   * - `undefined` - will use regular `JSON.stringify` to send data (it's the fast mode).
+   * - `false` - will handle also circular references.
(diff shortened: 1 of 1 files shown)


Bartel-C8:
@markerikson : Would it be possible to publish a new release containing this fix? Thanks 🙏

Issue #2477, opened 2022-07-01
Server side rendering with getRunningOperationPromises


JulienKode:
Hi 

I've try to take a look at the ssr-experiment exemple here:
- https://github.com/JulienKode/ssr-experiments/pull/new/chore/demo
- https://codesandbox.io/s/naughty-robinson-47e9jt?file=/package.json

I've:
- added console log on `getRunningOperationPromises`
- Create a fake long query (10s)
- try to fetch with few seconds of delay the two pages:
    - http://localhost:3000/pokemon/ssr/bulbasaur
    - http://localhost:3000/pokemon/ssr/beedrill

On the `beedrill` fetch this is what I can see when displaying running promises:
  ```
  {
  "getPokemonByName(\"beedrill\")": {
    "arg": "beedrill",
    "requestId": "070QxaL_aSi7MV3VbQzDQ",
    "queryCacheKey": "getPokemonByName(\"beedrill\")"
  },
  "getPokemonByName(\"bulbasaur\")": {
    "arg": "bulbasaur",
    "requestId": "FdsRw3PmBbDP5sR7-kKPe",
    "queryCacheKey": "getPokemonByName(\"bulbasaur\")"
  }
}
  ```

- Is it normal that we are waiting for `bulbasaur` when our page are just concern about `beedrill` ? 
- Should the `runningPromises` created in the api be unique to each page and not sharing across different page ? otherwise on large scale apps this could be an issue if we are waiting for the queries of another page
- Does the `runningPromises` should be isolated to each page on SSR ? 
- like `createListenerMiddleware` should we have a `createApiMiddleware` ? Then on server side, we can have a different middleware instance for each page that does not share values

This is the debug line that I've added in redux:
```
  function getRunningOperationPromises() {
    console.warn("[DEBUG] RUNNING OPERATION PROMISES ", {runningQueries})
    console.warn(JSON.stringify(runningQueries, null, 2))
```

@phryneas  @markerikson

phryneas:
Could you give some feedback over in #2481 please?

markerikson:
Marking this as done because it's in 1.9 beta.

JulienKode:
Thank you @markerikson @phryneas

PR #2481, opened 2022-07-02
[breaking fix]  replace `getRunningOperationPromise(s)` with `getRunning(Query|Queries|Mutation|Mutations)Thunk`

phryneas:
# Motivation

As reported in #2477, currently doing `await api.util.getRunningOperationsPromise()` in a SSR environment will not only await for queries and mutations of the page currently being rendered, but of all other pages being SSRed at the same time. When designing the api, I forgot that multiple SSR renders will likely share the same api instance, but use different stores for that.

Unforunately, with the existing api there is no way fixing this.
The only way of scoping these functions to a certain store is to change the way they are called:

```diff
-await Promise.all(api.util.getRunningOperationPromises())
+await Promise.all(dispatch(api.util.getRunningOperationPromises()))
```

We essentially need to do a "breaking fix", as much as it hurts me.

If doing this change directly on the original methods, old code (without `dispatch`) would keep running without any errors, but get even worse behaviour: all those `await` statements would immediately continue, resolving to the thunk and not waiting for any running queries and mutations.

As this is too much of a foot gun, this PR needs to introduce even bigger changes.

# Changes

* replace `getRunningOperationPromise` and `getRunningOperationsPromise` with stubs that immediately `throw` with a link to the documentation.
* change the type of `getRunningOperationPromise` and `getRunningOperationsPromise` to `never`, but keep them in the types so we can add a link to the documentation there as well (and mark them as @deprecated).
* create four new functions:
  * api.util.getRunningQueryThunk(endpointName, queryArgs)
  * api.util.getRunningMutationThunk(endpointName, fixedCacheKeyOrRequestId)
  * api.util.getRunningQueriesThunk()
  * api.util.getRunningMutationsThunk()

So both functions are split up into two separate functions for queries and mutations, and the word `Promise` has been replaced by `Thunk`.

That changes the usage like this:

```diff
-await Promise.all(api.util.getRunningOperationPromises())
+await Promise.all([
  ...dispatch(api.util.getRunningQueriesThunk()),
  ...dispatch(api.util.getRunningQueriesThunk()),
])
```
or more realistically like this
```diff
-await Promise.all(api.util.getRunningOperationPromises())
+await Promise.all(dispatch(api.util.getRunningQueriesThunk()))
```
as it should almost never happen that a mutation is initiated during SSR.

# The "breaking fix" compromise

I kept in `getRunningOperationPromises` for now. 
**The function will `throw` in development**, but keep it's old faulty behaviour in production builds so things will not immediately break beyond the state it was already broken before.

I could not create a version of `getRunningOperationPromise` that kept the old behaviour, so it throws in both development and production.
Luckily, that function was documented as "primarily added to add experimental support for suspense", so the impact of removing it should be quite low.

# Todos

* [ ] tests
* [ ] update the example SSR repo at https://github.com/phryneas/ssr-experiments (this has to be coordinated with the release of the next RTK version)
* [ ] maybe provide a codemod (that allows to choose between both refactorings above) - could you take a stab at that @markerikson ?

### DIFF
--- a/docs/rtk-query/api/created-api/api-slice-utils.mdx
+++ b/docs/rtk-query/api/created-api/api-slice-utils.mdx
@@ -249,53 +249,59 @@ Note that [hooks](./hooks.mdx) also track state in local component state and mig
 dispatch(api.util.resetApiState())
 ```
 
-## `getRunningOperationPromises`
+## `getRunningQueriesThunk` and `getRunningMutationsThunk`
 
 #### Signature
 
 ```ts no-transpile
-getRunningOperationPromises: () => Array<Promise<unknown>>
+getRunningQueriesThunk(): ThunkWithReturnValue<Array<QueryActionCreatorResult<any>>>
+getRunningMutationsThunk(): ThunkWithReturnValue<Array<MutationActionCreatorResult<any>>>
 ```
 
 #### Description
 
-A function that returns all promises for running queries and mutations.
+Thunks that (if dispatched) return either all running queries or mutations.
+These returned values can be awaited like promises.
 
-This is useful for SSR scenarios to await everything triggered in any way, including via hook calls,
+This is useful for SSR scenarios to await all queries (or mutations) triggered in any way, including via hook calls
 or manually dispatching `initiate` actions.
 
-```ts no-transpile title="Awaiting all currently running queries & mutations example"
-await Promise.all(api.util.getRunningOperationPromises())
+```ts no-transpile title="Awaiting all currently running queries example"
+await Promise.all(dispatch(api.util.getRunningQueriesThunk()))
 ```
 
-## `getRunningOperationPromise`
+## `getRunningQueryThunk` and `getRunningMutationThunk`
 
 #### Signature
 
 ```ts no-transpile
-getRunningOperationPromise: <EndpointName extends QueryKeys<Definitions>>(
+getRunningQueryThunk<EndpointName extends QueryKeys<Definitions>>(
   endpointName: EndpointName,
   args: QueryArgFrom<Definitions[EndpointName]>
-) =>
-  | QueryActionCreatorResult<Definitions[EndpointName]>
+): ThunkWithReturnValue<
+  | QueryActionCreatorResult<
+      Definitions[EndpointName] & { type: 'query' }
+    >
   | undefined
+>
 
-getRunningOperationPromise: <EndpointName extends MutationKeys<Definitions>>(
+getRunningMutationThunk<EndpointName extends MutationKeys<Definitions>>(
   endpointName: EndpointName,
   fixedCacheKeyOrRequestId: string
-) =>
-  | MutationActionCreatorResult<Definitions[EndpointName]>
+): ThunkWithReturnValue<
+  | MutationActionCreatorResult<
+      Definitions[EndpointName] & { type: 'mutation' }
+    >
   | undefined
+>
 ```
 
 #### Description
 
-A function that returns a single promise for a given endpoint name + argument combination,
-if it is currently running. If it is not currently running, the function returns `undefined`.
+Thunks that (if dispatched) return a single running query (or mutation) for a given
+endpoint name + argument (or requestId/fixedCacheKey) combination, if it is currently running.
+If it is not currently running, the function returns `undefined`.
 
-When used with mutation endpoints, it accepts a [fixed cache key](./hooks.mdx#signature-1)
-or request ID rather than the argument.
-
-This is primarily added to add experimental support for suspense in the future.
-It enables writing custom hooks that look up if RTK Query has already got a running promise
-for a certain endpoint/argument combination, and retrieving that promise to `throw` it.
+These thunks are primarily added to add experimental support for suspense in the future.
+They enable writing custom hooks that look up if RTK Query has already got a running query/mutation
+for a certain endpoint/argument combination, and retrieving that to `throw` it as a promise.
--- a/docs/rtk-query/api/created-api/overview.mdx
+++ b/docs/rtk-query/api/created-api/overview.mdx
@@ -53,20 +53,29 @@ type Api = {
       Array<TagTypes | FullTagDescription<TagTypes>>,
       string
     >
-    resetApiState: SliceActions['resetApiState']
-    getRunningOperationPromises: () => Array<Promise<unknown>>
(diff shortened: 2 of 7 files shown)


JulienKode:
I'm not as experience as you in open source and in this codebase but from what I saw:

For the fix I've try to take a look at the codebase and maybe value like `runningPromises` should be create inside a middleware but it seems a big refactoring step to do. As middleware are build with `buildMiddleware`

I think for a fix it's nice as:
- `getRunningOperationPromises` become a thunk (which is needed to be scope per store) so if refactoring needs to be made to move `runningPromises` inside a middleware this can be done without changing the API for the user.

I'll test the fix and let you know,
Thank you again for your help

JulienKode:
Hi @phryneas 

I did some test and it clearly seems to fix the issue, 
Thank you

3imed-jaberi (approved):
+1

markerikson:
@3imed-jaberi I appreciate the enthusiasm, but I would appreciate it if you _didn't_ go through all of our PRs and leave "+1 approve" comments on them :)  It adds to our notifications, and that's kind of annoying.

3imed-jaberi:
from time to time, I make some reviews/PRs/issues ...
I apologize for that! I just try to help ... 😅!

> It would be better to see labels that tell everyone which stage of the current issue/PR to prevent over actions

markerikson:
Yeah, I do appreciate the spirit :)  But, tbh generally other folks don't have the context to properly review these PRs.  Like, in this case, the changes in this PR aren't necessarily done, and there's a lot of complex nuance as to whether we even want to merge this PR in its current form.  So, a "+1 approved" from someone else isn't really helpful.

What we _could_ use help with is triaging other actual issues.

As far as labels and stuff go: yeah, those can be useful, but those also require a lot of time and effort to keep updated (and like I said, we don't generally expect folks who aren't one of the maintainers to be worrying about those anyway).

phryneas:
Okay, updated code & docs as well as the PR description itself - @markerikson do we want to go forward with it like I suggest above?

markerikson:
Looking at https://sourcegraph.com/search?q=context:global+getRunningOperationPromise+-repo:reduxjs+-repo:phryneas+-file:node_modules+-language:markdown&patternType=standard , I can only see 14 actual uses of `getRunningOperationPromise[s]` in the wild.  That's small enough that I'm okay with calling this a minor version change even though it's technically breaking.  Also that it's conceptually a "bug fix". 

We can file issues on those repos separately, or even do the PRs ourselves.

Also given that, I don't think it's worth putting the effort into a codemod.

nasir733:
can somebody help me by providing a complete example snippet for getserversideprops . I am confused about where is the dispatch coming from in the code snippet example in the docs `await Promise.all(dispatch(api.util.getRunningQueriesThunk()))`
@phryneas  
regards nasir

NikkiMather:
I think this has broken a feature I was about to release this week. @markerikson or @phryneas can you please confirm if this change is correct? The documentation around it provides absolutely no useful examples, which is a bit frustrating.

```
store.dispatch(retrieveContent.initiate());
await Promise.all(getRunningOperationPromises());
```

Becomes

```
store.dispatch(retrieveContent.initiate());
await Promise.all(dispatch(api.util.getRunningQueriesThunk()))
```

Is this correct? That change caused the following error; `ReferenceError: dispatch is not defined`

phryneas:
@nasir733 Well... `dispatch is not defined`
We assume that you have a function called `dispatch`, but apparently you were calling `store.dispatch` - do you have to use that.

browyn:
I'm trying to run rtkQuery during SSR, all goes well until it gets to dispatching the `getRunningQueriesThunk` and I get the following function `TypeError: _features_api_subApis_propertyApi__WEBPACK_IMPORTED_MODULE_4__.propertyApi.util.getRunningQueriesThunk is not a function`

```
export const getServerSideProps = reduxWrapper.getServerSideProps(
  (store) => async (context) => {
    const propertyId = context.params?.propertyId;

    store.dispatch(getProperty.initiate(propertyId));

    const response = await Promise.all(
      store.dispatch(propertyApi.util.getRunningQueriesThunk())
    );
    console.log(response);

    return {
      props: {},
    };
  }
);
```

I have tried looking for solution online, but I found none,
When I check the methods/functions of the `api.util`, I see the old `getRunningOperationPromises`  and cannot find `getRunningQueriesThunk`
Please how can I solve this?

markerikson:
@browyn what version of RTK do you have installed?

browyn:
It's version 1.8.6
The code base is legacy 🥲

browyn:
Perhaps, I should take my time to update the entire code base.
Been worried about that, so I don't get to break anything.

I just tried created a new project just to test out the `getRunningQueriesThunk` and it works

markerikson:
@browyn yeah, we release the fix in 1.9.0 :)  You just need to upgrade to 1.9.x.

browyn:
Thanks @markerikson

Problem solved!

shinhyogeun:
hi. @markerikson i'm in 1.9.5 version

and i make custom hook for Suspense

but i got error in this code.

```jsx
...

const info = useEndpointQuery(params, { ...(options || {}) });

if (info.isError) {
    throw info.error;
}

  if (Loading) {
    const promise = dispatch(api.util.getRunningQueryThunk(endpointName, params));

    throw promise
  }
```


because `const promise` is `undefined` in first time.

if i console.log promise value .

image


but i have to throw promise for suspense. 

Can you give me some advice? 🥲

markerikson:
@shinhyogeun that's expected, yes. The query doesn't actually start until a `useEffect` runs _after_ the render.

We don't have any official support for Suspense yet, although we're kicking around some ideas.

shinhyogeun:
thank you 🙏 @markerikson 

i found if i write code like this. i can get promise right from the start.

```jsx

if (info.isError) {
    throw info.error;
  }

  if (Loading) {
    dispatch(api.endpoints[`${endpointName}`].initiate(params));

    const promise = dispatch(api.util.getRunningQueryThunk(endpointName, params));

     throw promise;
  }

```

why is this possible?

and i also wonder if `getRunningQueryThunk` can give promise every time it polling.

i also check your code in https://github.com/replayio/devtools/pull/7205/files#diff-ffc9c20da71515be364ceb20ce82f542e4b96ae7fec51a2bb769d3dcc338ba84

phryneas:
> why is this possible?

Because `initiate` starts the request by hand. But be careful, that creates a subscription and will never be collected from cache that way.

PR #2493, opened 2022-07-06
feature : endpoint-specific args serializers

michal-kurz:
https://stackoverflow.com/questions/72800748/is-it-possible-to-scope-serializequeryargs-alongside-injected-endpoints

This seems to me to be the least overhead to achieve such thing - it seems to me I can't do this without such wrapper around the current `serializeQueryArgs`, beacuse `serializeQueryArgs` gets bound to modules at time of calling `createApi`.

I haven't found a convenient way of running this code inside an actual React app to confirm the keys look how I expect them to - the only confirmation I have that it works as expected, are the passing tests, which is not too convincing. Would someone please guide me to an easy way of running the code custom branch inside an actual React app? The in-repo examples didn't compile for me. I also tried using my fork repo [as an npm package source](https://stackoverflow.com/a/16350674/10467064) for `@reduxjs/toolkit`, but I couldn't get it to work.

Or, even better, is there a way to access cache keys from `api` object inside Jest, so I can include it in tests?


packages/toolkit/src/query/buildSerializeQueryArgs.ts
```diff
@@ -0,0 +1,34 @@
...
+    serializer: SerializeQueryArgs<any>
+  ) => {
+    endpointSpecificSerializers[endpointName] = serializer
+  }
+
+  return {
+    serializeQueryArgs,
+    registerArgsSerializerForEndpoint,
+  }
+}
```
michal-kurz, line 34: It would make sense to me to move the contents of `defaultSerializeQueryArgs` inside this file - but I haven't yet done that to get a clearer PR diff

packages/toolkit/src/query/tests/createApi.test.ts
```diff
@@ -819,3 +820,68 @@ describe('structuralSharing flag behaviors', () => {
...
+
+    const firstRef =
+      injectedApi.endpoints.injectedQueryWithCustomSerializer.select(5)(
+        storeRef.store.getState()
+      )
+    expect(serializer2).toHaveBeenCalled()
+
+    expect(firstRef.data).toEqual({ value: 'success' })
+  })
+})
```
michal-kurz, line 887: These tests don't seem very convincing to me, but I don't know how to test for anything more specific :/

packages/toolkit/src/query/buildSerializeQueryArgs.ts
```diff
@@ -0,0 +1,34 @@
...
```
michal-kurz, line 34: It would make sense to me to move the contents of defaultSerializeQueryArgs inside this file - but I haven't yet done that to get a clearer PR diff

packages/toolkit/src/query/createApi.ts
```diff
@@ -236,15 +236,18 @@ export function buildCreateApi<Modules extends [Module<any>, ...Module<any>[]]>(
...
     const optionsWithDefaults = {
       reducerPath: 'api',
-      serializeQueryArgs: defaultSerializeQueryArgs,
       keepUnusedDataFor: 60,
       refetchOnMountOrArgChange: false,
       refetchOnFocus: false,
       refetchOnReconnect: false,
       ...options,
       extractRehydrationInfo,
+      serializeQueryArgs,
```
phryneas, line 250: If you inline a `serializeQueryArgs` function here instead, it can access `context.endpointDefinitions[endpointName].serializeQueryArgs` here and you won't need to implement a `registerArgsSerializerForEndpoint` function and maintain that extra data.

phryneas:
Heyo, great job of digging in so deep already! 

I think it would be possible to implement this without having to maintain an extra `endpointSpecificSerializers` dictionary. 
Could you give that another look with the info from my in-code comment? :)

markerikson:
Picked this up.

I inlined the `serializeQueryArgs` function to use the endpoint-specific options, removed the "registration" handling, and updated the tests to check for the specific serializer result values as keys in the query state.

Thanks for getting this going!

PR #2500, opened 2022-07-07
Use useDebugValue in useQuery and useMutation hooks

### LINKED ISSUE #2492: RTK Query hooks should use `useDebugValue` to better display data
I'm looking at a `useQuery` hook's values in the React DevTools, and I have to dig down to inspect the contents:

image

We should use the `useDebugValue` hook to show this better - at a minimum `data`, and possibly some of the flags too.

kaankeskin:
Fixes #2492
This PR uses `useDebugValue` for better debugging in React DevTools.

This is how a `DebugValue` is shown for a useQuery hook in React DevTools:
image

and this is how a `DebugValue` is shown for a useMutation hook in React DevTools:
image

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -4,6 +4,7 @@ import type { Selector } from '@reduxjs/toolkit'
 import type { DependencyList } from 'react'
 import {
   useCallback,
+  useDebugValue,
   useEffect,
   useLayoutEffect,
   useMemo,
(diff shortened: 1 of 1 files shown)


markerikson:
Very nice!

Looking at the React docs, would it be worth using the "deferred formatting" option here?

https://reactjs.org/docs/hooks-reference.html#defer-formatting-debug-values

kaankeskin:
It does not worth I think. Because, I am passing an object to `useDebugValue`. I may use `JSON.stringify` with it but it makes debugging harder since it shows it as string and the object is not "collapsible".
Here like that:
image

markerikson:
Let's get this in!

kaankeskin:
Yay! Thanks for merging this. Having a commit in a project I love made me really happy :)

PR #2513, opened 2022-07-14
Downgrade msw to v40.2 in some usage examples

frankie303:
While I was going through Redux Toolkit Query documentation, I encountered some errors in some of the usage examples (kitchen sink, optimistic updates, authentication, authentication with extraReducers, mutations, pagination, prefetching, prefetching automatic, prefetching automatic waterfall). After some investigation, I realized that it's because of the version of msw. You can find related issue [here](https://github.com/mswjs/msw/issues/1252). After I downgraded the msw version to 40.2, the problem seems to be solved. You can find the screenshot of one of the broken examples below: 

Screenshot 2022-07-14 at 16 47 08

### DIFF
--- a/examples/query/react/authentication-with-extrareducers/package.json
+++ b/examples/query/react/authentication-with-extrareducers/package.json
@@ -11,7 +11,7 @@
     "@emotion/styled": "^11.3.0",
     "@reduxjs/toolkit": "^1.6.0-rc.1",
     "framer-motion": "^2.9.5",
-    "msw": "^0.41.1",
+    "msw": "^0.40.2",
     "react": "^18.1.0",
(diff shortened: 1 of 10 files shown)


markerikson:
Thanks!  In order for this to work, you also need to re-run `yarn` from the root of the repo to update the `yarn.lock` file, and commit that.  Can you make that change, and also double-check if any other examples need to have their `msw` version changed?

frankie303:
I updated `yarn.lock` and updated some of the other examples' msw version as well. I think it's now ready :) @markerikson

frankie303:
okay, they work fine. React 18 warnings were irrelevant.

markerikson:
Thank you!

Issue #2448, opened 2022-06-24
Remove abortcontroller polyfill 


JulienKode:
Currently there a polyfill of AbortController in createAsyncThunk but not in listeners middleware.

The goal of this issue is to remove the polyfill for the v2

Related to: https://github.com/reduxjs/redux-toolkit/pull/2446

phryneas:
Thanks for the issue! :)

JulienKode:
@phryneas I've created https://github.com/reduxjs/redux-toolkit/pull/2518 if you think this is useful

PR #2518, opened 2022-07-16
fix(2448): remove abort controller polyfill

JulienKode:
# Description 

In this PR we remove abort controller https://github.com/reduxjs/redux-toolkit/issues/2448 

Where should I add in the docs that ?
- For people that are using redux with a node.js server they:
    - need node.js 15 or have to polyfill `AbortController`

I'll update specs and add docs if we are ok

### DIFF
--- a/packages/toolkit/src/createAsyncThunk.ts
+++ b/packages/toolkit/src/createAsyncThunk.ts
@@ -553,36 +553,6 @@ export const createAsyncThunk = (() => {
         })
       )
 
-    let displayedWarning = false
-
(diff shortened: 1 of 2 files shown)


phryneas:
Let's target that for a 2.0 release :)

JulienKode:
@phryneas @markerikson let me know if I need to do any changes

markerikson:
Belated, but thank you!

JulienKode:
Thanks for merging it for 2.0

PR #2544, opened 2022-07-26
🏷️ Update createSlice type to account for upcoming TS changes

Shrugsy:
Experimentation at this stage - just want the CSB build

Not completely sure that the CI type-test is working


Shrugsy:
Covered under #2547 & #2548

PR #2547, opened 2022-07-26
`createSlice`: correctly infer `state` in TS 4.8

### LINKED ISSUE #2543: State argument on createSlice is no longer inferred with typescript beta 4.8 and they do not plan to fix it
See the bug on the typescript repo: https://github.com/microsoft/TypeScript/issues/49307#issuecomment-1195031370

It seems the typescript team will not fix this. Maybe the redux-toolkit types can be fixed to be compatible and still infer? else the examples and documentation need changing to say that the state argument will not be inferred and need to be explicitly provided.

cc/ @markerikson

phryneas:
fixes #2543

### DIFF
--- a/.github/workflows/test-codegen.yml
+++ b/.github/workflows/test-codegen.yml
@@ -6,27 +6,18 @@ defaults:
   run:
     working-directory: ./packages/rtk-query-codegen-openapi
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'packages/rtk-query-codegen-openapi/**'
+  pull_request:
+    paths:
+      - 'packages/rtk-query-codegen-openapi/**'
 
 jobs:
-  changes:
-    name: Check for changes
-    runs-on: ubuntu-latest
-    outputs:
-      codegen: ${{ steps.filter.outputs.codegen }}
-    steps:
-      - uses: actions/checkout@v2
-      - uses: dorny/paths-filter@v2
-        id: filter
-        with:
-          filters: |
-            codegen:
-              - 'packages/rtk-query-codegen-openapi/**'
-
   build:
-    needs: changes
-    if: ${{ needs.changes.outputs.codegen == 'true' }}
-
     runs-on: ubuntu-latest
 
     strategy:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -96,7 +96,8 @@ jobs:
       fail-fast: false
       matrix:
         node: ['16.x']
-        ts: ['4.1', '4.2', '4.3', '4.4', '4.5', '4.6', '4.7']
+        ts: ['4.1', '4.2', '4.3', '4.4', '4.5', '4.6', '4.7', 'next']
+    continue-on-error: ${{ matrix.ts == 'next' }}
     steps:
       - name: Checkout repo
         uses: actions/checkout@v2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,28 +1,20 @@
 name: CI
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'packages/toolkit/**'
+  pull_request:
+    paths:
+      - 'packages/toolkit/**'
+
 defaults:
   run:
     working-directory: ./packages/toolkit
 
 jobs:
-  changes:
-    name: Check for changes
-    runs-on: ubuntu-latest
-    outputs:
-      toolkit: ${{ steps.filter.outputs.toolkit }}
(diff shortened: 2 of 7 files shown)


phryneas:
This PR contains the fix - but right now I've commented the fix out and am still not getting the error in CI... I'll have to drop this. @Shrugsy do you maybe have resources to play around and find out why the CI isn't doing here?

Shrugsy:
> This PR contains the fix - but right now I've commented the fix out and am still not getting the error in CI... I'll have to drop this. @Shrugsy do you maybe have resources to play around and find out why the CI isn't doing here?

I have a suspicion that the GH actions yaml file may not be formatted correctly, so it may run the "tsc --version" line, but not actually run the type tests. I won't have a chance to look at it properly for a couple of days or so. It might be worth removing the "tsc --version" line temporarily to see whether this is the case.

Shrugsy:
> > This PR contains the fix - but right now I've commented the fix out and am still not getting the error in CI... I'll have to drop this. @Shrugsy do you maybe have resources to play around and find out why the CI isn't doing here?
> 
> I have a suspicion that the GH actions yaml file may not be formatted correctly, so it may run the "tsc --version" line, but not actually run the type tests. I won't have a chance to look at it properly for a couple of days or so. It might be worth removing the "tsc --version" line temporarily to see whether this is the case.

☝️ Looks like a dead end - not the issue

Meligy:
Hello,
I know you might want to wait until you fix the type tests runner etc., but thought it might be useful that I tried this in a real customer project just now, one that started failing when I tried the `next` version of TypeScript on it (and it resolved to ``). And I can confirm that the change to `ValidateSliceCaseReducers` in this PR fixed the problem. 👍

I tested it by modifying `node_modules/@reduxjs/toolkit/dist/createSlice.d.ts` directly, as I just wanted to test it and report here, and patching it in a custom declaration file didn't seem work, probably because I'll also need to to copy `createSlice` and `CreateSliceOptions` in the custom declaration file.

Which I also later did to see if I can use that as a temporary workaround. I added the following to the `react-app-env.d.ts` file that came with the create-react-app template:

```ts
// Workaround for `createSlice` reducer functions
//   not inferring the state type automatically in TypeScript 4.8
// Should be removed when this PR is released:
//   https://github.com/reduxjs/redux-toolkit/pull/2547
module "@reduxjs/toolkit" {
  export * from "@reduxjs/toolkit/dist";
  import type {
    CreateSliceOptions as original_CreateSliceOptions,
    Slice,
    SliceCaseReducers,
  } from "@reduxjs/toolkit/dist";

  // The main change in the PR
  export type ValidateSliceCaseReducers<
    S,
    ACR extends SliceCaseReducers<S>
  > = ACR & {
    [T in keyof ACR]: ACR[T] extends {
      reducer(s: S, action?: infer A): any;
    }
      ? {
          prepare(...a: never[]): Omit<A, "type">;
        }
      : ACR[T];
  };

  // The following types are copied from `@reduxjs/toolkit` with minimal changes
  //    as if we do not override them, `createSlice` will not use our custom `ValidateSliceCaseReducers`
  // So we override `createSlice`'s `CreateSliceOptions`
  //    and change the `reducers` option to point to our custom `ValidateSliceCaseReducers`

  export interface CreateSliceOptions<
    State = any,
    CR extends SliceCaseReducers<State> = SliceCaseReducers<State>,
    Name extends string = string
  > extends original_CreateSliceOptions<State, CR, Name> {
    reducers: ValidateSliceCaseReducers<State, CR>;
  }

  export declare function createSlice<
    State,
    CaseReducers extends SliceCaseReducers<State>,
    Name extends string = string
  >(
    options: CreateSliceOptions<State, CaseReducers, Name>
  ): Slice<State, CaseReducers, Name>;
}
```

Thanks a lot for providing the solution, and I hope it can be released soon.

phryneas:
@Meligy you can also always try the latest CodeSandbox build for a PR - in this case you find install instructions here: https://ci.codesandbox.io/status/reduxjs/redux-toolkit/pr/2547/builds

phryneas:
So with this change (which unfortunately adds a dependency), the patch will only be applies to TS versions >= 4.8, as it would break older TS versions.

Unfortunately we can't really test that at the moment as the current "next" release still labels itself as 4.7.

phryneas:
Now it works already with pre-release versions.

The type bugs that are still appearing are legit 4.8 TS bugs, but only occur in the "slice wrapped in generics" case, so with a whole level of complexity added on top. 

I'd suggest we get this fix out as a 1.8.* regardless of that as it will already alleviate 95% of errors people will have when upgrading to TS 4.8. 

As I'll be going on my vacation tomorrow I don't think I'll be able to look deep enough into the "generics error case".

PR #2409, opened 2022-06-12
[Chore][maintenance] upgrade several deps | use react 18 #2404

FaberVitale:
#### React 18
 
Upgrades react and react-dom to 18 in all workspaces except `website` and docs.

#### toolkit tests

`@reduxjs/toolkit` tests now use react 18 renderers thanks to `@testing-library/react 13`.

#### Examples

All examples now use react 18, react-redux 8, react-router-dom 6 & react-scripts 5.

#### Website

We could not upgrade react to 18 in website and docs because docusaurus doesn't support react 18 yet, see https://github.com/facebook/docusaurus/issues/7264. 

### Upgraded dependencies

- react `17 -> ^18.1.0`
- react-dom `17 -> ^18.1.0`
- react-scripts (cra) `4.0.3 -> ^5.0.1`
- react-redux `7.2.2 -> ^8.0.2`
- @types/react `17 -> ^18.0.12`
- @types/react-dom `17 -> ^18.0.5`
- react-router-dom `5.2.0 -> 6.3.0`
- @testing-library/react `^11.2.6 -> ^13.3.0`
- @testing-library/user-event `^13.1.5 -> ^14.2.0`

### Removed dependencies

- @types/react-redux 🎉 - see https://github.com/reduxjs/react-redux/releases/tag/v8.0.0
- @testing-library/react-hooks - see https://github.com/testing-library/react-hooks-testing-library/issues/849

Closes https://github.com/reduxjs/redux-toolkit/issues/2404

PR #2548, opened 2022-07-27
🐛 Fix type tests not checking files

Shrugsy:
Partially reverts some changes from #2409

#2409 Updated [@testing-library/user-event](https://github.com/testing-library/user-event) from `^13.1.5 -> ^14.2.0`. As part of the upgrade, `user-event` internally referenced the `ClipboardItem` type, which was not globally available for TS versions < 4.4 (see https://stackoverflow.com/a/61505395 for more info). @FaberVitale added a declaration file to address this, but it seems to have prevented any other files being included by `tsconfig` except for the declaration file itself.

Hence, the `type-tests` part of the github workflow hasn't actually been checking anything.

This PR reverts those changes, downgrading `user-event` back to `^13.1.5`, and reverting the changes to `tsconfig.test.json`, allowing the `type-tests` command to once again check all files.

Note: If anyone is able to get the `ClipboardItem` type declaration to be respected by `user-event`'s own declaration files, a further PR is welcome.


Shrugsy:
As a sanity check, here is a failing test run with TS `next` (Version 4.8.0-dev.20220726). Note that this is currently expected to fail, and will be addressed by #2547

https://github.com/Shrugsy/redux-toolkit/runs/7532322057?check_suite_focus=true

markerikson:
Huzzah!

Issue #2593, opened 2022-08-14
Rework subscription handling to batch "need to subscribe but entry exists" behavior


markerikson:
We've had several folks report that rendering a long list of components with RTKQ hooks can be very slow. My suspicion is that it's because we dispatch separate "need to subscribe" actions for every component.  The first component will fetch the data _and_ create the cache entry. All other components will see the cache entry exists, dispatch a `rejected` action, and the subscription logic handles that to add another subscription.

Additionally, as components unsubscribe, there's a lot of timer cancelation and resetting going on.

I think we need to figure out how to batch the "cache entry exists but need to subscribe" behavior into some debounced batched actions to cut down on the overhead.  We may also need to tweak the unsubscribe handling as well to cut down on timer manipulation.

Don't have a performance capture atm - I should set up an example.

markerikson:
Related: I just fixed `keepUnusedDataFor` timer counters in #2595 . Lenz also suggested:

> Maybe with that knowledge, we should also allow for some kind of "keep indefinitely" option with `Math.INFINITE` or just `-1`?

I like `Math.INFINITE` myself.  That kinda ties in here since I need to rethink the timer behavior anyway, so note to myself to add that as part of this work.

markerikson:
Both done in #2599 .

markerikson:
note for later reference - there's an example repo at https://github.com/nibblebot/rtkq-timer-perf :

> I setup a repo here https://github.com/nibblebot/rtkq-timer-perf that reproduces the issue. It simply renders a list of 100 components with queries bound to them and a button that toggles the whole list. If you do a chrome profile and filter the events to "Install Timer" or "Remove Timer" you can see the 100 timer install/remove when you toggle the list. However, the cost of Install/Remove Timer here appears to be 0ms so the impact is not the same as my other repro case in my private code base. Testing with the 1.9.0-alpha.0 shows these events are removed from the profile event log so I'd say the new version fixes the issue.

PR #2595, opened 2022-08-14
Prevent `keepUnusedDataFor` values from overflowing `setTimeout` counter

### LINKED ISSUE #2535: RTK-Query - keepUnusedDataFor has problems with very high values
I use RTK-Query for api fetches in my project, and I've met with an issue I can't fully understand.

Use Case:

I open a modal, once the modal mounts, I call a query endpoint by the auto generated query hook. So this will trigger a couple of actions, first "api/executeQuery/pending", then "api/executeQuery/fulfilled", so far so good, everything works as expected, data is in store, we are happy. Then I close the modal, component unmounts, this triggers yet another 2 actions, first "api/subscriptions/unsubscribeQueryResult", then "api/queries/removeQueryResult", this will do what it says, remove the data, because there are no more subscribers. 

So my goal is to prolong the cache time of that query result, because this modal is being opened frequently, so going through the docs I see "keepUnusedDataFor" option which in theory would do just I what I want. 

Reference from docs: 
**"Once the subscription is removed (e.g. when last component subscribed to the data unmounts), after an amount of time (default 60 seconds), the data will be removed from the cache. The expiration time can be configured with the keepUnusedDataFor"**

As I understand this, for my use-case, the query that ran when the modal was open, should be kept, for at least 60 seconds, till after the component unmounts. But in reality, the "api/subscriptions/unsubscribeQueryResult" and the "api/queries/removeQueryResult" actions run immediately after modal unmounts, there is no 60 seconds delay. I've tried setting it to longer, but was cheap shot obviously, I miss out on something or just not understanding clearly how this should behave.

Ideas?

Package versions:
    "redux": "^4.0.5",
    "react-redux": "^7.2.2",
    "redux-thunk": "^2.3.0",
    "@reduxjs/toolkit": "^1.8.3",

markerikson:
Fixes #2535

Turns out `setTimeout` counters, which are in milliseconds, are normally stored in a 32-bit counter.  So, 24.8 days is the longest you can safely provide (the usual 2.1 billion number, etc), otherwise it wraps and ends up executing _immediately_ instead - oops! 

Clamped it to be a max of 1s less, and also a min of 0.

### DIFF
--- a/packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
@@ -28,6 +28,13 @@ declare module '../../endpointDefinitions' {
   }
 }
 
+// Per https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#maximum_delay_value , browsers store
+// `setTimeout()` timer values in a 32-bit int. If we pass a value in that's larger than that,
+// it wraps and ends up executing immediately.
+// Our `keepUnusedDataFor` values are in seconds, so adjust the numbers here accordingly.
+export const THIRTY_TWO_BIT_MAX_INT = 2_147_483_647
+export const THIRTY_TWO_BIT_MAX_TIMER_SECONDS = 2_147_483_647 / 1_000 - 1
+
 export const build: SubMiddlewareBuilder = ({ reducerPath, api, context }) => {
   const { removeQueryResult, unsubscribeQueryResult } = api.internalActions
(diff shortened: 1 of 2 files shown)


phryneas:
Maybe with that knowledge, we should also allow for some kind of "keep indefinitely" option with `Math.INFINITE` or just `-1`?

markerikson:
Added a note to #2593 to add `Math.INFINITE` checks, as that task will need to rework timer behavior anyway.

PR #2604, opened 2022-08-16
add createAsyncThunk.withTypes

phryneas:
This might need some yak shaving with the name and I also want to maybe take another look at how `.forTypes` is called, because like this it gets more restrictive with every time it's called, but this should serve as a good prototype.

**Please test and give comments.**

### DIFF
--- a/packages/toolkit/src/tests/createAsyncThunk.typetest.ts
+++ b/packages/toolkit/src/tests/createAsyncThunk.typetest.ts
@@ -12,12 +12,13 @@ import type { ThunkDispatch } from 'redux-thunk'
 import type { AxiosError } from 'axios'
 import apiRequest from 'axios'
 import type { IsAny, IsUnknown } from '@internal/tsHelpers'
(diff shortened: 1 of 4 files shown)


packages/toolkit/src/tests/createAsyncThunk.typetest.ts
```diff
@@ -590,3 +596,54 @@ const anyAction = { type: 'foo' } as AnyAction
     async (_, api) => api.rejectWithValue(5, '')
   )
 }
+
+{
+  const typedCAT = createAsyncThunk.forTypes<{
+    state: RootState
+    dispatch: AppDispatch
+  }>()
```
phryneas, line 604: This seems to work without circular type problems!

markerikson:
I'd call it `.withTypes`, myself.  (Some parallel with `thunk.withExtraArgument`, for example.)

phryneas:
I think the types are now where I want them, wanna take another look? :)

PR #2629, opened 2022-08-25
Fix the `dispatch` type inference to correctly handle read-only middleware arrays

### LINKED ISSUE #2628: Weak type inference for the `EnhancedStore.dispatch` method
## Problem
When you decorate the `configureStore` function or pass a read-only middleware array in the options, the `EnhancedStore.dispatch` has the type `never`.

## Cause
The `ConfigureStoreOptions` interface with the default type parameters will have the following type of the [`middleware`](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/configureStore.ts#L59) property:
```typescript
middleware?: ((getDefaultMiddleware: CurriedGetDefaultMiddleware<S>) => Middlewares<S>) | Middlewares<S>;
```

The `Middlewares` macro is declared below:
```typescript
declare type Middlewares<S> = ReadonlyArray<Middleware<{}, S>>;
```

In this case, the `EnhancedStore` tries to [infer](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/configureStore.ts#L116) the `dispatch` method type from the `ReadonlyArray`. This will end up in [this](https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/tsHelpers.ts#L91) check:
```typescript
M extends Middleware[]
  ? ExtractDispatchFromMiddlewareTuple<[...M], {}>
  : never
```
The expression above will always return `never` for any `ReadonlyArray` because `Array` ⊂ `ReadonlyArray`.

### Decorated `configureStore`
```typescript
const someSlice = createSlice({
  name: 'something',
  initialState: null as any,
  reducers: {
    set(state) {
      return state;
    },
  },
});

function configureMyStore<S>(options: Omit<ConfigureStoreOptions<S>, 'reducer'>) {
  return configureStore({
    ...options,
    reducer: someSlice.reducer,
  });
}

store.dispatch(/* ... */); // error
```

### Read-only `middleware`
```typescript
const middleware = [/* ... */] as ReadonlyArray<Middleware>;
  middleware,
});

store.dispatch(/* ... */); // error
```

dokmic:
Resolves #2628.

All the details are in the linked issue.

The solution fixes the TypeScript helper to use a more generic array type to align with the `ConfigureStoreOptions` interface.

### DIFF
--- a/packages/toolkit/src/tests/configureStore.typetest.ts
+++ b/packages/toolkit/src/tests/configureStore.typetest.ts
@@ -13,6 +13,7 @@ import {
   configureStore,
   getDefaultMiddleware,
   createSlice,
+  ConfigureStoreOptions,
 } from '@reduxjs/toolkit'
 import type { ThunkMiddleware, ThunkAction, ThunkDispatch } from 'redux-thunk'
 import thunk from 'redux-thunk'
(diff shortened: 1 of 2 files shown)


dokmic:
@markerikson Thanks for looking into this! I've added a couple of tests covering the cases I brought in the issue.

markerikson:
Yeah, on my list of stuff to look at in the next few days when I've got time!

PR #2641, opened 2022-08-27
Consolidate RTKQ middleware to simplify stack size

markerikson:
Previously, the RTKQ middleware was made up of 7 "sub-middleware", each encapsulating a different responsibility (polling, caching, etc).  In practice, these _were_ each a real middleware, just pre-composed into a single middleware internally.

However, that meant that each middleware was called on _every_ action, even if it wasn't RTKQ-related. That adds to call stack size.

Almost all the RTKQ processing logic runs _after_ the action is handled by the reducers. So, I've reworked the files to create "handlers" (conceptually similar to the `(action) => {}` part of a middleware, with a couple extra args), and rewritten the top-level middleware to run those in a loop.  The middleware also only runs those handlers if the action appears to be related to this API slice.

This should speed up perf overall because there's fewer calls per action in all cases, as well as a shallower call stack.  (This will especially help if users are adding multiple API slices+middleware).

Note that much of the diff is just unindenting the body of each middleware/handler - you might want to turn off "Whitespace" in the diffs.

For now, I'm keeping the "batch similar subscription actions" behavior as-is, with an explicit check before the action goes to the reducers.  That may get obsoleted by the batching enhancer proposed over in #2621 .

### DIFF
--- a/packages/toolkit/src/query/core/buildMiddleware/devMiddleware.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/devMiddleware.ts
@@ -1,47 +1,34 @@
-import type { SubMiddlewareBuilder } from './types'
+import type { InternalHandlerBuilder } from './types'
 
-export const build: SubMiddlewareBuilder = ({
+export const buildDevCheckHandler: InternalHandlerBuilder = ({
   api,
   context: { apiUid },
   reducerPath,
 }) => {
-  return (mwApi) => {
-    let initialized = false
-    return (next) => (action) => {
-      if (!initialized) {
-        initialized = true
-        // dispatch before any other action
-        mwApi.dispatch(api.internalActions.middlewareRegistered(apiUid))
-      }
+  return (action, mwApi) => {
+    if (api.util.resetApiState.match(action)) {
+      // dispatch after api reset
+      mwApi.dispatch(api.internalActions.middlewareRegistered(apiUid))
+    }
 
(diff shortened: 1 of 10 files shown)


packages/toolkit/src/query/core/buildMiddleware/devMiddleware.ts
```diff
@@ -1,47 +1,34 @@
...
-    let initialized = false
-    return (next) => (action) => {
-      if (!initialized) {
-        initialized = true
-        // dispatch before any other action
-        mwApi.dispatch(api.internalActions.middlewareRegistered(apiUid))
-      }
-
-      const result = next(action)
-
```
phryneas, line 18: If all of this went into the initialization, should maybe the rest of it also go there? Like this, the logic is pretty split up between the files.

phryneas (commented):
It seems valid to me. I've reread it multiple times for cases where things should be initialized "once per store instance" and I think all of those are right, but if anything would break, it would probably be one of those - so we should keep an eye out for that.

packages/toolkit/src/query/core/buildMiddleware/devMiddleware.ts
markerikson: I could, yeah. Conceptually it felt like this file was split between "one-time init" and "dev check warning", so I left the dev warning in here.

PR #2643, opened 2022-08-28
Try fixing createAsyncThunk issues with TS 4.8

markerikson:


markerikson:
okay, this is nuts. The PR fixed the TS errors with `createAsyncThunk`...

but we're still _also_ seeing errors from files in `node_modules`, even though we've got `skipLibCheck: true` turned on:

```
Error: node_modules/msw/lib/glossary-eda04a15.d.ts(65,56): error TS2370: A rest parameter must be of an array type.
Error: node_modules/msw/lib/glossary-eda04a15.d.ts(115,76): error TS2344: Type 'ResponseType' does not satisfy the constraint 'Record<string | symbol, any>'.
Error: node_modules/type-fest/ts41/get.d.ts(95,37): error TS2344: Type 'BaseType' does not satisfy the constraint 'Record<string | number, any>'.
Error: node_modules/type-fest/ts41/get.d.ts(96,19): error TS2344: Type 'BaseType' does not satisfy the constraint 'Record<string | number, any>'.
Error: Process completed with exit code 2.
```

what in the world?

markerikson:
Okay, apparently we had `skipLibCheck: false` in our `tsconfig.test.json` file.

I don't know _why_ we ever set that to be `false`.

But turning it back to `true` fixes those errors locally.

I also tweaked a line where TS was complaining that "you're checking `window.document.createElement`, it'll always exist, right?".

markerikson:
More updates:

apparently our CI is configured to treat `@reduxjs/toolkit` as a lib, so that's why we have `skipLibCheck: false` in `tsconfig.test.json`.

Attempting to patch `msw` to add `// @ts-ignore` comments to the offending lines...

markerikson:
YAYYYY that compiles!

Issue #2389, opened 2022-06-07
rtkQuery lifecycle hook to invalidate cache from side-effects


schadenn:
Hi everyone,

I am currently experiencing a use-case that seems like it is not treated at all at the moment:
I have a query that may have its results invalidated by a side effect that is not within my control (as in: I can't call the invalidate function when the side effect happens).
Now I'm thinking it would be great to have something like a `onQueryInitiated` lifecycle event which will be triggered before the `reject` action would be triggered and therefore provide an integration point to check if the mentioned side effects happened in the meantime and if yes - call the `invalidateTags` function.

At the moment the only way to do this would be to wrap the autogenerated hook, which feels a little cumbersome.

Wdyt? I'd try implementing this if it is considered to be a good idea.

markerikson:
Sorry, not sure I'm following the described scenario here.  Can you give some concrete examples, and maybe show what the imaginary code for this feature might look like?

markerikson:
Hmm. Is the https://redux-toolkit.js.org/rtk-query/api/createApi#onquerystarted basically what you're looking for, or is there some difference in behavior / timing from that function?

PR #2663, opened 2022-09-01
Add `forceRefetch` to `QueryExtraOptions`

schadenn:
Resolves: #2389 

This PR is to further explain our use-case. We check for a timestamp of the last time a side-effect handled and if the timestamp is newer than the last time the endpoint was called (which means something that'll change the return of the query has happened in the meantime), then the call will be forced.

This still needs some testing and docs, but I wanted to make sure that the concept is something you could imagine being valuable.

### DIFF
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -279,20 +279,10 @@ Features like automatic cache collection, automatic refetching etc. will not be 
           endpointName,
         })
 
-        const endpointContext = context.endpointDefinitions[endpointName]
-        const sideEffectForced =
-          isQueryDefinition(endpointContext) &&
-          endpointContext.sideEffectForced?.({
-            getState,
-            endpointState: (
-              api.endpoints[endpointName] as ApiEndpointQuery<any, any>
-            ).select(arg)(getState()),
-          })
-
         const thunk = queryThunk({
           type: 'query',
           subscribe,
-          forceRefetch: forceRefetch || sideEffectForced,
+          forceRefetch: forceRefetch,
           subscriptionOptions,
           endpointName,
           originalArgs: arg,
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -5,11 +5,10 @@ import type {
   QueryArgFrom,
   ResultTypeFrom,
 } from '../endpointDefinitions'
-import { DefinitionType } from '../endpointDefinitions'
+import { DefinitionType, isQueryDefinition } from '../endpointDefinitions'
 import type { QueryThunk, MutationThunk, QueryThunkArg } from './buildThunks'
 import type { AnyAction, ThunkAction, SerializedError } from '@reduxjs/toolkit'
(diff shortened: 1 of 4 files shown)


markerikson:
Hmm. Skimming that issue, and this PR, I'm still not sure I understand the use case, and I _definitely_ don't understand the naming. What in the world is a `sideEffectForced` callback? What does that even _mean_ conceptually?

phryneas:
I *think* while the naming is a bit unfortunate having something like this could be a building block to solve https://github.com/reduxjs/redux-toolkit/discussions/2620#discussioncomment-3521798

schadenn:
Thanks for the quick feedback. 🙏
I was very unsure about the naming as well and just wanted to put something out here to try to explain our reasoning. In the issue I opened I was talking about a "lifecycle hook `onQueryInitiated`" but looking through the rtkQ codebase today I found that it probably wouldn't work as a lifecycle hook (or if it would it'd be way more complex).


Sadly I can't share the with you but I'll try to explain the use-case again. I'll try to set up a reproduction if you want me to.

In our setup, there's two sections in a big app. Both being built by different teams. Both are - to a certain extend - using the same underlying services though.
Now let's say user takes this path:
- Enters section of team 1, query is fired and result is cached.
- Navigates to section of team 2, here they perform actions that'll alter some data through their service.
- User comes back to section owned by team 1. Cached data is now invalid, but the request will not be performed again.

Having team 2 use team 1s `invalidateTags` functionality upon altering data, would solve the issue, but it's nothing team 1 should be reliant on as it'd be way too easy for team 2 to break team 1s functionality without them realizing. 
Which is why we'd want to "inverse control" by creating a way for team 1 to check for side effects (whatever they might be) and ignore the cached result if the side effects have happened.

phryneas:
@markerikson Maybe to explain how I see this: This would essentially be a `forceRefresh` option, but as a function instead of a boolean, and probably more as a config option than a value passed into the initiate thunk.

markerikson:
I'm... still really confused I'm afraid :(

I sorta get the idea of a different _lifecycle_ that runs at an earlier point, or when `onQueryStarted` doesn't.

But I'm not sure I understand what I'm seeing here.

Some actual usage examples would help, as would tests.

markerikson:
Looking at this again and I'm _still_ pretty confused.

Based on the implementation, this _sounds_ like it ought to be named something along the lines of `shouldForceRefetch`, but I still need to see some real examples/tests/docs to understand when and how you want to use this.

schadenn:
Sorry there's a lot going on right now. I refactored it to be closer to what @phryneas was suggesting, which absolutely makes sense for me. I'm trying to set up a repro/have testing in place soon.

schadenn:
Done with the repro. This is a little imaginary since I can't do real mutations on the Pokemon API. But you'll be able to see when you update the name in the "Manage" section, that - providing the forceRefetch - the query will happen again (this would obviously get new data in a real world scenario then).

https://codesandbox.io/s/examples-query-react-basic-forked-kl7bmf

markerikson:
Updated this to target `v1.9-integration`, changed the signature to be `{currentArg, previousArg, state, endpointState}`, and added a test.  This seems to make sense as part of an "infinite scrolling / pagination" use case.

PR #2696, opened 2022-09-14
docs: Clarify documentation about the overrideExisting flag in injectEndpoints

### LINKED ISSUE #2694: RTK Query: Possibly misleading documentation for `overrideExisting` option in endpoint injection
The [docs for code splitting and endpoint injection](https://redux-toolkit.js.org/rtk-query/api/created-api/code-splitting#description) say:

> Any endpoint definitions defined using that builder will be merged into the existing endpoint definitions for this API slice using a shallow merge, so any new endpoint definitions will override existing endpoints with the same name.

And a few lines after:

> The `overrideExisting` flag controls a development-only warning that notifies you if there is a name clash between endpoint definitions. When set to `true`, the warning will not be printed.

This might lead someone (such as me :p) to believe that endpoints get overridden no matter what and `overrideExisting` only controls if the warning is shown or not.

Looking at the code, it seems that the intended behavior is that `overrideExisting` actually controls whether existing endpoints are, in fact, overridden or not:

https://github.com/reduxjs/redux-toolkit/blob/master/packages/toolkit/src/query/createApi.ts#L307-L321

So, it looks like that section of the docs could be updated to clarify this. If you agree, i can open a PR for that.

PauloRSF:
# Proposed changes

Clarify in the `injectEndpoints` docs that the `overrideExisting` flag actually controls whether endpoints are overridden or not in development.

Closes #2694

### DIFF
--- a/docs/rtk-query/api/created-api/code-splitting.mdx
+++ b/docs/rtk-query/api/created-api/code-splitting.mdx
@@ -35,7 +35,7 @@ Accepts an options object containing the same `endpoints` builder callback you w
 
 Returns an updated and enhanced version of the API slice object, containing the combined endpoint definitions.
 
-The `overrideExisting` flag controls a development-only warning that notifies you if there is a name clash between endpoint definitions. When set to `true`, the warning will not be printed.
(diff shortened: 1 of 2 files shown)


markerikson:
Thanks!

Issue #2704, opened 2022-09-19
Typo issue in RTK Query docs


barzin144:
In rtk-query/overview page inside "Use Hooks in Components" section, the word "isFetching" has been used instead of "isLoading"

markerikson:
Appreciate the issue and PR, but that's actually intentional.

Per https://redux-toolkit.js.org/rtk-query/usage/queries#frequently-used-query-hook-return-values , `isLoading` is only true the _first_ time the query is running. `isFetching` is true _every_ time, so that's what you should use most of the time.

PR #2705, opened 2022-09-19
replace isFetching with isLoading

barzin144:
fix issue #2704


markerikson:
Per issue comment, that's actually intentional, but thanks for trying to help update things!

PR #2722, opened 2022-09-23
Fixed typo

sidsarasvati:


markerikson:
Hi! Appreciate the PR, but the second arg to a `createAsyncThunk` payload function really _is_ the `thunkApi` object.

In that particular example, we're pretending that there's some kind of `userApi` abstraction layer coming from elsewhere in the app, just to show "hey, make an async call here" without worrying about the details.

Given that, I'll close those, but thanks for trying to improve things!

PR #2740, opened 2022-10-01
Export the BaseQueryApi interface from rtk-query so it can be used as a type in TypeScript without importing from dist/.

nsthorat:
This PR adds the `BaseQueryApi` as an export to the public `src/query/index.ts` API from `baseQueryTypes`.


phryneas:
Good point - let's get this in :)

PR #2743, opened 2022-10-04
Fix retryCondition `error`.

phryneas:
Previously, the full baseQueryResult was passed into the `error` argument. This should fix it to only pass `error` in.


phryneas:
This works nicely, so let's get it in.

phryneas:
@kahirokunn as it seems this was not caught by any tests. If you have some spare time, I'd be very grateful if you could add a PR with a test for this.

kahirokunn:
okay

PR #2759, opened 2022-10-08
Speed up subscription behavior by tracking state in middleware

markerikson:
This PR:

- Rewrites our internal RTKQ subscription handling so that the primary state is kept internal to the RTKQ middleware:
  - The subscriptions data is now kept as a mutable object, initialized in the outer middleware and passed through to each "handler"
  - The `batchActions` file (which probably ought to be renamed) mutates that state directly for speed.  (Using an Immer-powered reducer added 2-3ms of noticeable overhead for every action.)  It then sets a `queueMicrotask`, and at the end of the tick, clones the subscriptions state, diffs it vs the previous state, and syncs it into the Redux store as Immer patches.
  - That `batchActions` handler also checks to see which actions should _not_ be forwarded to the reducers, and returns a flag indicating whether the action should continue or be halted.
  - Because the store state is no longer synchronously up to date with the latest subscription data immediately, the query hooks now dispatch an `internal_probeSubscription` action _while rendering_ instead of trying to read the state.  This is intercepted by the middleware, so it doesn't cause any renders or visible side effects.  The middleware then returns an `subscriptionExists` flag alongside the "should action continue" flag, so that the hook knows whether it needs to add a new subscription entry.  (Frankly this is disgustingly hacky, but it's _technically_ legal. or at least we can get away with it :)  )
  - The former `subscriptionSlice`, which was added as `state.api.subscriptions`, is now a plain function that literally-no-kidding _mutates_ the state it sees.  There's still a `subscriptionsSlice` variable, but I'm using it just for ease of generating the action  creators I need.  There's also `internalSubscriptionsSlice`, which is a real slice reducer, but its only action is receiving the Immer patches and updating the store copy of the subscriptions state.
- Finally adds that `delay` util I keep copy-pasting over into a utils files and updates files to import that instead


packages/toolkit/src/query/core/buildMiddleware/polling.ts
```diff
@@ -125,13 +132,15 @@ export const buildPollingHandler: InternalHandlerBuilder = ({
 
   function findLowestPollingInterval(subscribers: Subscribers = {}) {
     let lowestPollingInterval = Number.POSITIVE_INFINITY
-    for (const subscription of Object.values(subscribers)) {
-      if (!!subscription.pollingInterval)
+    for (let key in subscribers) {
```
markerikson, line 135: :speech_balloon: Saw that `findLowestPollingInterval` was showing up at the tail end of some of the perf capture flame charts. I don't expect this makes any real difference, but I don't think it _hurts_.

packages/toolkit/src/utils.ts
```diff
@@ -23,6 +23,10 @@ It is disabled in production builds, so you don't need to worry about that.`)
   }
 }
 
+export function delay(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms))
+}
+
```
phryneas, line 29: *cough*

https://github.com/reduxjs/redux-toolkit/blob/988c0499125c07e203794e6978e4796447c56579/packages/toolkit/src/query/tests/helpers.tsx#L39-L41

Maybe we should just rename that one? 😅

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -701,12 +701,21 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
 
       let { queryCacheKey, requestId } = promiseRef.current || {}
-      const currentRenderHasSubscription = useSelector(
-        (state: RootState<Definitions, string, string>) =>
-          !!queryCacheKey &&
-          !!requestId &&
-          !state[api.reducerPath].subscriptions[queryCacheKey]?.[requestId]
-      )
+
+      // HACK Because the latest state is in the middleware, we actually
```
phryneas, line 705: Because of this, it is not really problematic if the user forgets to add the middleware to the store.
Before, there was just no cleanup happening. Now, we get unexpected behaviour. Maybe we should change that "you forgot to add the middleware" warning to throwing an error at this points?

phryneas:
> The batchActions file (which probably ought to be renamed) mutates that state directly for speed. (Using an Immer-powered reducer added 2-3ms of noticeable overhead for every action.) It then sets a queueMicrotask, and at the end of the tick, clones the subscriptions state, diffs it vs the previous state, and syncs it into the Redux store as Immer patches.

I'm actually not seeing this - you are still using the immerized `subscriptionReducer` here. Is that intended?

phryneas:
Ugh I think I reviewed much of this on an outdated state 🤦

packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -380,75 +380,96 @@ export function buildSlice({
...
+      }
+      return true
+    } else if (unsubscribeQueryResult.match(action)) {
+      const { queryCacheKey, requestId } = action.payload
+      if (draft[queryCacheKey]) {
+        delete draft[queryCacheKey]![requestId]
+      }
+      return true
+    } else if (querySlice.actions.removeQueryResult.match(action)) {
+      delete draft[action.payload.queryCacheKey]
```
phryneas, line 438: return true

packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -380,75 +380,96 @@ export function buildSlice({
...
+    action: AnyAction
+  ) => {
+    if (updateSubscriptionOptions.match(action)) {
+      const { queryCacheKey, requestId, options } = action.payload
+
+      if (draft?.[queryCacheKey]?.[requestId]) {
+        draft[queryCacheKey]![requestId] = options
+      }
+      return true
+    } else if (unsubscribeQueryResult.match(action)) {
```
phryneas, line 431: with all the returns, we can remove the else statements here and save a few bytes

packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -505,10 +526,12 @@ export function buildSlice({
     ...configSlice.actions,
     ...querySlice.actions,
     ...subscriptionSlice.actions,
+    ...internalSubscriptionsSlice.actions,
     ...mutationSlice.actions,
     /** @deprecated has been renamed to `removeMutationResult` */
     unsubscribeMutationResult: mutationSlice.actions.removeMutationResult,
     resetApiState,
+    actuallyMutateSubscriptions,
```
phryneas, line 534: I'm not sure if we should have that on `actions` here

packages/toolkit/src/query/tests/buildHooks.test.tsx
```diff
@@ -2353,14 +2366,25 @@ describe('skip behaviour', () => {
...
     expect(subscriptionCount('getUser(1)')).toBe(1)
 
-    rerender([1, { skip: true }])
-    expect(result.current).toEqual(uninitialized)
+    await act(async () => {
+      rerender([1, { skip: true }])
+    })
+    expect(result.current).toEqual({
+      ...uninitialized,
+      currentData: undefined,
```
phryneas, line 2384: This is unexpected tbh. I'm pretty sure the hook should reset on changing to `skip`, as per the same logic that happens on a store reset. Something here is broken.
See
https://github.com/reduxjs/redux-toolkit/blob/dbcc556a2cd1ad6fd9776bc62131c9c43c72c4c0/packages/toolkit/src/query/react/buildHooks.ts#L608-L627

packages/toolkit/src/query/tests/buildHooks.test.tsx
```diff
@@ -2375,17 +2399,28 @@ describe('skip behaviour', () => {
...
+    expect(result.current).toMatchObject({ status: QueryStatus.fulfilled })
+    await delay(1)
     expect(subscriptionCount('getUser(1)')).toBe(1)
     expect(storeRef.store.getState().api.subscriptions).not.toEqual({})
 
-    rerender([skipToken])
-    expect(result.current).toEqual(uninitialized)
+    await act(async () => {
+      rerender([skipToken])
+    })
```
phryneas, line 2417: same

packages/toolkit/src/query/tests/buildSlice.test.ts
```diff
@@ -77,13 +75,26 @@ describe('buildSlice', () => {
...
+    await delay(1)
+
+    expect(storeRef.store.getState()).toEqual({
+      ...initialQueryState,
+      api: {
+        ...initialQueryState.api,
+        subscriptions: {
+          'getUser(1)': expect.any(Object),
+        },
+      },
```
phryneas, line 97: why did you remove that part?

packages/toolkit/src/query/tests/cleanup.test.tsx
```diff
@@ -177,24 +199,62 @@ test('Minimizes the number of subscription dispatches when multiple components a
...
+    'api/config/middlewareRegistered',
+    'api/executeQuery/pending',
+    'api/internalSubscriptions/subscriptionsUpdated',
+    'api/executeQuery/fulfilled',
+  ])
+
+  // Could be flaky in CI, but we'll see.
+  // Currently seeing 800ms in local dev, 6300 without the batching fixes
+  // console.log('Elapsed subscription time: ', timeElapsed)
+  expect(timeElapsed).toBeLessThan(1500)
```
phryneas, line 259: this file needs a good cleanup

phryneas (commented):
Some tidbits here and there, but already looking very good :)

Generally we probably have to do a big cleanup on the 1.9 branch before we release it - there is quite a lot of commented-out and debugging code in there now.

packages/toolkit/src/query/core/buildMiddleware/batchActions.ts
```diff
@@ -14,39 +16,75 @@ const queueMicrotaskShim =
...
+            JSON.stringify(internalState.currentSubscriptions)
+          )
+          // Figure out a smaller diff between original and current
+          const [, patches] = produceWithPatches(
+            previousSubscriptions,
+            () => newSubscriptions
+          )
+
+          // Sync the store state for visibility
+          mwApi.next(api.internalActions.subscriptionsUpdated(patches))
```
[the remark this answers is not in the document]
markerikson: My thought was that there's no point in doing so - this is _solely_ intended to sync the state update, no other middleware would care about this.

packages/toolkit/src/query/core/buildMiddleware/batchActions.ts
```diff
@@ -14,39 +16,75 @@ const queueMicrotaskShim =
...
       }
+
+      const isSubscriptionSliceAction =
+        !!action.type?.startsWith(subscriptionsPrefix)
+      const isAdditionalSubscriptionAction =
+        queryThunk.rejected.match(action) &&
+        action.meta.condition &&
+        !!action.meta.arg.subscribe
+
+      const actionShouldContinue =
```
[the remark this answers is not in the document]
markerikson: I still debate the "breakingness" of it. Sure, those actions were publicly visible, but I still see that as an implementation detail, not something we expect users to rely on.  (Which of course means that someone somewhere probably _is_ relying on them...)

My fear is that as soon as we start letting all those actions through we lose any perf improvement we've gotten out of this effort, because we're right back to running reducers and notifying subscribers each time.

packages/toolkit/src/query/core/buildMiddleware/batchActions.ts
```diff
@@ -14,39 +16,75 @@ const queueMicrotaskShim =
...
-  api,
-  queryThunk,
-}) => {
-  let abortedQueryActionsQueue: RejectedAction<QueryThunk, any>[] = []
+export const buildBatchedActionsHandler: InternalHandlerBuilder<
+  [actionShouldContinue: boolean, subscriptionExists: boolean]
+> = ({ api, queryThunk }) => {
+  const { actuallyMutateSubscriptions } = api.internalActions
+  const subscriptionsPrefix = `${api.reducerPath}/subscriptions`
+
```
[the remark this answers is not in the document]
markerikson: That could potentially work, yeah.

packages/toolkit/src/query/core/buildSlice.ts
markerikson: The other option would be to just move it straight over into the file where it's used. This was at least partly laziness :)

packages/toolkit/src/query/tests/buildSlice.test.ts
markerikson: Ah... good question. I think it was that the test seemed more focused on "was there _any_ state at all before, and did it all get cleared out by the reset action", rather than the specifics of exactly _which_ state fields existed, and I was getting tired of adding `delay()` calls to make things line up.

packages/toolkit/src/query/core/buildMiddleware/batchActions.ts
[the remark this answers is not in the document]
phryneas: If the actions don't actually change any state we won't notify subscribers though, right?

packages/toolkit/src/query/core/buildSlice.ts
phryneas: ^^

packages/toolkit/src/query/core/buildMiddleware/batchActions.ts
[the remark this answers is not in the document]
phryneas: Yeah, it's more a "correctness" thing at this point - usually I wouldn't recommend to use `next` for anything except the original (or replaced) action, so we might set a bad example doing it.

packages/toolkit/src/query/core/buildMiddleware/batchActions.ts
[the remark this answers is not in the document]
markerikson: Subscribers are _always_ notified, after _every_ action.

They normally do have an early bailout, like `if (newState !== prevState) return`, but that's still a bunch of unnecessary calls.  Similar for the reducers - they may no-op out, but they still had to get called.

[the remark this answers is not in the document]
phryneas: Ah, I somehow had convinced myself that there was a bailout before calling the subscribers (since the subscribers have no access to the action, calling them without a state change seemed unreasonable). In that case, yeah, that's a point for skipping it. I'd still prefer keeping them until 2.0, but I can also live with them going away :)

packages/toolkit/src/query/react/buildHooks.ts
markerikson: Addressed in #2792

packages/toolkit/src/query/tests/cleanup.test.tsx
markerikson: Done in #2789

PR #2767, opened 2022-10-11
docs(rtk-query): update axios base query example

michal-wrzosek:
It seems that actually, the params that are being sent to `axiosBaseQuery ` do not have a `data` but rather a `body`. This is at least happening in the project I'm working on which is using `@rtk-query/codegen-openapi`. Such change in `axiosBaseQuery` fixed problems I faced. Previous to that axios would send requests without any payload.

Anyone using axios can confirm that?

The example generated code looks like this for me:
```typescript
examplePostRequest: build.mutation<ExamplePostRequestApiResponse, ExamplePostRequestApiArg>({
  query: (queryArg) => ({
    url: `/example-post-request`,
    method: 'POST',
    body: queryArg.examplePostRequest,
  }),
}),
```

So when the `axiosBaseQuery` function is being used it receives `body` not `data`. Am I missing something here?

### DIFF
--- a/docs/rtk-query/usage/customizing-queries.mdx
+++ b/docs/rtk-query/usage/customizing-queries.mdx
@@ -252,13 +252,13 @@ const axiosBaseQuery =
     {
       url: string
       method: AxiosRequestConfig['method']
-      data?: AxiosRequestConfig['data']
+      body?: AxiosRequestConfig['data']
       params?: AxiosRequestConfig['params']
     },
     unknown,
     unknown
   > =>
-  async ({ url, method, data, params }) => {
+  async ({ url, method, body: data, params }) => {
     try {
       const result = await axios({ url: baseUrl + url, method, data, params })
       return { data: result.data }


raymondboswel:
I can confirm the issue. If you change the object passed to the `query` property to use data, instead of body it works. So basically documentation doesn't match up with the axiosBaseQuery implementation.  e.g
```
query: body => ({
        url: "user",
        method: 'post',
        data: body
      }),

```
I'm new to RTK query, so this took me quite a while to figure out..

phryneas:
You have a general misconception: 
In RTK, whatever you return from the `query` of an endpoint correlates to your `baseQuery`.

So, for `fetchBaseQuery`, your `query` function might return an object that has `body`. If you use `axiosBaseQuery`, your `query` function should return something with `data`. If you use `graphqlBaseQuery`, your `query` function should return an object with `query` and `variables`.  
So if you use `axiosBaseQuery`, you should `return { data: ... }` in your `query`, not `{ body: ... }`.    
We just showcase the `fetchBaseQuery` options in most docs examples because that is the `baseQuery` that we recommend for normal usage.  
But you could implement your own `baseQuery` that would accept `{ myCarColor: "red" }` - it's totally up to you. That's a "swappable" part.

Honestly, to this day I do not understand why anyone would want to ship an extra 5kb just to use `axios`, which has pretty much the same feature set as `fetch`, which is built into your browser for free. We have that example in the docs because people requested it, but there is really no good reason to use `axios` at all.

michal-wrzosek:
Thanks @phryneas. So you say the documentation for how to use axios is correct? If it's not correct I believe we should update the docs.

And about the use of axios, the reality is that there are some huge projects out there already using axios and it's just nice if you can easily integrate rtk-query with it - it was really helpful to use your docs to implement this library, I was just a bit confused if this example in docs ever worked with this `body`/`data` mismatch.

phryneas:
The documentation about axios is correct.

The point is just that if you add a custom `baseQuery` (like the axios one), you use RTK Query itself differently.

It's a bit hard to explain in text any more than I already did - but I have a video course that goes over this mechanism quite thoroughly - I'd recommend you spare the 20 minutes and watch [this free egghead course on RTK Query Basics: Query Endpoints, Data Flow and TypeScript](https://bit.ly/3YabCCG)

michal-wrzosek:
Ok, if I understand correctly in order to use the example from the docs "as is", I would need to change my queries, so that they pass payload as "data" not as a "body" property.

For me, the problem was, that I used `@rtk-query/codegen-openapi` to generate my queries (like the one in my first message) and the shape of the return object of those queries is not something I can configure (afaik it will always pass arguments as "body"). That is why there's this mismatch. But maybe I faced a really uncommon scenario and for most people this "Axios baseQuery" example from docs will work.

phryneas:
> Ok, if I understand correctly in order to use the example from the docs "as is", I would need to change my queries, so that they pass payload as "data" not as a "body" property.

To say it short: if you use `axiosBaseQuery`, you have to use RTK Query as if you were using `axios`.  
So if you would call `axios(someArg)`, write `query() { return someArg }`.  
That's also the only feasible reason why you would want to use `axiosBaseQuery`: you already are used to write axios options and want to keep using this style.  
Of course the docs don't reflect that - we assume that you are using `fetchBaseQuery`, since that is what we are shipping. `axiosBaseQuery` is only an example on how you could implement an alternative `baseQuery`.

> For me, the problem was, that I used `@rtk-query/codegen-openapi` to generate my queries (like the one in my first message) and the shape of the return object of those queries is not something I can configure (afaik it will always pass arguments as "body"). That is why there's this mismatch. 

Yeah, the codegen assumes that you use `fetchBaseQuery`.



Seeing that you are generating new code, and in fact not using any legacy axios code that you want to copy & paste over to RTK Query, could you please again try to explain why you want to use axios here? I really think that you are just wasting bundle size here - it doesn't seem that you are in any situation that would give you any benefits whatsoever by using axios (seeing that IE is phased out by Microsoft nowadays).

doktordirk:
>  could you please again try to explain why you want to use axios here

i'd like to say that myself. unfortunatelly `fetchBaseQuery`  makes some assumptions and doesn't offer the possibility to change those
 one example is the the `encodeURI()`, i presume, which `fetchBaseQuery` uses for search parameters (so `ids=1,2,4` becomes `ids=1%2C2%2C4`).  `axios` by default keeps it at  `ids=1,2,4`, but also in contrast to `fetchBaseQuery`, `axios` has 
```
  // `paramsSerializer` is an optional function in charge of serializing `params`
  // (e.g. https://www.npmjs.com/package/qs, http://api.jquery.com/jquery.param/)
  paramsSerializer: function (params) {
    return Qs.stringify(params, {arrayFormat: 'brackets'})
  },
```
to change that if if needed. 

though while it'll be better if `fetchBaseQuery` would have such option, atm one has to use an own `baseQuery` and than it's simplest to use `axios` to get all the configuration options one might need out of the box

phryneas:
`fetchBaseQuery` *literally* has a `paramsSerializer` option, and it you ever need additional functionality, you are not between the choice of your own `baseQuery` and axios, but between wrapping a function around `fetchBaseQuery` and axios - which is a *lot* less code to ship to your clients.

michal-wrzosek:
Thank you all for responding to my PR. Apparently, I was wrong and my suggestion here doesn't seem to make sense. Don't wanna create any further confusion so I'm closing it.

doktordirk:
> `fetchBaseQuery` _literally_ has a `paramsSerializer` option, and if you ever need additional functionality, you are not between the choice of your own `baseQuery` and axios, but between wrapping a function around `fetchBaseQuery` and axios - which is a _lot_ less code to ship to your clients.

ahh. perfect :heart: . missed that sowehow! well, that's gonna make my life easier

Issue #2411, opened 2022-06-14
Add a `skipToken` that resets the query hook if it's been initialized before (e.g. `skipOrResetToken`)


kasparkallas:
I'd like to see an easy way to reset the query to uninitialized without having to re-mount the component.

Example usage:
```ts
const accountData = useAccountDataQuery(accountId ? { id: accountId } : skipOrResetToken);
```
Passing the `skipOrResetToken` guarantees that there's no data (i.e. stale data) coming from the query. 

The use-case arose from the fact that the stale data with just the `skipToken` can be surprising when you have a user inputted conditional for whether to execute the query or not (e.g. "address", "item name", "username" etc).

markerikson:
Seems plausible.  I also don't think we have a way to reset a single cache entry imperatively atm - you can reset the _entire_ API state, but not just one entry.

phryneas:
I think the point here is more to reset the *hook* rather than the cache entry, right?

kasparkallas:
> I think the point here is more to reset the _hook_ rather than the cache entry, right?

Yes, it's about resetting the hook -- I'll edit the title.

Here's a more detailed example of the type of code I _currently_ end up writing:
```ts
const { connectedAccount } = useAccountContext();
const accountSettingsQuery = useAccountSettingsQuery(connectedAccount ? { account: connectedAccount } : skipToken);
const accountSettings = connectedAccount ? accountSettingQuery?.data : undefined;

// and then render something based on the preferences
```

1. Most of the time the component is landed on without a connected account so the query result is uninitialized as expected with `undefined` data. 
2. Then the account is connected and the query has the account related data.
3. And then the account is disconnected and I would like to see `undefined` data again but in actuality I get the last result.

There's an argument to be made that I should just re-mount the component for such a context change (fair). Or that the extra variable solution is simple (which it is) but the main issue with it is that it's _easy to miss_. I mistakenly tend to take for granted that when `skipToken` is passed then I will surely not have any results.

kasparkallas:
> @kasparkallas hmm. Would it help if you checked `accountSettingsQuery.currentData` instead? I _think_ that should be "the value from the absolute latest argument", whereas `data` is "the last _successful_ value".

Looks like that is indeed the case... :stuck_out_tongue: Thank you very much, sir!

PR #2779, opened 2022-10-15
fix `skipToken`  behaviour in `useQueryState`

phryneas:
This reverts a test change that I identified as problematic over in https://github.com/reduxjs/redux-toolkit/pull/2759/files/6975282e87ce98f9a548043c15a82aacf3c384fc#r990775186 and fixes the underlying bug.

When calling `useQueryState(skipToken)`, the result should always be a pristine `uninitialized` result and not contain previous data.

My theory: that only worked before because of a double-render, so during the first render the result changed a bit and during the second render it completely reset.
#2759 removed that additional render and so the behaviour broke.

This PR reintroduces that expected behaviour (and fixes some minor nitpicks I had over in #2759 - there is still stuff open over there though).

### DIFF
--- a/packages/toolkit/src/query/core/buildSlice.ts
+++ b/packages/toolkit/src/query/core/buildSlice.ts
@@ -428,15 +428,19 @@ export function buildSlice({
         draft[queryCacheKey]![requestId] = options
       }
       return true
-    } else if (unsubscribeQueryResult.match(action)) {
+    }
+    if (unsubscribeQueryResult.match(action)) {
       const { queryCacheKey, requestId } = action.payload
       if (draft[queryCacheKey]) {
         delete draft[queryCacheKey]![requestId]
       }
       return true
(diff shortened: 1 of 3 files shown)


kasparkallas:
Just wondering, is it now safe to assume that `skipToken` will always return `undefined` with `.data`? Because until now, `.currentData` had to be used because `.data` might have contained a result from a previous render without a `skipToken`. Related to this: https://github.com/reduxjs/redux-toolkit/issues/2411

phryneas:
It does. It is also the behaviour our tests always tested for, so I'm a bit confused that it comes up now. It seems that there was some inconsistent behaviour our tests did not catch correctly 🤔 

We might need to find a solution that allows for deliberately changing between both behaviours.

kasparkallas:
:+1: That always seemed like the more intuitive behaviour to me. I started using `.currentData` for these case but there's been a creeping feeling I'd miss it or a team member would.

PR #2781, opened 2022-10-16
experimental `use` support for suspense

phryneas:
This adds experimental `use` support. (For more context, see https://github.com/reactjs/rfcs/pull/229)

If this would land like this, it could eliminate all the "when do we start the request" problems we are having and supersede #2245.

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -728,23 +728,13 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
         promiseRef.current = undefined
       }, [subscriptionRemoved])
 
-      usePossiblyImmediateEffect((): void | undefined => {
+      const initiateQueryIfNeeded = useCallback(() => {
         const lastPromise = promiseRef.current
-        if (
-          typeof process !== 'undefined' &&
-          process.env.NODE_ENV === 'removeMeOnCompilation'
-        ) {
-          // this is only present to enforce the rule of hooks to keep `isSubscribed` in the dependency array
-          console.log(subscriptionRemoved)
-        }
-
         if (stableArg === skipToken) {
           lastPromise?.unsubscribe()
           promiseRef.current = undefined
           return
         }
-
-        const lastSubscriptionOptions = promiseRef.current?.subscriptionOptions
 
         if (!lastPromise || lastPromise.arg !== stableArg) {
           lastPromise?.unsubscribe()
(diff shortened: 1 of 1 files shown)


packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -775,21 +788,34 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+            )
+          return promiseRef.current?.refetch()
+        }, []),
+        /**
+         * a `.then` method that allows the return value to be used with the `use` hook like
+         * `const result = use(useSomeQuery(args))`
+         */
+        then: useCallback<Promise<unknown>['then']>(
+          (onfulfilled, onrejected) => {
+            initiateQueryIfNeeded()
```
phryneas, line 808: The question is still open if React will throw away the current component state & refs when suspending before the component was rendered at least once.

If it would, this would not work and we would have to try something with `getRunningQueryThunk` here instead. But let's hope that we won't have that problem in the future.

phryneas:
There are still open questions like "how would that look like with in-component query waterfalls?" and we would probably have to provide some kind of "memoizing Promise.all" for parallel suspending queries.

```js
const { data1, data2, data3 } = use(memoPromiseAll({
  data1: useSomeData(arg1),
  data2: useSomeMoreData(arg2),
  data3: useEvenModeData(arg3)
}))
```

But maybe even `use` could come with something like that in the long run, so it might become
```js
const { data1, data2, data3 } = use({
  data1: useSomeData(arg1),
  data2: useSomeMoreData(arg2),
  data3: useEvenModeData(arg3)
})
```
or a similar `useAll` hook:
```js
const { data1, data2, data3 } = useAll({
  data1: useSomeData(arg1),
  data2: useSomeMoreData(arg2),
  data3: useEvenModeData(arg3)
})
```

phryneas:
What this also does *not* take into account right now is "what if that cache value refreshes". `use` will always reference that one promise - so maybe for that we would either have to trick around in the `useQuery` hook itself and wiggle some data from `useQueryState` in - or we would have to fall back to a separate implementation.

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -775,21 +788,34 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
-          /**
-           * A method to manually refetch data for the query
-           */
-          refetch: () => {
-            if (!promiseRef.current)
-              throw new Error(
-                'Cannot refetch a query that has not been started yet.'
-              )
-            return promiseRef.current?.refetch()
+      return {
```
dai-shi, line 791: I don't understand the entire code, but `use` will attach `.value` prop, so you want to keep `useMemo` maybe?

phryneas: Hmm, good point, that might be very tricky.

The return value of this hook will be spread in yet another hook, so having a stable value here will not help a lot - and that other value can be memoized, but will change more often since it has other "changing values" in it.

We would probably have to manage `.value` on our own in that context 🤔

PR #2792, opened 2022-10-19
Convert "middleware not registered" warning into a full error

markerikson:


packages/toolkit/src/query/core/buildInitiate.ts
```diff
@@ -300,9 +300,9 @@ export function buildInitiate({
         ;(middlewareWarning as any).triggered = true
```
phryneas, line 300: This does not need any tracking to trigger only once any more, it should throw every time.
It should also do so in production I guess?

phryneas: Regarding this code:
```js
const registered =
        getState()[api.reducerPath]?.config?.middlewareRegistered
```
We can also do the check for `registered` better now by really dispatching an `internal_probeSubscription` action and seeing if the middleware picks it up. That `middlewareRegistered` field could also come from persistence being overactive.

PR #2804, opened 2022-10-21
Fill out v1.9 remaining docs

### LINKED ISSUE #2619: Document any additional 1.9 behavior
Don't think we have docs for `merge` or some of the other options.  Need to add those.

markerikson:
This PR:

- Updates Yarn to 3.2.4, as 3.1.0 didn't want to install an updated Docusaurus properly
- Updates Docusaurus to 2.1.0
- Exports `defaultSerializeQueryArgs` for now
- Documents the new RTKQ endpoint fields
- Documents the new `fetchBaseQuery` args and reworks that page in general
- Adds a page for "Codemods"

Fixes #2619 , or will once I'm done with it.

I'm most likely going to move the `defaultSerializeQueryArgs` export to its own branch, and also see about changing the `serializeQueryArgs` endpoint option to let you just return a non-string value if you want, but put it here for now because I needed to show an example of how `serializeQueryArgs` works _now_ and realized we don't export that.


docs/rtk-query/api/created-api/api-slice-utils.mdx
```diff
@@ -105,6 +110,43 @@ dispatch(api.endpoints.getPostById.initiate(1))
...
+  - `args`: an argument matching that used for a previous query call, used to determine which cached dataset needs to be updated
+  - `newEntryValue`: the value to be written into the corresponding cache entry's `data` field
+
+#### Description
+
+A Redux thunk action creator that, when dispatched, acts as an artificial API request to upsert a value into the cache.
+
+The thunk action creator accepts three arguments: the name of the endpoint we are updating (such as `'getPost'`), the appropriate query arg values to construct the desired cache key, and the data to upsert.
+
+If no cache entry for that cache key exists, a cache entry will be created and the data added. If a cache entry already exists, this will _overwrite_ the existing cache entry data.
```
ggaabe, line 136: If the value being upserted is `undefined`, will it remove the cache key entirely? If not, is there a way to remove a cache entry entirely? (ie, like `invalidateTags`, but without triggering a refetch)

markerikson: @ggaabe : no, it would replace the `.data` field with `undefined`.   It's the value, not the key.

There isn't currently a way to remove one, but a few folks have asked for something like this.  Could you comment over in #2820 and describe when and why you might want to do that?

markerikson:
Docs should be updated, pending me figuring out what to do with the `defaultSerializeQueryArgs` thing

markerikson:
Odd, no PR preview.

We'll, let's get it in.

PR #2813, opened 2022-10-24
[OpenAPI codegen] Use schema.title for body params

GeorchW:
Instead of falling back to `body`, request parameters now consider the schema title as an additional source of information.

Since it was not ideal to change the type or reference name to determine the name of the body parameter in my setup, I was lacking an ability to determine its name from the OpenAPI specification. Using the `title` field of the schema for this purpose seemed pretty reasonable to me. 

FastAPI in particular conveniently allows setting the title attribute of the embedded schema, which was what I was using.

While this is technically a breaking change, I believe it's very unlikely to actually break anyone's code, since setting an embedded type with a title different to the type name or ref name seems uncommon to me.

### DIFF
--- a/packages/rtk-query-codegen-openapi/src/generate.ts
+++ b/packages/rtk-query-codegen-openapi/src/generate.ts
@@ -168,13 +168,13 @@ export async function generateApi(
         ...apiGen.enumAliases,
         ...(hooks
           ? [
-              generateReactHooks({
-                exportName: generatedApiName,
-                operationDefinitions,
-                endpointOverrides,
-                config: hooks,
-              }),
-            ]
+            generateReactHooks({
+              exportName: generatedApiName,
+              operationDefinitions,
+              endpointOverrides,
+              config: hooks,
+            }),
+          ]
           : []),
       ],
       factory.createToken(ts.SyntaxKind.EndOfFileToken),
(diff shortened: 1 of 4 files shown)


phryneas:
I think it might be reasonable, but tbh. I've never worked with `title` - so I'll have to give this a thought.

Either way: could you please add a test showcasing this behaviour?

GeorchW:
@phryneas Here you go :-) I'm open for other approaches as well. I think using `title` here is in line with the JSON Schema specification:

> [9.1. ](http://json-schema.org/draft/2020-12/json-schema-validation.html#section-9.1)["title" and "description"](http://json-schema.org/draft/2020-12/json-schema-validation.html#name-title-and-description)
>
> [...] Both of these keywords can be used to decorate a user interface with information about the data produced by this user interface. A title will preferably be short, whereas a description will provide explanation about the purpose of the instance described by this schema.

GeorchW:
@phryneas wanna have another look?

GeorchW:
@phryneas ping?

msutkowski (approved):
LGTM. Can't see any downside - also ran this against multiple larger projects with no negative impact.

GeorchW:
Let's merge this then! It's been almost a year. @phryneas @markerikson @EskiMojo14 ?

phryneas:
@GeorchW it has, I'm sorry! 
Could you please resolve conflicts? I'd be happy to get this out as 1.1.2 in the next few days :)

GeorchW:
Rebased.

The oazapfts update actually required to change `schema.title` to `"title" in schema && schema.title` to make TypeScript happy, since e.g. arrays never have a `title`. Doesn't change any of the behavior, though.

phryneas:
Was lange währt, wird endlich gut, oder so. 
Danke :)

Issue #2385, opened 2022-06-04
Allow setting the default responseHandler at a higher level


taylorkline:
Because of #2354 (which is addressed in #2363, pending feedback), we have resorted to adding a custom `responseHandler` to each and every endpoint definition.

This gets tedious, and it's easy for one of the several developers on our team to miss a case until we run into a run-time `PARSING_ERROR` when we get a `"text/plain"` response.

It would be very convenient to be able to set the default `responseHandler` across all endpoints in the options to `createApi`.

phryneas:
A PR adding `responseHandler` to the `FetchBaseQueryArgs` argument to `fetchBaseQuery` would be welcome.

If you are not using TypeScript, it is probably even already possible to pass it into `fetchBaseQuery` directly at creation.

markerikson:
Slightly confused re-reading this, but I _think_ this is fixed by #2363 ?

markerikson:
I take it back. Re-re-reading our own types: I think you can pass `responseHandler` as part of an individual `query` object, but not to `fetchBaseQuery`?

taylorkline:
Short version: it would be nice to be able to specify `responseHandler` in `createApi`:

```ts
export const pokemonApi = createApi({
  reducerPath: 'pokemonApi',
  baseQuery: fetchBaseQuery({ baseUrl: 'https://pokeapi.co/api/v2/' }),
  responseHandler: 'content-type',
  endpoints: (builder) => ({
    getPokemonByName: builder.query<Pokemon, string>({
      query: (name) => `pokemon/${name}`,
    }),
  }),
})
```

Or similar.

phryneas:
I added a PR for that today - #2823

markerikson:
Merged into the 1.9 branch

PR #2823, opened 2022-10-26
allow for global `responseHandler` and `validateStatus` configuration

phryneas:
see #2385 
still needs tests


markerikson:
hah, nice! (I take it we're passing these through already by default, we just didn't say so in the types.)

phryneas:
At least I think so. I should probably still add tests, that's why this is a draft :)

taylorkline:
I can confirm, as per #3136, (and perhaps fixed by #3137), this is not working in `1.9.3`.

PR #2841, opened 2022-10-30
feat(docs): document react-native and expo starter templates

rahsheen:
Changes the Getting Started to have tabs for CRA, React Native CLI, and Expo template usage and splits the `bash` code blocks for easier copy-pasting.

### DIFF
--- a/docs/introduction/getting-started.md
+++ b/docs/introduction/getting-started.md
@@ -7,6 +7,8 @@ hide_title: true
 
 import LiteYouTubeEmbed from 'react-lite-youtube-embed';
 import 'react-lite-youtube-embed/dist/LiteYouTubeEmbed.css'
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
 
 &nbsp;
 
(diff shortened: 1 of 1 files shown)


phryneas:
Hm. Honestly I'm not 100% comfortable with a non-official template linked in the docs.

We could maybe move a RN template into the `reduxjs` namespace and maintain it there (WDYT @markerikson ?) though 🤔 .

As for the template itself, some immediate thoughts:
* this should probably go for the expo *managed* flow (no ios or android folder)
* some files there contain flow definitions and facebook copyright headers? (Header.tsx and LearnReduxLinks.tsx)
* I'm not sure if animated should be used in there - most ppl in the community would probably go for reanimated instead nowadays, so it might steer people in the wrong direction. (at the same time, I wouldn't want to showcase reanimated since it's unrelated)

markerikson:
Yeah, I've been debating how we should handle this.

Right now we have separate repos for the CRA+JS and CRA+TS templates.

I think it would make sense to have one monorepo with whatever templates we want to maintain.  Say, CRA-JS/TS, RN, and Vite.  There's also the Next template that's already over in the Next examples folder at https://github.com/vercel/next.js/tree/canary/examples/with-redux .

The biggest issues are:

- I certainly don't have any time to actively maintain these myself
- The Next template doesn't include `next-redux-wrapper` and there ought to be a page that actually discusses how/why to set up Redux + Next that way
- As Lenz pointed out, what the templates _should_ include

rahsheen:
> 

Sorry for the late reply.  😅 

* I created 4 templates. Expo TS/JS as well as React Native CLI JS/TS
* Some of the files are forked from the original templates, but I will check on the copyright / Flow headers 🤔 
* Animated is a native built-in feature so it made sense to use it here rather than add an unrelated third-party library

Let me know if there are any other questions. I created these quite a while back so not even sure how important it is to have JS templates to maintain. If there's some other repo these should live in, let me know.

PR #2846, opened 2022-10-30
Add an auto-batching enhancer that delays low-pri notifications and use with RTKQ

### LINKED ISSUE #2822: Review the `autoBatchEnhancer` concept and decide if it goes into 1.9
We played around with an enhancer that would tag actions with a priority level and only notify once all low-pri actions were done. Still worth reviewing for 1.9.

markerikson:
This PR:

- Adds a new `autoBatchEnhancer` that can cut down on the number of store subscription notifications, to improve UI perf
- Adds the corresponding `SHOULD_AUTOBATCH` field to `action.meta` in several key RTKQ actions, so that RTKQ gains the benefit of the batching if the enhancer is added to the store

Supersedes #2621 .

Fixes #2822 

## Description

A couple years ago I wrote a post on [A Comparison of Redux Batching Techniques](https://blog.isquaredsoftware.com/2020/01/blogged-answers-redux-batching-techniques/).  In that post I listed four major approaches:

- a higher-order reducer that accepts multiple actions nested inside one real action, and iterates over them together
- an enhancer that wraps `dispatch` and debounces the notification callback
- an enhancer that wraps `dispatch` to accept an array of actions
- React's `unstable_batchedUpdates()`, which just combines multiple queued renders into one but doesn't affect subscriber notifications

This enhancer is a variation of the "debounce" approach, but with a twist.

Instead of _just_ debouncing _all_ subscriber notifications, it watches for any actions with a specific `action.meta[SHOULD_AUTOBATCH]: true` field attached.

When it sees an action with that field, it queues a microtask.  The reducer is updated immediately, but the enhancer does _not_ notify subscribers right way.  If other actions with the same field are dispatched in succession, the enhancer will continue to _not_ notify subscribers.  Then, when the queued microtask runs at the end of the event loop tick, it finally notifies all subscribers, similar to how React batches re-renders.

The additional twist is also inspired by React's separation of updates into "low-priority" and "immediate" behavior (such as a render queued by an AJAX request vs a render queued by a user input that should be handled synchronously).

If some low-pri actions have been dispatched and a notification microtask is queued, then a _normal_ priority action (without the field) is dispatched, the enhancer will go ahead and notify all subscribers synchronously as usual, and _not_ notify them at the end of the tick.

This allows Redux users to selectively tag certain actions for effective batching behavior, making this purely opt-in on a per-action basis, while retaining normal notification behavior for all other actions.

Additionally, this now exports a `prepareAutoBatched()` util that can be used as the `prepare` callback inside of `createSlice`.

## Usage 

Usage of this might look like:

```ts
import { 
  createSlice,
  configureStore,
  autoBatchEnhancer,
  prepareAutoBatched
} from '@reduxjs/toolkit'

  name: 'counter',
  initialState: { value: 0 } as CounterState,
  reducers: {
    incrementBatched: {
      // Batched, low-priority
      reducer(state) {
        state.value += 1
      },
      prepare: prepareAutoBatched<void>(),
    },
    // Not batched, normal priority
    decrementUnbatched(state) {
      state.value -= 1
    },
  },
})
const { incrementBatched, decrementUnbatched } = counterSlice.actions

  enhancers: (existingEnhancers) => {
    return existingEnhancers.concat(autoBatchEnhancer())
  },
})
```

This PR also adds those batch meta fields to several key RTK Query action types.  This appears to significantly boost overall performance when rendering large lists of actions with unique query types.

## Benchmarks 

As a benchmark, I took https://github.com/phryneas/rtkq-performance-testbench and tweaked it slightly.  I then built it in prod mode using the `"react-dom/profiling"` entry point, opened the app, hit the "Record" button in the React Performance Profiler, and pasted `1000` into the "number of children" input.  Results:

### 1000 children, individual queries

- 1.8.5
  - 992 renders
  - initial: 81.5ms render, 7ms layout, 4355ms passive
- 1.9 local, no batching:
  - 992 renders
  - initial: 73ms render, 7ms layout, 3126ms passive
- 1.9 local, batched:
  - 992 renders
  - initial: 76ms render, 11ms layout, 1100ms passive
  
  
### 1000 children, same queries

- 1.8.5
  - 1 render
  - initial: 58ms render, 7ms layout, 3223ms passive
- 1.9.0 local, no batching:
  - 1 render
  - initial: 67ms render, 8ms layout, 167ms passive
- 1.9.0 local, batched:
  - 1 render
  - initial: 61ms render, 7ms layout, 144ms passive

My conclusions:

- The internal optimization work we've done to consolidate the RTKQ middleware and improve subscription handling drastically improves perf when many components/hooks request the same data at the same time
- The autobatch enhancer significantly improves perf when many components/hooks request _different_ data at the same time

## Summary

We plan to ship this in 1.9.  **It's purely opt-in, and we won't turn it on by default, but we'll likely update the docs to recommend adding this to the store as a standard approach**.

### DIFF
--- a/docs/api/autoBatchEnhancer.mdx
+++ b/docs/api/autoBatchEnhancer.mdx
@@ -1,0 +1,130 @@
+---
+id: autoBatchEnhancer
+title: autoBatchEnhancer
+sidebar_label: autoBatchEnhancer
+hide_title: true
+---
+
+&nbsp;
+
+# `autoBatchEnhancer`
+
+A Redux store enhancer that looks for one or more "low-priority" dispatched actions in a row, and delays notifying subscribers until either the end of the current event loop tick or when the next "normal-priority" action is dispatched.
+
+## Basic Usage
+
+```ts
+import {
+  createSlice,
+  configureStore,
+  autoBatchEnhancer,
+  prepareAutoBatched,
+} from '@reduxjs/toolkit'
+
+interface CounterState {
+  value: number
+}
+
+const counterSlice = createSlice({
+  name: 'counter',
+  initialState: { value: 0 } as CounterState,
+  reducers: {
+    incrementBatched: {
+      // Batched, low-priority
+      reducer(state) {
+        state.value += 1
+      },
+      // highlight-start
+      // Use the `prepareAutoBatched` utility to automatically
+      // add the `action.meta[SHOULD_AUTOBATCH]` field the enhancer needs
+      prepare: prepareAutoBatched<void>(),
+      // highlight-end
+    },
+    // Not batched, normal priority
+    decrementUnbatched(state) {
+      state.value -= 1
+    },
+  },
+})
+const { incrementBatched, decrementUnbatched } = counterSlice.actions
+
+const store = configureStore({
+  reducer: counterSlice.reducer,
+  // highlight-start
+  enhancers: (existingEnhancers) => {
+    // Add the autobatch enhancer to the store setup
+    return existingEnhancers.concat(autoBatchEnhancer())
+  },
+  // highlight-end
+})
+```
+
+## API
+
+### `autoBatchEnhancer`
+
+```ts title="autoBatchEnhancer signature" no-transpile
+export type SHOULD_AUTOBATCH = string
+export type autoBatchEnhancer = () => StoreEnhancer
+```
+
+Creates a new instance of the autobatch store enhancer.
+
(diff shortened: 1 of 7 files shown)


markerikson:
Gonna go ahead and merge this.

Lenz and I are up for a bit of bike-shedding on the `SHOULD_AUTOBATCH` field name, but if anyone has suggestions better make them _real_ soon - we'd like to publish 1.9 final in the next couple days :)

PR #2881, opened 2022-11-08
Fix Identifier/MemberExpression values in createSliceBuilder codemod

### LINKED ISSUE #2880: createSliceBuilder codemod deletes certain extra reducers
After running the codemod, some `extraReducers` got deleted (in my case those that just immediately delegated to an adapter).

Screenshot of a relevant part of the diff after running the `createSliceBuilder` codemod:

Screen Shot 2022-11-07 at 4 56 04 PM

kyletsang:
Fixes #2880

Reworked `wrapInAddCaseExpression` to handle different types of arguments for the `callExpression`.


markerikson:
Nice! Quick question: do we need to make similar changes to `createReducer` as well?

kyletsang:
> Nice! Quick question: do we need to make similar changes to `createReducer` as well?

It looks like we do.  I'll look at making a PR later in the day

PR #2888, opened 2022-11-09
fulfillWithValue should infer return value

### LINKED ISSUE #2886: `fulfillWithValue` creates wrong TS types for `action.payload`
Originally seen while working on #2884 .

If you do `return fulfillWithValue(someValue)` in `createAsyncThunk`, the TS typings end up thinking that the final return type _is_ the contents of that `FulfillWithMeta` object.  This results in TS thinking that `action.payload.payload.anyNestedField` and `action.payload.meta` are valid, even though _internally_ we do `fulfilled(action.payload)`.

Example:

```ts
export const getObj = createAsyncThunk(
  'slice/getObj',
  async (_: any, { fulfillWithValue, rejectWithValue }) => {
    try {
      return fulfillWithValue({ magic: 'object' });
    } catch (rejected: any) {
      return rejectWithValue(rejected?.response?.error || rejected);
    }
  }
);

export const mySlice = createSlice({
  name: 'slice',
  initialState,
  reducers: {},
  extraReducers: (builder) => {
    builder
      .addCase(getObj.pending, (state) => {
        state.loading = true;
      })
      .addCase(getObj.fulfilled, (state, action) => {
        state.loading = false;
        // ERROR This does not match
        state.obj = action.payload;
      })
      .addCase(getObj.rejected, (state) => {
        state.loading = false;
      });
  },
});
```

image

image

Looks like it's been broken since at least 1.8.

phryneas:
Closes #2886 

This should probably sit for a while to make sure it doesn't break anything or should be done somewhere else. 

So far, this was just not possible to be inferred - it is a bit nonsensical to use `returnWithValue` if no `fulfilledMeta` is declared in the `ThunkConfig` - that makes it impossible to add a `meta` as a second argument and defeats the purpose a bit.

### DIFF
--- a/packages/toolkit/src/createAsyncThunk.ts
+++ b/packages/toolkit/src/createAsyncThunk.ts
@@ -35,9 +35,7 @@ export type BaseThunkAPI<
   >
   fulfillWithValue: IsUnknown<
     FulfilledMeta,
-    <FulfilledValue>(
-      value: FulfilledValue
-    ) => FulfillWithMeta<FulfilledValue, FulfilledMeta>,
+    <FulfilledValue>(value: FulfilledValue) => FulfilledValue,
     <FulfilledValue>(
       value: FulfilledValue,
       meta: FulfilledMeta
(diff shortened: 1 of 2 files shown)


phryneas:
Okay, much more confident in this way of solving it.

PR #2919, opened 2022-11-17
Update axios example to use body instead of data

edvinerikson:
It seems that rtkq passes the data as `body` not `data`. This updates the example to send the body object as the data attribute

### DIFF
--- a/docs/rtk-query/usage/customizing-queries.mdx
+++ b/docs/rtk-query/usage/customizing-queries.mdx
@@ -303,13 +303,13 @@ const axiosBaseQuery =
     {
       url: string
       method: AxiosRequestConfig['method']
-      data?: AxiosRequestConfig['data']
+      body?: AxiosRequestConfig['data']
       params?: AxiosRequestConfig['params']
     },
     unknown,
     unknown
   > =>
-  async ({ url, method, data, params }) => {
+  async ({ url, method, body: data, params }) => {
     try {
       const result = await axios({ url: baseUrl + url, method, data, params })
       return { data: result.data }


phryneas:
Can you show an example of how you write one of your endpoints using that? I think you have a misconception there.

edvinerikson:
This is generated by the openapi codegen tool:
```
createOrder: build.mutation<CreateOrderApiResponse, CreateOrderApiArg>({
            query: queryArg => ({
                url: `/rest/ebs/orders/v1`,
                method: 'POST',
                body: queryArg.ebsCreateOrderRequestJson
            })
        }),
```

phryneas:
The OpenAPI codegen tool assumes that you use `fetchBaseQuery`. If you were using `axiosBaseQuery`, you would return different values from `query` - everything you return from `query` is input to your `baseQuery` and finally to axios or whatever other library you are using.

So a normal user using `axiosBaseQuery` (for **whatever reason**, sorry but I **really** cannot think of a good reason to use that instead of `fetch`, it just wasted 5kb of bundle size) would call
```js
createOrder: build.mutation<CreateOrderApiResponse, CreateOrderApiArg>({
            query: queryArg => ({
                url: `/rest/ebs/orders/v1`,
                method: 'POST',
                data: queryArg.ebsCreateOrderRequestJson
            })
        }),
```
because that is how axios is usually called.

=> In your case, this might be a workaround, but we don't want to show in the docs how to manipulate axios to work like fetch. At that point it starts becoming bizarre.

edvinerikson:
Ok, I understand. We can close this PR then. Axios is used for legacy reasons mostly. Can't get all shiny things at once unfortunately. I didn't realize that the codegen tool sets the data as body attribute hence my confusion.

Issue #2934, opened 2022-11-22
BUG: RTK Query: You can no longer set 'maxRetries' to 0


ritterb82:
In the latest update there was updates to the retry logic.  But the updates broke the ability to set maxRetries to 0.  I believe this is due to the `possibleMaxRetries` array that now exists and gets filtered by `Boolean`.  Which cause `0` to be tossed out.  

So then at that point, it just goes back to the default of 5.

https://github.com/reduxjs/redux-toolkit/blob/f7a828255e0612bed0856a0b6d1b9b8d6d6ac30d/packages/toolkit/src/query/retry.ts#L82-L87

OliverRadini:
Would it be alright if I picked this up? Looking to find some issues to work on and this seems interesting

markerikson:
Sure, please do!

OliverRadini:
@markerikson  thanks, hopefully the code in the PR properly resolves the issue

markerikson:
Out in https://github.com/reduxjs/redux-toolkit/releases/tag/v1.9.1

PR #2958, opened 2022-11-28
Retry now checks whether potential retry counts are undefined, rather than boolean, in order to avoid filtering out 0's

OliverRadini:
This will change the filter on possible max retries so that, instead of using `Boolean` (which may unintentionally match against false-y values) to checking that the properties are not undefined.

https://github.com/reduxjs/redux-toolkit/issues/2934

### DIFF
--- a/packages/toolkit/src/query/retry.ts
+++ b/packages/toolkit/src/query/retry.ts
@@ -83,7 +83,7 @@ const retryWithBackoff: BaseQueryEnhancer<
     5,
     ((defaultOptions as any) || EMPTY_OPTIONS).maxRetries,
     ((extraOptions as any) || EMPTY_OPTIONS).maxRetries,
-  ].filter(Boolean)
+  ].filter(x => x !== undefined)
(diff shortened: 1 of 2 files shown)


markerikson:
A bugfix PR with tests - wonderful! :)

OliverRadini:
@markerikson Thanks for merging, will take a look and see if there are any other issues I can take a look at

markerikson:
@OliverRadini : sure! there's some other stuff under https://github.com/reduxjs/redux-toolkit/milestone/11 that we could use help with.

OliverRadini:
@markerikson great, thanks. I may need to get up to speed on redux-toolkit, been a while since I've been able to do much in the redux world. Will do that and then take a look at the list you linked :)

markerikson:
Sure!  Also feel free to drop by the `#redux` channel in the Reactiflux Discord ( https://www.reactiflux.com) - we hang out there, mostly to answer questions.

PR #2992, opened 2022-12-10
Fix lint problems and enable linting on CI

thorn0:
Even though ESLint is configured in this project, it isn't called on CI, so the rules aren't enforced. Whenever I open the source code in my IDE, I see ESLint warnings and they're really distracting. This PR fixes the existing ESLint warnings and adds `yarn lint` to `build-ci` to enforce the rules in the future.

After fixing the lint issues (`yarn lint --fix`), I didn't run `yarn format` to reformat the code. Prettier is another tool that's configured but not enforced. Should I reformat the code and add `yarn format:check` to `build-ci` in this PR as well?

upd: I did reformat the code, but only in the files where lint issues have been fixed.

### DIFF
--- a/packages/toolkit/package.json
+++ b/packages/toolkit/package.json
@@ -77,13 +77,13 @@
     "yargs": "^15.3.1"
   },
   "scripts": {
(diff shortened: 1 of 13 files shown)


markerikson:
Thanks!

PR #3021, opened 2022-12-20
`raceWithSignal` method instead of `Promise.race`

### LINKED ISSUE #3020: User-reported memory leak in createListenerMiddleware
Pasted from https://stackoverflow.com/questions/74852458/memory-leak-with-redux-toolkits-createlistenermiddleware

### Original Report

I use Redux Toolkit, and in particular the new listener api, to perform tasks similar to what I could do with Redux-Saga.

Unfortunately, since a few days, I'm stuck with a memory leak and I can't find the cause.

I have reproduced a minimal example of the code that produces this memory leak, link to the example : https://github.com/MrSquaare/rtk-memory-leak

To observe this memory leak :

- I use Chromium, DevTools memory tool
- I trigger a garbage collector
- I make a heap memory snapshot
- I dispatch entity/load (via the UI button)
- I make several heap memory snapshots every 2-3 seconds
- I use the comparison tool, I notice that I have the array allocation size growing infinitely

And after dispatch entity/unload, then make a snapshot heap memory, we can observe that the allocations disappear...

Has anyone observed similar behavior? Or does anyone have an idea of the cause? Thanks!

EDIT 1:

I made an example with only the listener middleware ([`only-middleware` branch](https://github.com/MrSquaare/rtk-memory-leak/tree/only-middleware)), and compared it with different ways of doing :
- With `forkApi.pause` : Important leaks, especially of the generated entities
- Without `forkApi.pause` : I use directly `api.dispatch`, no more leaks of the generated entities, some leaks of other kinds, but maybe normal things (I am not qualified enough to pronounce on this)
- Without `api.dispatch` : I call directly the function that generates an entity, same result as with `api.dispatch`

It seems that the leak is related to `forkApi.pause`, but again I am not qualified enough to know the real cause...

### Other Notes

Lenz pointed to https://github.com/nodejs/node/issues/17469#issuecomment-349794909 - it may be related to use of `Promise.race()`:

```ts
export const createPause = <T>(signal: AbortSignal) => {
  return (promise: Promise<T>): Promise<T> => {
    return catchRejection(
      Promise.race([promisifyAbortSignal(signal), promise]).then((output) => {
        validateActive(signal)
        return output
      })
    )
  }
}
```

Lenz's note:

> So the problem is that promisifyAbortSignal will stick around for a very long time

phryneas:
This *should* fix #3020, but I have not tested it. 

@MrSquaare since you found this, could you please test the CodeSandbox CI build of this PR against your reproduction?

### DIFF
--- a/packages/toolkit/src/listenerMiddleware/index.ts
+++ b/packages/toolkit/src/listenerMiddleware/index.ts
@@ -39,10 +39,10 @@ import {
 } from './exceptions'
 import {
   runTask,
-  promisifyAbortSignal,
   validateActive,
   createPause,
   createDelay,
+  raceWithSignal,
 } from './task'
 export { TaskAbortError } from './exceptions'
 export type {
@@ -138,9 +138,9 @@ const createTakePattern = <S>(
     // Placeholder unsubscribe function until the listener is added
     let unsubscribe: UnsubscribeListener = () => {}
 
-    const tuplePromise = new Promise<[AnyAction, S, S]>((resolve) => {
+    const tuplePromise = new Promise<[AnyAction, S, S]>((resolve, reject) => {
       // Inside the Promise, we synchronously add the listener.
-      unsubscribe = startListening({
+      let stopListening = startListening({
         predicate: predicate as any,
         effect: (action, listenerApi): void => {
           // One-shot listener that cleans up as soon as the predicate passes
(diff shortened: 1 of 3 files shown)


MrSquaare:
Hello @phryneas, thanks for quickly proposing a solution

Indeed, the last commit ([4da043b](https://github.com/reduxjs/redux-toolkit/commit/4da043b682aab1d675d0de52e94184be759bba59)) fixes the memory leak problem!

Note: I tested using the CodeSandbox generated package for this commit as a dependency instead of using CodeSandbox directly.

PR #3051, opened 2022-12-30
Remove legacy object syntax for reducers

markerikson:
This PR:

- Removes the legacy object syntax for `createReducer` and `createSlice.extraReducers`

### DIFF
--- a/packages/toolkit/.size-limit.js
+++ b/packages/toolkit/.size-limit.js
@@ -19,30 +19,30 @@ function withRtkPath(suffix) {
         join(__dirname)
       ),
       new webpack.NormalModuleReplacementPlugin(
-        /rtk-query-react.esm.js/,
+        /rtk-query-react.modern.js/,
         (r) => {
           const old = r.request
           r.request = r.request.replace(
-            /rtk-query-react.esm.js$/,
+            /rtk-query-react.modern.js$/,
             `rtk-query-react.${suffix}`
           )
           // console.log(old, '=>', r.request)
         }
       ),
-      new webpack.NormalModuleReplacementPlugin(/rtk-query.esm.js/, (r) => {
+      new webpack.NormalModuleReplacementPlugin(/rtk-query.modern.js/, (r) => {
         const old = r.request
         r.request = r.request.replace(
-          /rtk-query.esm.js$/,
+          /rtk-query.modern.js$/,
           `rtk-query.${suffix}`
         )
         // console.log(old, '=>', r.request)
       }),
       new webpack.NormalModuleReplacementPlugin(
-        /redux-toolkit.esm.js$/,
+        /redux-toolkit.modern.js$/,
         (r) => {
           const old = r.request
           r.request = r.request.replace(
-            /redux-toolkit.esm.js$/,
+            /redux-toolkit.modern.js$/,
             `redux-toolkit.${suffix}`
           )
           // console.log(old, '=>', r.request)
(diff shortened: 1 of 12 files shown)


markerikson:
Note that I'm running into some TS issues here.

Removing the object overload is resulting in this error:

```ts
No overload matches this call.
  Overload 1 of 2, '(actionCreator: TypedActionCreator<string>, reducer: CaseReducer<number, Action<string>>): ActionReducerMapBuilder<number>', gave the following error.
    Argument of type 'string' is not assignable to parameter of type 'TypedActionCreator<string>'.
  Overload 2 of 2, '(type: "decrement", reducer: CaseReducer<number, Action<"decrement">>): ActionReducerMapBuilder<number>', gave the following error.
    Argument of type '(state: number, action: CounterAction) => number' is not assignable to parameter of type 'CaseReducer<number, Action<"decrement">>'.
      Types of parameters 'action' and 'action' are incompatible.
        Type 'Action<"decrement">' is not assignable to type 'CounterAction'.
          Property 'payload' is missing in type 'Action<"decrement">' but required in type '{ type: "decrement"; payload: number; }'.
```

In some cases this can be worked around by using `PayloadAction<number, "increment">` (ie specifying the literal type of `action.type`), but that's not always feasible.

This may need additional tweaking.

markerikson:
Interesting.  It seems to be specifically due to:

```ts
type CounterActions = 
  | {type: "increment", payload: number}
  | {type: "decrement", payload: number}

  const incrementHandler = (
    state: number,
    action: CounterActions
  ) => state + action.payload
```

If I change it to `action: {type: "increment", payload: number}`, it works okay.

markerikson:
Well this is lovely:

image

These expected errors _don't_ happen as of TS 4.7, and apparently again due to these changes.

markerikson:
Lovely.  Apparently as of TS 4.7, the reducer is getting inferred as `Reducer<any, AnyAction>'.  So of course that `state` type of `any` doesn't cause an error.

markerikson:
Hmm. Per Lenz, the `Reducer<any>` issue appears to already have been happening with TS 4.7+ on `master`.

phryneas:
Okay, so [this is our reproduction](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwMZQKYEMaoMoBthKoC+cAZlBCHAOQACaAJgK4AeAVgM4D0MEEuAa2AxqAWABQXAFQS4UuABVUHGAC44KDFjwFUACgCUcYADtSqKBzgwAFqjhoYTKCdQM4HfIWpWYATzBUADpZKS4JBFkNCBMVDy97AF4NNEwcBL1I8TgcuC4uXIA9AH4onJN0EFR1aiQIJhMsKGoAGjLjE2FgdFxsGDT1AAY27NzGJkJLdSzc3NNNKsb1PRUBuBMmEAAjCxa4dCQYYBijRIA+D36sOABqfcPjkyCwdD9cCHQGEdmchlQF1BLOArK7VdabHZQPYHI4nODnS5pOAAWnusKeLzeHy+7SI31yqBYMCg6AASm4JhYOMstkxgLg-lBThcZj9cfiiAYJEQJEA):

```ts
import { createSlice } from "@reduxjs/toolkit";
{
  const slice = createSlice({
    //   ^?
    name: "counter",
    initialState: 0,
    reducers: {
      increment: (state: number, action) => state + action.payload,
      decrement: (state: number, action) => state - action.payload,
    },
    extraReducers: (builder) => {},
  });
}
```
There are two conditions here: 
* `action` is untyped
* there is an `extraReducers` function

So it will probably not appear in the wild a lot, since `action` usually will not be untyped.

PR #3074, opened 2023-01-11
RFC: configurable `produce` implementation

phryneas:
With `immer` alternatives like [structura.js](https://giusepperaso.github.io/structura.js/) and [mutative](https://github.com/unadlib/mutative) popping up, it might make sense to allow the configuration of the `produce` implementation used by RTK.

This PR would, by default, still use `immer`, but expose `buildCreateReducer` and `buildCreateSlice` functions to create custom `createReducer` and `createSlice` functions. This should tree-shake well, and if only those other implementations are used, `immer` should not be bundled anymore (as soon as the polyfill in 2.0 is disabled, which is why I base this PR against the v2 version).

This would also theoretically allow `produce` to be swapped out for a stub, so people who *really* want to use RTK without any immutability helper could do so - even though I would highly recommend against that.

It would also prepare for future situations when Records and Tuples are widely available, and an immutability helper will not be necessary anymore.

There are still a few more places where we use `produce`, so additional changes would be required in
* createEntityAdapter
* createApi (here we need to configure with a version allowing for patches)

So far, I just wanted to put this out as a base for discussion - the code changes required for this are pretty small, so I think it might make sense at this point.

PS: we might need to do some docs updates, as, at the moment, the docs directly reference the types of `createSlice` and that has now moved into its own interface definition.

PS2: at this point, `mutative` does not allow for returning a new object, so that one might not work yet. [Let's see if we can convince the author to add that functionality](https://www.reddit.com/r/javascript/comments/100gsnx/comment/j3vpu8k/?utm_source=reddit&utm_medium=web2x&context=3).

### DIFF
--- a/packages/toolkit/src/createReducer.ts
+++ b/packages/toolkit/src/createReducer.ts
@@ -82,7 +82,8 @@ export type ReducerWithInitialState<S extends NotFunction<any>> = Reducer<S> & {
 
 let hasWarnedAboutObjectNotation = false
 
-/**
+export type CreateReducer = {
+  /**
  * A utility function that allows defining a reducer as a mapping from action
  * type to *case reducer* functions that handle these action types. The
  * reducer's initial state is passed as the first argument.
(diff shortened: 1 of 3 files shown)


unadlib:
> 

I will add support for return values to [mutative](https://github.com/unadlib/mutative). But from the user's point of view, we should suggest always returning the draft itself or no value at all, not only does it maintain optimal performance (no extra performance wasted traversing unpredictable return values), but it also maintains the full mutation feature.

phryneas:
Hi @unadlib! That's great news!
In [writing reducers with `immer`](https://redux-toolkit.js.org/usage/immer-reducers) (which we might need to rename to "writing mutable immutable reducers" or something like that), we generally recommend changing the existing state - but there are cases where the user might want to return something completely different.

Most common cases would be:
* replacing the full state with a completely new object (e.g. resetting state or taking a big chunk of data that came in, for example from an api call)
* the user wants to optimize things by hand

From our perspective it would be fine if the rule were "if the user returns a value and that value is not the draft, just return that whole new value without traversing and touching it in any way", but I'm not sure if there are additional things that you have to do and cannot skip.

unadlib:
> Hi @unadlib! That's great news! In [writing reducers with `immer`](https://redux-toolkit.js.org/usage/immer-reducers) (which we might need to rename to "writing mutable immutable reducers" or something like that), we generally recommend changing the existing state - but there are cases where the user might want to return something completely different.
> 
> Most common cases would be:
> 
> * replacing the full state with a completely new object (e.g. resetting state or taking a big chunk of data that came in, for example from an api call)
> * the user wants to optimize things by hand
> * they are just migrating over old Redux code that is not doing mutable updates yet.
> 
> From our perspective it would be fine if the rule were "if the user returns a value and that value is not the draft, just return that whole new value without traversing and touching it in any way", but I'm not sure if there are additional things that you have to do and cannot skip.

Cool proposal. Since the user may refer to the draft in the return of a completely new object, I will add a new option `strictReturn` , which we will suggest the user enable `strictReturn` in development mode and disable `strictReturn` in production mode. This will ensure safe returns and also keep good performance in the production build.

packages/toolkit/src/createReducer.ts
```diff
@@ -203,104 +204,118 @@ const counterReducer = createReducer(0, {
...
-      ...finalActionMatchers
-        .filter(({ matcher }) => matcher(action))
-        .map(({ reducer }) => reducer),
-    ]
-    if (caseReducers.filter((cr) => !!cr).length === 0) {
-      caseReducers = [finalDefaultCaseReducer]
+    // Ensure the initial state gets frozen either way (if draftable)
+    let getInitialState: () => S
+    if (isStateFunction(initialState)) {
+      getInitialState = () => freezeDraftable(initialState())
```
markerikson, line 255: :hand: This still relies on Immer imports over in `./utils`

packages/toolkit/src/createReducer.ts
```diff
@@ -203,104 +204,118 @@ const counterReducer = createReducer(0, {
...
+      }
 
-          return result as S
-        } else if (!isDraftable(previousState)) {
-          // If state is not draftable (ex: a primitive, such as 0), we want to directly
-          // return the caseReducer func and not wrap it with produce.
-          const result = caseReducer(previousState as any, action)
+      return caseReducers.reduce((previousState, caseReducer): S => {
+        if (caseReducer) {
+          if (isDraft(previousState)) {
```
markerikson, line 274: :hand: More Immer

unadlib:
@phryneas @markerikson Mutative v0.3.1 already supports returning values in the draft functions. It is already fully compliant with `produce` interface in `redux-toolkit` and maintains good performance.

markerikson:
@unadlib : nice!

One other question that came up is that we use several of Immer's utilities like `isDraft` and `isDraftable`, in order to check whether it's safe to pass certain values into `produce()`.

Does `mutative` have anything like those?

unadlib:
> @unadlib : nice!
> 
> One other question that came up is that we use several of Immer's utilities like `isDraft` and `isDraftable`, in order to check whether it's safe to pass certain values into `produce()`.
> 
> Does `mutative` have anything like those?

Yes, Mutative also provides `isDraft()` and `isDraftable()`.

markerikson:
I rebased this branch against `v2.0-integration`.

Also, for kicks, I tried swapping all our usages of Immer to use Mutative instead. Unfortunately, there's several critical test failures, including basic `createReducer` behavior:

https://github.com/reduxjs/redux-toolkit/tree/feature/try-mutative

unadlib:
> I rebased this branch against `v2.0-integration`.
> 
> Also, for kicks, I tried swapping all our usages of Immer to use Mutative instead. Unfortunately, there's several critical test failures, including basic `createReducer` behavior. Looks like it's hitting a `Cannot perform 'has' on a proxy that has been revoked` error.
> 
> https://github.com/reduxjs/redux-toolkit/tree/feature/try-mutative

hi, @markerikson, I will debug it later. Overall, regarding the integration with Reducer, Immer and Mutative are not exactly equivalent, Mutative supports more options.

markerikson:
Hmm. That definitely makes it harder to use `mutative` as a drop-in replacement, because we don't know ahead of time what users are going to do in their reducers.

I _think_ the right answer here is that we'd need to inject a wrapper that uses `safeReturn`, something like:

```ts
const createReducerWithMutative = buildCreateReducer({
  createNextState: (state, recipe) => {
    return create(state, (draft) => {
      return safeReturn(recipe(draft))
    })
  },
  isDraft,
  isDraftable
})
```

unadlib:
> Hmm. That definitely makes it harder to use `mutative` as a drop-in replacement, because we don't know ahead of time what users are going to do in their reducers.
> 
> I _think_ the right answer here is that we'd need to inject a wrapper that uses `safeReturn`, something like:
> 
> ```ts
> const createReducerWithMutative = buildCreateReducer({
>   createNextState: (state, recipe) => {
>     return create(state, (draft) => {
>       return safeReturn(recipe(draft))
>     })
>   },
>   isDraft,
>   isDraftable
> })
> ```

When the user uses `createReducerWithImmer`, if the reducer wants to return the value `undefined`, the reducer will have to return return Immer's `nothing`, and this implicit rule is similar to the case.

I also agree with you. Whether `createReducerWithMutative` returns `undefined` explicitly or not is left up to the user, like this.

```ts
const createReducerWithMutative = buildCreateReducer({
  createNextState: (state, recipe) => {
    return create(state, (draft) => {
      const value = recipe(draft);
      return value === undefined ? value : safeReturn(value);
    })
  },
  isDraft,
  isDraftable
})
```

unadlib:
@markerikson I have refactored the implementation of the return value, once you upgrade to mutative v0.5.0, you will migrate Immer to Mutative smoothly, the relevant tests have passed. 

The details are at [https://github.com/unadlib/mutative/issues/9](https://github.com/unadlib/mutative/issues/9)

BTW, I tested the Immer v10 branch. Mutative is still more than 15x faster than Immer.

giusepperaso:
Hi, I'm the author of [structurajs](https://giusepperaso.github.io/structura.js/). Thanks for this pull request!

After some work I think I finally managed to make the library fully compatible with redux toolkit

As an example I did setup a [try-structura branch](https://github.com/giusepperaso/redux-toolkit/tree/feature/try-structura) which passes all of the tests contained into the packages/toolkit folder

Note that in this example branch I removed Immer completely, and I put all of my imports inside the [exports.ts](https://github.com/giusepperaso/redux-toolkit/blob/feature/try-structura/packages/toolkit/src/exports.ts) file

Also, I called enableAutoFreeze and enableStandardPatches to make all of the tests pass, but they are not always necessary and could be disabled to gain some performance:
-  structura does not allow the produced state to be modified (at compile time via typescript), so relying on object freeze could be a bit redundant
- structura by default does not use standard json patches, but a proprietary format based on linked lists which is slightly faster; you probably don't need standard json patches unless you have to manipulate them with another library

If you see any problem or if you need help with something let me know! @phryneas @markerikson

markerikson:
@giusepperaso : thanks for working on that!

FWIW we really _do_ want freezing in place, because we _want_ users to see errors at runtime if they attempt to mutate outside of a reducer.  (Not everyone uses TS!)

EskiMojo14:
is it worth making a configurable version of createDraftSafeSelector too? that currently uses immer's isDraft() and current()

markerikson:
Found another similar lib: https://github.com/tnfe/limu

unadlib:
Hi, I'm not sure if there are any next steps planned for this proposal.

Currently, the performance test information related to Immer and Mutative is as follows:

- [https://github.com/instantdb/instant/pull/243](https://github.com/instantdb/instant/pull/243)
- [https://github.com/reduxjs/redux-toolkit/pull/4361#issuecomment-2110626022](https://github.com/reduxjs/redux-toolkit/pull/4361#issuecomment-2110626022)
- [https://mutative.js.org/docs/getting-started/performance/](https://mutative.js.org/docs/getting-started/performance/)

Perhaps performance improvement is not a high priority for the `redux-toolkit`. So maybe I could provide an additional Mutative lib  based on the `redux-toolkit`?

But, if there are further plans for this proposal, I would be very happy to participate.

markerikson:
@unadlib it's still an idea I'm interested in conceptually, but it's not on our roadmap to work on this any time soon. (I haven't had time to work on much RTK stuff lately, and when I do my focus will be on RTK Query features.)

PR #3095, opened 2023-01-16
Migrate the RTK package to be full ESM

markerikson:
This PR attempts to convert the RTK package from its existing "contains ESM and CJS modules, but not _fully_ ESM", to be fully ESM with `{type: "module"}` and still support CJS:

- **BREAKING**: Sets the main RTK `package.json` file to be `{type: "module"}`
- **BREAKING**: Updates all entry point `package.json` files to use `exports` to point to the types, ESM file, and CJS file
  - I still have `main` and `module` in there because WHO KNOWS WHETHER THOSE STILL END UP GETTING USED BY SOME TOOLS OR NOT 🤷‍♂️ 
- Updates the build script:
  - Fixed ESM compat on execution by replacing use of `__dirname` and fixing the Terser import
  - Switched all build targets to be `"esnext"`, to ensure the output is untouched other than TS transpilation
  - Moved all CJS build artifacts to be nested a level deeper in each entry point in a `./cjs/` folder, ie, `./dist/query/cjs/`
  - Added `{type: "commonjs"}` package files to those folders
  - Turned off the UMD build artifacts for now

I've done some hand-testing of a locally-built version of this.  I set up an example project that used all three entry points ( `@reduxjs/toolkit`, `/query`, and `/query/react`), then  that ported it to several different build tools and ran both dev and prod builds for each:

- Vite 4: worked
- CRA 4 and 5: worked
- Next 13: needs `redux-thunk` to be converted over to ESM as well, as the middleware array ended up as `[ {default: thunk()}, api1.middleware, api2.middleware]`. Installing a local build of `redux-thunk` with similar ESM changes made that work
- Parcel: worked
- RN Expo: it found the correct file ( `rtk-query-modern.js` ), but choked on the optional chaining syntax inside, and I don't feel like messing with RN further. Important thing is it found the right file.

I also tested this with a tiny Node script that imported all three entry points in both a CJS and an ESM file, and got that running without errors.

Finally, I tested this with the https://publint.dev/ CLI run against all three entry point `package.json` files, and it gave me a ✅ for each.

Given that, I'm tentatively hopeful that the `exports` configuration is actually sufficient.

### DIFF
--- a/packages/toolkit/jest.config.js
+++ b/packages/toolkit/jest.config.js
@@ -1,4 +1,4 @@
-module.exports = {
+export default {
   testEnvironment: 'jest-environment-jsdom',
   setupFilesAfterEnv: ['./jest.setup.js'],
   testMatch: ['<rootDir>/src/**/*.(spec|test).[jt]s?(x)'],
--- a/packages/toolkit/package.json
+++ b/packages/toolkit/package.json
@@ -22,10 +22,28 @@
   "publishConfig": {
     "access": "public"
   },
-  "main": "dist/index.js",
+  "type": "module",
   "module": "dist/redux-toolkit.modern.js",
-  "unpkg": "dist/redux-toolkit.umd.min.js",
+  "main": "dist/cjs/index.js",
   "types": "dist/index.d.ts",
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/redux-toolkit.modern.js",
+      "default": "./dist/cjs/index.js"
+    },
+    "./query": {
+      "types": "./dist/query/index.d.ts",
+      "import": "./dist/query/rtk-query.modern.js",
+      "default": "./dist/query/cjs/index.js"
+    },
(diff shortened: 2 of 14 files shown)


markerikson:
Well. The good news is I think the _runtime_ code works.

Bad news is Jest is being a pain.  In particular, something about the way it's importing `redux-thunk` makes the default import an object like `{default}`, which is not a middleware function, and thus the tests explode when we try to create a store.

I spent the last couple hours hacking around with the thunk exports and republishing it locally.  Switching the thunk package over to _not_ having a default export at all sorta helped, but now something about the "no dev middleware in prod" test is failing.

So, close, but can't even build this branch yet.

markerikson:
Hmm. So on the one hand, #3102 looks pretty good in that the migration is working and all RTK tests are passing.

On the other hand... surprisingly I'm not seeing the Immer default import issues here in CI that I was seeing locally.

I'm confused :)

markerikson:
Okay, let's do this!

Technically the OpenAPI test failed, but that's a timeout issue

PR #3102, opened 2023-01-19
Migrate RTK test suite from Jest to Vitest

markerikson:
While working on the RTK ESM migration in #3095 , I ran into issues with Jest and mixed ESM/CJS packages.

I've seen reports on Twitter that Jest struggles with default exports if a package is mixed ESM/CJS ( https://twitter.com/barelyreaper/status/1615297215312060417 ). In particular, both `redux-thunk` and `immer` export a function as their default export.  However, once the RTK package got turned into full ESM with `{type: "module"}`, Jest began importing both of those as `{default: whateverExport}` rather than just `whateverExport`, and this caused our tests to break because the library code was trying to call `whateverExport()` and getting "that's not a function" errors since it was now a `{default}` object.

In the case of `redux-thunk`, this manifested itself as `configureStore`'s "are these all really middleware?" check throwing an error.

We own `redux-thunk`, so I was able to put out a `3.0-alpha` build that migrated it to full ESM also.  But, I was still seeing the same errors.  I then changed `redux-thunk` to _only_ have named exports in `3.0.0-alpha.1`, and that fixed _those_ errors...

only to start seeing the same error come up with Immer.  I changed all of our `import produce/createNextState` to be `import { produce }`, but same issue.

Several folks suggested giving Vitest a shot instead, so I did.

> **Note**: This PR is _just_ the Jest->Vitest migration. See #3095 for the rest of the ESM conversion.

The majority of this migration is a straight 1:1 search-and-replace of `jest.someFunc()` -> `vi.someFunc()`.

The use of fake timers got simple, as I was able to delete some of the use of various wait utils we had.

I did run into some problems with the `console-testing-library` package, in three ways.  One is that it uses `__filename` internally, which doesn't exist in a full ESM environment.  Second, it does `import from "./pure"`, which doesn't have a `.js` extension, and that's needed with ESM.  Finally, it tries to define/override its own `toMatchInlineSnapshot`, and that broke Vitest's version - something about differences in stack traces and error serialization.  I ended up defining a `yarn patch` to fix the import and comment out the entire `.toMatchInlineSnapshot` section (which also had the `__filename` usage).

I also saw some very odd behavior with `.toMatchInlineSnapshot()`, both before and after I fixed up `console-testing-library`.  I'd run the tests with `yarn test run -u`, and everything would pass.  But, some of the snapshots in the entity adapter files got updated incorrectly, which is disturbing.  I later had to go back, reset those files to pre-Vitest-migration versions, and then do search-and-replace to change the snapshots to be `.toEqual(someObject)` instead.

Overall, the tests pass locally and seem to run fairly past (15s):

```
  Snapshots  19 updated
 Test Files  50 passed | 2 skipped (52)
      Tests  779 passed | 30 skipped (809)
   Start at  21:06:13
   Duration  15.81s (transform 5.93s, setup 12.76s, collect 52.91s, tests 33.03s)
```

The last problems I ran into were related to how we actually do a local `yarn pack` + `yarn install` of RTK, and then run our tests against the "published" package.  `configureStore.test.ts` tries to spy on the Redux package exports to verify that `configureStore` really calls them, but that wasn't working with the "test vs published package" job after the Vitest switch. I had to spend a couple hours randomly shuffling around various forms of `vi.mock()` and `vi.doMock()` until I finally found something that worked.

I _have_ occasionally seen Node crash with v8 errors that seem memory-related, but maybe that's a Windows thing?

At this point I'm mostly inclined to include this in the 2.0 ESM migration.  It's green, we know Jest has issues with ESM overall, and it does run faster.

One concern I do have is that the RTK 2.0 ESM + Immer 9 default import issue might be something that affects end users using Jest, not just us and our repo.  That will need some more investigation.


markerikson:
Did some quick benchmarking out of curiosity.

On `master`, we have `jest --runInBand` for some reason.  Not sure why.

Times:

- jest inBand: 105s
- jest threaded: 28s
- vitest: 18s (wall time)

So, Vitest threaded seems faster than Jest threaded in our case, and _way_ faster than Jest inBand

packages/toolkit/src/query/tests/cacheCollection.test.ts
```diff
@@ -8,15 +9,17 @@ import {
...
 })
 
-const onCleanup = jest.fn()
+const onCleanup = vi.fn()
 
 beforeEach(() => {
   onCleanup.mockClear()
 })
 
+test('abc', () => {})
```
Shrugsy, line 21: ❓ 

was this intended to be a temporary thing?

packages/toolkit/src/query/tests/cacheLifecycle.test.ts
```diff
@@ -1,10 +1,17 @@
 import { createApi } from '@reduxjs/toolkit/query'
 import type { FetchBaseQueryMeta } from '@reduxjs/toolkit/query'
+import { vi } from 'vitest'
 import { fetchBaseQuery } from '@reduxjs/toolkit/query'
-import { expectType, fakeTimerWaitFor, setupApiStore, waitMs } from './helpers'
+import {
+  expectType,
+  fakeTimerWaitFor,
+  setupApiStore,
+  // waitMs,
```
Shrugsy, line 9: 🗑️ ❓ 

are we fine to just delete this line now?

packages/toolkit/src/query/tests/cacheLifecycle.test.ts
```diff
@@ -121,11 +129,12 @@ describe.each([['query'], ['mutation']] as const)(
...
 
-      promise.unsubscribe(), await waitMs()
+      promise.unsubscribe()
+      await vi.advanceTimersByTimeAsync(DEFAULT_DELAY_MS)
       if (type === 'query') {
-        jest.advanceTimersByTime(59000), await waitMs()
+        await vi.advanceTimersByTimeAsync(59000) //, await waitMs()
         expect(onCleanup).not.toHaveBeenCalled()
-        jest.advanceTimersByTime(2000), await waitMs()
+        await vi.advanceTimersByTimeAsync(2000) //, await waitMs()
```
Shrugsy, line 137: ❓ 

are we fine to delete the commented out `await Ms`  portions?

packages/toolkit/src/query/tests/cleanup.test.tsx
```diff
@@ -55,11 +55,11 @@ test('data stays in store when component stays rendered', async () => {
...
 
-  jest.advanceTimersByTime(120000)
+  vi.advanceTimersByTime(120000)
 
-  await waitFor(() =>
-    expect(getSubStateA()?.status).toBe(QueryStatus.fulfilled)
-  )
+  // await waitFor(() =>
+  expect(getSubStateA()?.status).toBe(QueryStatus.fulfilled)
+  // )
```
Shrugsy, line 62: 🗑️ ❓ 

are we fine to delete the commented out `waitFor` portion?

packages/toolkit/src/query/tests/helpers.tsx
```diff
@@ -1,13 +1,15 @@
...
   EnhancedStore,
   Middleware,
   Store,
+  Reducer,
 } from '@reduxjs/toolkit'
 import { configureStore } from '@reduxjs/toolkit'
 import { setupListeners } from '@reduxjs/toolkit/query'
-import type { Reducer } from 'react'
-import React, { useCallback } from 'react'
+// import type { Reducer } from 'react'
```
Shrugsy, line 11: 🗑️ ❓

packages/toolkit/src/query/tests/queryLifecycle.test.tsx
```diff
@@ -53,7 +54,7 @@ describe.each([['query'], ['mutation']] as const)(
             async onQueryStarted(arg, { queryFulfilled }) {
               onStart(arg)
               // awaiting without catching like this would result in an `unhandledRejection` exception if there was an error
-              // unfortunately we cannot test for that in jest.
+              // unfortunately we cannot test for that in vi.
```
Shrugsy, line 57: 💁 

this seems like a victim of search and replace - it's probably better to refer to it as `vitest` here rather than `vi`

Shrugsy (approved):
👍 

awesome! I've left some notes on things that are mainly commented lines which look like they could be deleted now

Issue #885, opened 2021-02-03
Only run serializableCheck on actions and not on state


ksjogo:
The serializableCheck is quite valuable, but with our state architecture it often takes up to 100ms for every action of which a lot might be dispatched.
It would be great to have an option to only run the check on the current action and not the whole state.
This seems to be quite easy to add around
https://github.com/reduxjs/redux-toolkit/blob/ba8c911c53e67365499ac00982f640b369360a64/src/serializableStateInvariantMiddleware.ts#L193
Could I create a PR?

markerikson:
Yeah, that seems reasonable.  Please go ahead.

phryneas:
While you're at it: it might also be interesting to skip out of the check if the part of the state is frozen. That will usually be the case with data that was autofrozen by immer, so most of the time the check would be pretty obsolete anyways.

markerikson:
Yeah, would it be worth having the mutation check middleware look for frozen state as well?

I could see that speeding up both of these middleware noticeably, which would actually improve the value.

PR #3115, opened 2023-01-23
Add caching to serializableStateInvariantMiddleware

GeorchW:
The serializableStateInvariantMiddleware is currently scanning the entire state for non-serializable values whenever an action is dispatched. Dispatching a bunch of actions thus takes about `O(numDispatchedActions * stateSize)` time once the state is populated, independent of the size of the action or the complexity of the reducer. This massively slows down the application performance in dev mode for big state trees. Such trees can quickly occur when downloading a bunch of data and putting it into the store, e.g. using rtk-query.

In #885, it was discussed to not check the serializability of the state at all (and only the actions) as an optional addition, which was implemented in #943. It has the disadvantage of giving up detecting non-serializable objects created in a reducer, and I suppose that that's the reason it isn't enabled by default. It can also be avoided by excluding some parts of the store from the middleware using `ignoredPaths`. All of these options, however, require profiling and googling to realize why rtk/rtk-query is slow and what to do about it.

This PR introduces `WeakSet`-based caching to the middleware, making the performance independent of the state size as long as the state is frozen, which happens automatically when using the toolkit-provided functions for creating slices.

When enabled by default, developers should experience the same time complexity in dev and prod, and they would need to profile/google much less.

## Open questions

- [ ] How should we determine the frozen-ness of an object? Can we assume that all descendants of an object are frozen if the object itself is frozen? Do we need to imitate the `getEntries`-based method of finding descendants of `findNonSerializableValue`?

### DIFF
--- a/packages/toolkit/src/serializableStateInvariantMiddleware.ts
+++ b/packages/toolkit/src/serializableStateInvariantMiddleware.ts
@@ -38,7 +38,8 @@ export function findNonSerializableValue(
   path: string = '',
   isSerializable: (value: unknown) => boolean = isPlain,
   getEntries?: (value: unknown) => [string, any][],
-  ignoredPaths: IgnorePaths = []
+  ignoredPaths: IgnorePaths = [],
+  cache?: WeakSet<object>
 ): NonSerializableValue | false {
   let foundNestedSerializable: NonSerializableValue | false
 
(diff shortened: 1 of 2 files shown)


markerikson:
Oh this is neat! And yeah, the very first question that came to mind was "what about older browsers without `WeakSet`?", and I see that's handled.

(I don't _want_ to keep supporting IE11, and RTK 2.0 will explicitly drop support for older browsers, but shipping code that _always_ uses `WeakSet` in default behavior would have been pretty close to a breaking change by itself, and thus not really suitable for a minor or patch.)

Issue #3105, opened 2023-01-20
Race conditions when invalidations come in quick succession


GeorchW:
There is a simple race condition in the library right now:

1. A query that provides a tag starts.
2. A mutation invalidates the same tag.
3. The query finishes.

The query is not rerun in this case, despite its tags having been invalidated. I actually noticed this in my application when the query took a bit too long while mutations were applied.

The problem stated here is similar to the problem stated in #2203, but the title/framing is different in this issue.

A simple fix could be to add the option to not invalidate any tags while any queries (or mutations) are running, and only invalidate them after all pending queries are settled. This would also solve #2203. It could, however, cause problems for people using extremely long-running queries (since no tags are invalidated while they run), so it might make sense to allow disabling this behavior.

Implementation-wise, we could add another field to the API slice named `pendingTagInvalidations`, which contains a list of tags that will be invalidated once all queries are settled. Each time that a mutation settles, it checks if there are any queries/mutations that are in the `pending` state. If that is the case, it adds the tags to `pendingTagInvalidations`, else it invalidates them directly. When a query settles, it does the same check and invalidates all tags in `pendingTagInvalidations`, then clears the list.

Pseudocode:

```ts
// api is our API slice
function hasPendingRequests(api: ApiSlice) {
    const hasPendingQueries = Object.values(queries)
        .some(query => query?.status === QueryStatus.pending)
    const hasPendingMutations = Object.values(mutations)
        .some(mutation => mutation?.status === QueryStatus.pending);
    return hasPendingQueries || hasPendingMutations;
}
// When a mutation settles:
if (hasPendingQueries(api)) {
    api.pendingTagInvalidations.push(...invalidatedTags);
} else {
    invalidateTags(invalidatedTags);
}
// When a query settles:
if (!hasPendingQueries(api)) {
    invalidateTags(api.pendingTagInvalidations);
    api.pendingTagInvalidations = [];
}
```

The overhead of this shouldn't be too high (in `O(#endpoints)` ), but if required, we could save the list of pending requests separately (making it `O(1)`).

I'd be happy to contribute!

markerikson:
Yeah, we're definitely open to PRs if you think you have a solution!

markerikson:
Out in https://github.com/reduxjs/redux-toolkit/releases/tag/v2.0.0-beta.4

PR #3116, opened 2023-01-23
Delayed tag invalidations

### LINKED ISSUE #2203: Document how to batch RTK Query mutations properly
I feel that the documentation needs to cover the following scenario.

I have an array of objects that I need to send to the API. The API only allows mutating one object per request. What is a good way to do this while:

- avoiding a race condition where the final cache invalidation is ignored because the second to last cache provide comes later.
- even better, firing just one cache invalidation after all mutations have been sent.

I understand mutations can probably use a `queryFn` to do this, but the docs only cover how to do multiple fetches.

GeorchW:
I've described the approach in this PR earlier in #3105.
Resolves #2203, #3105

## Problem

There are race conditions when a mutation is finishing while a query is running. More concretely, this happens:

1. The query Q starts.
2. Mutation M finishes. It invalidates the tag T. There is currently no query that provides the tag T, so no query is invalidated.
3. The query Q finishes. It provides the tag T.

If the data from Q was generated before M arrived on the server, then the data is outdated. This is rare in normal circumstances, but when many mutations/queries are being started within a short time frame, it becomes likely.

## Solution

If a mutation finishes while another query or mutation is running, don't invalidate tags. Instead, invalidate them later after all queries/mutations are settled. 

In the much more common case that the mutation isn't running at the same time as a query, this won't change anything.

For the bug described above, it would be sufficient if the tag invalidation would only be delayed until all queries are settled (as opposed to queries + mutations). However, waiting for the mutations as well has the advantage of solving issue #2203 on the go: when multiple mutations are fired at the same time, queries will only start after all the mutations settled. In a way, they are batched automatically.

## Open Questions

- [ ] Does this change require an update to the documentation?
- [ ] Should the purpose of `pendingTagInvalidations` be documented somewhere in the code? If yes, where?
- [ ] I'm not sure whether I've put my code in the right places. Feel free to request changes :-)

### DIFF
--- a/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
@@ -85,6 +85,11 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
 
     const state = rootState[reducerPath]
 
+    if (state.config.invalidateImmediately) {
+      handleInvalidatedTags(tags, mwApi)
+      return
+    }
+
     const hasPendingQueries = Object.values(state.queries).some(
       (x) => x?.status === QueryStatus.pending
     )
--- a/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
@@ -108,6 +108,7 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
     const state = rootState[reducerPath]
 
     const tags = [...state.pendingTagInvalidations, ...newTags]
+    if (tags.length === 0) return
 
     const toInvalidate = api.util.selectInvalidatedBy(rootState, tags)
 
--- a/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
@@ -39,10 +39,8 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
   )
 
   const isQueryEnd = isAnyOf(
-    isFulfilled(mutationThunk),
-    isRejected(mutationThunk),
-    isFulfilled(queryThunk),
-    isRejected(queryThunk)
+    isFulfilled(mutationThunk, queryThunk),
+    isRejected(mutationThunk, queryThunk)
   )
 
   let pendingTagInvalidations: FullTagDescription<string>[] = []
--- a/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
@@ -5,9 +5,12 @@ import {
   isRejectedWithValue,
 } from '../rtkImports'
 
-import type { FullTagDescription } from '../../endpointDefinitions'
+import type {
+  EndpointDefinitions,
+  FullTagDescription,
+} from '../../endpointDefinitions'
 import { calculateProvidedBy } from '../../endpointDefinitions'
-import type { QueryCacheKey } from '../apiState'
+import type { CombinedState, QueryCacheKey } from '../apiState'
 import { QueryStatus } from '../apiState'
 import { calculateProvidedByThunk } from '../buildThunks'
 import type {
@@ -29,11 +32,7 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
   refetchQuery,
   internalState,
 }) => {
-  const {
-    removeQueryResult,
-    addPendingTagInvalidations,
-    clearPendingTagInvalidations,
-  } = api.internalActions
+  const { removeQueryResult } = api.internalActions
   const isThunkActionWithTags = isAnyOf(
     isFulfilled(mutationThunk),
     isRejectedWithValue(mutationThunk)
@@ -46,22 +45,23 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
     isRejected(queryThunk)
   )
 
+  let pendingTagInvalidations: FullTagDescription<string>[] = []
+
   const handler: ApiMiddlewareInternalHandler = (action, mwApi) => {
     if (isThunkActionWithTags(action)) {
-      processPendingTagInvalidations(
+      invalidateTags(
         calculateProvidedByThunk(
           action,
           'invalidatesTags',
           endpointDefinitions,
           assertTagType
         ),
-        mwApi,
-        internalState
+        mwApi
       )
     } else if (isQueryEnd(action)) {
-      processPendingTagInvalidations([], mwApi, internalState)
+      invalidateTags([], mwApi)
     } else if (api.util.invalidateTags.match(action)) {
-      processPendingTagInvalidations(
+      invalidateTags(
         calculateProvidedBy(
           action.payload,
           undefined,
@@ -70,57 +70,40 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
           undefined,
           assertTagType
         ),
-        mwApi,
-        internalState
+        mwApi
       )
     }
   }
 
-  function processPendingTagInvalidations(
-    tags: readonly FullTagDescription<string>[],
-    mwApi: SubMiddlewareApi,
-    internalState: InternalMiddlewareState
+  function hasPendingRequests(
+    state: CombinedState<EndpointDefinitions, string, string>
   ) {
-    const rootState = mwApi.getState()
-
-    const state = rootState[reducerPath]
-
-    if (state.config.invalidateImmediately) {
-      handleInvalidatedTags(tags, mwApi)
-      return
-    }
-
-    const hasPendingQueries = Object.values(state.queries).some(
-      (x) => x?.status === QueryStatus.pending
-    )
-    const hasPendingMutations = Object.values(state.mutations).some(
-      (x) => x?.status === QueryStatus.pending
-    )
-
-    if (hasPendingQueries || hasPendingMutations) {
-      if (tags && tags.length > 0)
-        mwApi.dispatch(addPendingTagInvalidations(tags))
-    } else {
-      handleInvalidatedTags(tags, mwApi)
-    }
+    return Object.values({
+      ...state.queries,
+      ...state.mutations,
+    }).some((x) => x?.status === QueryStatus.pending)
   }
 
-  function handleInvalidatedTags(
+  function invalidateTags(
     newTags: readonly FullTagDescription<string>[],
     mwApi: SubMiddlewareApi
   ) {
     const rootState = mwApi.getState()
     const state = rootState[reducerPath]
 
-    const tags = [...state.pendingTagInvalidations, ...newTags]
+    pendingTagInvalidations.push(...newTags)
+
+    if (!state.config.invalidateImmediately && hasPendingRequests(state)) {
+      return
+    }
+
+    const tags = pendingTagInvalidations
+    pendingTagInvalidations = []
     if (tags.length === 0) return
 
     const toInvalidate = api.util.selectInvalidatedBy(rootState, tags)
 
     context.batch(() => {
-      if (state.pendingTagInvalidations.length > 0)
-        mwApi.dispatch(clearPendingTagInvalidations())
-
       const valuesArray = Array.from(toInvalidate.values())
       for (const { queryCacheKey } of valuesArray) {
         const querySubState = state.queries[queryCacheKey]
--- a/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
@@ -84,11 +84,11 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
 
     pendingTagInvalidations.push(...newTags)
 
-    if (!state.config.invalidateImmediately) {
-      const hasPendingRequests = Object.values({
-        ...state.queries,
-        ...state.mutations,
-      }).some((x) => x?.status === QueryStatus.pending)
+    if (state.config.invalidationBehavior === 'delayed') {
+      const hasPendingRequests = [
+        ...Object.values(state.queries),
+        ...Object.values(state.mutations),
+      ].some((x) => x?.status === QueryStatus.pending)
 
       if (hasPendingRequests) return
     }
--- a/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
@@ -75,15 +75,6 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
     }
   }
 
-  function hasPendingRequests(
-    state: CombinedState<EndpointDefinitions, string, string>
-  ) {
-    return Object.values({
-      ...state.queries,
-      ...state.mutations,
-    }).some((x) => x?.status === QueryStatus.pending)
-  }
-
   function invalidateTags(
     newTags: readonly FullTagDescription<string>[],
     mwApi: SubMiddlewareApi
@@ -93,8 +84,13 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
 
     pendingTagInvalidations.push(...newTags)
 
-    if (!state.config.invalidateImmediately && hasPendingRequests(state)) {
-      return
+    if (!state.config.invalidateImmediately) {
+      const hasPendingRequests = Object.values({
+        ...state.queries,
+        ...state.mutations,
+      }).some((x) => x?.status === QueryStatus.pending)
+
+      if (hasPendingRequests) return
     }
 
     const tags = pendingTagInvalidations
--- a/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
@@ -1,4 +1,9 @@
-import { isAnyOf, isFulfilled, isRejectedWithValue } from '../rtkImports'
+import {
+  isAnyOf,
+  isFulfilled,
+  isRejected,
+  isRejectedWithValue,
+} from '../rtkImports'
 
 import type { FullTagDescription } from '../../endpointDefinitions'
 import { calculateProvidedBy } from '../../endpointDefinitions'
@@ -18,20 +23,32 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
   context,
   context: { endpointDefinitions },
   mutationThunk,
+  queryThunk,
   api,
   assertTagType,
   refetchQuery,
   internalState,
 }) => {
-  const { removeQueryResult } = api.internalActions
+  const {
+    removeQueryResult,
+    addPendingTagInvalidations,
+    clearPendingTagInvalidations,
+  } = api.internalActions
   const isThunkActionWithTags = isAnyOf(
     isFulfilled(mutationThunk),
     isRejectedWithValue(mutationThunk)
   )
 
+  const isQueryEnd = isAnyOf(
+    isFulfilled(mutationThunk),
+    isRejected(mutationThunk),
+    isFulfilled(queryThunk),
+    isRejected(queryThunk)
+  )
+
   const handler: ApiMiddlewareInternalHandler = (action, mwApi) => {
     if (isThunkActionWithTags(action)) {
-      invalidateTags(
+      processPendingTagInvalidations(
         calculateProvidedByThunk(
           action,
           'invalidatesTags',
@@ -41,10 +58,10 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
         mwApi,
         internalState
       )
-    }
-
-    if (api.util.invalidateTags.match(action)) {
-      invalidateTags(
+    } else if (isQueryEnd(action)) {
+      processPendingTagInvalidations([], mwApi, internalState)
+    } else if (api.util.invalidateTags.match(action)) {
+      processPendingTagInvalidations(
         calculateProvidedBy(
           action.payload,
           undefined,
@@ -59,7 +76,7 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
     }
   }
 
-  function invalidateTags(
+  function processPendingTagInvalidations(
     tags: readonly FullTagDescription<string>[],
     mwApi: SubMiddlewareApi,
     internalState: InternalMiddlewareState
@@ -68,9 +85,36 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
 
     const state = rootState[reducerPath]
 
+    const hasPendingQueries = Object.values(state.queries).some(
+      (x) => x?.status === QueryStatus.pending
+    )
+    const hasPendingMutations = Object.values(state.mutations).some(
+      (x) => x?.status === QueryStatus.pending
+    )
+
+    if (hasPendingQueries || hasPendingMutations) {
+      if (tags && tags.length > 0)
+        mwApi.dispatch(addPendingTagInvalidations(tags))
+    } else {
+      handleInvalidatedTags(tags, mwApi)
+    }
+  }
+
+  function handleInvalidatedTags(
+    newTags: readonly FullTagDescription<string>[],
+    mwApi: SubMiddlewareApi
+  ) {
+    const rootState = mwApi.getState()
+    const state = rootState[reducerPath]
+
+    const tags = [...state.pendingTagInvalidations, ...newTags]
+
     const toInvalidate = api.util.selectInvalidatedBy(rootState, tags)
 
     context.batch(() => {
+      if (state.pendingTagInvalidations.length > 0)
+        mwApi.dispatch(clearPendingTagInvalidations())
+
       const valuesArray = Array.from(toInvalidate.values())
       for (const { queryCacheKey } of valuesArray) {
         const querySubState = state.queries[queryCacheKey]
--- a/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
@@ -73,6 +73,16 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
     }
   }
 
+  function hasPendingRequests(state: CombinedState<EndpointDefinitions, string, string>) {
+    for (const key in state.queries) {
+      if (state.queries[key]?.status === QueryStatus.pending) return true;
+    }
+    for (const key in state.mutations) {
+      if (state.mutations[key]?.status === QueryStatus.pending) return true;
+    }
+    return false;
+  }
+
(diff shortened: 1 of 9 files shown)


Shrugsy:
Awesome work 👍 

Regarding this behaviour: 

> If a mutation finishes while another query or mutation is running, don't invalidate tags. Instead, invalidate them later after all queries/mutations are settled.

In #3105 you raised the following concern: 

> It could, however, cause problems for people using extremely long-running queries (since no tags are invalidated while they run), so it might make sense to allow disabling this behavior.

I think that is a valid concern, and is a bit of a tradeoff rather than a strict upgrade.

Let me know if I've misinterpreted any of the behaviour in the below example.


e.g.
- Query A has previously run, providing tag `'User'`
- Query B starts, which would provide tag `'Post'`, but takes a long time to resolve (say 3 seconds after the following mutation)
- A mutation is fired & resolves, which intends to invalidate tags `'Post'` & `'User'`

Here is where the behaviour diverges.

**Old behaviour**: Query A is refetched. 3 seconds later, Query B resolves, but does not refetch. It has the potential to resolve with stale data.

**New behaviour**: 3 seconds later, Query A is refetched, Query B resolves, then refetches. If it had stale data, it would be refetched with up to date data.

i.e. under the new behaviour, refetching Query A is unnecessarily delayed until the other query resolves. Just something to consider.

GeorchW:
@Shrugsy Right, will add an option to disable it. Do you think it's fine to enable the new behavior by default? I think that would make sense, since it provides correctness of the query invalidations, and I would consider correctness to be more important than performance in edge cases, but I'm open to other arguments.

phryneas (changes requested):
I like the general idea, but have some change suggestions.
Generally, I think we can slim this down a lot if yet-to-be-invalidated-tags are not saved to a separate slice, but just be kept in a per-store middleware-scoped variable.

I have added quite a lot of comments, I hope they make sense :)

packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -391,6 +392,20 @@ export function buildSlice({
...
 
+  const pendingTagInvalidationsSlice = createSlice({
+    name: `${reducerPath}/pendingTagInvalidations`,
+    initialState: [] as FullTagDescription<string>[],
+    reducers: {
+      addPendingTagInvalidations: (
+        state,
+        action: PayloadAction<readonly FullTagDescription<string>[]>
+      ) => {
+        state.push(...action.payload)
```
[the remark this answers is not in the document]
GeorchW: No, the QueryCacheKeys that are invalidated are deduplicated using a set later: https://github.com/GeorchW/redux-toolkit/blob/delayed-tag-invalidations/packages/toolkit/src/query/core/buildSelectors.ts#L209

Shrugsy:
> @Shrugsy Right, will add an option to disable it. Do you think it's fine to enable the new behavior by default? I think that would make sense, since it provides correctness of the query invalidations, and I would consider correctness to be more important than performance in edge cases, but I'm open to other arguments.

Having a config option and two different versions seems like a bit much

Ideally the behaviour would be roughly:

When a mutation is performed _while a query is still pending_:
1. Invalidate tags immediately, refetching any 'settled' queries that provide those tags
2. _Don't_ clear the "to be invalidated" tags yet, hold onto them
3. When the pending query resolves, if it provides any of the 'held' tags, refetch that query
4. Clear the "to be invalidated" tags

How feasible does it look to achieve step 3 - refetching a specific query if it provides the corresponding tags, WITHOUT doing so by merely invalidating the tags again? Noting that if it was done by invalidating tags again, that would cause an additional refetch of the settled queries

phryneas:
@Shrugsy tbh., I think an ability to turn this on and off might be a good call since it significantly differs from current behavior. It might make sense to evaluate your suggestion, but I feel that the current approach has its merits. It might even be better to refetch all or none to minimize inconsistencies.

The only negative thing I can see with the current approach is that on invalidation, things might not immediately go into a "pending" state, but wait for current pending queries/mutations to finish, so a loading indicator might appear late. 🤔

packages/toolkit/src/query/core/buildSlice.ts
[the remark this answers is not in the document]
phryneas: Ah right, I forgot about that part :)

GeorchW:
@Shrugsy I was thinking about this, too. We would need to store the "pending tag invalidations" per query then, instead of globally. I see the following pros and cons:

- Pro: Long-running queries are not an issue any more
- Con: The implementation as a whole would be significantly more complicated
   - In particular, answering the question "has this query been invalidated by these tags" is something that neither the data structures nor the code is designed for right now -- we are asking "which queries have been invalidated by these tags" instead. We could just compute the entire list of invalidated queries and look if "our" query is in it
- Con: Mutation batching is currently "included" with barely any additional code, which would need a separate implementation (#2203)

I would suggest keeping it simple and see how it's working out in practice, maybe adding a "Troubleshooting" entry that tells people to disable the new behavior if they have issues with long-running queries. If we get bug reports for which disabling the setting is not a viable workaround, we can still implement the "per-query" logic.

GeorchW:
> The only negative thing I can see with the current approach is that on invalidation, things might not immediately go into a "pending" state, but wait for current pending queries/mutations to finish, so a loading indicator might appear late. :thinking:

Ugh, this didn't occur to me, good point.

I *think* the most likely situation for the delay to trigger at all is just a couple milliseconds long, so this issue would be hard to observe. Again, I think it makes sense to see if this is an actual problem in practice before jumping to conclusions.

GeorchW:
@phryneas I've applied a few changes, want do do another round of review?

Related: The implementation changed back and forth a bit due to the suggestions. Should I rebase to clean up the history?

phryneas (dismissed):
This starts to look pretty good :)

I'm still a bit concerned about the need for a third, "combined" mode in the future, so let's directly pave the way for that.

No need to rewrite any commits, since we are going to squash that anyways, but please rebase this whole PR against the [v2.0-integration](https://github.com/reduxjs/redux-toolkit/tree/v2.0-integration) branch.

packages/toolkit/src/query/tests/fetchBaseQuery.test.tsx
```diff
@@ -175,7 +175,7 @@ describe('fetchBaseQuery', () => {
       expect(res.meta?.response).toBeInstanceOf(Object)
       expect(res.error).toEqual({
         status: 'PARSING_ERROR',
-        error: 'SyntaxError: Unexpected token h in JSON at position 1',
+        error: expect.stringMatching(/^SyntaxError: .*$/i),
```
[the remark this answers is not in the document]
GeorchW:
```
$ node
Welcome to Node.js v19.4.0.
Type ".help" for more information.
> JSON.parse("whatever")
Uncaught SyntaxError: Unexpected token 'w', "whatever" is not valid JSON
```

Btw, Deno 1.29.4 gives the same message, as does current Chrome -- I guess it's just the V8 implementation that changed. Firefox/Gecko gives this:

```
SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
```

And Epiphany (WebKit, so likely the same as Safari) does this:

```
SyntaxError: JSON Parse error: Unexpected identifier "whatever"
```

markerikson:
Just to check: is this something we want to change as _default_ behavior? or should it be opt-in?

phryneas:
In 2.0, I could see this as a useful default behavior, yes.

markerikson:
Okay, so maybe we ship this in 1.9.x with it defaulted to the existing behavior, and then flip the default in 2.0.

GeorchW:
Instead of flipping defaults back and forth, I'd suggest simply rebasing this onto 2.0 and release it a couple weeks later.

dorsharonfuse:
Any update on this?
We badly need this...

packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
```diff
@@ -16,19 +21,31 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
...
   const isThunkActionWithTags = isAnyOf(
     isFulfilled(mutationThunk),
     isRejectedWithValue(mutationThunk)
   )
 
+  const isQueryEnd = isAnyOf(
+    isFulfilled(mutationThunk),
+    isRejected(mutationThunk),
+    isFulfilled(queryThunk),
+    isRejected(queryThunk)
```
EskiMojo14, line 42:
```suggestion
    isFulfilled(mutationThunk, queryThunk),
    isRejected(mutationThunk, queryThunk)
```
nitpick

EskiMojo14:
> Any update on this? We badly need this...

it's still marked as having changes requested by @phryneas, though by the looks of it all of those are resolved

other than the nitpick i've added, everything else looks good to me.

i don't know how likely we are to release another 1.9 before 2.0, however - so this may end up being merged into master and then being worked into 2.0's codebase 🤔 that's not something i decide though.

packages/toolkit/src/query/createApi.ts
```diff
@@ -242,6 +252,7 @@ export function buildCreateApi<Modules extends [Module<any>, ...Module<any>[]]>(
       refetchOnMountOrArgChange: false,
       refetchOnFocus: false,
       refetchOnReconnect: false,
+      invalidationBehavior: 'delayed',
```
EskiMojo14, line 255: based on conversations regarding this MR, this new default behaviour is the main sticking point for this PR.

if the default was switched back to the existing behaviour (though obviously keeping the ability to enable the new behaviour), this would very easily fit into another 1.9 release - we could then switch to the new behaviour as a default in 2.0, where changes in behaviour like this could be more expected.

alternatively, this entire change could be held off until 2.0, but considering we don't expect to be releasing 2.0 stable for a good while (realistically, multiple months), that seems like a disservice to the people interested in this change.

GeorchW:
```suggestion
      invalidationBehavior: 'immediately',
```

Okay, so we should just apply the above suggestion and then just merge already?

EskiMojo14: yeah - CI would hopefully check that there aren't any tests etc that would fail with the default being changed back

dorsharonfuse:
I'm wondering if there's any way to be more specific about which query we need to wait to finish before pushing out the pending tag invalidations.

In my own implementation of this (since we're still waiting for this feature), we attached every pending tag invalidation to an endpoint name and args, and our middleware would compare the fulfilled/rejected action to see if it matches the endpoint name and args and only then would invalidate their specific attached tag .

Meaning, if I want to invalidate fetchX once it's done running, I don't want to wait for all other running queries to finish, as they can easily be unrelated, and i'll just end up waiting longer for no good reason.
I'd want to wait only for fetchX to finish.

Any chance of addressing this?
I think waiting for ALL queries to finish could be a lot to ask in some cases...

GeorchW:
@dorsharonfuse 

> I'm wondering if there's any way to be more specific about which query we need to wait to finish before pushing out the pending tag invalidations.

I wondered that too, I don't think it's possible in general. I don't quite understand your approach. For example, you could have this situation:

```
query A starts
query A ends, provides tag Y
query A starts again
mutation B starts
mutation B ends, invalidates tag X
query A ends, provides tag X
```

How do you know that query A is going to provide tag X the next time?

We can of course make the assumption for delayed tag invalidations that a query is going to provide the same tags as last time, but since this is not necessarily correct, I wouldn't make it the default.

> I think waiting for ALL queries to finish could be a lot to ask in some cases...

Tbh, my gut feeling is that in most situations, you won't constantly have those long-running queries, and that the solution should be good enough. Queries usually finish in a couple hundred ms at most, and unrelated mutations are usually triggered by user interaction, which is further apart.

The correctness issue is a bit more pressing IMO, so I'd like to get this merged soon-ish (whenever maintainers react), but if you have a good idea, I'm absolutely open to make another MR or support you with making one

dorsharonfuse:
@GeorchW I get what you mean.
I guess I did kind of assume in my own implementation which tag is going to be provided, since it's a bit static in our case.

dorsharonfuse:
@GeorchW Hey, I recently had to implement a similar mechanism since we have really bad race conditions on our app.
I ended up using sort of 'overriding' the `invalidateTags` action and the mutations' `invalidatesTags` so that whenever tags are about to get invalidated, it checks the affected endpoints using `selectInvalidateBy`.
Then, I check the `QueryStatus` of each of these endpoints/args combos and if any of them are `pending` I push to a pending tags array, simliarly to how you did it.

This allowed to only wait for the relevant queries to finish, instead of ALL queries, which could result in longer delays.

Don't know if that helps you in any way.
Either way, keeping an eye on this PR :)

packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
```diff
@@ -55,12 +71,27 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
...
+    newTags: readonly FullTagDescription<string>[],
     mwApi: SubMiddlewareApi
   ) {
     const rootState = mwApi.getState()
     const state = rootState[reducerPath]
 
+    pendingTagInvalidations.push(...newTags)
+
+    if (state.config.invalidationBehavior === 'delayed') {
+      const hasPendingRequests = [
```
markerikson, line 83: 💬 I have a partial concern here that calling `Object.values()` isn't entirely efficient. Might be an over-micro-optimization concern, but I could foresee that with a lot of queries this could get vaguely expensive.  (Only immediate alternative I have is some `for...in` loops, though.)

packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
```diff
@@ -55,12 +71,27 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
...
+        ...Object.values(state.queries),
+        ...Object.values(state.mutations),
+      ].some((x) => x?.status === QueryStatus.pending)
```
markerikson, line 86: ❓ so the intent here is to delay invalidation until _every_ outstanding request is done, not just ones that might be overlapping?

packages/toolkit/src/query/createApi.ts
```diff
@@ -151,6 +151,16 @@ export interface CreateApiOptions<
...
+  /**
+   * Defaults to `'immediately'`. This setting allows you to control when tags are invalidated after a mutation.
+   *
+   * - `'immediately'`: Queries are invalidated instantly after the mutation finished, even if they are running.
+   *   If the query provides tags that were invalidated while it ran, it won't be re-fetched.
+   * - `'delayed'`: Invalidation only happens after all queries and mutations are settled.
+   *   This ensures that queries are always invalidated correctly and automatically "batches" invalidations of concurrent mutations.
+   *   Note that if you constantly have some queries (or mutations) running, this can delay tag invalidations indefinitely.
+   */
+  invalidationBehavior?: 'delayed' | 'immediately'
```
markerikson, line 163: :question: so this is configured at the full API level, not just per-endpoint?

packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
```diff
@@ -55,12 +71,27 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
...
 
   function invalidateTags(
-    tags: readonly FullTagDescription<string>[],
+    newTags: readonly FullTagDescription<string>[],
     mwApi: SubMiddlewareApi
   ) {
     const rootState = mwApi.getState()
     const state = rootState[reducerPath]
 
+    pendingTagInvalidations.push(...newTags)
```
markerikson, line 80: :hand: This looks like we could end up with the same tags showing up in `pendingTagInvalidations` multiple times, which I think also means we'd end up running some extra logic when we try to invalidate.  Can `pendingTagInvalidations` be a `Set` instead?

markerikson (changes requested):
Okay, _finally_ was able to take a meaningful look at this PR.

Good news is that it's a fairly small set of changes, which makes it easier to review and also suggests we're not radically changing things.

I do have some immediate questions / concerns:

- It looks like we're only able to set the invalidation behavior once for the entire API.  This _feels_ like something we might want to configure on a per-endpoint basis. How feasible is that?
- I _think_ the current logic in the PR could end up with the same tags in `pendingTagInvalidations` multiple times, in which case I think we might try to process them multiple times as well
- I'm not sure I understand why we're trying to wait for _every_ pending query or mutation to complete before doing more work.  At a guess: because any pending query could provide more tags, and any pending mutation could invalidate those tags?

The RTKQ side of things is admittedly not my strong point - I haven't worked on the code as much, and I don't have a deep understanding of data fetching needs the way Lenz does.  But at first glance, something about this _feels_ just a bit off, and I can't pinpoint exactly how.  Like, I think I see the general intent of these changes, but I'm not immediately convinced this is the best answer to the batching/invalidation questions being raised.  (I also don't have _better_ suggestions atm either 🤷‍♂️ .)

@GeorchW : can you at least make a couple tweaks around ensuring we only process each tag once, and see if it even makes sense to specify the behavior per-endpoint?  And we can discuss further from there.

packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
GeorchW: No, since tags are objects and not strings in general. More specifically, their type is 

```ts
{
    type: TagType;
    id?: string | number | undefined;
}
```

The set of invalidated endpoints is already deduplicated in [`selectInvalidatedBy`](https://github.com/GeorchW/redux-toolkit/blob/cfa9fe3c46affd21aef996ce10ec1cf61a76603a/packages/toolkit/src/query/core/buildSelectors.ts#L209). Now, we could try de-duplicating here again by converting these tag objects into canonical string forms, but I think this is creating more overhead than it alleviates except in extreme cases.

packages/toolkit/src/query/createApi.ts
GeorchW: Yes. 

I think this is a sane solution to a rare-ish case. For a use-case where one would note the delay, you would need to have an application where you have very long-running queries constantly, and mutations firing at the same time that they run. This tastes unhealthy for performance to begin with and I would strongly suggest the developer to fix the queries, but if that's not possible there's this escape hatch.

Or maybe I'm wrong. Do you have a particular use case in mind where a per-endpoint setting would have a strong advantage?

packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
GeorchW: Will fix

packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
GeorchW: Hmm. I just wrote it up. It's like 20 lines more, but it introduces quite a bit of complexity and its a bit harder to optimize. I think I'd advocate for keeping it simple and looking if people have a problem with it. (I suspect they won't.)

packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
GeorchW: Oh, and while its true that tags can end up in `pendingTagInvalidations` multiple times, the amount of processing required is no different than if they were invalidated immediately. So this is not a performance regression. (It's not like this code will show up in the profiler anyway.)

markerikson:
Okay. Rebased this against `v2.0-integration`.  Updated a couple bits of test data that needed the new field included.  Went ahead and flipped the default to `"delayed"`.

I'm going to go ahead and merge this now, then put out another 2.0 beta with this included. Thank you for the change, and for your patience!

markerikson (approved):

GeorchW:
Oh, thank you so much!

mjwvb:
Coming here from the 2.0.0-beta.4 releasenotes. Although a delayed solution works for most "perfect" server APIs, our API is far from perfect and has some queries that are taking longer than usual (out of my control unfortunately). A couple of seconds is not an exception for some of our endpoints, with extremes here and there. If I understand correctly we cannot use the delayed behavior and have to fallback to immediate.

I don't know the internals, so only based on gut feeling: why not have a third behavior that does the following?

1. Get all currently running queries when invalidateTags is called
2. Push/merge the tags into the pendingTagInvalidations in the state of each running query
3. After a query finished, check the pendingTagInvalidations and refetch accordingly

Is this what is meant by per-endpoint as stated above? Otherwise I'm curious about @dorsharonfuse's workaround, as it sounds exactly what I need.

markerikson:
@mjwvb : when I said "per-endpoint", I was suggesting that the new `invalidationBehavior` field could be defined at both the `createApi` level and also at the `build.query()` level, in case there was more flexibility needed.

And yeah, @dorsharonfuse , could you share more details / example code of what you did?

dorsharonfuse:
@markerikson What I basically did is utilize the Module system in RTK and implemented an extra module of my own where I overridden the `invalidateTags` action and the mutataions' `invalidatesTags` behavior like so:
`export const pendingInvalidationsModuleName = Symbol('pendingInvalidationsModule');
export type PendingInvalidationsModule = typeof pendingInvalidationsModuleName;

declare module '@reduxjs/toolkit/query' {
	// eslint-disable-next-line @typescript-eslint/no-unused-vars,@typescript-eslint/naming-convention
	export interface ApiModules<BaseQuery extends BaseQueryFn, Definitions extends EndpointDefinitions, ReducerPath extends string, TagTypes extends string> {
		[pendingInvalidationsModuleName]: {
			util: {
				immediatelyInvalidateTags: ActionCreatorWithPayload<TagDescription<ETagType>[]>;
			};
		};
	}
}

export const pendingInvalidationsModule = (): Module<PendingInvalidationsModule> => ({
	name: pendingInvalidationsModuleName,
	init: (
		api
	): {
		injectEndpoint(endpointName: string, definition: EndpointDefinition<any, any, any, any>): void;
	} => {
		const anyApi = api as any as Api<BaseQueryFn, EndpointDefinitions, string, ETagType>;

		const { invalidateTags: immediatelyInvalidateTags, selectInvalidatedBy } = anyApi.util;

		// Override the `invalidateTags` action to push pending invalidations if needed
		const invalidateTags = createAction<PrepareAction<TagDescription<ETagType>[]>>(immediatelyInvalidateTags.type, (tags: TagDescription<ETagType>[]) => {
			const state = store.getState();
			const queriesToInvalidate = selectInvalidatedBy(state, tags);

			// If the tags that we're trying to invalidate have pending queries,
			// we need to wait for them to finish before we can invalidate the tags
			if (hasPendingQueries(state, queriesToInvalidate)) {
				store.dispatch(
					addPendingInvalidation({
						tags,
						endpoints: queriesToInvalidate.map(({ endpointName, originalArgs }) => ({ endpointName, args: originalArgs })),
					})
				);

				return { payload: [], meta: 'Pushed to pending invalidations' };
			}

			return immediatelyInvalidateTags(tags);
		});

		assign(api.util, { invalidateTags, immediatelyInvalidateTags });

		return {
			injectEndpoint(endpoint, definition): void {
				if (!isNil(definition.invalidatesTags)) {
					const endpointInvalidatesTags = definition.invalidatesTags;

					definition.invalidatesTags = (response, error, request, meta): readonly TagDescription<ETagType>[] => {
						let tagsToInvalidate: readonly TagDescription<ETagType>[];

						if (isFunction(endpointInvalidatesTags)) {
							tagsToInvalidate = endpointInvalidatesTags?.(response, error, request, meta) ?? [];
						} else {
							tagsToInvalidate = endpointInvalidatesTags ?? [];
						}

						const state = store.getState();
						const queriesToInvalidate = selectInvalidatedBy(state, tagsToInvalidate);

						// If the tags that we're trying to invalidate have pending queries,
						// we need to wait for them to finish before we can invalidate the tags
						if (hasPendingQueries(state, queriesToInvalidate)) {
							logger.debug('Pushing pending invalidations...', { queriesToInvalidate });

							store.dispatch(
								addPendingInvalidation({
									tags: tagsToInvalidate as TagDescription<ETagType>[],
									endpoints: queriesToInvalidate.map(({ endpointName, originalArgs }) => ({ endpointName, args: originalArgs })),
								})
							);

							return [];
						}

						return tagsToInvalidate;
					};
				}
			},
		};
	},
});
`

Then, I added another middleware that upon a `fulfilled/rejected` action goes over the `pendingInvalidations` array and checks if any of the tags should be invalidated like so:
`export const pendingInvalidationsHandler: Middleware =
	({ dispatch, getState }) =>
	(next) => {
		return (action) => {
			const isQueryEnd = isFulfilled(action) || isRejected(action);
			const pendingInvalidations = selectPendingInvalidations(getState());

			if (isQueryEnd && pendingInvalidations.length) {
				const state = getState();

				// Find any pending invalidations that are relevant to the current query and don't have any more relevant queries that are pending.
				const [relevantInvalidations, otherInvalidations] = partition(pendingInvalidations, ({ endpoints }) => {
					const [relevantEndpoints, otherEndpoints] = partition(
						endpoints,
						({ endpointName, args }) => isEqual(action?.meta?.arg?.endpointName, endpointName) && isEqual(action?.meta?.arg?.originalArgs, args)
					);

					const isInvalidationRelevantForCurrentQuery = relevantEndpoints.length > 0;
					const hasPendingOtherQueries = hasPendingQueries(
						state,
						otherEndpoints.map(({ endpointName, args }) => ({ endpointName, originalArgs: args }))
					);

					return isInvalidationRelevantForCurrentQuery && !hasPendingOtherQueries;
				});

				if (relevantInvalidations.length) {
					const tagsToInvalidate = relevantInvalidations.reduce((result, { tags }) => [...result, ...tags], []);

					// Remove the relevant invalidations from the pending invalidations array
					dispatch(setPendingInvalidations(otherInvalidations));

					// If this is a successful request, maintain the current data in the store and don't update it with the new data
					// since a new request will be made soon anyway, and we don't want to show stale data in the UI.
					if (isFulfilled(action)) {
						const query = state?.api?.queries?.[action?.meta?.arg?.queryCacheKey];

						set(action, 'payload', query?.data);
					}

					// Let the fulfilled/rejected action finish running
					const result = next(action);

					// Invalidate the tags
					dispatch(immediatelyInvalidateTags(tagsToInvalidate));

					return result;
				}

				return next(action);
			}

			return next(action);
		};
	};
`

Note that `immediatelyInvalidateTags` is simply the regular `invalidateTags` from RTK.

Hope I haven't committed any sins by using this module, I haven't this kind of usage anywhere really.
For us, it works as it only waits for relevant queries to finish instead of all of them.

GeorchW:
@mjwvb TBH, it depends on the way your app works. The invalidation behavior only triggers when queries are running while mutations are applied. If your typical use case is like "load a bunch of data and allow the user to edit it, then save it when the user clicks 'save'", then it might be unlikely that queries are still running when the user clicks "save". But yeah, in general, you might suffer delays.

@dorsharonfuse So, if I understand you correctly, you assume that the tags generated by a query are the same in each request?

GeorchW:
@dorsharonfuse ah right, I re-read the top part of the conversation again and we talked about this already a couple months back. 

Maybe we could "officially" support a mode that assumes this: queries always return the same tags for the same args, so we only need to delay invalidation for queries that provided now-invalidated tags last time they ran or that are running for the first time. I think this should not be the default behavior, but it could be useful for anyone with long-running queries and simple tags setup. E.g. with openapi-codegen, the assumption is guaranteed to be true.

We'd need to make sure the documentation on this is clear, though.

minlare:
@GeorchW I'd like to see a mode that supports the above, as openapi-codegen is exactly the scenario I have.

PR #3120, opened 2023-01-25
fix docs rtk-query/usage/queryies typo

knyga:
image
2nd argument has been described


markerikson:
Hi. Appreciate the PR, but based on the examples it looks like it _should_ be "2nd".  Thanks anyway!

PR #3134, opened 2023-01-29
Update Examples to use React Router 6.8 Patterns

seanmcquaid:
I have updated all of the examples that use react-router-dom to use the latest current version (6.8) along with some pattern updates that reflect React Router's docs.

In the linked issue there is mention of a desire for an example that use's the new loaders pattern from react-router-dom, should I also include that example in this PR?


seanmcquaid:
Not totally certain why the build is failing. My assumption is that bumping the examples to 6.8 caused some other inherent dependency issues?

PR #3146, opened 2023-02-01
Update rtk-query redux-persist Example

sambolias:
Found that when rehydration is set up per the existing docs there is a runtime error when trying to load first time, or after clearing storage. Optionally accessing reducer path seems to work well, and was published on stack overflow https://stackoverflow.com/a/72959139. Updating these docs should save others the trouble I had.


kasparkallas:
Trying out `redux-persist`, the store seems to rehydrate without specifing any `extractRehydrationInfo` logic. Do you know if there's any difference whether I call it or not?

phryneas:
The job of `extractRehydrationInfo` here is to remove data that is otherwise restored "too much", so yes, you need it.

PR #3147, opened 2023-02-01
Bug: cache invalidation of queries that use serializeQueryArgs

SebastienGllmt:
If you specify serializeQueryArgs it seems the queries don't get invalidated properly when you call invalidateTags.

Notably, the endpoint does get invalidated and rerun as expected, but the arguments passed in are the old arguments when the cache key was created and not what they are after the invalidation

I added a test that shows this bug in this PR, but I don't have a fix for it.

Here is pseudo code if you prefer that:
```typescript
getFoo: {
    query: request => {
       console.log(request); // this will always print the `request` value when initially cached
    },
    serializeQueryArgs: ({ queryArgs, endpointDefinition, endpointName }) => {
        return defaultSerializeQueryArgs({
          endpointName,
          queryArgs: {
            foo: queryArgs.foo,
          },
          endpointDefinition,
        });
    },
    providesTags: ["FooTag"]
}

query({ foo: 5, bar: 10})
invalidateTags(["FooTag"])
query({ foo: 5, bar: 15}) // this triggers the rtk query, but "request" still has bar=10
```

### DIFF
--- a/packages/toolkit/src/query/tests/createApi.test.ts
+++ b/packages/toolkit/src/query/tests/createApi.test.ts
@@ -876,8 +876,13 @@ describe('custom serializeQueryArgs per endpoint', () => {
     },
   }
 
+  const TagTypes = {
+    ListItemsTag: 'ListItemsTag',
+  } as const;
+
   const api = createApi({
     baseQuery: fetchBaseQuery({ baseUrl: 'https://example.com' }),
+    tagTypes: Object.values(TagTypes),
     serializeQueryArgs: ({ endpointName, queryArgs }) =>
       `base-${endpointName}-${queryArgs}`,
     endpoints: (build) => ({
@@ -943,6 +948,16 @@ describe('custom serializeQueryArgs per endpoint', () => {
         forceRefetch({ currentArg, previousArg }) {
           return currentArg !== previousArg
         },
+      }),
+      listItems3: build.query<string[], number>({
+        query: (pageNumber) => {
+          console.log(pageNumber);
+          return `/listItems3?page=${pageNumber}`
+        },
+        serializeQueryArgs: ({ endpointName }) => {
+          return endpointName
+        },
+        providesTags: [TagTypes.ListItemsTag],
       }),
     }),
   })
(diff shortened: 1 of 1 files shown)


SebastienGllmt:
Debugging this a bit more, it seems like the following is happening:

1. First query runs with `bar: 10` and the result is cached
2. `invalidateTags` invalidates the query, but the invalidation causes the query to rerun right away using `bar: 10` again and that gets cached
3. When calling with `bar: 15`, nothing is run because it simply pulls from the cache created in (2)

SebastienGllmt:
I tried looking into adding something like

```typescript
if ('serializeQueryArgs' in endpointDefinitions[endpointName]) {
  mwApi.dispatch(
    removeQueryResult({
      queryCacheKey: queryCacheKey as QueryCacheKey,
    })
  )
}
```

inside the `invalidateTags` function, but I fear there is no way to get this to work properly given other parts of the library expect the `refetchQuery` to execute as expected inside `invalidateTags` if there is a subscriber.

Maybe you can think of a way to solve this, but I also feel like maybe my use-case of having arguments that are only important enough to trigger a re-query after invalidation may be an abuse of API. Instead, it may make more sense to mention this behavior in the docs and recommend people to instead use [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#caching)-based caching if you want to get a unique ID to put inside your `serializeQueryArgs` that either can't be serialized (or is too big to be serializable)

packages/toolkit/src/query/tests/createApi.test.ts
```diff
@@ -1083,4 +1098,35 @@ describe('custom serializeQueryArgs per endpoint', () => {
...
+
+    const initialEntry = selectListItems(storeRef.store.getState())
+    expect(initialEntry.data).toEqual(['a', 'b', 'c'])
+
+    await storeRef.store.dispatch(api.util.invalidateTags([TagTypes.ListItemsTag]));
+
+    // TODO: due to bug, the call to listItems3 still uses `1` instead of `2` as expected
+    await storeRef.store.dispatch(api.endpoints.listItems3.initiate(2))
+    const updatedEntry = selectListItems(storeRef.store.getState())
+    expect(updatedEntry.data).toEqual(['a', 'b', 'c', 'd', 'e', 'f'])
```
SebastienGllmt, line 1130: This is a copy-paste mistake and should have been `'d', 'e', 'f'`

phryneas:
I'm just skimming over this, but couldn't this be solved here?
https://github.com/reduxjs/redux-toolkit/blob/0ce2ec02b079a6dedec2b6d5206c56ef6eebb653/packages/toolkit/src/query/core/buildSlice.ts#L161-L163

SebastienGllmt:
> recommend people to instead use [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#caching)-based caching if you want to get a unique ID to put inside your serializeQueryArgs

I tried this and unfortunately this isn't an option either. This is because if you have `JSON.stringify(queryArgs1) === JSON.stringify(queryArgs2)` but `cache(queryArgs1) !== cache(queryArgs2)` it will cause your query to be stuck forever.

It seems this is because in `useStableQueryArgs`, we have
```typescript
useEffect(function () {
        if (cache.current.serialized !== incoming.serialized) {
            cache.current = incoming;
        }
    }, [incoming]);
```

and if since serialized is the JSON stringification of queryArgs and not the cache key, `cache.current` will never update.

SebastienGllmt:
Alright, yet another attempt at this: use `forceRefetch` on the endpoint to force a refetch when `bar: 15` even if the cache key is different

```typescript
forceRefetch: ({ currentArg, previousArg }) => { 
    return currentArg !== previousArg
}
```

This works in practice, but doesn't work in the test case because `forceRefetch` isn't even called during the `query({ foo: 5, bar: 15})` query. It looks like this is because of the following line in buildThunks

```typescript
// Don't retry a request that's currently in-flight
if (requestState?.status === 'pending') {
    return false
}
```

markerikson:
@SebastienGllmt : thanks for the PR! I'll try to look at it "soon" (although I've got other things going on the next few days).

That said, can you clarify what actual use cases you're trying to solve atm that led to this?

SebastienGllmt:
We have an RTK Query where the input to the query is large enough that the serialization logic to create the cache key affects performance, so we want to remove it from the cache. The combination of `serializeQueryArgs` and `forceRefetch` seems to have done the trick for our application because now we can keep the cache key short, yet still refetch the data when necssary

When I created this PR initially, I didn't realize I could use `forceRefetch` for this, so I was trying to keep the cache key short and manually trigger refetching by invalidating tags

markerikson:
Gotcha.

Hmm. I wonder if we could do some basic `WeakMap`-based caching of args for `defaultSerializeQueryArgs`?

markerikson:
@SebastienGllmt : could you do me a favor and either generate a JS devtools CPU perf of your app (can send it to me privately on Discord if you want), or put together a quick example that shows the stringification taking up time?  

Also, could you try out the CodeSandbox CI build from #3193 and see if that helps improve perf for your app?

PR #3158, opened 2023-02-07
docs(rtk-query): Add missing keys to `BaseQueryApi`

petermekhaeil:
The `BaseQueryApi` type is missing keys in the [documentation](https://redux-toolkit.js.org/rtk-query/api/createApi#basequery). I've added a short description and I am happy to reword it if anyone has better suggestions.

Thank you for all the hard work you've done on redux-toolkit!


markerikson:
Thanks!

PR #2901, opened 2022-11-12
Feature/add type parameter for id in create entity adapter

Matt-Ord:
Closes #2098
Re making the pr #2099 after was closed

PR #3187, opened 2023-02-18
Add Id type parameter in createEntityAdapter

Matt-Ord:
Closes #2099 

Redo of #2901 for the 2.0 branch


Matt-Ord:
Yarn tsc is failing for old Typescript versions, but I think it has nothing to do with the changes I've made?

markerikson:
Probably best to rebase this against `v2.0-integration` instead of merging

Matt-Ord:
@markerikson I've rebased, the type tests are failing due to something unrelated to this PR. Is there an easy way to fix it?

Matt-Ord:
@markerikson I've rebased - any chance this could be merged soon?

packages/toolkit/src/tests/injectableCombineReducers.example.ts
EskiMojo14, line 1: could you remove this file? it was deleted once combineSlices was fully implemented 🙂

packages/toolkit/src/entities/unsorted_state_adapter.ts
```diff
@@ -158,7 +161,7 @@ export function createUnsortedStateAdapter<T>(
         0
 
       if (didMutateIds) {
-        state.ids = Object.keys(state.entities)
+        state.ids = Object.keys(state.entities) as Id[]
```
EskiMojo14, line 164: hmm - this isn't actually a safe assertion.

Object.keys *always* returns string[], even if numbers were used as keys:
image

not entirely sure how to handle this better though

Matt-Ord: Fixed for now. I think the 'proper' way is to enforce ids to be strings only - as that is what is actually going on here. Not sure what @markerikson thinks about that, but the current way clearly has some issues.

markerikson:
Thanks, and sorry for the delay on this!

PR #3193, opened 2023-02-20
Use a WeakMap cache for query arg serialization for perf

### LINKED ISSUE #3153: Investigate caching of args for `defaultSerializeQueryArgs`
Per https://github.com/reduxjs/redux-toolkit/pull/3147#issuecomment-1415011755 :

> We have an RTK Query where the input to the query is large enough that the serialization logic to create the cache key affects performance...
> We're still getting some performance hit because it looks like RTK Query does some JSON stringification of the arguments internally even if it's not part of the cache key, but this is manageable (about 50ms of our app startup is spent on this stringification of arguments)

I wonder if we could do some basic caching of the serialization, maybe by adding a `WeakMap` usage to `defaultSerializeQueryArgs`?

markerikson:
Per https://github.com/reduxjs/redux-toolkit/pull/3147#issuecomment-1415011755 :

> We're still getting some performance hit because it looks like RTK Query does some JSON stringification of the arguments internally even if it's not part of the cache key, but this is manageable (about 50ms of our app startup is spent on this stringification of arguments)

This hopefully addresses that issue by using a `WeakMap` to cache results in `defaultSerializeQueryArgs`.

Fixes #3153 .

### DIFF
--- a/packages/toolkit/src/query/defaultSerializeQueryArgs.ts
+++ b/packages/toolkit/src/query/defaultSerializeQueryArgs.ts
@@ -2,21 +2,38 @@ import type { QueryCacheKey } from './core/apiState'
 import type { EndpointDefinition } from './endpointDefinitions'
 import { isPlainObject } from '@reduxjs/toolkit'
 
+const cache: WeakMap<any, string> | undefined = WeakMap
+  ? new WeakMap()
+  : undefined
+
 export const defaultSerializeQueryArgs: SerializeQueryArgs<any> = ({
   endpointName,
   queryArgs,
 }) => {
+  let serialized = ''
+
+  const cached = cache?.get(queryArgs)
+
+  if (typeof cached === 'string') {
+    serialized = cached
+  } else {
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/query/defaultSerializeQueryArgs.ts
```diff
@@ -2,21 +2,38 @@ import type { QueryCacheKey } from './core/apiState'
...
+        ? Object.keys(value)
+            .sort()
+            .reduce<any>((acc, key) => {
+              acc[key] = (value as any)[key]
+              return acc
+            }, {})
+        : value
+    )
+    if (isPlainObject(queryArgs)) {
+      cache?.set(queryArgs, stringified)
```
SebastienGllmt, line 31: Makes sense. There is a slight worry that two objects with the same shape will get two different entries in the map since the map is based off memory address and not value, but this shouldn't blow up in memory since I would assume old entries can get garbage collected

PR #3207, opened 2023-02-24
Allow inference of enhancer state extensions, and fix inference when using callback form

### LINKED ISSUE #3206: Enhancer types are lost when using callback form, and state extensions not reflected
#2550 made a great start on getting enhancers typed properly! 

Unfortunately, the inference still wasn't there when using the callback form, e.g.

```ts
const store = configureStore({ reducer: () => 0, enhancers: (defaultEnhancers) => [...defaultEnhancers, myEnhancer] });
```

The above wouldn't pick up any store extensions.

Additionally, the StoreEnhancer type has a `StateExt` generic that's not used by RTK - presumably it would make sense for this type to be reflected in the final getState return.

EskiMojo14:
Fixes #3206

### DIFF
--- a/docs/api/configureStore.mdx
+++ b/docs/api/configureStore.mdx
@@ -18,7 +18,7 @@ to the store setup for a better development experience.
 
 ```ts no-transpile
 type ConfigureEnhancersCallback = (
-  defaultEnhancers: StoreEnhancer[]
+  defaultEnhancers: EnhancerArray<[StoreEnhancer]>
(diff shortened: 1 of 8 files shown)


markerikson:
Tagging @Methuselah96 - curious how much this overlaps with the 2.0 work you're doing?

Methuselah96:
There's no overlap with my 2.0 work besides the fact that they both involve enhancer types. I was able to integrate them into the v2.0 with a few minor changes.

PR #3211, opened 2023-02-25
fix: Correct build outputs

### LINKED ISSUE #3208: [RTK Alpha]: Invalid exports config for CJS consumption w/ Node
I skimmed the issues but couldn't find anything matching, so apologies if it's a duplicate and/or intentional.

The current config in the Alpha does not allow for consumption of the CJS bundle in modern version of Node/any tool that follows Node's module resolution spec, as it uses `.js` to refer to a CJS module despite the package setting `"type": "module"`. If `"type": "module"` is set, `.cjs` is necessary in `"exports"`.

Some bundlers do work around this and are more forgiving (whether or not that's a good thing is something else entirely), but this config will not work in Node and/or any environments that follow its resolution mechanism.

To reproduce:

```js
// index.cjs
const { createSlice } = require('@reduxjs/toolkit');
console.log(typeof createSlice);
```

Attempting to execute this file (`node index.cjs`) will result in the following expected error: 

```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /home/ryun/Projects/example/node_modules/redux-thunk/lib/index.js
require() of ES modules is not supported.
require() of /home/ryun/Projects/example/node_modules/redux-thunk/lib/index.js from /home/ryun/Projects/example/node_modules/@reduxjs/toolkit/dist/cjs/redux-toolkit.cjs.development.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /home/ryun/Projects/example/node_modules/redux-thunk/package.json.
```

Generally, it should be pretty safe to use `.cjs` in `"exports"`. Both are relatively new, and any environment that supports consuming from `"exports"` should support `.cjs`. This may mean you'll want to duplicate your CJS bundles, however (provide as `.js` for `"main"` and the spec-adherent `.cjs` for `"exports"`), assuming you're not opposed to larger install size.

rschristian:
Closes #3208

### DIFF
--- a/packages/toolkit/package.json
+++ b/packages/toolkit/package.json
@@ -109,6 +109,7 @@
   },
   "files": [
     "dist/**/*.js",
+    "dist/**/*.mjs",
     "dist/**/*.js.map",
     "dist/**/*.d.ts",
     "dist/**/package.json",
--- a/packages/toolkit/package.json
+++ b/packages/toolkit/package.json
@@ -22,25 +22,24 @@
   "publishConfig": {
     "access": "public"
   },
-  "type": "module",
-  "module": "dist/redux-toolkit.modern.js",
   "main": "dist/cjs/index.js",
+  "module": "dist/redux-toolkit.modern.js",
   "types": "dist/index.d.ts",
   "exports": {
     "./package.json": "./package.json",
     ".": {
       "types": "./dist/index.d.ts",
-      "import": "./dist/redux-toolkit.modern.js",
+      "import": "./dist/redux-toolkit.modern.mjs",
       "default": "./dist/cjs/index.js"
     },
     "./query": {
       "types": "./dist/query/index.d.ts",
-      "import": "./dist/query/rtk-query.modern.js",
+      "import": "./dist/query/rtk-query.modern.mjs",
       "default": "./dist/query/cjs/index.js"
     },
     "./query/react": {
       "types": "./dist/query/react/index.d.ts",
-      "import": "./dist/query/react/rtk-query-react.modern.js",
+      "import": "./dist/query/react/rtk-query-react.modern.mjs",
       "default": "./dist/query/react/cjs/index.js"
     }
   },
--- a/packages/toolkit/scripts/build.ts
+++ b/packages/toolkit/scripts/build.ts
@@ -146,7 +146,7 @@ async function bundle(options: BuildOptions & EntryPointOptions) {
   }
 
   const outputFolder = path.join(...folderSegments)
-  const outputFilename = `${prefix}.${name}.js`
+  const outputFilename = `${prefix}.${name}.${format === 'cjs' ? 'js' : 'mjs'}`
 
   await fs.ensureDir(outputFolder)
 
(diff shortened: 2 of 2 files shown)


rschristian:
Came back to fix up any remaining test issues (sorry for the long delay), but looks like it was just an issue with my Node version. Didn't realize the project targets v16+, so nullish coalescing syntax errors aren't actually an issue.

I'll mark this as ready for review, but feel free to copy/paste or fold into anything else, obviously this change is quite tiny and is just implementing what we talked about in the linked issue.

markerikson:
The build errors all say it can't find the `@reduxjs/toolkit` entry point.  So, most likely something about this change _is_ breaking the package definition :(

rschristian:
> The build errors all say it can't find the `@reduxjs/toolkit` entry point. So, most likely something about this change _is_ breaking the package definition :(

Oh god dammit. It's the `"files"` in the `package.json`, the `.mjs` entries aren't white listed.

I tested with [`yalc`](https://npm.im/yalc), but as the issue was with the CJS specifically, I never switched to test the ESM

markerikson:
OOOOO-kay!

After a _very_ long day today (and a bunch more hours on previous weekends), I have _finally_ come up with a set of packaging changes that appears to be as "correct" as I think we're gonna get, and I've got it being verified via a battery of CI tests that build and run a set of example apps using the PR build artifact.  There's also a CI test that uses `are-the-types-wrong` to verify that the entry points actually seem to work as expected.

I've published new alphas of `redux` and `redux-thunk` using the package config you have here as a starting point.

Here's what I ended up with for `redux`:

```ts
  "main": "dist/cjs/redux.cjs",
  "module": "dist/redux.legacy-esm.js",
  "types": "dist/redux.d.ts",
  "exports": {
    "./package.json": "./package.json",
    ".": {
      "types": "./dist/redux.d.ts",
      "import": "./dist/redux.mjs",
      "default": "./dist/cjs/redux.cjs"
    }
  },
```

Note the use of `module` with a `.js` extension for Webpack 4 compat.  I found that one out the hard way - the `cra4` test example failed to find imports when RTK and Redux had a `.mjs` file in the `module` key :(

Similarly, for RTK:

```ts
  "module": "dist/redux-toolkit.legacy-esm.js",
  "main": "dist/cjs/index.js",
  "types": "dist/index.d.ts",
  "exports": {
    "./package.json": "./package.json",
    ".": {
      "types": "./dist/index.d.ts",
      "import": "./dist/redux-toolkit.modern.mjs",
      "default": "./dist/cjs/index.js"
    },
    "./query": {
      "import": "./dist/query/rtk-query.modern.mjs",
      "default": "./dist/query/cjs/index.js"
    },
    "./query/react": {
      "types": "./dist/query/react/index.d.ts",
      "import": "./dist/query/react/rtk-query-react.modern.mjs",
      "default": "./dist/query/react/cjs/index.js"
    }
  },
```

And then I'm still having to use the fallback `query/package.json`-type files for Webpack 4 as well.

Those changes are over in #3318 , which I'm about to merge.

Thank you for helping sort out what the _apparently_-right keys are and pointing me in the right direction!

PR #3239, opened 2023-03-07
Add `createApiProvider` function to allow for `configureStore` options to be set for `ApiProvider`

Harris-Miller:
In RTK-query, the included `<ApiProvider>` does not allow for any customization of the `configureStore()` that is created in a `useState()`

This MR wraps the `ApiProvider` in a factory function that accepts `Omit<ConfigureStoreOptions, 'reducer'>` to allow for that customization.

I'm particularly interested in allowing for `devTools` to be set to `process.env.NODE_ENV === 'production'`, a general use-case, but I figured that I would allow for all `ConfigureStoreOptions` (except `reducer` of course)

I choose to do this as a factory function since the `ConfigureStoreOptions` I don't see as something that should be mutable via react-props. It's something you set up in advance to be shared across usages of the resultant `<ApiProvider>`

the previous `ApiProvider` is now simply `export const ApiProvider = createApiProvider({});`

### DIFF
--- a/packages/toolkit/src/query/react/ApiProvider.tsx
+++ b/packages/toolkit/src/query/react/ApiProvider.tsx
@@ -8,19 +8,27 @@ import { setupListeners } from '@reduxjs/toolkit/query'
 import type { Api } from '@reduxjs/toolkit/dist/query/apiTypes'
 import type { ConfigureStoreOptions } from '../../configureStore';
 import { configureStore } from '../../configureStore'
-import type { ThunkMiddlewareFor } from '../../getDefaultMiddleware'
+import type { CurriedGetDefaultMiddleware, ThunkMiddlewareFor } from '../../getDefaultMiddleware'
 
 // copied from configureStore
 type Middlewares<S> = ReadonlyArray<Middleware<{}, S>>
 
 type Enhancers = ReadonlyArray<StoreEnhancer>
 
+/**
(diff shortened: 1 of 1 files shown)


phryneas:
To be honest, I'm not so sure what to make of this.  
`ApiProvider` is a shortcut that we provide if you want the most basic setup possible - it's biggest use is not in real applications but as a helper for people wanting to get set up quickly when following a tutorial.  
As soon as you want to configure anything, we recommend setting up your store using `configureStore` and the normal `Provider`. When you start implementing (or even importing) a `createApiProvider`, you are already reaching a level of complexity where you could also just do the manual setup.  
In that light, I'm not really in favor of adding this; I'm sorry :/

On a side note on that `devTools` `production` check: if someone wants to look into the internals of your Redux store, they will always be able to do so. Turning off the dev tools just makes it more complicated for you to debug things, but it will not lock anyone out. For that reason, we usually recommend not turning them off.

Harris-Miller:
> As soon as you want to configure anything, we recommend setting up your store using configureStore and the normal Provider

The use case that I had is for MFEs and isolating a few different RTK-Querys, but hearing that said out loud, I realize that I can do that for my use case as well, seems very obvious actually

phryneas:
It's great that that works as a solution for you!

I'm sorry that the PR didn't fit in, and I still very much appreciate the work you put in here!

Issue #3251, opened 2023-03-10
When adding new entities into a query's cache using updateQueryData, "api.provided" isn't updated (UPDATED)


dutziworks:
When calling `updateQueryData`, is there a way to update the provided tags of the query whose cache was updated?

Our optimistic update needs to opt. CREATE an entity (that didn’t exist before) and add it into a query’s cache. 

That new entity is now inside the cache, but not listed inside `api.provided`, since it didn’t go through the “regular” flow (fetched through a list/get endpoint). 

Is there a way to also update that list/get endpoint’s provided tags?

dutziworks:
Looking at [the source](https://github.com/reduxjs/redux-toolkit/blob/a636751f12e4272391c32e7b3672a6d65a59b5d0/packages/toolkit/src/query/core/buildSlice.ts#L319) it seems like there isn't a way to update the provided slice (or "invalidation slice").

So to achieve our goal, we'll probably have to maintain an "extendedProvided" slice, shape it as RTKQ's provided slice and update that one instead, which will work, but feels a bit weird...

The reason we need to update the provided tags of a (list) query is that we later use that data to "know" which queries to optimistically update later on, when mutations occur.

Our app is built in a way where all the data is fetched once, and from that point on, no invalidations ever occur. Everything is optimistically/pessimistically updated. We also set up DB subscriptions for all "active" queries (all mounted `useList/Get...Query()` hooks) so that if other users make changes to the DB everyone gets synced.

Will it be accepted if I make a PR that adds 2 case-reducers into `invalidationSlice`: `providedTagAdded` and `providedTagRemoved`, which accept somethings like `{ tagType: TagTypes, id: EntityId, cacheKey: QueryCacheKey }`, and adds/remove those entities from the provided slice for that tagType/cacheKey.

phryneas:
You are perfectly right, that should never have been missing.

Unfortunately, I just don't have the capacity to see this through myself right now fully - but I would be very open to a PR to add `provided` calculation to `updateQueryData`.
I have taken a short stab at this, and it seems plausible - so you would not start from zero, this is an outline on how such a fix would look like: https://github.com/reduxjs/redux-toolkit/commit/3c34511f2893b909e68f35bb0b69090db719839a

Of course, this doesn't come without it's own problems: This would call `providedBy` with a different `meta` than usually, so it has to be toggleable if `providedBy` should be called in an `updateQueryData` case.
[...]

PR #3255, opened 2023-03-12
add option to update provided tags

dutzi:
See #3251 for context.

### DIFF
--- a/docs/rtk-query/api/created-api/api-slice-utils.mdx
+++ b/docs/rtk-query/api/created-api/api-slice-utils.mdx
@@ -29,7 +29,8 @@ Some of the TS types on this page are pseudocode to illustrate intent, as the ac
 const updateQueryData = (
   endpointName: string,
   args: any,
-  updateRecipe: (draft: Draft<CachedState>) => void
+  updateRecipe: (draft: Draft<CachedState>) => void,
+  updateProvided?: boolean
 ) => ThunkAction<PatchCollection, PartialState, any, AnyAction>;
 
 interface PatchCollection {
@@ -43,6 +44,7 @@ interface PatchCollection {
   - `endpointName`: a string matching an existing endpoint name
   - `args`: an argument matching that used for a previous query call, used to determine which cached dataset needs to be updated
   - `updateRecipe`: an Immer `produce` callback that can apply changes to the cached state
+  - `updateProvided`: a boolean indicating whether the endpoint's provided tags should be re-calculated based on the updated cache. Defaults to `false`.
 
 #### Description
 
@@ -155,7 +157,8 @@ await dispatch(
 const patchQueryData = (
   endpointName: string,
   args: any
-  patches: Patch[]
+  patches: Patch[],
+  updateProvided?: boolean
 ) => ThunkAction<void, PartialState, any, AnyAction>;
 ```
 
@@ -163,6 +166,7 @@ const patchQueryData = (
   - `endpointName`: a string matching an existing endpoint name
   - `args`: a cache key, used to determine which cached dataset needs to be updated
   - `patches`: an array of patches (or inverse patches) to apply to cached state. These would typically be obtained from the result of dispatching [`updateQueryData`](#updatequerydata)
+  - `updateProvided`: a boolean indicating whether the endpoint's provided tags should be re-calculated based on the updated cache. Defaults to `false`.
 
 #### Description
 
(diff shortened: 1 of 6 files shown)


packages/toolkit/src/query/core/buildThunks.ts
```diff
@@ -236,41 +237,58 @@ export function buildThunks<
...
+        api.internalActions.queryResultPatched({ queryCacheKey, patches })
+      )
+
+      if (!updateProvided) {
+        return
+      }
+
+      const newValue = api.endpoints[endpointName].select(args)(getState())
+
+      const providedTags = calculateProvidedBy(
```
phryneas, line 259: Hmm, I see the merit of "we have only one code path" for moving this into `patchQueryData`. 
Still, it has a significant downside: it will always dispatch two actions, whereas the approach I had suggested dispatches only one for an update and two for `undo` (which should happen far less often).

Every of those dispatches *could* result in a rerender - and as it stands, we cannot mark them as `batchable` since we need to read the updated state immediately after dispatch.

In that light, could you please keep it with the "mostly 1 dispatch, sometimes 2" approach?

phryneas: That was a brainfart, of course we can mark them batchable - batching will not delay reducer execution, but only the resulting rerender.

In that case, could you please make sure that `queryResultPatched` and `updateProvidedBy` are marked with `prepareAutoBatched` ?

phryneas:
Apart from that: looking good! Could you please also add tests for this functionality?

dutzi:
Sure, I'll add tests soon

dutzi:
It seems like TS 4.1  complains about something, but when I tried checking out the commit (6a7879f) previous to yours and then mimicking that test, I still get that error:

image

Other than that, I think I'm done!

phryneas (approved):
Don't sweat the TS4.1 test too much - we'll drop support for that with the next major or minor.

Generally, this would probably ship with our next minor with a default of `updateProvided=false`, or we could even consider `updateProvided=true` as default, but then it would need to go into the next major as it is technically breaking (and we should probably offer a codemod).

As it looks right now, we'll probably only do a 2.0 next anyways, so we could consider going that route.
Either way, this will take some time until we ship it.

Meanwhile, you can use the build from this PR via your package manager by following the instructions from CI on this page: https://ci.codesandbox.io/status/reduxjs/redux-toolkit/pr/3255 (If you do, please report back if everything works for you as expected.)

As for this PR: please don't be discouraged if it is open for a bit - we usually merge PRs shortly before releases, so this could stay open for a while. I intend to give this some more local testing, and at this point, I am pretty heavily backlogged - I'll need to reserve some specific time for RTK work at some point.

docs/rtk-query/api/created-api/api-slice-utils.mdx
```diff
@@ -155,14 +157,16 @@ await dispatch(
...
+  patches: Patch[],
+  updateProvided?: boolean
 ) => ThunkAction<void, PartialState, any, AnyAction>;
 ```
 
 - **Parameters**
   - `endpointName`: a string matching an existing endpoint name
   - `args`: a cache key, used to determine which cached dataset needs to be updated
   - `patches`: an array of patches (or inverse patches) to apply to cached state. These would typically be obtained from the result of dispatching [`updateQueryData`](#updatequerydata)
+  - `updateProvided`: a boolean indicating whether the endpoint's provided tags should be re-calculated based on the updated cache. Defaults to `false`.
```
phryneas, line 169: Big :heart: for the docs commit!

dutziworks:
Great!

Thanks for helping me out and guiding me through this process :)

packages/toolkit/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@reduxjs/toolkit",
-  "version": "1.9.2-11",
+  "version": "1.9.1",
```
markerikson: :hand: Not sure why this has been changed - any changes here should be reverted

dutziworks: test

dutzi (commented):
Sorry for the late response, I didn’t know I needed to submit the review for my comments to become visible.

packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -319,7 +320,36 @@ export function buildSlice({
...
+          const subscribedQueries = ((draft[type] ??= {})[
+            id || '__internal_without_id'
+          ] ??= [])
+          const alreadySubscribed = subscribedQueries.includes(queryCacheKey)
+          if (!alreadySubscribed) {
+            subscribedQueries.push(queryCacheKey)
+          }
+        }
+      },
+    },
```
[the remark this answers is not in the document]
markerikson: Yeah, was re-reviewing this and was about to ask about copy-pasted code, but calling the case reducer again is a nice trick.

markerikson:
Okay, we'll give this a shot.

PR #3256, opened 2023-03-12
Experiment with createDynamicMiddleware idea

EskiMojo14:
see #3252 - opening PR to get a package to play with

### DIFF
--- a/packages/toolkit/api-extractor-react.json
+++ b/packages/toolkit/api-extractor-react.json
@@ -1,0 +1,356 @@
+/**
+ * Config file for API Extractor.  For more info, please visit: https://api-extractor.com
+ */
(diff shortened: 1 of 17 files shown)


EskiMojo14:
having a dist/react folder seems to cause interference with RTKQ's 'react' imports - as suggested by @markerikson I've tried this on the v2.0 integration branch and this no longer occurs. As such I think this MR will need to be rebuilt to aim at that branch, so a @reduxjs/toolkit/react entry point can be added cleanly.

Technically the main part of this is not react specific, but I think a large reason to add this instead of just using existing solutions is *for* the ability to get a useDispatch hook that knows about injected middleware's type signature.

PR #3259, opened 2023-03-13
Merge Node CI tests and re-enable Node ESM checks

markerikson:


markerikson:
I _think_ this error is a good thing - it seems to be correctly identifying that there's an ESM mismatch in RTK's package setup, and I really should try removing `type: "module"`:

image

markerikson:
And removing `type: "module"` from RTK causes errors the _other_ way with `redux-thunk`, I think:


> Error [ERR_REQUIRE_ESM]: `require()` of ES Module `/home/runner/work/redux-toolkit/redux-toolkit/examples/publish-ci/node-standard/node_modules/redux-thunk/lib/index.js` from `/home/runner/work/redux-toolkit/redux-toolkit/examples/publish-ci/node-standard/node_modules/@reduxjs/toolkit/dist/cjs/redux-toolkit.cjs.development.js` not supported.
> 
> `index.js` is treated as an ES module file as it is a `.js` file whose nearest parent package.json contains "type": "module" which declares all `.js `files in that package scope as ES modules.
> 
> Instead rename index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change `"type": "module"` to `"type": "commonjs"` in `/home/runner/work/redux-toolkit/redux-toolkit/examples/publish-ci/node-standard/node_modules/redux-thunk/package.json` to treat all `.js` files as CommonJS (using `.mjs` for all ES modules instead).
```

So I guess the next step is to go rework the `redux-thunk@3.0-alpha` packaging per all the discussions, then come back here.

markerikson:
Superseded by #3318 , which appears to work!

PR #3277, opened 2023-03-17
Update slice.ts

modex98:


phryneas:
That code is perfectly fine as it is. 

What makes you think it wouldn't work - do we need to document something better?

PR #3279, opened 2023-03-18
Chore: Remove out-dated build script comment

### LINKED ISSUE #3210: [RTK Alpha]: Is ESNext output for all modules intended?
At the moment, all output files seem to be targeting `esnext`, which limits the support to the past couple years at a maximum. Fair enough if this was intended, but thought I'd ask, as there's a few comments in the build script which seem to indicate the idea (for the ESM at least) was to target ES2017: https://github.com/reduxjs/redux-toolkit/blob/ef49075675eacd75a80a03dc703d717e8b0e9f19/packages/toolkit/scripts/build.ts#L54

For example, here's the CJS output with optional chaining, an ES2020 feature: https://www.runpkg.com/?@reduxjs/toolkit@2.0.0-alpha.2/dist/cjs/redux-toolkit.cjs.development.js#1097

Edit: Generally, for maximum compatibility, you'd want to limit the output to ES2017 and only provide that via `"exports"`. There's no guarantees that something consuming `"main"` or `"module"` can consume ES2017+.

Edit2: I know optional chaining in particular will trip Webpack 4's parser, so users would need to make sure rtk is transpiled (which is very possible, of course): https://github.com/webpack/webpack/issues/10227

rschristian:
Closes #3210

(I think this is the right branch, certainly let me know if it's not and I'll get it fixed)

Hopefully what you had in mind; figure there's little point in specifying ES format if everything will be ESNext.

### DIFF
--- a/packages/toolkit/scripts/build.ts
+++ b/packages/toolkit/scripts/build.ts
@@ -51,7 +51,7 @@ const buildTargets: BuildOptions[] = [
     minify: true,
     env: 'production',
   },
-  // ESM, embedded `process`, ES2017 syntax: modern Webpack dev
+  // ESM, embedded `process`: modern Webpack dev
   {
     format: 'esm',
     name: 'modern',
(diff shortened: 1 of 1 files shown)


markerikson:
Yep, thanks!

PR #3281, opened 2023-03-19
fix post put mock handler

### LINKED ISSUE #3217: data bug in example app
I am trying to update a Post in the mutations example app but get this error:

```
TypeError: comparatorFn is not a function
    at http://localhost:3000/static/js/bundle.js:32276:16
    at Array.reduce (<anonymous>)
    at http://localhost:3000/static/js/bundle.js:32245:41
    at Array.map (<anonymous>)
    at http://localhost:3000/static/js/bundle.js:32235:40
    at http://localhost:3000/static/js/bundle.js:32342:12
    at http://localhost:3000/static/js/bundle.js:32719:9
    at forEach (http://localhost:3000/static/js/bundle.js:32701:7)
    at Object.filter (http://localhost:3000/static/js/bundle.js:32718:3)
    at queryByPrimaryKey (http://localhost:3000/static/js/bundle.js:32340:24)
```

I think this is a problem with the data package, but [the code to reproduce it is here.](https://github.com/reduxjs/redux-toolkit/blob/master/examples/query/react/mutations/src/mocks/db.ts#L53)

here is a screenshot indicating the comparatorName is `"0"`

Screen Shot 2023-02-26 at 10 38 40 AM

fpetrakov:
fixes #3217 

also you can check out the sandbox fork, it's the same as in the docs but with dp.post.update changed a bit 

https://codesandbox.io/s/wandering-cache-54h3ey?file=/src/mocks/db.ts

### DIFF
--- a/examples/query/react/mutations/src/mocks/db.ts
+++ b/examples/query/react/mutations/src/mocks/db.ts
@@ -51,7 +51,11 @@ export const handlers = [
     const id = Array.isArray(req.params.id) ? req.params.id[0] : req.params.id
 
     const post = db.post.update({
-      where: { id },
+      where: {
+        id: {
+          equals: id,
+        },
+      },
       data: { name },
     })
 


EskiMojo14:
thanks!

PR #3295, opened 2023-03-25
Allow injecting case reducers into the result of createReducer and createSlice

EskiMojo14:


EskiMojo14:
closing as unplanned - it was fun to implement and play around with, but it provides a very large footgun and makes reducer behaviour less predictable.

injecting whole slices is being investigated over at #3297.

PR #3296, opened 2023-03-25
remove "alternative" from descriptions of builder callback

EskiMojo14:


packages/toolkit/src/createSlice.ts
```diff
@@ -100,9 +100,6 @@ export interface CreateSliceOptions<
    * A callback that receives a *builder* object to define
    * case reducers via calls to `builder.addCase(actionCreatorOrType, reducer)`.
    * 
-   * Alternatively, a mapping from action types to action-type-specific *case reducer*
```
phryneas, line 103: Did we at this point only have the sentence, but no more examples of it?

EskiMojo14: yeah, the sentence was left in while the syntax itself was removed

phryneas: Then let's get this merged :)

PR #3320, opened 2023-04-03
Bump Immer to 10.0-beta

markerikson:


markerikson:
Bunch of discussion about Immer's packaging in https://github.com/immerjs/immer/issues/992 , and the size changes in https://github.com/immerjs/immer/pull/1028 .

Copying my perf comment from that PR:

Just ran a quick perf test with an RTK Query perf stress test scenario I've used previously.

I set it to render 1000 RTKQ-connected list items, then invalidated all the cache entries to force them to reload.

RTKQ dispatches when subscriptions are initiated, and again as data comes in.

With RTK 2.0.0-alpha.2 and Immer 9.x, average dispatch time during the load phase was 1.5ms.  With an RTK CSB CI build and Immer 10.0-beta.1, average dispatch time was 0.12ms (which is the lower limit of Chrome's function timing resolution):

image

When I hit "Invalidate", the Immer 9.x build had a solid 2.93 _seconds_ of dispatching.

With Immer 10, it was "only" 930ms:

image

The only change there was the Immer 10 bump.

So, clearly some pretty significant perf improvements!

That said, the Vite test project I'm using went _up_ in size with Immer 10.  The bundle went from 386K min (Immer 9.x) to 392K min (Immer 10).  Haven't looked into that yet, but very surprising.

PR #3334, opened 2023-04-07
Docs: Add a note that .take and .condition resolve only after an action is dispatched

chmac:
This was just a quick and dirty attempt to get the message across. Maybe it doesn't warrant so many words? Feel free to edit to your heart's content. :-)

Thanks for redux toolkit btw, it saves us a ton of boilerplate writing redux code. 🧡 🧡 🧡


markerikson:
Seems reasonable, thanks!

Issue #629, opened 2020-06-21
Add support for defining thunks in `createSlice`


markerikson:
Here's one that's been asked for a lot since we came up with `createSlice`: adding some kind of syntax for defining thunks directly inside of the `createSlice` call.

Now that I've actually used `createAsyncThunk` a few times in the new Redux tutorial I'm writing, I can see that there's still some pain points there:

- hand-declaring the thunk separately from the slice
- having to duplicate the slice `name`  in your action type prefix arg
-  having to always define `extraReducers` and defining the handlers for each result

We're going to ignore the issue of defining loading state and stuff for now, and focus on what a possible syntax for this might look like.

The biggest complicating factor here is the TS usage. `createAsyncThunk` may require several possible generic args, including `{dispatch, state, extra}`. We can't know, at the time the slice is defined, what the `state` type might be.

@phryneas listed several possible syntaxes in chat discussion, along with why they won't work.  Pasting that here for reference:

> So it seems to work without direct circularity problems. But, we don't really have a way to inject any generics there, so we don't really have any way to specify types for dispatch, extra & getState. So currently we can really only accept any in there :/
Probably the only way around that would be to define generic inferfaces for state/dispatch/extra for the RTK module that could be extended by module augmentation. So,
> 
>```ts
> declare module `@reduxjs/toolkit` {
>  interface AppRootState extends ReturnType<typeof store.getState> {}
>  // not sure if that would even work
>  interface AppDispatch extends typeof store.dispatch 
>   interface AppExtra //...
> }
> ```
> 
> Not really a fan of that thought/
> Or use some kind of second-level builder that could allow for generics like
> 
> ```ts
> createSlice({
>   asyncThunks: {
>     thunk1: createAsyncThunk => createAsyncThunk</*...*/>(/*...*/),
>     thunk2: createAsyncThunk => createAsyncThunk</*...*/>(/*...*/),
>   }
> })
> ```
> also, not a fan.  Also, it would have to be nested that deep down to still be able to extract the keys to create action creators.
>
> ```ts
> createSlice({
>   asyncThunks: buildThunks => buildThunks<ApiSignature>({
>     thunk1: {
>       payloadCreator(...){},
>       fulfilled(state, action) {} 
>     },
>     thunk2: {
>       payloadCreator(...){},
>       fulfilled(state, action) {} 
>     }
>   })
> })
> ```
> 
[...]

PR #3388, opened 2023-04-19
Add "creator callback" syntax to slice reducer field, to allow for async thunk creation

### LINKED ISSUE #1045: Suggestion: Define thunk on createSlice
I'm sure this has already been discussed elsewhere, but perhaps not with the following reasoning in mind, so please bear with me :)

One thing that I often come across is the desire to reuse a whole reducer across different parts of the state, so different slices would be the correct term I guess. 
This can of course easily be done by creating a higher order reducer etc. What cannot so easily be done however is to reuse the async actions. You need to somehow point the action back to the different part of the state, and you also need to make sure any getState calls inside any async actions select from the correct part of the state.

Couldn't this perhaps somehow be baked into createSlice? Apologies for not having a fully fledged idea here, but I do think this could be a nice improvement to already excellent DX of RTK. Especially if those thunk actions could also somehow easily have the "local" state available, so that they don't need to know where in the global state they are located.

Some other things that would be improved by defining thunks via createSlice (which I'm sure you are of course alredy aware of) are:
- Remove the need to define the name in a way like this:
      const addGroup = createAsyncThunk(`${sliceName}/addGroup`, async ({group}, {getState}) => {
    });
- Automatically export all thunks as well via reducer.actions

Would be interested to hear your thoughts on this. And thanks again for the awesome library 👍

EskiMojo14:
reuses code from #637, with some changes to alleviate previous concerns:

- create.asyncThunk now has .withTypes, to allow for defining common config within a slice
- config passed must **not** include state or dispatch, as those will cause circular type issues
  - instead, they should be cast **inside** the payload creator, in the same way that RTKQ requires.

remaining concerns:
- no way to only import cAT *if* we need it
- because we can't use the state and dispatch type, the created async thunk will not have knowledge of what state/dispatch type it expects
  - this means it would technically be possible to dispatch the thunk into a dispatch that would not give it the state/dispatch signatures it expects
  - as a counterpoint, it seems like a rare case where an app would have more than one rootstate/appdispatch type

fixes #1045 and #629

### DIFF
--- a/docs/api/createSlice.mdx
+++ b/docs/api/createSlice.mdx
@@ -301,15 +301,11 @@ reducers: (create) => {
 
   return {
     fetchTodo: createAThunk<Todo, string>(async (id, thunkApi) => {
-      const state = thunkApi.getState() as RootState
-      const dispatch = thunkApi.dispatch as AppDispatch
       throw thunkApi.rejectWithValue({
         error: 'Oh no!',
       })
     }),
     fetchTodos: createAThunk<Todo[], string>(async (id, thunkApi) => {
-      const state = thunkApi.getState() as RootState
-      const dispatch = thunkApi.dispatch as AppDispatch
       throw thunkApi.rejectWithValue({
         error: 'Oh no, not again!',
       })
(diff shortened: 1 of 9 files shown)


markerikson:
Okay, let's do this!

PR #3393, opened 2023-04-20
types: make it easier to wrap createAsyncThunk

shrouxm:
Not exporting `GetThunkAPI` requires code dup when wrapping `createAsyncThunk` (in order to get top level access to the `ThunkAPI`'s type). There also was a minor typo making the default value of the `RejectedValue` type parameter `undefined` instead of `unknown`.


markerikson:
Thanks!

PR #3407, opened 2023-04-27
Implement auto fork joining

### LINKED ISSUE #3313: Support listenerApi.joinForks() or similar.
I am examining converting a lot of existing sagas to the listener middleware and I am very excited.

However, I have come across a few gaps that make things challenging. One of those gaps, is that forks are attatched in sagas and there is no such attachment in the listeners. So where no management of forks is needed when writing sagas, you now need to keep track of your forked results so you can await them explicitly.

```js
function* doThingSaga(action) {
  // lots of code
  yield fork(doAnotherThing);
  // lots of code
  yield fork(doAnotherThing2);
  // lots of code
  if (someCondition) {
    yield fork(doAnotherThing3);
  }
  // lots of code
  if (someCondition2) {
    yield fork(doAnotherThing4);
  }
}
```

This effectively requires the code: 
```js
function doThingEvent(action, api) {
  const forks = [];
  // lots of code
  forks.push(api.fork((forkApi) => doAnotherThing(forkApi, api)));
  // lots of code
  forks.push(api.fork((forkApi) => doAnotherThing2(forkApi, api)));
  // lots of code
  if (someCondition) {
    forks.push(api.fork((forkApi) => doAnotherThing3(forkApi, api)));
  }
  // lots of code
  if (someCondition2) {
    forks.push(api.fork((forkApi) => doAnotherThing4(forkApi, api)));
  }

  await api.pause(Promise.all(forks.map(f => f.result)));
}
```

In a perfect world, the forks would not stop just because the parent completed, but given how breaking that is, a more helpful alternative could be:
```js
function doThingEvent(action, api) {
  // lots of code
  api.fork((forkApi) => doAnotherThing(forkApi, api));
  // lots of code
  api.fork((forkApi) => doAnotherThing2(forkApi, api));
  // lots of code
  if (someCondition) {
    api.fork((forkApi) => doAnotherThing3(forkApi, api));
  }
  // lots of code
  if (someCondition2) {
    api.fork((forkApi) => doAnotherThing4(forkApi, api));
  }

  await api.joinForks();
}
```

This is much cleaner/easier to read.

ericanderson:
fixes #3313

Adds support for options to `api.fork` to allow for `api.fork(myForkFunc, { autoJoin: true })` to cause the current listener to wait for all `autoJoin`ed forks to complete/cancel before completing itself.

This change is backwards compatible and can be merged into 1.x.

That said, I think that anyone coming from sagas is going to assume that `fork()` behaves as `{autoJoin: true}` and this probably should just be the default in 2.x

Example:

```ts
addListener({
      actionCreator: increment,
      effect: async (_, listenerApi) => {
        const forkedTask = listenerApi.fork(async (forkApi) => {
          await forkApi.pause(delay(1_000))
        }, { autoJoin: true});
      }, // effect will not be marked as complete until the pause completes
})
```

### DIFF
--- a/packages/toolkit/src/listenerMiddleware/index.ts
+++ b/packages/toolkit/src/listenerMiddleware/index.ts
@@ -23,6 +23,7 @@ import type {
   TaskResult,
   AbortSignalWithReason,
   UnsubscribeListenerOptions,
+  ForkOptions,
 } from './types'
 import {
   abortControllerWithReason,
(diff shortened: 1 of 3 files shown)


markerikson:
Looks good, thanks!

PR #3408, opened 2023-04-28
[Docs] Update link to react-query comparison

adamlindqvist:
Update link to the react-query comparison in "Comparison with Other Tools"


markerikson:
Thanks!

PR #3411, opened 2023-04-30
Link to new typescript docs of type predicates

nilshartmann:
The current docs point to the deprecated TS docs about type predicates.


phryneas:
Danke!

PR #3414, opened 2023-05-02
create action creator middleware

### LINKED ISSUE #3413: Dev check middleware to warn about action creators being dispatched
A semi-common mistake is to dispatch an action creator without calling it first, e.g. `dispatch(actionCreator)`.

This doesn't currently raise any errors:
- In Typescript, RTK's action creators having a static `type` property mean they conform to the Action type.
- In runtime, dispatching a function is usually picked up by the thunk middleware, which would end up returning `actionCreator(dispatch, getState, extra)`.
An error would only be raised with no thunk middleware.

From a Typescript perspective, this would likely only be solveable by adding some sort of `[notAnAction)?: never` property to the base Action type, and adding a `[notAnAction]: true` property to the action creators, making it incompatible. (this also wouldn't prevent action creators without parameters being valid thunks, unless something similar was done for ThunkAction)

From a runtime perspective, this could be fairly simple: a dev-only middleware similar to the serializability and immutable middlewares looks for a unique property we assign to RTK action creators, and warns in console if it sees one.

EskiMojo14:
fixes #3413

### DIFF
--- a/packages/toolkit/src/actionCreatorInvariantMiddleware.ts
+++ b/packages/toolkit/src/actionCreatorInvariantMiddleware.ts
@@ -13,7 +13,7 @@ export function getMessage(type?: unknown) {
   return `Detected an action creator with type "${
     type || 'unknown'
   }" being dispatched. 
-Make sure you're calling the action before dispatching, i.e. \`dispatch(actionCreator())\` instead of \`dispatch(actionCreator)\`. This is necessary even if the action has no payload.`
+Make sure you're calling the action creator before dispatching, i.e. \`dispatch(actionCreator())\` instead of \`dispatch(actionCreator)\`. This is necessary even if the action has no payload.`
 }
 
 export function createActionCreatorInvariantMiddleware(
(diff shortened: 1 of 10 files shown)


packages/toolkit/src/actionCreatorInvariantMiddleware.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export interface ActionCreatorInvariantMiddlewareOptions {
+  /**
+   * The function to identify whether a value is an action creator.
+   * The default checks for a function with a static type property and match method.
+   */
+  isActionCreator?: (action: unknown) => action is Function & { type?: unknown }
+}
+
+export function getMessage(type?: unknown) {
```
markerikson, line 12: :question: My only question here is whether this is going to get properly tree-shaken.  We switched from Terser to ESBuild for minification, and I was seeing that the other dev check middleware needed to have message functions like this _inside_ the middleware, and inside an `else` in order to shake properly.

Let me do a prod build of a test app and see what happens.

markerikson: Okay, looks good:

image

Issue #930, opened 2021-03-19
Dictionary type - why is undefined a possible entity value


GlynL:
```
export declare interface Dictionary<T> extends DictionaryNum<T> {
    [id: string]: T | undefined;
}
```
Getting the chance to use rtk for the first time, and using `createEntityAdapter`. Just coming across this and wondering why the possible value of an entity is undefined. The entity keys are created off the entity object so I can't imagine a scenario where it would be undefined. Not a big deal, but means accounting/coding for undefined every time you access an entity.

Thanks.

markerikson:
Because if you were to access `state.entities["abcd1234"]`, there's no guarantee there's an entity with that ID loaded.

So, rather than go with a TS type like `Record<K, V>` that says "we'll just assume that every field you access is valid", we opted to reflect reality.  Agreed that it means doing a bit more bookkeeping, but consider it a safety precaution.

GlynL:
that's fair enough thanks for the reply. makes sense.

markerikson:
Sure.

The other answer is, "We ported `createEntityAdapter` from `@ngrx/entity`, that's the way they wrote it, and we just copied it" :)

diegodlh:
Just came up with the same question while experimenting with redux.

I'd like to add here that as a side effect of this there seem to be no way of iterating through all entities without checking whether they are undefined. I mean, following the example [here](https://redux.js.org/tutorials/essentials/part-6-performance-normalization#converting-the-notifications-slice), if one wants to do something for all entities using `Object.values(state.entities).forEach()`, one has to make sure that every entity is not `undefined`.

I assume using the `selectAll` selector (which returns an array of not-undefined entities) could not be used here instead to do something on all entities, correct?

I wonder whether eventually changing the `entities` object from a dictionary to a map would solve this.

markerikson:
FWIW, the difference in behavior here in terms of both runtime and types is because `selectAll` iterates over the IDs array.  So, we _know_ that every lookup of `state.entities[id]` _will_ have a valid result.

We _could_ hypothetically change this, but at that point we're also sorta lying with the types. Then again, that's also what the built-in `Record<K,V>` type does.

@phryneas , thoughts ?

phryneas:
This type is a lot older than `noUncheckedIndexedAccess`, but was created with a similar intent. We could probably remove it in that context, but I'd wait for 2.0 for that.

markerikson:
[...]

PR #3424, opened 2023-05-07
Use Record<EntityId, T> instead of Dictionary<T>

EskiMojo14:
as per #930, uses Record instead of Dictionary type, so user can choose whether `entities[id]` is possibly undefined via `noUncheckedIndexedAccess`

`selectById` uses a type helper I've published to match the user's setting.


markerikson:
I know this is kinda picky, but I'd really rather not add another dependency just for this.  

Can we inline it?

EskiMojo14:
yes, I can switch back to importing a TS file from src if you'd prefer.

markerikson:
yes please :)

Matt-Ord:
This conflicts with #2901 - I would you rather I remove the Dictionary type in that Pr instead?

EskiMojo14:
if all goes to plan this should probably be in the next alpha, so if you're happy to fix any conflicts and include the changes in #3187 that would work - otherwise I'm happy to deal with any conflicts here should your MR go in first :)

cc: @markerikson

Matt-Ord:
If possible it would be great if my pr could be merged first - it's been sitting around for a long time and I don't want to waste my time with yet another rebase

Matt-Ord:
I think the use of 
```
Compute<UncheckedIndexedAccess<T>> 
```
is needlessly complex @markerikson would you merge it if I made the changes I suggested?

EskiMojo14:
@Matt-Ord I'm confused what you're asking - #3187 is already merged?

Matt-Ord:
> @Matt-Ord I'm confused what you're asking - #3187 is already merged?

Yeah but I'd suggested something that could have be improved in the review - if I did the improvement as another pr is that something that would be merged?

Matt-Ord:
My point is it's not necessary if you replace it with Record<Id,T>[Id], which is a lot simpler

EskiMojo14:
feel free to open a PR demonstrating that working - we'd be happy to have a less complex approach.

PR #3425, opened 2023-05-07
Remove toString override from action creators, in favour of explicit .type field.

### LINKED ISSUE #3370: Deprecate/remove getType (and possibly actionCreator.toString override?)
Currently RTK exports a function called getType, which uses the .toString override from an action creator to return that action's type.

This was likely useful when the toString override (which was mainly for the now-removed object syntax for extraReducers) was the only method of getting the type, but action creators now have a static .type property which does this in a much simpler fashion.

As a result, it could be worth removing getType and possibly the toString override for 2.0.

EskiMojo14:
fixes #3370

### DIFF
--- a/docs/api/createAction.mdx
+++ b/docs/api/createAction.mdx
@@ -31,7 +31,7 @@ const action = increment(3)
 // { type: 'counter/increment', payload: 3 }
 ```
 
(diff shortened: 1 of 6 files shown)


docs/api/createAction.mdx
```diff
@@ -103,21 +100,23 @@ const counterReducer = createReducer(0, (builder) => {
 })
 ```
 
+<!-- TODO: how do we handle this? -->
```
markerikson, line 103: :question: I dunno, what "this" are we referring to here? :)

EskiMojo14: I don't think we have anything remaining that actually depends on the type being a string - but I'm assuming we still want to advise that they should be strings?

PR #3428, opened 2023-05-08
Fixed Stackoverflow bug if children prop is a ref to root/parent object

cheprasov:
Fixed Stackoverflow bug if children prop is a ref to root/parent object

Fixed Bug:
RangeError: Maximum call stack size exceeded at trackProperties (.../node_modules/@reduxjs/toolkit/dist/redux-toolkit.cjs.development.js:322:25)

### DIFF
--- a/packages/toolkit/src/immutableStateInvariantMiddleware.ts
+++ b/packages/toolkit/src/immutableStateInvariantMiddleware.ts
@@ -92,11 +92,13 @@ function trackProperties(
   isImmutable: IsImmutableFunc,
   ignorePaths: IgnorePaths = [],
   obj: Record<string, any>,
-  path: string = ''
+  path: string = '',
+  checkedObjects: Set<Record<string, any>> = new Set()
 ) {
   const tracked: Partial<TrackedProperty> = { value: obj }
 
-  if (!isImmutable(obj)) {
+  if (!isImmutable(obj) && !checkedObjects.has(obj)) {
+    checkedObjects.add(obj);
     tracked.children = {}
 
     for (const key in obj) {


markerikson:
@cheprasov Interesting, thanks.  Two questions:

- Can you add a new unit test that checks this case?
- What scenario happened that led you to finding this? Normally I would not expect the same object reference to appear in multiple places in a state tree.

cheprasov:
> @cheprasov Interesting, thanks. Two questions:

Hi @markerikson 
Thanks a lot.
> * Can you add a new unit test that checks this case?
* Unfortunately, I do not have enough time for writing the tests. I just copied the fix from my project. 

> * What scenario happened that led you to finding this? Normally I would not expect the same object reference to appear in multiple places in a state tree.
> I did complex thing, but will try to explain simple way. I added to the state Express Response object that has a lot of child props with ref to some parents objects.

I really need the fix in my project, it allows me to use Redux state for more complex things :)

markerikson:
@cheprasov : I hate to say it, but that _sounds_ like a misuse of Redux.

It's fine to use Redux on the server.  But, a Redux store really should only have plain non-serializable data inside, and an Express `Response` object is definitely not "plain data":

- https://redux.js.org/style-guide/#do-not-put-non-serializable-values-in-state-or-actions

If you are really going to do that, I'd suggest just turning off this middleware during store setup.

cheprasov:
> @cheprasov : I hate to say it, but that _sounds_ like a misuse of Redux.
> 
> It's fine to use Redux on the server. But, a Redux store really should only have plain non-serializable data inside, and an Express `Response` object is definitely not "plain data":
> 
> * https://redux.js.org/style-guide/#do-not-put-non-serializable-values-in-state-or-actions
> 
> If you are really going to do that, I'd suggest just turning off this middleware during store setup.

Yep, thanks a lot. I have disabled serialization already.

markerikson:
That said, the tweak here doesn't hurt, so I'll go ahead and merge it. Thanks!

PR #3451, opened 2023-05-16
export build hooks

lucashm:


lucashm:
Usage example:
- Federated module (remote entry)
image

- micro-frontend
image

lucashm:
Usage example:
- Federated module (remote entry)
image

- micro-frontend
image

Issue #3474, opened 2023-05-25
[RED-19] Create callback doesn't handle actions without payload properly


EskiMojo14:
When no action is specified for a reducer, the action creator becomes `ActionCreatorWithPayload<any>` instead of the intended `ActionCreatorWithoutPayload`.

[example](https://codesandbox.io/s/slice-selector-circular-type-forked-exzu6c?file=/src/Counter.tsx)

<sub>[RED-19](https://linear.app/redux/issue/RED-19/create-callback-doesnt-handle-actions-without-payload-properly)</sub>

PR #3475, opened 2023-05-25
Fix inference for reducers without actions specified

EskiMojo14:
fixes https://github.com/reduxjs/redux-toolkit/issues/3474

### DIFF
--- a/packages/toolkit/src/createSlice.ts
+++ b/packages/toolkit/src/createSlice.ts
@@ -353,6 +353,9 @@ interface AsyncThunkCreator<
 }
 
 export interface ReducerCreators<State> {
+  reducer(
+    caseReducer: CaseReducer<State, PayloadAction>
+  ): CaseReducerDefinition<State, PayloadAction>
   reducer<Payload>(
     caseReducer: CaseReducer<State, PayloadAction<Payload>>
   ): CaseReducerDefinition<State, PayloadAction<Payload>>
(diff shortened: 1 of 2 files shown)


EskiMojo14:
@einarq would you be able to try out the package built from this PR, see if it fixes your issue? 🙂

einarq:
Sure, how do I reference it?

EskiMojo14:
There are install instructions [here](https://ci.codesandbox.io/status/reduxjs/redux-toolkit/pr/3475/builds/381165) 😄

einarq:
Ok. It solves the issue, but it introduces another issue (which might be
fine?).

I now get a TS error if I use create.reducer, but I don't supply any types
on the action.

[image: image.png]

EskiMojo14:
hmm, that image doesn't seem to display on github. Could you post a snippet, or adjust the codesandbox with an example?

einarq:
Didnt work when coming in via an email I guess, trying again.

image


Weird thing is that I'm unable to repro this on the sandbox

einarq:
Got it now

https://codesandbox.io/s/slice-selector-circular-type-forked-exzu6c?file=/src/counterSlice.ts

EskiMojo14:
ahh, I see - I'm tempted to say this is a good thing though, if you want a payload other than void you should provide it.

either by using the usual `PayloadAction<Payload>` or by doing `create.reducer<Payload>(`

EskiMojo14:
in the meantime if you want the previous behaviour of using `any` you can always do something like 
```ts
export type TODOFIXME = any;

create.reducer<TODOFIXME>((state, { payload: { orderDir } }) => {})
```

einarq:
We’ll just fix them as we go, but thanks!On 25 May 2023, at 12:22, Ben Durrant ***@***.***> wrote:﻿
in the meantime if you want the previous behaviour of using any you can always do something like
export type TODOFIXME = any;

create.reducer<TODOFIXME>((state, { payload: { orderDir } }) => {})

PR #3483, opened 2023-05-28
Remove Unused Imports

adamaveray:
A follow-up to #2246, removing some new imports that are causing `tsc` errors when the library is used in a project with `--noUnusedLocals` enabled.

<details>
<summary>Some example error output from `tsc`</summary>

```
node_modules/@reduxjs/toolkit/src/query/core/buildInitiate.ts:8:1 - error TS6192: All imports in import declaration are unused.

8 import { DefinitionType, isQueryDefinition } from '../endpointDefinitions'
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@reduxjs/toolkit/src/query/core/buildInitiate.ts:11:36 - error TS6196: 'RootState' is declared but never used.

11 import type { SubscriptionOptions, RootState } from './apiState'

[...]
```
</details>

### DIFF
--- a/packages/toolkit/scripts/build.ts
+++ b/packages/toolkit/scripts/build.ts
@@ -14,7 +14,7 @@ import yargs from 'yargs/yargs'
 
 import { extractInlineSourcemap, removeInlineSourceMap } from './sourcemap'
 import type { BuildOptions, EntryPointOptions } from './types'
-import { appendInlineSourceMap, getLocation } from './sourcemap'
+import { appendInlineSourceMap } from './sourcemap'
 
 const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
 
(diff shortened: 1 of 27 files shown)


adamaveray:
An update: These errors are being caused by directly importing files from within `@reduxjs/toolkit/src/...`, which is being auto-imported by my IDE instead of those within `dist`. Ensuring all imports use the `dist` subdirectory instead of `src` correctly prevents `tsc` from flagging issues with this library (maybe it would be worth only shipping the `dist` subdirectory in the bundled module to prevent the source files being referenced instead of the proper dist versions).

phryneas:
It's a bit of a tradeoff - if we ship the source, people can accidentally import it (but it seems rather rare); if we don't, people can't directly look up source code. Especially since VSCode now supports to jump directly to the uncompiled source of a library (if it ships it), it's usually worth shipping the sources.

markerikson:
Seems like this is an end-user linting issue?

PR #2730, opened 2022-09-25
split `enhanceEndpoints` into `addTagTypes` and `enhanceEndpoint`

phryneas:
(at the moment this is TS only and missing runtime code)

This would deprecate `enhanceEndpoints` and add new methods `addTagTypes` and `enhanceEndpoint`.

These are easier to type from our side and would allow the argument and return type of existing endpoints to be changed.

PR #3485, opened 2023-05-28
Re-investigate splitting `enhanceEndpoints` into `addTagTypes` and `enhanceEndpoints`

EskiMojo14:
see #2730


packages/rtk-codemods/transforms/enhanceEndpoint/test.js
```diff
@@ -0,0 +1,9 @@
+'use strict'
+
+const { runTransformTest } = require('codemod-cli')
+
+runTransformTest({
+  name: 'enhanceEndpoint',
+  path: require.resolve('./index.ts'),
+  fixtureDir: `${__dirname}/__testfixtures__/`
+})
```
aryaemami59, line 9:
```suggestion
import path from 'node:path'
import { runTransformTest } from '../../transformTestUtils.js'
import transform, { parser } from './index.js'

runTransformTest(
  'enhanceEndpoints',
  transform,
  parser,
  path.join(__dirname, '__testfixtures__')
)
```

We should probably get rid of [**`codemod-cli`**](https://github.com/rwjblue/codemod-cli#readme) since it hasn't been updated in about 4 years and does not work with later versions of TypeScript.

Issue #3443, opened 2023-05-15
transformResponse in enhanceEndpoints returning Promise type when trying to transform the type of the response


cini9:
Following the update to V1.9.4, with the ability to override TS type of the original data in transformResponse of enhanceEndpoints, the response type I am receiving in my components are being additionally transformed to add a Promise<> type of my desired response type. 

I am using codegen to generate types from my backend, but need to transform that response to extract lower-level data from the raw graphql query response.

```
type TransformedGetUser = GetUserQuery['getUser']
type TransformedGetUsers = GetUsersQuery['getUsers']

type Definitions = DefinitionsFromApi<typeof generatedUserApi>
type TagTypes = TagTypesFromApi<typeof generatedUserApi>

type UpdatedDefinitions = Omit<Definitions, 'getUsers' | 'getUser'> & {
  getUsers: OverrideResultType<Definitions['getUsers'], TransformedGetUsers>
  getUser: OverrideResultType<Definitions['getUser'], TransformedGetUser>
}

const userApi = generatedUserApi.enhanceEndpoints<TagTypes, UpdatedDefinitions>(
  {
    endpoints: {
      getUsers: {
        providesTags: ['User'],
        transformResponse: (response: GetUsersQuery): TransformedGetUsers =>
          response.getUsers,
      },
      getUser: {
        providesTags: ['User'],
        transformResponse: (response: GetUserQuery): TransformedGetUser =>
          response.getUser,
      },
    },
  }
)
```
This gives me the typing in the components in the screen shot below, with my correct type (id: string, firstName: string, etc) followed by the Promise<...> type.

![image](https://github.com/reduxjs/redux-toolkit/assets/56114997/4df6c49a-3ac6-45d8-b354-a02054ec989a)

When I try to use the data returned from these queries, I'm getting a 'property id does not exist on type Promise< .... >'

![image](https://github.com/reduxjs/redux-toolkit/assets/56114997/57bf894d-e854-4bcb-b64d-e6150f9077d6)

Is there any way to tell RTK to not add the Promise type when we are overriding the response type in transformResponse ?

Thanks in advance for your help !

quentingirard:
@dmitrigrabov did you find any solution ?

Is there a way that I can help ?

Mozzarella123:
The main problem is that transformResponse can return Promise in types. And when we use query data it become `MaybePromise`

[...]

PR #3500, opened 2023-06-02
Fix TransformedResponse type to unwrap promise

EskiMojo14:
fixes https://github.com/reduxjs/redux-toolkit/issues/3443


phryneas (approved):
Looks good to me!

PR #3518, opened 2023-06-11
Allow createAsyncThunk to infer type prefix as literal string

EskiMojo14:


EskiMojo14:
closing for now - this adds extra complexity and will only actually work if the generics are either all provided or all inferred.

might be worth revisiting once typescript ships some form of partial inference, such as the _ sigil

Issue #2657, opened 2022-08-30
File extension doesn't change in the docs according to the selected language


imrishabh18:
- The [Quick Start](https://redux-toolkit.js.org/tutorials/quick-start) docs page doesn't change the file extension from `js` to `ts` on selection.

https://user-images.githubusercontent.com/38923768/187497604-4e0e0626-cda8-45f9-9c5a-b34bcb56f60d.mov

imrishabh18:
I would like to make a PR to resolve this issue.

markerikson:
I'm actually not sure there's anything we can do about this, tbh. If you look at the markdown, that filename is directly written as part of the codeblock metadata.

If you can figure out a way to do it, great, but I'm honestly not sure the Docusaurus docs tool will let us change that dynamically.

imrishabh18:
I am not that familiar with Docusaurus, tbh. I was going through the [typescript.md](https://github.dev/reduxjs/redux-toolkit/blob/master/docs/tutorials/typescript.md) file and thought there might be a way to show these codeblocks with the correct title on the docs page. 

I even tried to find some Docusaurus plugins to dynamically change the `title` of the codeblock, but couldn’t find any. 

Anyways, if you have directions where I can look into to resolve this, would be happy too.

PR #3535, opened 2023-06-18
Fixed filename extensions (ts/js) in code examples

gitman69420:
Trying to fix this issue: #2657


markerikson:
Afraid the right answer is going to involve a more technical change, but thanks!

PR #3555, opened 2023-06-23
Fix example not using correct accessor

marlonschlosshauer:
Hey while following the docs I found that the RTK-Query example was using the wrong accessor. 

Desctructuring just`pokemonApi` did not allow me access the endpoints. I had to append `.endpoints` to be able to get them. Could be an error on my end though, sorry if that's the case. 

Here some examples from my editor showing TypeScript complaining:


marlonschlosshauer:
Ups, my IDE didn't show the import correctly, my bad, sorry.

PR #3562, opened 2023-06-28
Update queries.mdx, added apostrophe - "posts" to "post's"

flyingace:
Added an apostrophe, changing `...the given posts data...` to `...the given post's data...`. An extremely minor change to be sure, but when writing about a post in a group of posts being able to easily distinguish between the two is important.


EskiMojo14:
thanks!

PR #3569, opened 2023-06-29
fix(RTKQuery): Resolves type portability with createApi. fixes #3568 

### LINKED ISSUE #3568: [RTK v2.0]: Function `createApi` Types are not Portable
For those experimenting with **Redux Toolkit v2 (currently 2.0.0-Beta.0)** and must export their own type declarations, there are quite a few issues with **the portability of types** from Redux Toolkit Query’s function `createApi`..

(There are also issues with functions like `createSlice` too, but I haven’t delved as deep into it, so I’m just focusing on `createApi` right now.)

Using `createApi`...
```typescript
import { createApi } from '@reduxjs/toolkit/query';
export const myApi = createApi({ /** ... */ });
```
results in an error like so…
```
The inferred type of 'myApi' cannot be named without a reference to '../node_modules/@reduxjs/toolkit/dist/query/core/module.js'. This is likely not portable. A type annotation is necessary. ts(2742)
```

### Significant TypeScript configurations:
```json
{
   "moduleResolution": "Node16",
   "declaration": true,
}
```

### Minimal Example with Error
https://codesandbox.io/p/sandbox/rtk-type-portability-j99lyh?file=%2Fsrc%2Frtk%2FmyApi.ts%3A1%2C1

_Note: I believe this error does not occur for a majority of developers using RTK because it's normally included in applications that don’t need compiled declarations (`declarations: false` in `tsconfig.json`). This may be why this issue isn’t frequently brought to attention._

eric-crowell:
Exports additional types to resolve the `ts2742` error when using the `createApi` function from RTK Query.

Resolves issue: #3568


packages/toolkit/src/query/index.ts
```diff
@@ -26,5 +29,6 @@ export type { CreateApi, CreateApiOptions } from './createApi'
 export { buildCreateApi } from './createApi'
 export { fakeBaseQuery } from './fakeBaseQuery'
 export { copyWithStructuralSharing } from './utils/copyWithStructuralSharing'
-export { createApi, coreModule } from './core'
+export { createApi, coreModule, coreModuleName } from './core'
 export { defaultSerializeQueryArgs } from './defaultSerializeQueryArgs'
+export * from './tsHelpers';
```
phryneas, line 34: Please try to keep it to a minimum and check what is actually needed. From what I gather, that might actually only be `symbol` types. 
We don't want to export too much, since at that point implementation details become a public API that we have to try to keep stable and keep supporting pretty much indefinitely.

eric-crowell: Makes sense; you got it! I'll pin point it.

Maybe you're right about the symbols and only `coreModuleName` needs to be exposed. TypeScript has a hard time making it clear.

Here's a very simple reproduction of all the **portability errors** I was seeing in my project:
https://codesandbox.io/p/sandbox/rtk-type-portability-j99lyh?file=%2Fsrc%2Frtk%2FmyApiBaseQuery.ts%3A11%2C3

^ Additional comments in the sandbox to see what might've lead me astray.

eric-crowell:
You are correct @phryneas, it is just the symbol type that is necessary to resolve this issue.

phryneas:
While you're at it, could you make sure to export the `reactHooksModuleName` from `query/react` in the same way?

eric-crowell:
> While you're at it, could you make sure to export the `reactHooksModuleName` from `query/react` in the same way?

Sure thing!

eric-crowell:
I found additional problems inferring types when using `configureStore`. This is only when API reducers (from `createApi`) are included.

Reproduction of the error on [Code Sandbox](https://codesandbox.io/p/sandbox/rtk-type-portability-j99lyh?file=%2Fsrc%2Frtk%2Fstore.ts%3A10%2C4)

Exporting the exact types `QueryCacheKey, RootState, CombinedState` from `./core/apiState` resolves the issue.

markerikson:
This PR had conflicts and I didn't have push permissions, so I duped and rebased over into #3678 . Thanks!

PR #3572, opened 2023-06-30
Throw error when type is empty in builder.addCase

### LINKED ISSUE #3570: Throw error when custom action creator does not have `type` value
It'd be great to have a guard for passing in an action creator that **doesn't** have a truthy `type` string value.

I encountered this issue when migrating over to RTK. We built some custom action creators that we use in combination with a custom api middleware (we're not ready to switch to RTQ just yet - baby steps 😅!). While these have the `toString()` override, the builder `addCase` method is not using that _and_ it seems like this override is [getting deprecated in v2](https://github.com/reduxjs/redux-toolkit/pull/3425).

The following led to a fairly subtle bug, since it fails silently.
```js
extraReducers: (builder) => {
  builder.addCase(apiActions.fetchTodo, todoHandler) // fetchTodo.type === undefined
}
```
Since I'm working in JS, this issue is easier to run into. Although, I believe a guard may still be helpful to prevent accidental assignments of an empty string to the `type` attribute in the implementation of a custom action creator in TS.
https://github.com/reduxjs/redux-toolkit/blob/24fc34392d249573fd14088db8daea10e848de41/packages/toolkit/src/mapBuilders.ts#L158-L168

chawes13:
Resolves #3570 

Some of the thrown errors in `mapBuilders` are only executed when `NODE_ENV !== 'production'`. I decided to not use this guard because this functionality seemed to closely match the intent of the `type in actionsMap` check, which also does not have the aforementioned guard applied. I don't have strong opinions on it though!

### DIFF
--- a/packages/toolkit/src/mapBuilders.ts
+++ b/packages/toolkit/src/mapBuilders.ts
@@ -140,7 +140,7 @@ export function executeReducerBuilderCallback<S>(
     ) {
       if (process.env.NODE_ENV !== 'production') {
         /*
-         to keep the definition by the user in line with actual behavior, 
+         to keep the definition by the user in line with actual behavior,
          we enforce `addCase` to always be called before calling `addMatcher`
          as matching cases take precedence over matchers
          */
@@ -159,6 +159,11 @@ export function executeReducerBuilderCallback<S>(
         typeof typeOrActionCreator === 'string'
           ? typeOrActionCreator
           : typeOrActionCreator.type
+      if (!type) {
+        throw new Error(
+          '`builder.addCase` cannot be called with an empty action type'
+        )
+      }
       if (type in actionsMap) {
         throw new Error(
           'addCase cannot be called with two reducers for the same action type'
--- a/packages/toolkit/src/mapBuilders.ts
+++ b/packages/toolkit/src/mapBuilders.ts
@@ -166,7 +166,7 @@ export function executeReducerBuilderCallback<S>(
       }
       if (type in actionsMap) {
         throw new Error(
-          'addCase cannot be called with two reducers for the same action type'
+          '`builder.addCase` cannot be called with two reducers for the same action type'
         )
       }
       actionsMap[type] = reducer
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/mapBuilders.ts
```diff
@@ -140,7 +140,7 @@ export function executeReducerBuilderCallback<S>(
     ) {
       if (process.env.NODE_ENV !== 'production') {
         /*
-         to keep the definition by the user in line with actual behavior, 
+         to keep the definition by the user in line with actual behavior,
```
chawes13, line 143: Snuck in when I ran formatting on this file – can revert if you'd like

packages/toolkit/src/mapBuilders.ts
```diff
@@ -159,6 +159,11 @@ export function executeReducerBuilderCallback<S>(
         typeof typeOrActionCreator === 'string'
           ? typeOrActionCreator
           : typeOrActionCreator.type
+      if (!type) {
+        throw new Error(
+          '`builder.addCase` cannot be called with an empty action type'
```
chawes13, line 164: I was torn on whether to use just `addCase` to match the actionsMap error or to use `builder.<method>` to match the pattern of the rest of the error messages in this file

EskiMojo14: matching the rest makes sense - in fact it would be good to update the actionsMap error to match too

chawes13: Happy to do that! Would you like me to create a separate pr?

EskiMojo14: nah, it seems like it would be fine to do here imo

EskiMojo14: looks good to me - slightly odd that the CI tests haven't run though

chawes13: Weird! Anything I can do to help?

PR #3588, opened 2023-07-08
fix/kitchen-sink-isAuthenticated: fixed isAuthenticated state change on login fullfilled

shrijan00003:
-  Examples/ kitchen-sink:  Set isAuthenticated state on login API fullfilled


EskiMojo14:
thanks!

PR #3621, opened 2023-07-26
docs: use type declaration for type-safe when declare initialState

Seung-wan:
i think Type Declaration is type-safe than Type Assertion.

```ts
interface CounterState {
  value: number
}

// Type Declaration
const initialState: CounterState = { value: 0 }
const initialState: CounterState = { value: 0, id: 1 } // Error: ...

// Type Assertion
const initialState = { value: 0 } as CounterState
const initialState = { value: 0, id: 1 } as CounterState // Not error, but id is not type of CounterState.

```

### DIFF
--- a/docs/api/createSlice.mdx
+++ b/docs/api/createSlice.mdx
@@ -25,7 +25,7 @@ interface CounterState {
   value: number
 }
 
-const initialState: CounterState = { value: 0 }
+const initialState = { value: 0 } satisfies CounterState as CounterState
 
 const counterSlice = createSlice({
   name: 'counter',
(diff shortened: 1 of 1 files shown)


phryneas:
No, we show that pattern very deliberately as TypeScript's flow analysis otherwise sometimes gives weird results. See #827

If you want to carry this forward, we'd be happy for a PR that adds a `satisfies`, but the `as` is required.
```diff
-const initialState = { value: 0 } as CounterState
+const initialState = { value: 0 } satisfies CounterState as CounterState
```

Seung-wan:
@phryneas 
Okay. I think I'll just send out a new PR when I'm done.

PR #3623, opened 2023-07-27
docs: add satisfies keyword with type assertion

Seung-wan:
Add typescript `satisfies` keyword with type assertion.


phryneas:
Looking good, thanks!

markerikson:
This unfortunately broke our docs build. Reverting it.

PR #3624, opened 2023-07-27
docs: fix typo

Seung-wan:
I think this is typo.


phryneas:
The second one surely, on the first I'm not entirely sure - @markerikson ?

markerikson:
Eh, it works either way.

Issue #3596, opened 2023-07-15
[RED-25] createEntityAdapter actions cannot be passed to reducers in createSlice call from generic enhancer


Olesj-Bilous:
[Here](https://codesandbox.io/s/hidden-sun-j4cr65?file=/src/problemSlice.tsx) is a sandbox with a minimal example. It follows below as well.

I tried to stay as close as possible to the example provided in the API [reference](https://redux-toolkit.js.org/api/createEntityAdapter), switching Book for a generic TModel.

``` ts
import { createSlice, createEntityAdapter } from "@reduxjs/toolkit";

function modelSliceEnhancer<TModel extends { id: string }>(name: string) {
  const modelAdapter = createEntityAdapter<TModel>({
    selectId: (model) => model.id // just checking
  });

  return createSlice({
    name: name + "/state",
    initialState: modelAdapter.getInitialState(),
    reducers: {
      /*
        Type '{ <S extends EntityState<TModel>>(state: IsAny<S, EntityState<TModel>, S>, entity: TModel): S; <S extends EntityState<TModel>>(state: IsAny<...>, action: { ...; }): S; }' is not assignable to type 'CaseReducer<EntityState<TModel>, { payload: any; type: string; }>'.
          Types of parameters 'state' and 'state' are incompatible.
            Type 'WritableDraft<EntityState<TModel>>' is not assignable to type 'EntityState<TModel>'.
              Types of property 'entities' are incompatible.
                Type 'WritableDraft<Dictionary<TModel>>' is not assignable to type 'Dictionary<TModel>'.
                  'string' index signatures are incompatible.
                    Type 'Draft<TModel> | undefined' is not assignable to type 'TModel | undefined'.
                    [...]
                      Type 'WritableDraft<TModel>' is not assignable to type 'TModel'.
                        'WritableDraft<TModel>' is assignable to the constraint of type 'TModel', but 'TModel' could be instantiated with a different subtype of constraint '{ id: string; }'.
                          Type 'Draft<TModel[K]>' is not assignable to type 'TModel[K]'.
                            'TModel[K]' could be instantiated with an arbitrary type which could be unrelated to 'Draft<TModel[K]>'.
      */
      oneAdded/* <- ERROR */: modelAdapter.addOne
    }
  });
}
```

I found [this](https://github.com/reduxjs/redux-toolkit/issues/663) issue where seemed to be a similar predicament. The "extends any" magic suggested by a commenter there does not appear to work here, however.

[...]

PR #3669, opened 2023-08-23
[RED-24] Feature/entity adapter draftable entity state

Olesj-Bilous:
I got the go-ahead for this feature in [this issue](https://github.com/reduxjs/redux-toolkit/issues/3596)

Changes were restricted to types and typing in the `src/entities` directory

I introduced two new types:

```ts
DraftableEntityState<T> = EntityState<T> | Draft<EntityState<T>>

DraftableIdSelector<T> = IdSelector<T | Draft<T>>
```

### Draftable Entity State

This type was used to update the methods of `EntityStateAdapter<T>` interface, e.g. `addOne<S extends DraftableEntityState<T>>`

_(Perhaps we could lift S up to EntityStateAdapter, as_ `S extends DraftableEntityState<T> = EntityState<T>`_? I left it unchanged for now)_

`createStateOperator` was updated to reflect this change, which eliminated the need for a ts-ignore

```ts
const isDraftTyped = isDraft as <T>(value: any) => value is Draft<T>

function createStateOperator<V, R>(
  mutator: (arg: R, state: DraftableEntityState<V>) => void
) {
  return function operation<S extends DraftableEntityState<V>>(
    state: S,
    arg: R | PayloadAction<R>
  ): S {
    //
    // ...
    //
    if (isDraftTyped<EntityState<V>>(state)) {
      runMutator(state)
      return state
    }
    return createNextState(state, runMutator)
  }
}
```

### Draftable Id Selector

The draftable id selector type seems necessary because `createUnsortedStateAdapter<T>(selectId: IdSelector<T>)` returns an adapter with method `updateMany`

`updateMany` is generated within `createUnsortedStateAdapter` by a `createStateOperator(updateManyMutably)` call whereby `updateManyMutably` is set up to invoke a local function `takeNewKey(keys, update, state): boolean`

`takeNewKey` gets the original entity from `state` to see if `update` specifies a new id value:

```ts
const original = state.entities[update.id]
const updated: T = Object.assign({}, original, update.changes)
const newKey = selectIdValue(updated, selectId)
const hasNewKey = newKey !== update.id
```

However, since I changed the type of state to extend DraftableEntityState<T>, the second line needs to be `const updated: T | Draft<T>` and the `selectId` parameter to `createUnsortedStateAdapter` should be `DraftableIdSelector<T>`

`selectId` parameter type to upstream functions `createSortedStateAdapter` and `createEntityAdapter` was updated accordingly

### Conclusion

While the `DraftableEntityState` addition seems harmless, the `DraftableIdSelector` addition seems as though it may force clients to account for the possibility that the `model` parameter to `selectId` is `Draft<T>` when passing `selectId` as a parameter to `createEntityAdapter`

However, as evidenced in the existing `tests/entity_state.test.ts`, in effect:

```ts
createEntityAdapter({
  selectId: (book: BookModel) => book.id,
})
```

The type-updated `createEntityAdapter` correctly recognizes selectId as `DraftableIdSelector<BookModel>`

Only if we try to call `createEntityAdapter` from a _generic_ entitySliceEnhancer do we get an error `'IdSelector<TModel>' is not assignable to type 'DraftableIdSelector<TModel>'`

That is to say, only those who opt in to this newly available functionality - by implementing a generic entitySliceEnhancer - will have to take into account that `selectId` needs to be `DraftableIdSelector<T>`

If such an entitySliceEnhancer were to be implemented in accordance with the above, a concretely typed IdSelector will be valid TypeScript again

```ts
const slice = entitySliceEnhancer({
  name: 'book',
  selectId: (book: BookModel) => book.id
}) // slice is inferred as Slice<EntityState<BookModel>>

function entitySliceEnhancer<T>({
  selectId
}: {
  selectId: DraftableIdSelector<T>
}) {
  const modelAdapter = createEntityAdapter<T>({
    selectId: selectId
  })
  return createSlice({name: '', initialState: modelAdapter.getInitialState()})
}
```

<sub>[RED-24](https://linear.app/redux/issue/RED-24/featureentity-adapter-draftable-entity-state)</sub>


Olesj-Bilous:
`CI / Test against dist (16.x)` failed because I didn't provide an actual test to the suite, I fixed that in the last commit. I am not sure where the other errors are coming from, or how I can get the API extractor to work

markerikson:
@Olesj-Bilous yeah, that docs build issue was due to another PR that got merged, which I've reverted.  Also, the `api-extractor` stuff is stale - we haven't used that in a while :)

markerikson:
@Olesj-Bilous After some discussion, we'd like to put this into 2.0 instead of 1.9.x.  That does mean this will need to be re-targeted at the `v2.0-integration` branch, and also updated to deal with the other entity adapter ID changes in that branch.  Could you update this accordingly?

Olesj-Bilous:
> @Olesj-Bilous After some discussion, we'd like to put this into 2.0 instead of 1.9.x. That does mean this will need to be re-targeted at the `v2.0-integration` branch, and also updated to deal with the other entity adapter ID changes in that branch. Could you update this accordingly?

@markerikson Alright I'll take care of that over the course of this week

markerikson:
Okay, looks good!

PR #3671, opened 2023-08-23
Add selector factories to createSlice

EskiMojo14:


markerikson:
Deferring to past 2.0.

EskiMojo14:
the conflicts on this are pretty nasty, so I think it would be simpler to close this and reimplement the changes later if we revisit the idea

Issue #3707, opened 2023-09-10
Support for explicit resource management "using statements" on RTK Query subscriptions


jared-ca:
## Background
RTK Query is built around the concept of consumers subscribing and unsubscribing from endpoints as required. The subscription count is used to determine when data is in use and caches can be cleared.

This is quite nice in React when using hooks as the hook lifetime is directly coupled to the subscription lifetime.

RTK Query supports being used in other frameworks and outside of React components where the subscription must be manually unsubscribed when finished with. 

This makes it easy to introduce subtle bugs especially when dealing with exceptions so ideally every manual dispatch of an `initiate` action should be unsubscribed in a finally block which can get messy when dealing with multiple subscriptions.

## Explicit resource management
Explicit resource management is a stage 3 TC39 proposal to add `using` statements to Javascript to solve this exact problem. It was recently shipped as part of Typescript 5.2 and is available in Javascript using babel.

See:
 - https://github.com/tc39/proposal-explicit-resource-management
 - https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/#using-declarations-and-explicit-resource-management
 - https://babeljs.io/docs/babel-plugin-proposal-explicit-resource-management

I would like to propose adding a `[Symbol.dispose]` method to the object returned from `dispatch(endpoint.initiate()` so it can cleaned up with a using block.

## What does that look like for consumers?

In this snippet we're cleaning up the subscriptions manually. I'm trying to simulate a real world scenario where the second subscription isn't always created so we need multiple finally blocks.

```
  const subscription1 = store.dispatch(pokemonApi.endpoints.getPokemonByName.initiate('bulbasaur'));
  try {
    const { data : pokemon1 } = await subscription1;
    processPokemon(pokemon1);
    if (shouldFetchPicachu()) {
      const subscription2 = store.dispatch(pokemonApi.endpoints.getPokemonByName.initiate('pikachu'));
      try {
        const { data : pokemon2 } = await subscription2;
        processPokemon(pokemon2);
      } finally {
        subscription2.unsubscribe();
      }
    }
  }
  finally {
    subscription1.unsubscribe();
  }
```

This is the same example again with using blocks demonstrating how much more ergonomic explicit resource management is.
```
  using subscription1 = store.dispatch(pokemonApi.endpoints.getPokemonByName.initiate('bulbasaur'));
  const { data : pokemon1 } = await subscription1;
  processPokemon(pokemon1);

  if (shouldFetchPicachu()) {
[...]

PR #3708, opened 2023-09-10
Support for explicit resource management "using statements" on RTK Query subscriptions

jared-ca:
Issue: #3707

This PR proposes the addition of a `[Symbol.dispose]: () => void` to be added to the object returned from `dispatch(endpoint.initiate())` to enable explicit resource management with using statements.

I've put effort into ensuring that none of the exported types reference `Symbol.dispose` when it's not available so consumers with older versions of Typescript won't have any type errors.

There are some explicit any casts within `disposableIfAvailable.ts` that allows this change to be compiled without needing to update the whole library to use Typescritp 5.2+ We're currently on 4.2.4. These any types are only internal to the module.

## Testing
I've manually tested this change using the example code snippet in #3707

### DIFF
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -16,6 +16,8 @@ import type { BaseQueryError, QueryReturnValue } from '../baseQueryTypes'
 import type { QueryResultSelectorResult } from './buildSelectors'
 import type { Dispatch } from 'redux'
 import { isNotNullish } from '../utils/isNotNullish'
+import type { DisposableIfAvailable } from '../utils/disposableIfAvailable'
+import { disposableIfAvailable } from '../utils/disposableIfAvailable'
 
 declare module './module' {
   export interface ApiEndpointQuery<
@@ -65,7 +67,7 @@ export type QueryActionCreatorResult<
   refetch(): QueryActionCreatorResult<D>
   updateSubscriptionOptions(options: SubscriptionOptions): void
   queryCacheKey: string
-}
+} & DisposableIfAvailable
 
 type StartMutationActionCreator<
   D extends MutationDefinition<any, any, any, any>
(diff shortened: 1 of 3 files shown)


phryneas:
While I generally love `using`, I'm not sure if it's actually useful, or if this case would just hide another bug.

If you really want to expose of the store value within a scope, chances are you don't really care about a store subscription in the first place, but more about the promise return value.

Right now we have the bug that if you call `initiate` with `{ subscribe: false }`, and no cache entry exists yet, the Promise returns a bogus value.
That would be solved with #3709 - so I'd rather have people using `initiate(x, {subscribe: false})` instead of adding `using` here to create something "long-lived" where we don't really care about it being "long-lived" at all.

jared-ca:
Cheers @phryneas,

I think you're probably right that most people only care about the promise!

In my case I actually care about the subscription and want to ensure the response stays in the cache for the lifetime of the scope because I have a bunch of computationally expensive business logic in memoized selectors that I want to make use of.

So in practice I'm writing listener middleware that looks something like this:
```
const listenerMiddleware = createListenerMiddleware()
listenerMiddleware.startListening({
  actionCreator: someAction,
  effect: async (action, api) => {
    using request1 = api.endpoints.request1();
    using request2 = api.endpoints.request2();
    using request3 = api.endpoints.request3();
    using request4 = api.endpoints.request4();

    await (Promise.all([request1, request2, request3, request4]));

    const shouldPerformAction = selectExpensiveUserIsCool(api.getState());
    if (shouldPerformAction) {
      await doSomethingElse();
    }
  }
})
```

Maybe I'm doing things a bit oddly but I love the pattern of having a flat redux store and then doing the cross store complex data wrangling in selectors, leveraging re-select memozing them all and re-evaluating them only when the dependent state changes.

Selectors are the hammer I like and everything looks like a nail perhaps but I end up with quite a large tree of memoized re-select selectors 😂 

From the React side it's great, just `use()` all the dependent apis and then the `selector` but using the same selectors outside of React, in this case in a middleware is a bit more awkward.

EskiMojo14:
closing as the linked issue has been closed - thanks for the PR regardless!

PR #3709, opened 2023-09-10
keep subscription on data while query is running

### LINKED ISSUE #3706: BUG: UseLazyXXXQuery Returning undefined On Long Running Queries
I am in the process of optimizing my website to handle efficiently querying from tables with potentially hundreds of millions of records, and for those queries that are presently not optimized I have noticed the occasional odd error from RTK Query. In particular, I have noticed that when the queries start taking minutes to resolve (and they eventually do resolve successfully) that sometimes RTK Query returns undefined from the promise. IE...

```
const [trigger] = useLazyXXXQuery();
...
const query = trigger(...);
query.then(response => {
  try {
    const { field1, field2, ... } = response.data;
    ...
  } catch(e) {
    console.error(e); // => logs the error: "TypeError: response is undefined"
    console.log(`response status: ${response.status}, isSuccess? ${response.isSuccess}, isError? ${response.isError}, isLoading? ${response.isLoading}, isUninitialized? ${response.isUninitialized}`);
    console.log(`response.Error: ${JSON.stringify(response.error)}`);
  }
});
```

When the error occurs, here is the log trace: 
```
TypeError: response.data is undefined
response status: uninitialized, isSuccess? false, isError? false, isLoading? false, isUninitialized? true
response.Error: undefined
```

NOTE: I have confirmed that the requests do resolve properly (with the expected data) by monitoring the requests via the Network tab in the Developer Tools on Firefox. Also, there are usually several similar requests being sent at once (via the 3rd party component AgGrid), but with different arguments. Some resolve successfully, others are resolving prematurely and getting the above error.

-----------------------
EDIT: I removed the unwrap statement and added some more trace
EDIT 2: Possibly the same bug as https://github.com/reduxjs/redux-toolkit/issues/3662

phryneas:
This would fix #3706 and solve a generally long-existing problem: Prior to this PR, unsubscribed `initiate` calls without an existing cache entry don't have a return value.

The downside is that now every `initiate` call creates a cache entry, even with `subscribe: false` - and trigger lifecycle actions.  
It might make sense to target 2.0 with this.

### DIFF
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -20,6 +20,7 @@ import type { BaseQueryError, QueryReturnValue } from '../baseQueryTypes'
 import type { QueryResultSelectorResult } from './buildSelectors'
 import type { Dispatch } from 'redux'
 import { isNotNullish } from '../utils/isNotNullish'
+import { countObjectKeys } from '../utils/countObjectKeys'
 
 declare module './module' {
   export interface ApiEndpointQuery<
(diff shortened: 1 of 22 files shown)


EskiMojo14 (approved):

markerikson:
I'm admittedly hesitant on this one because I don't like anything that involves dispatching _more_ actions.  But both of you seem to think this is a good idea, so we'll go with it. Retargeted for 2.0 and rebased.

markerikson:
I tried doing a perf check on this, and I'm pretty concerned.

- Setup:
  - https://github.com/phryneas/rtkq-performance-testbench
  - Updated to first build and run with 2.0.0-beta.3, then this PR
- Steps:
  - loaded the page and hit "Record"
  - changed the "num components" counter from 10 -> 100 -> 1000
  - scrolled to the bottom of the list
  - waited for them to finish loading
  - hit "Invalidate" and waited for them to finish loading

here's the "Invalidate" reload activity section for each, beta.3 on the left , PR on the right.  Note that beta.3 is like 2800 ms (WHICH IS STILL REALLY BAD) and PR is 6000+ seconds (ACK):

image

The issue here is that every dispatching is running in a microtask and taking up another 2-5ms.  I haven't counted the number of actions, but the amount of work here is going up significantly somehow.

This is sort of tied into the way we're mutating the subscription data in the RTKQ middleware, and then trying to track that and generate diffs via `JSON.parse(JSON.stringify())` before and after so that we can `dispatch(subscriptionsUpdated(patches))`.

At this point I'm inclined to kick this to "Post 2.0" (and likely to RTK 3.0) on the grounds that this needs more time to be thought through, and I'm trying to be pretty ruthless about cutting scope to get this out the door.

markerikson:
Did a bunch of work on top of this in #3824 around changing how often we sync the subscription state, and also reworked how hooks access the current subscription data.  Perf seems equivalent now.

PR #3728, opened 2023-09-16
[RED-23] fix: Updated type references to resolve portable types issue

tdurnford:
- Tries to fix all of our uses of `/dist/` in type imports, by re-exporting needed types

<sub>[RED-23](https://linear.app/redux/issue/RED-23/fix-updated-type-references-to-resolve-portable-types-issue)</sub>


jonchurch:
FYI it appears the imports that reference dist instead of src was added in #1093

markerikson:
Well, the type tests and CI checks all pass.  Let's give this a shot.

PR #3738, opened 2023-09-23
Assorted outstanding docs improvements

### LINKED ISSUE #1423: actions don't get inferred when the type's fields are all optional
Hi, thanks for the library.

Probably, some mistake from my side. But, I don't get the action inferred when all the type's fields are optional. Shouldn't toolkit infer the action type by the generic of `PayloadAction`?

ok:
image

with error:
image

### LINKED ISSUE #2130: Assorted docs tweaks suggestions
Per https://www.reddit.com/r/reactjs/comments/tfjz8i/my_opinion_on_reduxtoolkit_rtk/i0wuvkf/ :

> While my memories from reading the docs are still fresh, here are two things I was struggling with yesterday:

> 1. I was reading through an example in RTKQ docs section about [mutations](https://redux-toolkit.js.org/rtk-query/usage/mutations), which had the following code: updatePost: build.mutation<Post, Partial<Post> & Pick<Post, 'id'>>. It wasn't obvious from the generic which of the two parameter types was the type of the data sent to the server, and which of the value returned from the endpoint's thunk. It looked, judging by the name of the function, that the first type parameter to the generic was the return type and the second parameter was the type of the argument of the updatePost function; but it took me a while to locate the section in the docs that would confirm this. It was on a different page altogether.
> 2. Reading (or rather, skimming) through the listener middleware page, it was hard to discover the full list of functions exposed by the listenerApi. Re-reading this page now, I realize that I made a mistake — I assumed that the functions takeLatest or takeLeading mentioned in the beginning of the page were the part of the exposed api; searched for them in the page, and didn't find their description, which left me puzzled. I can now see that all functions are listed below in the "Listener API" section, and that takeLatest was a bit of a distraction, an api from redux saga that can be emulated via other methods of the listener middleware. Perhaps listing all methods of the listener api before describing each in more detail would help. Or maybe it's fine as it is.

### LINKED ISSUE #2238: Clarify `configureStore` intent and behavior in API ref
The API ref for `configureStore` ( https://redux-toolkit.js.org/api/configureStore ) doesn't give any real details on what it actually _does_, how it compares to `createStore`, or why you should use it instead.  It just says it's a "friendly abstraction".

We should fill in a lot more details to explain the technical differences and intent.

Lenz has a good explanation in a Twitter thread starting here that we should copy-paste as a starting poing:

https://twitter.com/phry/status/1514338477810360322

(While we're at it, we should review the other docs pages to see if similar explanations should be added.)

### LINKED ISSUE #2926: Improve `extraReducers` docs to explain purpose
Per https://www.reddit.com/r/reactjs/comments/yzcegz/when_using_redux_it_is_better_to_go_basic_than/ix0g24j/ , the current `extraReducers` docs explain syntax, but not anything about the purpose or use cases for `extraReducers`.  We should rework this.

### LINKED ISSUE #3113: Add a better example of writing an RTKQ `queryFn`
The RTKQ `queryFn` option is meant for making arbitrary async calls instead of calling through `baseQuery`.  This is most commonly used for making requests with a user-provided async function or an external SDK, like Supabase or Firebase.

However, the "Customizing Queries" page ( https://redux-toolkit.js.org/rtk-query/usage/customizing-queries ) doesn't show any examples along those lines.  It talks about the arguments, and shows an example with `Math.random()`, and then later at the bottom it shows a couple other examples, but they don't feel like they cover the most common use case, along the lines of:

```ts
createApi({
  baseQuery,
  endpoints: (build) => ({
    myCustomEndpoint: build.query({
      queryFn: async (someId) => {
        const result = someAsyncMethod(id);
        return {data: result};
      }
    })
  })
})
```

We should add a couple examples like this, ideally one with a user-provided method, and one with a real third-party SDK.

markerikson:
This PR:

- Updates the `configureStore` API reference to clarify its purpose and what it does
- Updates the `createSlice` API reference to clarify what `extraReducers` does, and warn about the removal of the object form
- Fixes broken links in `matching-utilities`
- Updates the "Customizing Queries" page to give a couple simple examples of using `queryFn` with a random async API and a third-party SDK, and emphasizes _returning `{error}` from both base queries and query functions
- Clarifies the `<ReturnType, ArgType>` generics for RKTQ queries and mutations
- Documents the `AtLeastOne<T>` workaround for partial payload fields

Fixes #2926
Fixes #2238 
Fixes #3113 
Fixes #2130 
Fixes #1423


docs/api/createSlice.mdx
```diff
@@ -162,6 +163,14 @@ See [the "Builder Callback Notation" section of the `createReducer` reference](.
 
 ### The `extraReducers` "map object" notation
 
+:::caution
+
+The "map object" notation is deprecated and will be removed in RTK 2.0. Please migrate to the "builder callback" notation, which offers much better TypeScript support and more flexibility.
```
EskiMojo14, line 168: worth linking to the codemod page?

PR #3749, opened 2023-09-25
Try fixing some lockfile issues

markerikson:


markerikson:
And that broke going the other way.

Going to just tell Yarn to allow lockfile changes when running React-Redux's CI checks, not like the changes are critical.

PR #3766, opened 2023-10-02
Add example to refetch request in `usage without react hooks` documentation page

Azeirah:
I experienced the issue that my endpoint wouldn't get refetched when I subscribed to the cache using the code example on the `usage without react hooks` page. The problem was that subscribing isn't enough, you need to call `refetch()` on the subscription.

This PR adds a short comment to make clear that the subscription is the object you can _use_ to interact with the cache, not sending the request itself.

I believe that wanting to re-send the request is a common enough use-case to be mentioned directly in the first example.

The wording can of course be changed if preferred.


markerikson:
Thanks!

Issue #3765, opened 2023-10-02
rtk-query-codegen-openapi has stopped working. 


kahirokunn:
I ran test with the following command and got an error. 😢 
I tried yarn versions 2~4, but I couldn't figure it out.
Do you know what could be the cause?
I'm not sure when I last ran it, but I think it happened after `rm -rf node_modules` and then `yarn install` again. 😭 

```sh
$ cd redux-toolkit/packages/rtk-query-codegen-openapi
$ yarn install
$ yarn test
 FAIL  test/cli.test.ts (8.155 s)
  ● CLI options testing › generation with `config.example.js`

    expect(received).toEqual(expected) // deep equality

    - Expected  -  3
    + Received  + 21

      Object {
    -   "error": null,
    -   "stderr": "",
    +   "error": Error {
    +     "cmd": "/Users/kahiro/Documents/redux-toolkit/packages/rtk-query-codegen-openapi/node_modules/ts-node/dist/bin.js -T -P /Users/kahiro/Documents/redux-toolkit/packages/rtk-query-codegen-openapi/tsconfig.json /Users/kahiro/Documents/redux-toolkit/packages/rtk-query-codegen-openapi/src/bin/cli.ts ./config.example.js",
    +     "code": 1,
    +     "killed": false,
    +     "signal": null,
    +   },
    +   "stderr": "DeprecationWarning: 'createTypeAliasDeclaration' has been deprecated since v4.8.0. Decorators are no longer supported for this function. Callers should switch to an overload that does not accept a 'decorators' parameter.
    + DeprecationWarning: 'createParameterDeclaration' has been deprecated since v4.8.0. Decorators have been combined with modifiers. Callers should switch to an overload that does not accept a 'decorators' parameter.
    + DeprecationWarning: 'createIndexSignature' has been deprecated since v4.8.0. Decorators are no longer supported for this function. Callers should switch to an overload that does not accept a 'decorators' parameter.
    + TypeError: Cannot read properties of undefined (reading 'length')
    +     at emitUnparsedSourceOrPrepend (/Users/kahiro/Documents/redux-toolkit/packages/rtk-query-codegen-openapi/src/generate.ts:90429:61)
    +     at pipelineEmitWithHintWorker (/Users/kahiro/Documents/redux-toolkit/packages/rtk-query-codegen-openapi/src/generate.ts:90150:24)
[...]

PR #3771, opened 2023-10-03
chore: Fixed oazapfts typescript version in rtk-query-codegen-openapi

kahirokunn:
https://github.com/reduxjs/redux-toolkit/issues/3765

The reason for this is that codegen's TypeScript version was 4.5, but oazapfts' TypeScript version was 4.9.2, which broke compatibility.

To fix this problem, it would be best to use the resolution field and fix the oazapfts TypeScript version to the user's TypeScript version.

The test timed out, so the test time was extended.
https://github.com/reduxjs/redux-toolkit/actions/runs/6388673287/job/17338896802

### DIFF
--- a/packages/rtk-query-codegen-openapi/package.json
+++ b/packages/rtk-query-codegen-openapi/package.json
@@ -63,6 +63,9 @@
     "swagger2openapi": "^7.0.4",
     "typescript": ">=4.1 <=4.5"
   },
+  "resolutions": {
+    "typescript": ">=4.1 <=4.5"
+  },
   "husky": {
     "hooks": {
       "pre-commit": "pretty-quick --staged"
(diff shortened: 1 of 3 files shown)


kahirokunn:
@phryneas 
Hello!
I would be happy if you could review this when you have time.

phryneas:
Hey @kahirokunn, resolutions won't help here for the npm package, as package mangers will ignore that field when installing a dependency. It seems like we'd need to upgrade the codegen to support newer TS versions. I know that @msutkowski wanted to go through a bunch of open PRs, but I'm not sure if that was also on his codegen-related TODO list. Maybe the two of you could coordinate a bit?

kahirokunn:
@msutkowski Hello.
How do you plan to tackle this issue?

phryneas:
Seems like the TS update has already happened - closing :)

PR #3774, opened 2023-10-03
docs(rtk-query): add headers argument to base query example using axios

reza-sadeghzadeh:
- include the `headers` in the example


markerikson:
Thanks!

Issue #3683, opened 2023-08-29
Add additional cancellation primitives to `listenerApi`


markerikson:
We should add `listenerApi.throwIfCanceled()`.  Trying to rework some Replay listener logic that does a bunch of "if we ever moved to a different pause, cancel the current workflow in progress" stuff.  we do have `listenerApi.signal.throwIfAborted()`, but I'm not 100% sure I'm hitting or using that right, and I think having a top-level method would be nice.

Also, we should add `listenerApi.cancel()` too.  `cancelActiveListeners()` cancels _other_ instances, but I could see use cases for canceling _this_ one, like from deeply nested logic (a forked task or something)

PR #3775, opened 2023-10-04
Created listenerApi.cancel()

julian-ford:
#3683

Created `listenerApi.cancel()`, a counterpart to `cancelActiveListeners()`.

`throwIfCanceled()` not included.


markerikson:
I'll try to look at this over the weekend!

markerikson:
Looks nice, thank you!

(Only other suggestion I can immediately think of would be having the test do the cancelation from within a forked child task, to help show that it bails out, but this seems to show that the cancelation did indeed happen.)

Issue #3710, opened 2023-09-10
Cache handling not working on failed requests


nikischin:
I have two APIs. One to log in with my credentials and one to access data.

The second API for the data requires the token from the first request. In some cases, the second API might be called without the user being logged in. In that case it does return a `401 Unauthorized` HTTP Status Code.

```
export const generalApi = createApi({
    reducerPath: 'api',
    tagTypes: [ 'data' ],
    baseQuery: fetchBaseQuery({
        baseUrl: 'https://api.mydomain.com/v1/',
        prepareHeaders: (headers, { getState })=> {
            const token = getState().session.token;

            if (token) {
                headers.set('authorization', `Bearer ${token}`);
            }

            return headers;
        }
    }),
    endpoints: (builder)=> ({
        getData: builder.query({
            query: ({ id })=> 'getData/' + encodeURIComponent(JSON.stringify({ id: id })),
            transformResponse: (response)=> response.dataContent,
            transformErrorResponse: (response)=> response.error,
            providesTags: [ 'data' ],
        }),
        login: builder.mutation({
            query: (body)=> ({
                url: 'login',
                method: 'POST',
                body: body
            }),
            invalidatesTags: [ 'data' ]
        }),
    }),
});
```

Where the token is stored in a separate slice
```
const sessionSlice = createSlice({
    name: 'session',
    initialState: { token: null },
    reducers: {},
    extraReducers: (builder)=> {
        builder.addMatcher(
            generalApi.endpoints.login.matchFulfilled,
            (state, { payload })=> {
                state.token = payload.token ?? null;
            }
        );
    },
});
```

**Desired behavior:**
After successful login I expect the tag 'data' to be invalidated and the getData API to be called again. However, as it ran into a `401` before, nothing does happen and my component which implemented the `useGetDataQuery()` doesn't see any update. If I disable the 401 on server side, the caching invalidation works as expected, though I personally feel like this is a rather ugly implementation. Another workaround would be to implement a `skip` condition, though there might be cases where I cannot predict and avoid some `401` errors on client side.

nikischin:
For anyone running into the same issue: 

As the query does not run successfully on the first run, it is not being cached. This kind of makes sense indeed if you know about it. So the providedTags are not being considered and cannot be invalidated. Also there is no other way the query would be triggered again automatically.

However, if you use a callback on the provided tags like this it does also provide tags on failed requests, which can be invalidated.
```
//Also provide Tags on failed Queries
[...]

PR #3776, opened 2023-10-04
Add Info for cache tags

nikischin:
PR for #3710

### DIFF
--- a/docs/rtk-query/usage/automated-refetching.mdx
+++ b/docs/rtk-query/usage/automated-refetching.mdx
@@ -38,7 +38,8 @@ A _query_ can have its cached data _provide_ tags. Doing so determines which 'ta
 
 The `providesTags` argument can either be an array of `string` (such as `['Post']`), `{type: string, id?: string|number}` (such as `[{type: 'Post', id: 1}]`), or a callback that returns such an array. That function will be passed the result as the first argument, the response error as the second argument, and the argument originally passed into the `query` method as the third argument. Note that either the result or error arguments may be undefined based on whether the query was successful or not.
 
-Please note, tags provided in an simple array (such as `['Post', 'User']`) and not a callback are only considered for successful requests. In order to provide tags also for failed requests the callback has to be used. With the callback it is also possible providing different tags for failed or successful requests.
+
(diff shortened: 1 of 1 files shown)


phryneas:
I'm pretty sure that's incorrect - the plain string array would be considered in success and error cases, and you need the function if you want those to return different values depending on success or not.

nikischin:
> I'm pretty sure that's incorrect - the plain string array would be considered in success and error cases, and you need the function if you want those to return different values depending on success or not.

In #3710 I experienced differently. However, I am happy to be proofed wrong. I wasn't able to find this part in the code, do you know which part of the code is responsible for it?

phryneas:
That would be 

https://github.com/reduxjs/redux-toolkit/blob/9c0d553c574691244e25ddcc262170e8744883b1/packages/toolkit/src/query/core/buildThunks.ts#L662-L678

 and 

https://github.com/reduxjs/redux-toolkit/blob/9c0d553c574691244e25ddcc262170e8744883b1/packages/toolkit/src/query/endpointDefinitions.ts

`providesTags` will only not behave like this if your queryFn throws an error instead of returning `{ error: ... }`

docs/rtk-query/usage/automated-refetching.mdx
```diff
@@ -38,6 +38,8 @@ A _query_ can have its cached data _provide_ tags. Doing so determines which 'ta
 
 The `providesTags` argument can either be an array of `string` (such as `['Post']`), `{type: string, id?: string|number}` (such as `[{type: 'Post', id: 1}]`), or a callback that returns such an array. That function will be passed the result as the first argument, the response error as the second argument, and the argument originally passed into the `query` method as the third argument. Note that either the result or error arguments may be undefined based on whether the query was successful or not.
 
+Please note, tags provided in an simple array (such as `['Post', 'User']`) and not a callback are only considered for successful requests. In order to provide tags also for failed requests the callback has to be used. With the callback it is also possible providing different tags for failed or successful requests.
```
markerikson, line 41:
```suggestion

Note that tags provided in an simple array (such as `['Post', 'User']`) and not a callback are only considered for successful requests. In order to provide tags for failed requests the callback has to be used. With the callback it is also possible to provide different tags for failed or successful requests.
```

markerikson:
Hmm. Sounds like there's some confusion going on here, actually.

EskiMojo14:
here's a quick sandbox showing that failed queries do provide tags when using the array form for providesTags: https://codesandbox.io/p/sandbox/failed-query-providing-tags-t4yjv7

PR #3781, opened 2023-10-06
Add an enhancer to combineSlices which allows it to dispatch an action upon inject

EskiMojo14:
This "solves" the main quirk with `combineSlices`, being that the injected state doesn't actually show up until an action is dispatched.


markerikson:
Deferring past 2.0 to see if this is really needed.

EskiMojo14:
closing as we're still not sure this is needed

PR #2854, opened 2022-10-31
Adds the option of generating real TS enums instead of string unions

labmorales:
Added an option of generating real TS enums on [oazapfts](https://github.com/cellular//pull/308) and was hoping to add the feature here as well 😄. I think a lot of people may benefit from this optional behavior.

It seems like the oazapfts version has all the features included on the oazapfts patched version therefore I replaced it with the new one that has the generate enums feature.

Right now it's just a config option, but I can create a CLI option for it as well. To use it the config file should pass the useEnumType option like this:

```ts
import type { ConfigFile } from '@rtk-query/codegen-openapi';

const config: ConfigFile = {
  schemaFile: './fixtures/petstore.yaml',
  apiFile: './fixtures/emptyApi.ts',
  outputFile: './tmp/example.ts',
  useEnumType: true,
};

export default config;
```

This fixes this [issue](https://github.com/reduxjs/redux-toolkit/issues/1889).

PR #3785, opened 2023-10-09
fix: adjust snake case regex

juliengbt:
Fix the snake case regex, related to #2854


GeorchW:
@juliengbt could you also update the snapshot?

PR #3780, opened 2023-10-06
Codegen: better handling of duplicate param names

kahirokunn:
When we generated this swagger, the names of the path parameter and query parameter were covered and broken, so we fixed it.

```json
"/api/v1/namespaces/{namespace}/pods/{name}/proxy/{path}": {
  "get": {
    "tags": ["core_v1"],
    "description": "connect GET requests to proxy of Pod",
    "operationId": "connectCoreV1GetNamespacedPodProxyWithPath",
    "responses": {
      "200": {
        "description": "OK",
        "content": { "*/*": { "schema": { "type": "string" } } }
      },
      "401": { "description": "Unauthorized" }
    },
    "x-kubernetes-action": "connect",
    "x-kubernetes-group-version-kind": {
      "group": "",
      "version": "v1",
      "kind": "PodProxyOptions"
    }
  },
  "put": {
    "tags": ["core_v1"],
    "description": "connect PUT requests to proxy of Pod",
    "operationId": "connectCoreV1PutNamespacedPodProxyWithPath",
    "responses": {
      "200": {
        "description": "OK",
        "content": { "*/*": { "schema": { "type": "string" } } }
      },
      "401": { "description": "Unauthorized" }
    },
    "x-kubernetes-action": "connect",
    "x-kubernetes-group-version-kind": {
      "group": "",
      "version": "v1",
      "kind": "PodProxyOptions"
    }
  },
  "post": {
    "tags": ["core_v1"],
    "description": "connect POST requests to proxy of Pod",
    "operationId": "connectCoreV1PostNamespacedPodProxyWithPath",
    "responses": {
      "200": {
        "description": "OK",
        "content": { "*/*": { "schema": { "type": "string" } } }
      },
      "401": { "description": "Unauthorized" }
    },
    "x-kubernetes-action": "connect",
    "x-kubernetes-group-version-kind": {
      "group": "",
      "version": "v1",
      "kind": "PodProxyOptions"
    }
  },
  "delete": {
    "tags": ["core_v1"],
    "description": "connect DELETE requests to proxy of Pod",
    "operationId": "connectCoreV1DeleteNamespacedPodProxyWithPath",
    "responses": {
      "200": {
        "description": "OK",
        "content": { "*/*": { "schema": { "type": "string" } } }
      },
      "401": { "description": "Unauthorized" }
    },
    "x-kubernetes-action": "connect",
    "x-kubernetes-group-version-kind": {
      "group": "",
      "version": "v1",
      "kind": "PodProxyOptions"
[...]

PR #3430, opened 2023-05-09
Support OpenAPI readOnly/writeOnly properties

cvereterra:
Addresses https://github.com/reduxjs/redux-toolkit/issues/2742 in `rtk-query-codegen-openapi`

`oazapfts` added support for readOnly/writeOnly in [v4.7.1](https://github.com/oazapfts/oazapfts/releases/tag/v4.7.1). Relevant changes of this PR that must be looked at with care:

- Upgraded `oazapfts` to `4.7.1`
- Upgraded `typescript` to `^5.0.4`  
- Removed typescript check that enforced typescript version to be between `4.1` and `4.5`. I'm sure that this check was there for a reason, I'm open to help.
- `@rtk-query/oazapfts-patched` is not used anymore. If you need to, I can update the package to use `oazapfts@4.7.1`
- Added a test for the readOnly/writeOnly API spec

Cheers

PR #3798, opened 2023-10-13
Create RTK Query Code Generator for Open API Changelog

elliot-huffman:
Create the initial changelog.
Other changes are to be included in the future.

### DIFF
--- a/packages/rtk-query-codegen-openapi/ChangeLog.md
+++ b/packages/rtk-query-codegen-openapi/ChangeLog.md
@@ -1,0 +1,27 @@
+# Changelog
+
+All notable changes to the `RTK Query - Code Generator` for `Open API` project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.1.3] - 10-11-2023
+
+### Added
+- Support for Read Only Properties in the Open API spec. Previously, this property was ignored.   
+ Now if the readOnly property is present and set to `true` in a schema, it will split the type into two types: one with the read only property suffixed as 'Read' and the other without the read only properties, using the same type name as before.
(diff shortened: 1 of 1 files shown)


packages/rtk-query-codegen-openapi/ChangeLog.md
```diff
@@ -0,0 +1,27 @@
...
+All notable changes to the `RTK Query - Code Generator` for `Open API` project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/), and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.1.3] - 10-11-2023
+
+### Added
+- Support for Read Only Properties in the Open API spec. Previously, this property was ignored.   
+ Now if the readOnly property is present and set to `true` in a schema, it will split the type into two types: one with the read only property suffixed as 'Read' and the other without the read only properties, using the same type name as before.
+This may cause issues if you had your OpenAPI spec properly typed/configured, as it will remove the read onyl types from your existing type. You will need to switch to the new type suffixed as 'Read' to avoid missing property names.
```
phryneas, line 12: There is a bit more of a change :) 

Let me try to get it together:

## 1.1.0 
* Adds the option of generating real TS enums instead of string unions #2854 
* Also ensures compatibility with TypeScript 5.0 versions - as a result also needs a higher TypeScript version to work with (old version range was 4.1-4.5)
* Also changes depenendcy from a temporarily patched old version of `oazapfts` back to the current upstream version

## 1.1.1
* Codegen: better handling of duplicate param names (https://github.com/reduxjs/redux-toolkit/pull/3780)
If a parameter name is both used in a query and a parameter, it will be prefixed with `query`/`param` now to avoid conflicts

## 1.1.2
* Support OpenAPI readOnly/writeOnly properties (https://github.com/reduxjs/redux-toolkit/pull/3430)

## 1.1.3
* adds a temporary workaround for https://github.com/oazapfts/oazapfts/issues/491

elliot-huffman:
Added the details you provided, I missed the implications of those while reading the commit messages.

phryneas:
Looking good, thanks a ton! 😊

PR #3800, opened 2023-10-13
Update error message: name of action type

### LINKED ISSUE #3799: Clarifying error message about type in addCase
I am newbie in RTK and when I met error `'builder.addCase' cannot be called with two reducers for the same action type` I cannot recognize which one of my types is duplcated.
I solved my problem with devtools logpoint but it's not convenient at all
image
image

It would be great if error will contain type. For example:
`'builder.addCase' cannot be called with two reducers for the same action type ${type}`

Kalibryyy:
Resolves #3799

### DIFF
--- a/packages/toolkit/src/mapBuilders.ts
+++ b/packages/toolkit/src/mapBuilders.ts
@@ -166,7 +166,7 @@ export function executeReducerBuilderCallback<S>(
       }
       if (type in actionsMap) {
         throw new Error(
-          '`builder.addCase` cannot be called with two reducers for the same action type ' + type
+          `\`builder.addCase\` cannot be called with two reducers for the same action type '${type}'`
         )
       }
       actionsMap[type] = reducer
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/mapBuilders.ts
```diff
@@ -166,7 +166,7 @@ export function executeReducerBuilderCallback<S>(
       }
       if (type in actionsMap) {
         throw new Error(
-          '`builder.addCase` cannot be called with two reducers for the same action type'
+          '`builder.addCase` cannot be called with two reducers for the same action type ' + type
```
markerikson, line 169: :hand: Can you make a couple tweaks here?

- Change the error string to be a template with backticks
- Use that to put quotes around the `type` variable in the string

(Since this error string _already_ has backticks in it, you'll have to escape those too.)

That way it's more clear what the actual string type value is.

PR #3820, opened 2023-10-24
Basic starting point for docs for the App Router

jherr:
A starting point for documentation on how to use Redux with the NextJS App Router and RSCs (a common question). This is very much a work in progress.


docs/tutorials/nextjs.mdx
```diff
@@ -0,0 +1,186 @@
...
+- Knowledge of React terminology: [JSX](https://reactjs.org/docs/introducing-jsx.html), [State](https://reactjs.org/docs/state-and-lifecycle.html), [Function Components, Props](https://reactjs.org/docs/components-and-props.html), and [Hooks](https://reactjs.org/docs/hooks-intro.html)
+- Understanding of [Redux terms and concepts](https://redux.js.org/tutorials/fundamentals/part-2-concepts-data-flow)
+- Working through the [Quick strt guide](./quick-start.mdx) is recommended
+
+:::
+
+## Introduction
+
+NextJS is a server side rendering framework for React that presents some unique challenges for using Redux properly. There are two architectures for a NextJS application; the Pages Router and the App Router. The Pages Router is the original architecture for NextJS. Using Redux with the Pages Router is well understood and handled primarily by the (next-redux-wrapper)[https://github.com/kirill-konshin/next-redux-wrapper]. This tutorial will focus on the App Router architecture as it is the new default architecture option for NextJS and it presents some unique challenges for using Redux properly.
```
markerikson, line 29: Early thoughts here:

- "Unique challenges" - like what?  How does this differ from a classical CRA/Vite SPA?  (maybe this gets answered later?)
- 
- "Well understood" - should probably link or point to more details here.  Might still be worth having a section somewhere in this page that talks about the challenges involved with the Pages router (multiple pages,  SSR pass,  creating multiple store instances,  passing data from server to client, etc.  There are still a lot of folks using Pages (maybe even in new projects?)

Not sure how that content should best be organized - we can iterate on that.

Conceptually, I think I'm sort of picturing a "why is any of this necessary / how is Next different than an SPA?" type section that lays out the differences, as context.

docs/tutorials/nextjs.mdx
```diff
@@ -0,0 +1,186 @@
...
+
+### How to Read This Tutorial
+
+This page assumes that you already have an exisiting NextJS application based on the App Router architecture.
+
+## The App Router Architecture and Redux
+
+The primary new feature of the NextJS App Router is the addition of support for React Server Components (RSCs). RSCs are a special type of React component that only renders on the server, as opposed to "client" components that render on **both** the client and the server. RSCs can be defined as `async` functions and return promises during rendering as they make async requests for data to render.
+
+RSCs abilitiy to block for data means that with the App Router you no longer have `getServerSideProps` to fetch data for rendering. Any component in the tree can make asychronous requests for data. While this is very convenient it also means thats if you define global variables (like the Redux store) they will be shared across requests. This is a problem because the Redux store could be contaminated with data from other requests.
```
markerikson, line 39: 💬 yep, the "shared across requests" bit is huge.

docs/tutorials/nextjs.mdx
```diff
@@ -0,0 +1,186 @@
...
+
+### Creating a Redux Store per Request
+
+Following along with the (Quick-Start guide)[./quick-start.mdx] we need to make some changes to the `app/store.js` file. The first change is to move from defining store as a global to defining a `createStore` function that returns a new store for each request.
+
+```ts title="src/app/store.ts"
+import { configureStore } from '@reduxjs/toolkit'
+
+export const createStore = () =>
```
markerikson, line 48: :hand: I'm hesitant to use `createStore` here because of the name collision with the Redux core API.

`makeStore`, maybe?

docs/tutorials/nextjs.mdx
```diff
@@ -0,0 +1,186 @@
...
+  configureStore({
+    reducer: {},
+  })
+
+// Infer the type of createStore
+export type StoreType = ReturnType<typeof createStore>
```
markerikson, line 54: :speech_balloon: Might be worth pointing out the "hey, we can pull off this inference trick without having the actual store instance variable yet" aspect.

docs/tutorials/nextjs.mdx
```diff
@@ -0,0 +1,186 @@
...
+// Infer the `RootState` and `AppDispatch` types from the store itself
+export type RootState = ReturnType<StoreType['getState']>
+// Inferred type: {posts: PostsState, comments: CommentsState, users: UsersState}
+export type AppDispatch = StoreType['dispatch']
+```
+
+To use this new `createStore` function we need to create a new "client" component that will create the store and share it using the React-Redux `Provider` component.
```
markerikson, line 61: :hand: Add some context (HAH!) for _why_ this has to be in a "client component" ?

docs/tutorials/nextjs.mdx
```diff
@@ -0,0 +1,186 @@
...
+import { createStore } from './store'
+// highlight-end
+
+export default function StoreProvider({
+  children,
+}: {
+  children: React.ReactNode
+}) {
+  // highlight-start
+  const storeRef = useRef<ReturnType<typeof createStore>>(createStore())
```
markerikson, line 97: :question: do we need to do the `if (!storeRef.current)` check here too?  Does this technically get rendered more than once?

markerikson (commented):
Looks like a pretty solid start!

Left some thoughts on where it would be useful to expand and provide more details.

I do think it would be helpful to provide some further context around _why_ all these changes are necessary, and if possible to add at least some info on how/why to do this with the Pages router as well.

I'm sorta split on whether this ought to be a "Tutorial" or "Usage Guide" page in terms of category and concepts.

Also, a few minutes ago I said "let's move this to the core docs site", but it might make sense to copy-paste this in both repos.  No strong feelings about where it goes first, we can move it around after it's complete.

Overall, _very_ useful and I like where this is going - THANK YOU!

docs/tutorials/nextjs.mdx
EskiMojo14, line 97: any reason this couldn't import the StoreType for here?

docs/tutorials/nextjs.mdx
```diff
@@ -0,0 +1,186 @@
...
+// highlight-end
+
+export default function StoreProvider({
+  count,
+  children,
+}: {
+  count: number
+  children: React.ReactNode
+}) {
+  const storeRef = useRef<ReturnType<typeof createStore> | null>(null)
```
EskiMojo14, line 160: same here re: StoreType

docs/tutorials/nextjs.mdx
```diff
@@ -0,0 +1,186 @@
...
+}) {
+  // highlight-start
+  const storeRef = useRef<ReturnType<typeof createStore>>(createStore())
+  // highlight-end
+
+  return <Provider store={storeRef.current}>{children}</Provider>
+}
+```
+
+If you need to intialize the store with data from the parent component then define that data as a property and use an action on the slice to set the data in the store as shown below.
```
EskiMojo14, line 104: my concern with this approach is that it essentially endorses setter actions, which we explicitly discourage

my first approach would be to allow makeStore to receive a preloadedState option - dispatching actions like this would be okay as an alternative but i'd prefer something a little less "setter"-y for the example if possible

EskiMojo14 (commented):
a few nitpicks, but the overall article makes sense to me and would be a good addition

docs/tutorials/nextjs.mdx
phryneas: I'll just point out that that is also true for the `pages` directory - there it's just not as much of a common problem, because `getServerSideProps` and `next-redux-wrapper` give you a pattern that doesn't involve globals. But if you use them, you're equally screwed.

docs/tutorials/nextjs.mdx
```diff
@@ -0,0 +1,186 @@
...
+// highlight-start
+export type StoreType = ReturnType<typeof createStore>
+// highlight-end
+// Infer the `RootState` and `AppDispatch` types from the store itself
+export type RootState = ReturnType<StoreType['getState']>
+// Inferred type: {posts: PostsState, comments: CommentsState, users: UsersState}
+export type AppDispatch = StoreType['dispatch']
+
+// file: app/StoreProvider.tsx
+;('use client')
```
phryneas, line 84: can we get prettier to work without changing the pragma into this form? It's probably not reasily recognizable like this.

docs/tutorials/nextjs.mdx
phryneas: +1 from my side, it would be a bit more consistent.

docs/tutorials/nextjs.mdx
phryneas: I'd prefer the action, as it could open the door for an interesting data prefetching pattern - but I agree, it should be less `setter`-y.

docs/tutorials/nextjs.mdx
```diff
@@ -0,0 +1,186 @@
...
+    // highlight-start
+    storeRef.current.dispatch(setCount(count))
+    // highlight-end
+  }
+
+  return <Provider store={storeRef.current}>{children}</Provider>
+}
+```
+
+The next step is to include the `StoreProvider` in your layout component. This will ensure that the store is created for each request and that the store is not shared across requests. Use the store exactly as you would normally using the hooks provided by `react-redux`.
```
phryneas, line 172:
```suggestion
The next step is to include the `StoreProvider` in your layout component. This will ensure that the store is created for each request and that the store is not shared across requests. In all client components further down the tree, you can use the store exactly as you would normally using the hooks provided by `react-redux`.
```

docs/tutorials/nextjs.mdx
```diff
@@ -0,0 +1,186 @@
...
+
+## What You've Learned
+
+That was a brief overview of how to set up and use Redux Toolkit with the App Router:
+
+:::tip Summary
+
+- **Create a Redux store per request by using `configureStore` wrapped in a `createStore` function**
+- **Provide the Redux store to the React application components** using a "client" component
+- **Use the store as you normally would using the hooks provided in react-redux**
```
phryneas, line 182: Again, could need a mention that this is only possible in client components.

docs/tutorials/quick-start.mdx
```diff
@@ -7,7 +7,7 @@ hide_title: true
 
 &nbsp;
 
-# Redux Toolkit Quick Start
+# NextJS Integration Tutorial
```
phryneas, line 10: Not sure if this change was intentional?

phryneas:
Our comments here might seem like we're nitpicking left and right, but I just want to say that I *really* appreciate the work you're doing here, @jherr!

jherr:
Nitpick away! This was really just an early early PR so that Mark could see
the basics of the approach.

One question though; what practically would be “less setter-ey”.

Also dispatching an action is an option here but not when setting state in
a pre-existing “global“ store on a per-route basis.

docs/tutorials/nextjs.mdx
markerikson: Yeah, I've done the "partial preloaded state" bit too in the one Next app I worked on, but I don't have access to that code any more. I'm sure I could find an example if I looked hard enough.

docs/tutorials/nextjs.mdx
```diff
@@ -0,0 +1,337 @@
...
+- **Server caching friendly** - Recent versions of NextJS (specifically applications using the App Router architecture) support aggressive server caching. The ideal store architecture should be compatible with this caching.
+
+There are two architectures for a NextJS application; the Pages Router and the App Router. The Pages Router is the original architecture for NextJS.
+
+Using Redux with the Pages Router is well understood and handled properly by the (next-redux-wrapper)[https://github.com/kirill-konshin/next-redux-wrapper]. This tutorial will focus on the App Router architecture as it is the new default architecture option for NextJS.
+
+### How to Read This Tutorial
+
+This page assumes that you already have an exisiting NextJS application based on the App Router architecture.
+
```
A7med3bdulBaset, line 43:
```suggestion

If you don't have one. run the following command to follow along 
```bash
npx create-next-app next-with-redux
```
```

markerikson:
Rebased this against `v2.0-integration`.  Will do some more editing and get this in shortly!  _THANK YOU_ for putting this together!

docs/tutorials/nextjs.mdx
markerikson: Given that this is more of a setup guide, maybe it's better to say "create a blank Next app, and also see the `next-with-redux` example as well".

jherr:
> Rebased this against `v2.0-integration`. Will do some more editing and get this in shortly! _THANK YOU_ for putting this together!

You mentioned at one point moving it out of RTK docs at all. Just putting that idea back into play. I'm good either way.

markerikson:
@jherr yeah, I'm gonna go ahead and land it in both docs sites.  Will nail down the content here first, then copy-paste it over to the core site.

markerikson:
Ironically, I just looked over at https://github.com/vercel/next.js/tree/canary/examples/with-redux/lib/redux , and they don't do the kinds of things we talk about here :)  There's a singleton store, etc.

We'll have to get that example updated to match this guide.

jherr:
Interesting. The React and Apollo folks were adamant about not using globals. And it makes sense to me because of the fact that you can have rendering suspended on a promise and have potential collision on the stores. Honestly, I think it's a good practice and in the case of Redux there is almost no additional overhead to creating a store per request. Because of the Redux provider there are no changes beyond when/where you create the store.

docs/tutorials/nextjs.mdx
A7med3bdulBaset: just found that `create-next-app` has [an official example for redux with the App Router with id `with-redux`](https://github.com/vercel/next.js/tree/canary/examples/with-redux)

It can be installed by `npx create-next-app --example with-redux`

markerikson:
Yeah.  I'm going to get this page done, merge it here and in the core docs, and then file a PR to update Next's `with-redux` example to match what this guide says :)

I'm doing some fiddling with folder structures and variable names.  Mostly going for the Redux logic in `/lib` (since that seems to maybe be a Next convention?), changing `StoreType` to `AppStore`, stuff like that.  

I'll ping you once I've got this updated - would appreciate a sanity check / review, since I've never tried any of this before! :)

markerikson:
@jherr okay, I think this is ready! Any feedback before I merge this?

Here's the previous page:

- https://deploy-preview-3820--redux-starter-kit-docs.netlify.app/usage/nextjs

markerikson:
Y'know what, let's just get this in. We can always edit it later!

jherr:
Sorry, was on a hike LGTM.

We might need to do a followup on how to use Server Actions in conjunction with Redux which, I'm not actually sure how to do. Might be something to send out into the community.

markerikson:
Hah, yes, we can absolutely iterate.  I'm just _so_ happy to finally have this topic covered!

**THANK YOU!!!!**

PR #3824, opened 2023-10-26
Rewrite RTKQ internal subscription lookups and subscription syncing

markerikson:
- Changed `subscriptionUpdated` from a microtask to a 500ms throttle
- ~~Exposed the RTKQ internal middleware state to be returned via an
  internal action, so that hooks can read that state directly without
  needing to call `dispatch()` on every render.~~
  Updated the `batchActions` middleware to return a `SubscriptionSelectors` object with `{getSubscriptions, getSubscriptionCount, isRequestSubscribed}`, and used that in the hook (and tests)
- Reworked tests to completely ignore `subscriptionsUpdated` in any
  action sequence checks due to timing changes and irrelevance
- Fixed case where `invalidateTags` was still reading from store state

### DIFF
--- a/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
@@ -9,6 +9,7 @@ import type {
   SubMiddlewareApi,
   InternalHandlerBuilder,
   ApiMiddlewareInternalHandler,
+  InternalMiddlewareState,
 } from './types'
 
 export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
@@ -19,6 +20,7 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
   api,
   assertTagType,
   refetchQuery,
+  internalState,
 }) => {
   const { removeQueryResult } = api.internalActions
   const isThunkActionWithTags = isAnyOf(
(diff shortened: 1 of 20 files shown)


packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -681,6 +682,27 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
         Definitions
       >
       const dispatch = useDispatch<ThunkDispatch<any, any, UnknownAction>>()
+      const internalStateRef = useRef<InternalMiddlewareState | null>(null)
```
phryneas, line 685: This logic in the hooks hile hurts a bit. Maybe the middleware could return a `hasSubscription` function instead of the internal state itself?

packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
```diff
@@ -67,7 +73,8 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
       const valuesArray = Array.from(toInvalidate.values())
       for (const { queryCacheKey } of valuesArray) {
         const querySubState = state.queries[queryCacheKey]
-        const subscriptionSubState = state.subscriptions[queryCacheKey] ?? {}
+        const subscriptionSubState =
+          internalState.currentSubscriptions[queryCacheKey] ?? {}
```
phryneas, line 77: Thank you for catching this one!

PR #3830, opened 2023-10-28
Update to Yarn 4

phryneas:
Some things still break in some examples, but that seems to be independent? Let's see what CI says now.


examples/publish-ci/cra4/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@reduxjs/toolkit": "^2.0.0-beta.2",
+    "@reduxjs/toolkit": ">=2.0.0-beta.3 || >=2.0.0-rc.0 || ^2.0.0",
```
jomarquez21, line 6: Why is this change necessary if the contribution is to update the Yarn version?
It's just a doubt.

aryaemami59:
I believe this is superseded by #4222.

PR #3833, opened 2023-10-29
Allow passing selector creators with different memoize options to getSelectors

EskiMojo14:
also fixes a few misplaced generics in tests

### DIFF
--- a/packages/toolkit/src/entities/state_selectors.ts
+++ b/packages/toolkit/src/entities/state_selectors.ts
@@ -2,9 +2,15 @@ import type { CreateSelectorFunction, Selector, createSelector } from 'reselect'
 import { createDraftSafeSelector } from '../createDraftSafeSelector'
 import type { EntityState, EntitySelectors, EntityId } from './models'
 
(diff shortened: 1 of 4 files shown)


packages/toolkit/src/entities/state_selectors.ts
```diff
@@ -20,7 +25,9 @@ export function createSelectorsFactory<T, Id extends EntityId>() {
     selectState?: (state: V) => EntityState<T, Id>,
     options: GetSelectorsOptions = {}
   ): EntitySelectors<T, any, Id> {
-    const { createSelector = createDraftSafeSelector } = options
+    let { createSelector } = options
+    createSelector ??= createDraftSafeSelector
```
EskiMojo14, line 29: for some reason typescript infers differently doing this vs the default assignment

with the default assignment it ends up as a union of the two types, rather than keeping the more generic type
image

the alternative would be 
```suggestion
    const { createSelector = createDraftSafeSelector as AnyCreateSelectorFunction } = options
```
i suppose

PR #3836, opened 2023-10-29
Explore simpler selector factory idea

EskiMojo14:


markerikson:
Seems like we're skipping this for now.  Can always resurrect later.

PR #3855, opened 2023-11-07
Throw an error if ApiProvider is nested inside a normal Provider.

### LINKED ISSUE #3852: Update `<ApiProvider>` to throw an error if it's used inside of another `<Provider>`
`<ApiProvider>` is _only_ meant for apps that don't already have a Redux store being created.  But, we've repeatedly seen people try to use something like `<Provider store={store}><ApiProvider>`, and then wonder why the app breaks.

We should add a debug check inside of `<ApiProvider>` that checks for the existence of `const store = useStore()` (or similar), and throw an error if it finds that above, to try to help users avoid that error.

EskiMojo14:
fixes #3852

### DIFF
--- a/packages/toolkit/src/query/react/ApiProvider.tsx
+++ b/packages/toolkit/src/query/react/ApiProvider.tsx
@@ -1,9 +1,10 @@
 import { configureStore } from '@reduxjs/toolkit'
 import type { Context } from 'react'
+import { useContext } from 'react'
 import { useEffect } from 'react'
 import React from 'react'
 import type { ReactReduxContextValue } from 'react-redux'
-import { Provider } from 'react-redux'
+import { Provider, ReactReduxContext } from 'react-redux'
 import { setupListeners } from '@reduxjs/toolkit/query'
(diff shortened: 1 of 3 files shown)


packages/toolkit/src/query/react/ApiProvider.tsx
```diff
@@ -37,6 +38,13 @@ export function ApiProvider<A extends Api<any, {}, any, any>>(props: {
   setupListeners?: Parameters<typeof setupListeners>[1] | false
   context?: Context<ReactReduxContextValue>
 }) {
+  const context = props.context || ReactReduxContext
+  const existingContext = useContext(context)
+  if (existingContext) {
+    throw new Error(
+      'Existing Redux context detected. If you already have a store set up, please use the traditional Redux setup.'
```
markerikson, line 45: :hand: This should result in a new error code being added.  Can you run `yarn build` locally, see if it updated `errors.json`, and push that if there is a change?

EskiMojo14: yep, done

Issue #3868, opened 2023-11-12
Update Deprecated `createStore` Usage in `createSlice.mdx`


PaoloDiBello:
### Problem
The current implementation in `createSlice.mdx` relies on the deprecated `createStore` method for Redux store configuration.

### Solution
Update the store configuration to use the recommended `configureStore` method, as `createStore` is now deprecated in the latest Redux versions.

### Steps to Reproduce
1. Navigate to `/docs/api/createSlice.mdx`.
2. Observe the usage of `createStore` for store configuration.

### Expected Behavior
The Redux store should be configured using the `configureStore` method to align with the latest best practices.

### Solution 

```diff

-import { createStore, combineReducers } from 'redux'
+import { configureStore, combineReducers } from 'redux'

-const reducer = combineReducers({
-  counter: counter.reducer,
-   user: user.reducer,
- })
-const store = createStore(reducer)
+const store = configureStore({
+  reducer: {
+    counter: counter.reducer,
+    user: user.reducer,
+  },
+})
```

### Additional Information
- Redux version: 4.2.1

### Note
This change is essential to maintain compatibility with the latest Redux updates and ensure a smooth migration path for future releases.

PaoloDiBello:
Issue resolved with merged pull request: https://github.com/reduxjs/redux-toolkit/pull/3869

PR #3869, opened 2023-11-12
Update store configuration using configureStore in createSlice.mdx

PaoloDiBello:
This pull request addresses the deprecation of createStore by updating the store configuration in createSlice.mdx to use the recommended configureStore method.

Closes issue: https://github.com/reduxjs/redux-toolkit/issues/3868

### DIFF
--- a/docs/api/createSlice.mdx
+++ b/docs/api/createSlice.mdx
@@ -235,7 +235,7 @@ Result's function `getInitialState` provides access to the initial state value g
 ```ts
 import { createSlice, createAction } from '@reduxjs/toolkit'
 import type { PayloadAction } from '@reduxjs/toolkit'
-import { createStore, combineReducers } from 'redux'
+import { configureStore, combineReducers } from 'redux'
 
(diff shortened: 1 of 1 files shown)


docs/api/createSlice.mdx
```diff
@@ -235,7 +235,7 @@ Result's function `getInitialState` provides access to the initial state value g
 ```ts
 import { createSlice, createAction } from '@reduxjs/toolkit'
 import type { PayloadAction } from '@reduxjs/toolkit'
-import { createStore, combineReducers } from 'redux'
+import { configureStore, combineReducers } from 'redux'
```
EskiMojo14, line 238: configureStore needs to be imported from @reduxjs/toolkit instead of the core package

EskiMojo14 (changes requested):

docs/api/createSlice.mdx
PaoloDiBello: Resolved

PR #3872, opened 2023-11-14
Create standardised methods of modifying reducer handler context.

EskiMojo14:


packages/toolkit/src/createSlice.ts
```diff
@@ -630,6 +630,43 @@ export function buildCreateSlice({ creators }: BuildCreateSliceConfig = {}) {
...
+          )
+        }
+        context.sliceCaseReducersByType[type] = reducer
+        return contextMethods
+      },
+      addMatcher(matcher, reducer) {
+        context.sliceMatchers.push({ matcher, reducer })
+        return contextMethods
+      },
+      exposeAction(name, actionCreator) {
```
EskiMojo14, line 660: the expose methods could also throw errors if there's name conflicts

PR #3885, opened 2023-11-16
Migration page

EskiMojo14:
Should be reusable for core site


website/src/css/custom.css
```diff
@@ -256,3 +256,9 @@ table.checkbox-table tbody td {
 .diagonal-cell--bottomLeft {
   grid-column-start: 1;
 }
+
+.migration-guide .typescript-only h4:after {
```
markerikson, line 260: :speech_balloon: clever :)

EskiMojo14: taken from the style guide, for consistency 😛

PR #3887, opened 2023-11-16
Create code gen module imports absolute paths

FabianFrank:
Use absolute paths for imports and module declarations to reflect how this would be used in an actual project that depends on redux toolkit, as opposed to how it is used inside the rtk monorepo.


FabianFrank:
Just struggled with updating my own custom code gen module for a few minutes to catch up with the changes from https://github.com/reduxjs/redux-toolkit/pull/3728 so I thought it might make sense to update the docs to reflect how the code would actually look like in a project that consumes rtk as a dependency.

markerikson:

PR #3894, opened 2023-11-18
Implement custom slice creators

### LINKED ISSUE #3837: Custom reducer creators
courtesy of @phryneas 

- creates an async thunk

EskiMojo14:
see https://github.com/reduxjs/redux-toolkit/issues/3837


packages/toolkit/src/tsHelpers.ts
```diff
@@ -205,3 +205,7 @@ export type Id<T> = { [K in keyof T]: T[K] } & {}
 export type Tail<T extends any[]> = T extends [any, ...infer Tail]
   ? Tail
   : never
+
+export type KeysMatching<T, V> = {
```
phryneas, line 209:
```suggestion
export type KeysForValueOfType<T, V> = {
```
`KeysMatching` sounded to me like `keyof T & keyof V`

packages/toolkit/tsconfig.base.json
```diff
@@ -32,18 +35,37 @@
...
+      ], // @remap-prod-remove-line
+      "@reduxjs/toolkit/query/react": [
+        "src/query/react/index.ts"
+      ], // @remap-prod-remove-line
       // for type imports in tests only
-      "@reduxjs/toolkit/dist/*": ["src/*"], // @remap-prod-remove-line
+      "@reduxjs/toolkit/dist/*": [
+        "src/*"
+      ], // @remap-prod-remove-line
       // for type imports in tests only
```
phryneas, line 58: These moved comments might prove problematic in some tests

EskiMojo14: prettier moment - will fix

packages/toolkit/src/createSlice.ts
```diff
@@ -585,17 +849,167 @@ function getType(slice: string, actionKey: string): string {
...
+        },
+      }[caseReducer.name],
+      {
+        _reducerDefinitionType: ReducerType.reducer,
+      } as const,
+    )
+  },
+  handle({ type, reducerName }, reducer, context) {
+    context
+      .addCase(type, reducer as any)
```
markerikson, line 870: :hand: this seems like a common sequence to do. Worth building it in?

packages/toolkit/src/createSlice.ts
```diff
@@ -673,30 +1089,51 @@ export function buildCreateSlice({ creators }: BuildCreateSliceConfig = {}) {
...
-        )
+    } else {
+      for (const [reducerName, reducerDefinition] of Object.entries(
+        options.reducers as SliceCaseReducers<State>,
+      )) {
+        const reducerDetails: ReducerDetails = {
+          reducerName,
+          type: getType(name, reducerName),
+        }
+        if ('reducer' in reducerDefinition) {
```
markerikson, line 1122: :hand: silly byte-shaving - can we do `const handler = 'reducer' in reducerDefinition ? preparedReducerCreator : reducerCreator`, and then unconditionally do `handler(reducerDetails, reducerDefinition, contextMethods)` ?

packages/toolkit/src/entities/slice_creator.ts
markerikson, line 1: :hand: For readability purposes, can you break all the entity creator stuff out into a separate PR?  I'd like this PR to focus on the core mechanism itself.

packages/toolkit/src/createSlice.ts
```diff
@@ -25,19 +30,317 @@ import type {
...
+          any
+        >
+          ? ReducerName
+          : never]: CaseReducers[ReducerName] extends { reducer: infer Reducer }
+          ? Reducer
+          : never
+      }
+    }
+  >
+  [ReducerType.asyncThunk]: ReducerCreatorEntry<
```
markerikson, line 140: :question: does it make any sense to break out the async thunk creator code and definitions into a separate file?  It'd still be built into RTK, but if there's already the ability to define custom creators anyway, it feels like this is both an example of a custom creator and something that could be pulled out from the `createSlice.ts` file

packages/toolkit/src/createReducer.ts
```diff
@@ -152,14 +152,7 @@ export function createReducer<S extends NotFunction<any>>(
   let [actionsMap, finalActionMatchers, finalDefaultCaseReducer] =
     executeReducerBuilderCallback(mapOrBuilderCallback)
 
-  // Ensure the initial state gets frozen either way (if draftable)
```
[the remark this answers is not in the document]
EskiMojo14: yep - currently we get `getInitialState` by building the reducer, but in order to expose it to the handler context it needs to be moved out so we can have it before the reducer is built

packages/toolkit/src/createSlice.ts
EskiMojo14: hmm, what would it be called?
currently a custom creator can just import reducerCreator and use it directly:

```ts
reducerCreator.handle(
  { type, reducerName },
  reducerCreator.create(reducer),
  context,
);
```

I do this in my history-adapter creator: https://github.com/EskiMojo14/history-adapter/blob/main/src/creator.ts#L122-L126

packages/toolkit/src/createSlice.ts
```diff
@@ -585,17 +849,167 @@ function getType(slice: string, actionKey: string): string {
...
-interface BuildCreateSliceConfig {
-  creators?: {
-    asyncThunk?: typeof asyncThunkCreator
+export const reducerCreator: ReducerCreator<ReducerType.reducer> = {
+  type: ReducerType.reducer,
+  create(caseReducer: CaseReducer<any, any>) {
+    return Object.assign(
+      {
+        // hack so the wrapping function has the same name as the original
+        // we need to create a wrapper so the `reducerDefinitionType` is not assigned to the original
```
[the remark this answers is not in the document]
EskiMojo14: we need to assign the _reducerDefinitionType brand to the reducer we return, so it can be passed to the right handler. however, we probably don't want to assign it to the function reference passed to the creator. this creates a new function reference with the same name as the one passed in (for trace purposes) and assigns the brand to that instead

docs/api/createSlice.mdx
```diff
@@ -662,3 +506,1305 @@ console.log(counter.actions.decrement.type)
...
+        },
+      },
+    ),
+  }),
+})
+
+export const { addTodo, deleteTodo, fetchTodo } = todosSlice.actions
+```
+
+#### RTK Creators
```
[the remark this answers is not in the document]
EskiMojo14: for reference, [this discussion](https://github.com/reduxjs/redux-toolkit/discussions/4020#discussioncomment-8057488) was the feedback I was trying to account for - currently we just throw `buildCreateSlice` out there without really explaining what it's doing, whereas ideally we'd explain how `buildCreateSlice` works *before* introducing our custom creators (`asyncThunk`, `entityMethods`) that you need `buildCreateSlice` for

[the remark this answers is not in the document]
markerikson: My thought was that the API reference page first focuses on introducing the callback syntax with just the built-in capabilities (`create.reducer/preparedReducer`), and briefly mentions "hey look you can customize this too" and points to the additional details.

packages/toolkit/src/createSlice.ts
EskiMojo14: ya - originally this made typescript happier but since we're `as any`ing anyway, might as well

packages/toolkit/src/createSlice.ts
EskiMojo14: done - SliceReducerCreators part needed to stay in createSlice.ts, but the rest has been moved

packages/toolkit/src/createSlice.ts
```diff
@@ -460,33 +742,18 @@ type SliceActionType<
  * @public
  */
 export type CaseReducerActions<
-  CaseReducers extends SliceCaseReducers<any>,
+  CaseReducers extends CreatorCaseReducers<State>,
   SliceName extends string,
-> = {
-  [Type in keyof CaseReducers]: CaseReducers[Type] extends infer Definition
```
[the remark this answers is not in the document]
EskiMojo14: you [can](https://github.com/reduxjs/redux-toolkit/pull/3894/commits/5f5fe7952630acc7b28de4d0e371522a23922cba), but you lose the ergonomics of `ReducerNamesOfType`. You end up always having to do
```ts
{
  [ReducerName in keyof CaseReducers]: CaseReducers[ReducerName] extends ReducerDefinition<typeof myCustomType> ? MyActualLogicHere : never
}
```

Definitions that aren't recognised by any handlers also get mapped to `unknown` instead of not being exposed at all: 
```ts
const slice = createSlice({
  name,
  initialState,
  reducers: {
    foo: "banana!"
  }
})

slice.actions.foo // previously not exposed as a key, now is `unknown`
```

the types are also having to do more work because there's 1) more intersections and 2) we're converting any `never`s into `unknown` ("foo" & never = never, "foo" & unknown = "foo")

packages/toolkit/src/entities/slice_creator.ts
EskiMojo14: done: https://github.com/reduxjs/redux-toolkit/pull/4223

packages/toolkit/src/createSlice.ts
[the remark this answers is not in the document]
EskiMojo14: removed `ReducerNamesOfType` and updated docs with conditional needed instead

docs/api/createSlice.mdx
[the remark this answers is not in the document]
EskiMojo14: yeah, my concern is that i don't want to bury create.asyncThunk or create.entityMethods too much, i think they'll be widely useful but only if known about

docs/api/createSlice.mdx
```diff
@@ -134,14 +134,18 @@ const todosSlice = createSlice({
 })
 ```
 
-### The `reducers` "creator callback" notation
+:::tip Slice creators and the `reducers` "creator callback" notation
```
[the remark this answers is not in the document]
EskiMojo14: yeah, the idea was to have a short "hey here's a preview of cool stuff, go here for more details" so `create.asyncThunk` wouldn't get buried

docs/api/createSlice.mdx
```diff
@@ -662,3 +506,1305 @@ console.log(counter.actions.decrement.type)
...
+
+const { addOneGoose, upsertManyGeese, removeAllGeese, ...etc } =
+  gooseSlice.actions
+```
+
+### Writing your own creators
+
+In version v2.2.0 (TODO), we introduced a system for including your own creators.
+
+The below documentation will cover how to write your own creators, and how to use them with `createSlice`.
```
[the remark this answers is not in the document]
EskiMojo14: should create.asyncThunk and create.entityMethods be documented here or on the usage guide?

PR #3906, opened 2023-11-23
fix isFSA import and avoid unnecessary type param

EskiMojo14:


packages/toolkit/src/listenerMiddleware/types.ts
```diff
@@ -581,10 +582,7 @@ export type FallbackAddListenerOptions = {
  */
 
 /** @public */
-export type GuardedType<T> = T extends (
-  x: any,
-  ...args: unknown[]
-) => x is infer T
+export type GuardedType<T> = T extends (x: any, ...args: any[]) => x is infer T
```
EskiMojo14, line 585: arguments are contravariant so this type wouldn't have worked for guards with other arguments
image

Issue #3854, opened 2023-11-07
rtk-query-codegen-openapi: circular reference definition


yizumi1012xxx:
An error occurs when generating a client for a definition that includes the following schema.
(schema has circular reference)
```yaml
    CategoryDto:
      type: object
      properties:
        parent:
          type: string
        sub:
          type: array
          items:
            $ref: '#/components/schemas/CategoryDto'
        id:
          type: string
        title:
          type: string
```

1.0.0 don't occur error, but 1.1.3 occur following error.
```txt
RangeError: Maximum call stack size exceeded
    at [Symbol.replace] (<anonymous>)
    at String.replace (<anonymous>)
    at /Users/y.izumi/workspace/karakuri-web/node_modules/oazapfts/src/codegen/generate.ts:129:29
    at Array.map (<anonymous>)
    at refPathToPropertyPath (/Users/y.izumi/workspace/karakuri-web/node_modules/oazapfts/src/codegen/generate.ts:129:6)
    at ApiGenerator.resolve (/Users/y.izumi/workspace/karakuri-web/node_modules/oazapfts/src/codegen/generate.ts:322:18)
    at ApiGenerator.checkSchemaOnlyMode (/Users/y.izumi/workspace/karakuri-web/node_modules/oazapfts/src/codegen/generate.ts:781:41)
    at ApiGenerator.checkSchemaOnlyMode (/Users/y.izumi/workspace/karakuri-web/node_modules/oazapfts/src/codegen/generate.ts:794:16)
    at /Users/y.izumi/workspace/karakuri-web/node_modules/oazapfts/src/codegen/generate.ts:808:14
    at Array.some (<anonymous>)
```

markerikson:
This is an issue with the upstream `oazapfts` lib - can you file or look for an issue there? Thanks!

milotoor:
For what it's worth, this used to work with version 1.0 of `@rtk-query/codegen-openapi`. I imagine it has to do with [this commit](https://github.com/reduxjs/redux-toolkit/commit/acca228c) which switched from using the `@rtk-query/oazapfts-patched` package to "updated `oazapfts`".

milotoor:
[...]

PR #3915, opened 2023-11-27
Update oazapfts to 4.10.1 to fix infinite recursion of checkSchemaOnlyMode

jarben:
This PR fixes an infinite recursion of `checkSchemaOnlyMode` by updating `oazapfts` dependency to `4.10.1`

More info: 
- https://github.com/reduxjs/redux-toolkit/issues/3854
- https://github.com/oazapfts/oazapfts/pull/489

### DIFF
--- a/packages/rtk-query-codegen-openapi/package.json
+++ b/packages/rtk-query-codegen-openapi/package.json
@@ -58,7 +58,7 @@
   "dependencies": {
     "@apidevtools/swagger-parser": "^10.0.2",
     "commander": "^6.2.0",
-    "oazapfts": "^4.8.0",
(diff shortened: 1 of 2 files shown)


phryneas:
This should not be necessary on our side. We have a depencency on `^4.8.0`, which translates to "any version that is major version 4 and at least 4.8".

You can just update `oazapfts`. It doesn't need any new release on our side.

jarben:
> This should not be necessary on our side. We have a depencency on ^4.8.0, which translates to "any version that is major version 4 and at least 4.8".
 
Yeah, of course, closing :)

Issue #3916, opened 2023-11-28
Promise.allSettled is not universally available


bever1337:
The Promise.allSettled API has 95% browser adoption according to caniuse. Promise.allSettled was not publicly exposed in earlier hermes-engine releases nor did RN expose that until RN ~72. Because redux-toolkit does not require allSettled, I believe redux-toolkit could replace allSettled with all for better compat.

aryaemami59:
Looks like this was resolved by #3917.

PR #3917, opened 2023-11-28
Task/remove all settled

bever1337:
See #3916


bever1337:
For some anecdotal evidence, I've already shipped this change as a patch in react-native environments. So far so good.

PR #3931, opened 2023-12-01
Avoid passing an identity function to createSelector

### LINKED ISSUE #3930: v2.0.0-rc.2 inundates console with warnings from runIdentityFunctionCheck
v2.0.0-rc.2 inundates consoles with warnings from [runIdentityFunctionCheck](https://github.com/reduxjs/reselect/blob/master/src/devModeChecks/identityFunctionCheck.ts). This is evident in the console output for the tests that were ran as part of #3928 (https://github.com/reduxjs/redux-toolkit/actions/runs/7055909073/job/19207110065)

EskiMojo14:
fixes #3930

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -1013,16 +1013,16 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
       )
 
       const { requestId } = promise || {}
-      const initialSelector = useMemo(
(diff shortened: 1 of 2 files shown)


EskiMojo14:
found it - useQuery passes noPendingQueryStateSelector as a default selector, but it's technically an identity function except if the selected state is uninitialised

this is good enough reason to disable the check in this case, i think - ultimately an identity function for selectFromResult isn't going to be the end of the world

PR #3935, opened 2023-12-02
Try switching EntityState back to default ID types

markerikson:


EskiMojo14:
posted this on discord but copying over here so it doesn't get lost

the vast majority of entities used with createEntityAdapter will not have the ID parameter as string | number, which means that this default will be incompatible with most cases

i'd argue having the type parameter required is a more understandable error rather than having type errors when a more specific type is required

PR #3948, opened 2023-12-05
Update docs to avoid circular type

### LINKED ISSUE #3947: Typing for predicate for `isHydrateAction` without causing circular reference to `RootState`
Hello!

Firstly, congrats on the v2 release 🎉

Been upgrading one of our sites to RTK 2.0 this morning and came across a classic circular reference issue as I updated the action handling in our api slice's `extractRehydrationInfo`:

- In `store/api.ts`, we have a `const api = createApi({...})`
- In `store/index.ts`, we create a reducer with `const reducer = combineSlices(api)`, and define the store type with `type RootState = ReturnType<typeof reducer>`
-  In `store/api.ts`, my `api`'s `extractRehydrationInfo` uses a type predicate to determine `action is PayloadAction<???>`

I can't type my predicate with `PayloadAction<RootState>` as this causes a circular reference. [The docs suggest that the type would be inferred](https://redux-toolkit.js.org/rtk-query/api/createApi#extractrehydrationinfo) and use `any` but is there a better way I'm missing to give the `PayloadAction` the generic it needs?

If this was a `createSlice({ name: 'foo', ... })` I'd use a partial like `PayloadAction<{ foo: FooState }>`, but obviously the state is much more complex with `createApi` and I don't see a clean way to do the same with the generic `CreateApi` type 🤔

EskiMojo14:
fixes #3947

### DIFF
--- a/docs/rtk-query/usage/persistence-and-rehydration.mdx
+++ b/docs/rtk-query/usage/persistence-and-rehydration.mdx
@@ -37,7 +37,7 @@ box with the `autoMergeLevel1` or `autoMergeLevel2` [state reconcilers](https://
 when persisting the root reducer, or with the `autoMergeLevel1` reconciler when persisting just the api reducer.
 
 ```ts title="redux-persist rehydration example"
-import type { Action, PayloadAction } from '@reduxjs/toolkit'
+import type { Action } from '@reduxjs/toolkit'
 import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react'
 import { REHYDRATE } from 'redux-persist'
 
--- a/docs/rtk-query/usage/persistence-and-rehydration.mdx
+++ b/docs/rtk-query/usage/persistence-and-rehydration.mdx
@@ -43,22 +43,27 @@ import { REHYDRATE } from 'redux-persist'
 
 type RootState = any // normally inferred from state
 
-function isHydrateAction(action: Action): action is PayloadAction<RootState> {
+function isHydrateAction(action: Action): action is Action<typeof REHYDRATE> & {
+  key: string
+  payload: RootState
+  err: unknown
+} {
   return action.type === REHYDRATE
 }
 
 export const api = createApi({
   baseQuery: fetchBaseQuery({ baseUrl: '/' }),
   // highlight-start
-  extractRehydrationInfo(action, { reducerPath }) {
+  // to prevent circular type issues, the return type needs to be annotated as any
+  extractRehydrationInfo(action, { reducerPath }): any {
     if (isHydrateAction(action)) {
-      if ((action as any).key === 'key used with redux-persist') {
-        // when persisting the api reducer
+      // when persisting the api reducer
+      if (action.key === 'key used with redux-persist') {
         return action.payload
       }
 
       // When persisting the root reducer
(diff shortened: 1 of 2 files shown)


docs/rtk-query/usage/persistence-and-rehydration.mdx
```diff
@@ -43,22 +43,27 @@ import { REHYDRATE } from 'redux-persist'
...
   baseQuery: fetchBaseQuery({ baseUrl: '/' }),
   // highlight-start
-  extractRehydrationInfo(action, { reducerPath }) {
+  // to prevent circular type issues, the return type needs to be annotated as any
+  extractRehydrationInfo(action, { reducerPath }): any {
     if (isHydrateAction(action)) {
-      if ((action as any).key === 'key used with redux-persist') {
-        // when persisting the api reducer
+      // when persisting the api reducer
+      if (action.key === 'key used with redux-persist') {
```
harry-gocity, line 61: It may be worth also mentioned `next-redux-cookie-wrapper` here, as I don't think this is unique to `redux-persist`

(`redux-persist` may be behind the scenes of `next-redux-cookie-wrapper` but is invisible from an end user pov)

docs/rtk-query/usage/persistence-and-rehydration.mdx
```diff
@@ -37,28 +37,33 @@ box with the `autoMergeLevel1` or `autoMergeLevel2` [state reconcilers](https://
...
 ```ts title="redux-persist rehydration example"
-import type { Action, PayloadAction } from '@reduxjs/toolkit'
+import type { Action } from '@reduxjs/toolkit'
 import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react'
 import { REHYDRATE } from 'redux-persist'
 
 type RootState = any // normally inferred from state
 
-function isHydrateAction(action: Action): action is PayloadAction<RootState> {
+function isHydrateAction(action: Action): action is Action<typeof REHYDRATE> & {
```
harry-gocity, line 46: For me it seems that simply using `any` on the return type of `extractRehydrationInfo` is enough to break the circular reference, and then the type check can be done with
```tsx
function isHydrateAction(action: Action): action is PayloadAction<RootState> {
    return action.type === HYDRATE;
}
```

EskiMojo14: yeah, i was just adjusting the type guard to what the hydrate action actually looks like, to avoid the `action as any` when accessing action.key

packages/toolkit/src/query/createApi.ts
```diff
@@ -175,14 +175,18 @@ export interface CreateApiOptions<
...
    * type RootState = any; // normally inferred from state
    *
-   * function isHydrateAction(action: Action): action is PayloadAction<RootState> {
-   *   return action.type === HYDRATE
+   * function isHydrateAction(action: Action): action is Action<typeof REHYDRATE> & {
+   *   key: string
+   *   payload: RootState
+   *   err: unknown
+   * } {
+   *   return action.type === REHYDRATE
```
EskiMojo14, line 183:
```suggestion
   * function isHydrateAction(action: Action): action is PayloadAction<RootState> {
   *   return action.type === HYDRATE
```

PR #3951, opened 2023-12-05
Remove abort event listner for AbortController

Fonger:
Call removeEventListener() to prevent potential memory leak.

In theory, new AbortController() should be removed by GC per action.
However it's always safe to remove the event listener explicitly when we no longer need it.

If we somehow

### DIFF
--- a/packages/toolkit/src/createAsyncThunk.ts
+++ b/packages/toolkit/src/createAsyncThunk.ts
@@ -577,6 +577,7 @@ export const createAsyncThunk = /* @__PURE__ */ (() => {
           : nanoid()
 
         const abortController = new AbortController()
+        let abortHandler: (() => void) | undefined
         let abortReason: string | undefined
 
         function abort(reason?: string) {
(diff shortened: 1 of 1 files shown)


packages/toolkit/src/createAsyncThunk.ts
```diff
@@ -600,14 +601,14 @@ export const createAsyncThunk = /* @__PURE__ */ (() => {
...
-              abortController.signal.addEventListener('abort', () =>
+            const abortedPromise = new Promise<never>((_, reject) => {
+              abortHandler = () => {
                 reject({
                   name: 'AbortError',
                   message: abortReason || 'Aborted',
                 })
-              )
-            )
+              }
```
EskiMojo14, line 610:
```suggestion
              }
              abortController.signal.addEventListener('abort', abortHandler)
```

EskiMojo14:
as you say, can't hurt - thanks!

PR #3984, opened 2023-12-16
Add React Native demo app to CI workflow

aryaemami59:
This PR:

- [X] Adds React Native example app to CI workflow.


markerikson:
We need both of these to actually do some kind of a real build step, so that we know that things at least compile.  Can you update them to do an actual build instead of `echo "Done"`?

aryaemami59:
> We need both of these to actually do some kind of a real build step, so that we know that things at least compile. Can you update them to do an actual build instead of `echo "Done"`?

Yes I will work on that, tbh I just had to take a break from it since RN is not my favorite thing to work with :)

aryaemami59:
I will update this as soon as Reselect 5.0.2 with the new `createSelector.withTypes<RootState>()` API is released.
I will probably have to submit another PR to do the same for all the other examples.

markerikson:
Any idea why it errored with `gradlew not found`?

aryaemami59:
@markerikson I think so, I'll see if I can fix it.

examples/publish-ci/react-native/src/constants/TypedColors.ts
```diff
@@ -0,0 +1,13 @@
...
+  primary: string
+  white: string
+  lighter: string
+  light: string
+  dark: string
+  darker: string
+  black: string
+}
+
+export const TypedColors = Colors as AllColors
```
EskiMojo14, line 13: out of curiosity, why not just add a declaration for the original Colors variable?

aryaemami59: Oh I just remembered, it wasn't missing a declaration, the `Colors` variable was just typed as `any`. And this was my way of overriding it to make it more type safe. Is there a better a way to accomplish that without doing what I did there?

EskiMojo14: you're already providing types for `"react-native/Libraries/NewAppScreen"` in `globals.d.ts`, just add the Colors export to it 

```diff
declare module "react-native/Libraries/NewAppScreen" {
  import type { FC } from "react"
  export const HermesBadge: FC
+
+  export const Colors: {
+    primary: string;
+    white: string;
+    lighter: string;
+    light: string;
+    dark: string;
+    darker: string;
+    black: string;
+  };
}
```

aryaemami59: @EskiMojo14 I get `Cannot redeclare block-scoped variable 'Colors'.ts(2451) LaunchScreen.d.ts(15, 16): 'Colors' was also declared here.`

EskiMojo14: ah, interesting - apparently my dependencies hadn't actually installed 

then yeah, i guess this is necessary - shame the original types are so rubbish

Issue #4028, opened 2023-12-31
`selectFromResult` runs on any change to the store


thisjeremiah:
## Problem

I've noticed that the `selectFromResult` callback in `useQuery` seems to be running any time any data in the store changes. 

I'm trying to understand if this behavior is necessary.

## Expected Behavior

I'd prefer `selectFromResult` to only run when the relevant data changes (on changes to loading state or  the response data).

## Benefit

The primary benefit of improving this behavior is that if someone (intentionally or accidentally) returns an unstable response from `selectFromResult`, the component will only re-render twice for every request that gets made (`isLoading` is set to `true` then `false`).

## Reproduction

[...]

PR #4029, opened 2024-01-01
fix: improve selectFromResult memoization

thisjeremiah:
Improve `selectFromResult` memoization to only run the callback when relevant data is changed.

Fixes this issue: https://github.com/reduxjs/redux-toolkit/issues/4028.

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -915,7 +915,12 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
               (_: ApiRootState, lastResult: any) => lastResult,
               (_: ApiRootState) => stableArg,
             ],
-            queryStatePreSelector
+            queryStatePreSelector,
+            {
+              memoizeOptions: {
+                resultEqualityCheck: shallowEqual,
+              },
+            }
           ),
         [select, stableArg]
       )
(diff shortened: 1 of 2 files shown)


phryneas:
Thanks a ton! Will merge this once CI has gone through all tests :)

denniscual:
Can we backport this fix into v1? We currently have project that is still using v1 due to legacy React version dependency. Right now, we just patch this fix. But it would be better to add this in RTK support version. If this is fine with you guys. I can create PR, most likely will just copy and paste the changes into this PR.

phryneas:
> Can we backport this fix into v1? We currently have project that is still using v1 due to legacy React version dependency. Right now, we just patch this fix. But it would be better to add this in RTK support version. If this is fine with you guys. I can create PR, most likely will just copy and paste the changes into this PR. 

I believe apart from the dependency, there's nothing in V2 stopping you from using it with an outdated React version. Could you please try that first?

denniscual:
V2 requires react-redux v9 which uses React v18. Its on the [migration guide](https://redux-toolkit.js.org/usage/migrating-rtk-2#requires-react-18).

phryneas:
@denniscual Hm, that is a point, we dropped that shim after 2 years of React 18.
You could try package.json overrides to run it with react-redux 8 - I believe the only thing colliding there might be types, and I'm not even sure about those. @EskiMojo14 do you maybe know that?

EskiMojo14:
yeah, the types changed so you'd probably have to patch some things too

Issue #4038, opened 2024-01-04
rtk-query-codegen-openapi has outdated prettier 2.x peer dependency


FabianFrank:
Prettier 3 has been out for half a year, it is probably time to update, see https://prettier.io/blog/2023/07/05/3.0.0.html.

See also #3604

FabianFrank:
Are there thoughts about removing the prettier dependency and support altogether and having people reformat the code _after_ code gen? A peer dep on specific prettier versions seems pretty rough as projects I assume are usually on their own prettier update schedule.

phryneas:
This honestly sounds like primarily a problem with your package manager - it should be standard procedure for your package manager to install prettier 2 in `node_modules/@rtk-query/codegen-openapi/prettier` (and for that to be used by the codegen) while your project uses prettier 3 from `node_modules/prettier`.

Did you enable some form of module hoisting?

But generally, this is an Open Source project maintained by volunteers in our spare time and we are open for Pull Requests.

FabianFrank:
> This honestly sounds like primarily a problem with your package manager - it should be standard procedure for your package manager to install prettier 2 in `node_modules/@rtk-query/codegen-openapi/prettier` (and for that to be used by the codegen) while your project uses prettier 3 from `node_modules/prettier`.
> 
> Did you enable some form of module hoisting?

I did suspect and investigate the package manager (npm in my case) angle at first as well, but it appears to me that npm behaves correctly and instead we are ending up with an incompatible combination of prettier and prettier plugins. I am using npm with default settings and from what I can observe it is working as intended, i.e. the `@rtk-query/codegen-openai` has its own 2.x version of prettier installed, which I verify as follows:
```
% jq < node_modules/prettier/package.json .version
"3.1.1"
% jq < node_modules/@rtk-query/codegen-openapi/node_modules/prettier/package.json .version
"2.8.8"
```

In addition there are also prettier plugins `prettier-plugin-embed@0.3.2` and `prettier-plugin-sql@0.18.0`. These are prettier 3.x plugins and they thus can be (and are) ESM modules. Now when the CJS `prettier@2.8.8` that ships with `@rtk-query/codegen-openai` loads the default prettier config that references the ESM module plugins it then fails to load them because you can't load ESM from CJS:
```
[...]

PR #4041, opened 2024-01-05
rtk query code gen: add optional prettierrc override

FabianFrank:
Allow users to optionally provide the explicit location of the prettier config to use. For example unblocks users that are using an incompatible/different major version of prettier as part of their project that breaks the prettier version rtk query code gen depends on.

Example:
```typescript
import type { ConfigFile } from "@rtk-query/codegen-openapi"

const config: ConfigFile = {
  schemaFile: "http://localhost:8080/openapi",
  ...,
  prettierConfigFile: "./rtk-codegen.prettierrc.json",
}

export default config
```

see #4038

### DIFF
--- a/packages/rtk-query-codegen-openapi/src/index.ts
+++ b/packages/rtk-query-codegen-openapi/src/index.ts
@@ -15,11 +15,14 @@ export async function generateEndpoints(options: GenerationOptions): Promise<str
     const { generateApi } = await import('./generate');
     return generateApi(schemaAbsPath, options);
   });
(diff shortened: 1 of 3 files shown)


johanvaneck:
Any estimate on when this could be merged?

EskiMojo14:
could some tests be added? I generally don't touch codegen PRs because I don't use/write them, but some tests would possibly increase my confidence enough to review properly.

Reichenberg:
Still an issue, would love to see this merged @FabianFrank ❤️

PR #4042, opened 2024-01-05
Add slices overload to configureStore

EskiMojo14:


EskiMojo14:
closing, as this isn't possible without an overload, and the decrease in UX for reported errors isn't worth it when the user can just call `configureStore({ reducer: combineSlices(...slices) })`.

PR #4048, opened 2024-01-05
Allow customising createSelector instance used by RTKQ

EskiMojo14:
Can now create modules using different memoiser:
```ts
const createApi = buildCreateApi(coreModule({ createSelector }), reactHooksModule({ createSelector }))
```

### DIFF
--- a/packages/toolkit/src/query/core/buildSelectors.ts
+++ b/packages/toolkit/src/query/core/buildSelectors.ts
@@ -1,4 +1,5 @@
-import { createNextState, createSelector } from './rtkImports'
+import type { createSelector as _createSelector } from './rtkImports'
+import { createNextState } from './rtkImports'
 import type {
   MutationSubState,
   QuerySubState,
@@ -123,9 +124,11 @@ export function buildSelectors<
 >({
   serializeQueryArgs,
   reducerPath,
+  createSelector,
 }: {
   serializeQueryArgs: InternalSerializeQueryArgs
   reducerPath: ReducerPath
+  createSelector: typeof _createSelector
 }) {
   type RootState = _RootState<Definitions, string, string>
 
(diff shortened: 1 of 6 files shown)


packages/toolkit/src/query/react/buildHooks.ts
psychedelicious, line 1: `import { createSelector } from '@reduxjs/toolkit'` can be removed from this file now.

psychedelicious, line 1: On L908 in this file in the definition of `selectDefaultResult`, some options are passed to `createSelector`. I think this will override the configuration of the custom `createSelector`.

On one hand, maybe the options specified in `selectDefaultResult` are objectively correct, but on the other, this is an unexpected behavior.

packages/toolkit/src/query/core/buildSelectors.ts
```diff
@@ -1,4 +1,5 @@
-import { createNextState, createSelector } from './rtkImports'
+import type { createSelector as _createSelector } from './rtkImports'
```
psychedelicious, line 1:
```suggestion
import { createSelector as _createSelector } from './rtkImports'
```

psychedelicious (commented):
Does what it says on the tin, thanks!

packages/toolkit/src/query/core/buildSelectors.ts
psychedelicious: Oh, nvm. TIL - you can use `import type` for objects you intended to use only for typing purposes (e.g. `typeof _createSelector`). Cool~

packages/toolkit/src/query/react/buildHooks.ts
EskiMojo14: yeah - it'll override `memoizeOptions`, but not `memoize` or `argsMemoize`. It was added to fix https://github.com/reduxjs/redux-toolkit/issues/4028 - check that thread out for more context

psychedelicious: Thanks for the context, makes sense and will prevent footguns when using a custom createselector

Issue #978, opened 2021-04-03
Website - visited link color


x4th:
Would you consider changing/removing the [visited link style](https://github.com/reduxjs/redux-toolkit/blob/087d4b62906025f04262cb5f1f24e8bb86b5c4e3/website/src/css/custom.css#L120)? Having it be the same color as the active section makes it unclear on which section the user is currently on.

x4th:
Here's a userscript in the meantime while this is discussed.

```
(function() {
    'use strict';
    const styleSheet = document.createElement('style')
    styleSheet.type = 'text/css'
    styleSheet.innerHTML = `
    a:visited {
        color: unset;
    }

    .menu__link--active:not(.menu__link--sublist) {
        color: white;
        background-color: var(--ifm-color-primary);
    }
    `
    document.head.appendChild(styleSheet)
})();
```

DouglasPds:
I would like to solve this.

DouglasPds:
What the expected behavior?

I think that if change the active link to color white and background purple, it's good to maintain the color for visited links.

![maintain-color](https://user-images.githubusercontent.com/49247431/122779829-a45ce700-d284-11eb-9466-e25037d58b50.png)

Here are the same visited links that the photo above, but with no color for visited links.

![removed-color](https://user-images.githubusercontent.com/49247431/122780113-e7b75580-d284-11eb-9673-cb77c1568424.png)

And this is the actual behavior, for a comparison

![actual-behavior](https://user-images.githubusercontent.com/49247431/122780698-775d0400-d285-11eb-8408-e1a888211010.png)

savita-kumari13:
Hey @markerikson created a PR - https://github.com/reduxjs/redux-toolkit/pull/3544.
Let me know of any issues

Fargrim:
Hello, I was looking for "good first issue" tags here and I found this but appears to be completed in the above PR and Redux Toolkits current API page reflects the suggested behavior above. Is there any additional work that needs to be done here?

PR #4050, opened 2024-01-07
Website/Docs - changed styling on active Table Of Contents link to differentiate from visited

riqts:
This bug annoyed me while reading the docs so I decided to try fix it.

Open issue #978 - discussed and sidebar fix was implemented, so I am extending a similar implementation to TOC on each page

Not sure if there is a preferred solution to the styling here, don't mind changing it, I just picked underline and the previously existing font-weight that had incorrect css targetting.

https://github.com/reduxjs/redux-toolkit/assets/79554296/898fb1e9-94c2-40c1-be16-54ec1dd42f50

Summary:

Moved font-weight from previous unstable css targeting using infima to new working class and added an underline.
Made a copy of the Sidebar Menu's behavior that stops the flash of white when hovering over visited items (that are not active)

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -134,8 +134,9 @@ a.contents__link > code {
   color: inherit;
 }
 
-a.contents__link.contents__link--active {
-  font-weight: 600;
+.table-of-contents__link--active {
+  font-weight: 500;
+  text-decoration: underline;
 }
 
 .table-of-contents__link:hover > code,
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -138,6 +138,11 @@ a.contents__link.contents__link--active {
   font-weight: 600;
 }
 
+.table-of-contents__link:hover > code,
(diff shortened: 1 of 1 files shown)


EskiMojo14:
image

i'm fine with the changes to active/visited but my concern is that there's not much difference between normal and hovered in dark mode - any way that could be tweaked?

riqts:
@EskiMojo14 
> i'm fine with the changes to active/visited but my concern is that there's not much difference between normal and hovered in dark mode - any way that could be tweaked?

I am happy to do that, it does re-introduce an annoying css thing here in exchange though.  docusaurus has a default `a:hover` color transition back to white property on these links and I just chose that color because it looks much nicer running your cursor down the list without that transition. So just lmk which you'd prefer I show both in this screen recording


https://github.com/reduxjs/redux-toolkit/assets/79554296/35257703-7cfe-4ff7-a0fe-2f576cf333cb

PR #4055, opened 2024-01-09
Option for queries to pause polling when unfocused

### LINKED ISSUE #2516: Option to pause Polling when window Unfocused
Sorry if this is a silly question, but is it possible to pause polling on a query if the window loses focus.

I currently have a situation where i need to manually `initiate` a query due to some other bad code where i have used prefetch in a horrible hacky way to bypass cache timeout (it's a pagination/inifinate scroll issue that is first on my list for a refactor!!).

```typescript
  useEffect(() => {
    let unsubscribe: () => void;

    if (id) {
      const result = dispatch(
        getChatHistory.initiate(
          { id, page: 1 },
          {
            subscribe: true,
            subscriptionOptions: {
              pollingInterval: 10000,
            },
          }
        )
      );

      unsubscribe = result.unsubscribe;
    }

    return () => {
      unsubscribe && unsubscribe();
    };
  }, [id, dispatch]);
```

If there is no "native" way to pause polling, I think it should be possible to create a listener for the focus/unfocus action dispatches and set some state value accordingly.

I have posted here instead of SO in case the functionality could be useful (please close/delete the issue if there is no merit in the idea and I will ask on SO instead).

riqts:
fixes #2516 

~PR will be in a draft state as I plan on finishing tests around it but wanted to get the PR up with the initial implementation in case there is some input~

- `skipPollOnFocusLost` option added to SubscriptionOptions
- Extended `findLowestPollingInterval` function in pollingHandler to also return the skipPollOnFocusLost flag that is stored in each subscriptions data alongside pollingInterval. Figured a per subscription setting gave the most granular control if I was going to pick one place for the first setting.

```ts
function findLowestPollingInterval(subscribers: Subscribers = {}) {
    let skipPollOnFocusLost: boolean | undefined = false
    let lowestPollingInterval = Number.POSITIVE_INFINITY
    for (let key in subscribers) {
      if (!!subscribers[key].pollingInterval) {
        lowestPollingInterval = Math.min(
          subscribers[key].pollingInterval!,
          lowestPollingInterval
        )
        skipPollOnFocusLost = subscribers[key].skipPollOnFocusLost
      }
    }

    return {
      lowestPollingInterval,
      skipPollOnFocusLost,
    }
  }
  ```
  
  - Added tests to cover base functionality
  
Notes for review:

- I used document.hasFocus() in the focus check rather than using the focus state that requires the listenerMiddleware, but I can change that if it aligns better with the vision of RTKQ.

- Existing functionality in polling is that it takes the lowest interval of all subscriptions set for the queryCacheKey, where as I have the current functionality set as taking the most recently mounted skipPollOnFocusLost flag, but it still takes the lowest poll interval

- Also review the name it probably sucks and let me know what docs I need to change

### DIFF
--- a/packages/toolkit/src/query/core/buildMiddleware/polling.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/polling.ts
@@ -53,38 +53,36 @@ export const buildPollingHandler: InternalHandlerBuilder = ({
     { queryCacheKey }: QuerySubstateIdentifier,
     api: SubMiddlewareApi
   ) {
-    const state = api.getState()[reducerPath];
-    const querySubState = state.queries[queryCacheKey];
-    const subscriptions = internalState.currentSubscriptions[queryCacheKey];
+    const state = api.getState()[reducerPath]
+    const querySubState = state.queries[queryCacheKey]
+    const subscriptions = internalState.currentSubscriptions[queryCacheKey]
 
     if (!querySubState || querySubState.status === QueryStatus.uninitialized)
-      return;
+      return
 
-    const { lowestPollingInterval, skipPollOnFocusLost } = findLowestPollingInterval(subscriptions);
-    if (!Number.isFinite(lowestPollingInterval)) return;
+    const { lowestPollingInterval, skipPollOnFocusLost } =
+      findLowestPollingInterval(subscriptions)
+    if (!Number.isFinite(lowestPollingInterval)) return
 
-    const currentPoll = currentPolls[queryCacheKey];
+    const currentPoll = currentPolls[queryCacheKey]
 
     if (currentPoll?.timeout) {
-      clearTimeout(currentPoll.timeout);
-      currentPoll.timeout = undefined;
+      clearTimeout(currentPoll.timeout)
+      currentPoll.timeout = undefined
     }
 
-    const nextPollTimestamp = Date.now() + lowestPollingInterval;
+    const nextPollTimestamp = Date.now() + lowestPollingInterval
 
-    // Always update the polling interval
     currentPolls[queryCacheKey] = {
       nextPollTimestamp,
       pollingInterval: lowestPollingInterval,
       timeout: setTimeout(() => {
-        // Conditionally dispatch the query
         if (document.hasFocus() || !skipPollOnFocusLost) {
-          api.dispatch(refetchQuery(querySubState, queryCacheKey));
+          api.dispatch(refetchQuery(querySubState, queryCacheKey))
         }
-        // Regardless of dispatch, set up the next poll
-        startNextPoll({ queryCacheKey }, api);
+        startNextPoll({ queryCacheKey }, api)
       }, lowestPollingInterval),
-    };
+    }
   }
 
   function updatePollingInterval(
--- a/packages/toolkit/src/query/core/buildMiddleware/polling.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/polling.ts
@@ -53,35 +53,38 @@ export const buildPollingHandler: InternalHandlerBuilder = ({
     { queryCacheKey }: QuerySubstateIdentifier,
     api: SubMiddlewareApi
   ) {
-    const state = api.getState()[reducerPath]
-    const querySubState = state.queries[queryCacheKey]
-    const subscriptions = internalState.currentSubscriptions[queryCacheKey]
+    const state = api.getState()[reducerPath];
+    const querySubState = state.queries[queryCacheKey];
+    const subscriptions = internalState.currentSubscriptions[queryCacheKey];
 
     if (!querySubState || querySubState.status === QueryStatus.uninitialized)
-      return
+      return;
 
-    const lowestPollingInterval = findLowestPollingInterval(subscriptions)
-    if (!Number.isFinite(lowestPollingInterval)) return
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/toolkit/src/query/core/buildMiddleware/polling.ts [hunk 1]
```diff
@@ -72,16 +73,16 @@ export const buildPollingHandler: InternalHandlerBuilder = ({
...
-    const currentInterval: typeof currentPolls[number] = (currentPolls[
-      queryCacheKey
-    ] = {
+    currentPolls[queryCacheKey] = {
       nextPollTimestamp,
       pollingInterval: lowestPollingInterval,
       timeout: setTimeout(() => {
-        currentInterval!.timeout = undefined
-        api.dispatch(refetchQuery(querySubState, queryCacheKey))
+        if (document.hasFocus() || !skipPollOnFocusLost) {
```
EskiMojo14, line 80: this is a web-specific API, so won't work for React Native users (for one example)

instead, you'd need to make use of the state that `setupListeners` modifies:
```suggestion
        if (api.getState()[reducerPath].config.focused || !skipPollOnFocusLost) {
```

riqts: Thanks, was definitely curious why the refetch behaviours opted into that.
Updated PR to reflect this change with:
- Tests changed
- Documentation in apiState.ts changed
- Behaviour changed

packages/toolkit/src/query/tests/polling.test.tsx
```diff
@@ -3,6 +3,7 @@ import { createApi } from '@reduxjs/toolkit/query'
 import { setupApiStore, waitMs } from './helpers'
 import { delay } from '../../utils'
 import type { SubscriptionSelectors } from '../core/buildMiddleware/types'
+import { createListenerMiddleware } from '@reduxjs/toolkit'
```
EskiMojo14, line 6: you don't need createListenerMiddleware for setupListeners 😛

riqts: I see, I made my life harder than it had to be trying to set up a new store with it and apparently just didn't test to see if it was in the current storeRef in that test file... and then was busy with work for a day or two.

Anyway should reflect this change in most recent revision.

packages/toolkit/src/query/core/buildMiddleware/polling.ts
```diff
@@ -126,17 +127,23 @@ export const buildPollingHandler: InternalHandlerBuilder = ({
...
   function findLowestPollingInterval(subscribers: Subscribers = {}) {
+    let skipPollOnFocusLost: boolean | undefined = false
     let lowestPollingInterval = Number.POSITIVE_INFINITY
     for (let key in subscribers) {
       if (!!subscribers[key].pollingInterval) {
         lowestPollingInterval = Math.min(
           subscribers[key].pollingInterval!,
           lowestPollingInterval
         )
+        skipPollOnFocusLost = subscribers[key].skipPollOnFocusLost
```
EskiMojo14, line 138: unsure of what the intent here is - currently this just means that skipPollOnFocusLost will be whatever the last subscription in the map has it set to.

if the desired approach is to skip if at least one of the subscriptions says to, you'd need something like 
```suggestion
        skipPollOnFocusLost = subscribers[key].skipPollOnFocusLost || skipPollOnFocusLost
```

riqts: Yeah absolutely was not sure what the right way to handle that behavior was and don't have a very deep understanding of the subscription logic. So I just chose to use the last subscription and wrote that intent in the test behavior, but I can see how this is better.

Most recent commit should reflect this change + test and test name

EskiMojo14: looks good, thanks!

EskiMojo14 (approved):
at some point we'll want to get this added to the docs but I think that can be done in a separate PR

riqts:
Sorry about the infrequency.

@EskiMojo14  > to bike shed a little, what about `skipPollingWhileUnfocused`? the current name sounds like it only skips once when the focus is first lost, whereas my understanding is that it's completely skipped until focus is gained again

I changed it to `skipPollingIfUnfocused` because once I got in there I thought that it was potentially more representative of the actual behavior it has where the poll stays synced but simply doesn't run if unfocused. However, now that I am back here typing it I am not so convinced. Let me know if you still think I should go with while instead of if.

packages/toolkit/src/query/core/buildMiddleware/polling.ts
```diff
@@ -72,16 +73,16 @@ export const buildPollingHandler: InternalHandlerBuilder = ({
...
+    currentPolls[queryCacheKey] = {
       nextPollTimestamp,
       pollingInterval: lowestPollingInterval,
       timeout: setTimeout(() => {
-        currentInterval!.timeout = undefined
-        api.dispatch(refetchQuery(querySubState, queryCacheKey))
+        if (state.config.focused || !skipPollingIfUnfocused) {
+          api.dispatch(refetchQuery(querySubState, queryCacheKey))
+        }
+        startNextPoll({ queryCacheKey }, api)
```
EskiMojo14, line 83: hmm, how come this was added?

it looks like startNextPoll is called below, so why does it need calling here?

EskiMojo14:
@riqts `skipPollingIfUnfocused` is good enough for me :)

Issue #4054, opened 2024-01-09
`injectedSlice.selectors` seems doesn't handle uninjected problem like `.getSelectors()` & `.selectSlice()`


whitetrefoil:
```ts
export interface State {
  initializing?: true
}

export const lazySlice = createSlice({
  name: 'friends',
  initialState: {} as State,
  reducers: create => ({
    INIT: create.reducer(s => {
      s.initializing = true
    }),
    // ...
  }),
  selectors: {
    $isInitializing: s => s.initializing === true,
  },
})

export const injectedSlice = lazySlice.injectInto(rootReducer)

console.log(injectedSlice.selectSlice({}))
// => {}
console.log(injectedSlice.getSelectors(s => s.friends).$isInitializing({}))
// => false
console.log(injectedSlice.selectors.$isInitializing({}))
// => Uncaught Error: selectSlice returned undefined for an uninjected slice reducer
```

EskiMojo14:
thanks for the report! I've replicated the issue in tests, and raised #4056 to fix it. that'll create a codesandbox build you can use in the meantime, and we'll try to get it merged and released when we can :)

PR #4056, opened 2024-01-09
fix selectors not recognising they've been injected

EskiMojo14:
fixes https://github.com/reduxjs/redux-toolkit/issues/4054


EskiMojo14:
closing as I think #4071 is a better fix for this

PR #4061, opened 2024-01-10
fix inference and order of async thunk generics

### LINKED ISSUE #4060: Wrong `AsyncThunkAction` args type when creating a thunk without args using `create.asyncThunk`
Sometimes thunks doesn't require any args.
When creating a thunk without args using `create.asyncThunk`, the action still expects one argument.

Example

```ts
import { buildCreateSlice, asyncThunkCreator } from "@reduxjs/toolkit";

const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
const createSliceWithThunks = buildCreateSlice({
  creators: { asyncThunk: asyncThunkCreator },
});
export const counterSlice = createSliceWithThunks({
  name: "counter",
  initialState: { value: 0 },
  reducers: (create) => ({
    increment: create.asyncThunk(() => sleep(1000)),
    decrement: create.asyncThunk((_: undefined) => sleep(1000)),
  }),
});

export const { increment, decrement } = counterSlice.actions;

increment(); // ERROR: Expected 1 arguments, but got 0.
increment(undefined); // ok
decrement(); // ok

export default counterSlice.reducer;
```

Codesandbox: https://codesandbox.io/p/sandbox/quirky-sunset-8qsxf4?file=%2Fsrc%2FcounterSlice.ts%3A20%2C1

```json
"@reduxjs/toolkit": "2.0.1",
"react-redux": "9.0.4"
```

EskiMojo14:
fixes #4060

### DIFF
--- a/packages/toolkit/src/createSlice.ts
+++ b/packages/toolkit/src/createSlice.ts
@@ -350,7 +350,7 @@ interface AsyncThunkCreator<
   State,
   CurriedThunkApiConfig extends PreventCircular<AsyncThunkConfig> = PreventCircular<AsyncThunkConfig>
 > {
-  <ThunkArg extends any, Returned = unknown>(
+  <Returned, ThunkArg = void>(
     payloadCreator: AsyncThunkPayloadCreator<
       Returned,
       ThunkArg,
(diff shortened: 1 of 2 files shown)


phryneas (approved):
Unfortunate this slipped through but better we fix it now than later.

PR #4081, opened 2024-01-15
Fix codemods to work with TypeScript 4.7+

aryaemami59:
This PR:

  - [X] Copies over setup and configuration from Reselect codemod.
  - [X] Removes `codemod-cli` as it does not support TypeScript 4.7+ syntax.
  - [X] Refactor transform files
  - [X] Eliminates `@ts-ignore` directives to improve TypeScript compliance.
  - [X] Configures `lineTerminator` in `.toSource.options` for consistent end-of-line across platforms, overriding `jscodeshift`'s default OS-specific EOL.
  - [X] Removes unnecessary `test.js` file.
  - [X] Refactores `__fixtures__` files to contain slightly more realistic and practical examples. 
  
  This new setup while somewhat flawed, works with TS 4.7+. I'm sure there is room for improvement and I am open to suggestions.


EskiMojo14:
overall this looks good - my only concern is that the READMEs have become significantly more verbose without any real benefit for the reader

i'm fine with the test fixtures including extra things, but i think the READMEs should only really demonstrate what the codemod does rather than having to be a fully functional example

PR #4082, opened 2024-01-15
cacheLifecycle fix for fixedCacheKey in mutationThunk

### LINKED ISSUE #2943: onCacheEntryAdded doesn't work with fixedCacheKey
**onCacheEntryAdded** is not called when **fixedCacheKey** is specified for mutation ([getCacheKey doesn't handle that case](https://github.com/reduxjs/redux-toolkit/blob/f7a828255e0612bed0856a0b6d1b9b8d6d6ac30d/packages/toolkit/src/query/core/buildMiddleware/cacheLifecycle.ts#L254))

```
const [fetchList, list] = rtkApi.useListMutation({
    fixedCacheKey: 'FIXED_CACHE_KEY',
});

...

list: builder.mutation({
  query: () => ({
    method: "POST",
    url: `/some-url`,
  }),
  async onCacheEntryAdded(arg, { cacheDataLoaded }) {
    console.log("log me please");
    await cacheDataLoaded;
    console.log("log again");
  })
})
```

riqts:
fixes #2943 

This PR fixes the initiation of the cacheLifecycle when using fixedCacheKey in mutationThunk and adds a test to verify the creation of a new cache entry when using fixedCacheKey. The getCacheKey function has been refactored to return the fixedCache Key if it exists, otherwise it returns the requestId. A new test has been added to ensure that a new cache entry is created when using fixedCache Key.

### DIFF
--- a/packages/toolkit/src/query/tests/useMutation-fixedCacheKey.test.tsx
+++ b/packages/toolkit/src/query/tests/useMutation-fixedCacheKey.test.tsx
@@ -20,6 +20,9 @@ describe('fixedCacheKey', () => {
     endpoints: (build) => ({
       send: build.mutation<string, string | Promise<string>>({
         query: (arg) => arg,
+        onCacheEntryAdded: (arg) => {
+          onNewCacheEntry(arg)
+        },
       }),
     }),
   })
@@ -359,18 +362,6 @@ describe('fixedCacheKey', () => {
   })
 
   test('using fixedCacheKey should create a new cache entry', async () => {
-    api.injectEndpoints({
-      overrideExisting: true,
-      endpoints: (build) => ({
-        send: build.mutation<string, string | Promise<string>>({
-          query: (arg) => arg,
-          onCacheEntryAdded(arg, {}) {
-            onNewCacheEntry(arg)
-          },
-        }),
-      }),
-    })
-
     render(<Component name="C1" fixedCacheKey={'testKey'} />, {
       wrapper: storeRef.wrapper,
     })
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/query/tests/useMutation-fixedCacheKey.test.tsx
```diff
@@ -354,4 +357,34 @@ describe('fixedCacheKey', () => {
...
+      overrideExisting: true,
+      endpoints: (build) => ({
+        send: build.mutation<string, string | Promise<string>>({
+          query: (arg) => arg,
+          onCacheEntryAdded(arg, {}) {
+            onNewCacheEntry(arg)
+          },
+        }),
+      }),
+    })
```
EskiMojo14, line 372:
```suggestion
    api.enhanceEndpoints({
      endpoints: {
        send: {
          onCacheEntryEnded: (arg) => onNewCacheEntry(arg)
        }
      }
    })
```

EskiMojo14: sorry, my suggestion had a typo - edited

the problem with what you have now is that your mock isn't being reset between tests

image

riqts: All good, I thought the ended was a typo.

Seeing as you're active, I'll ask what the preference here is so I can know for future as I am not an expert on unit testing practices really. Is it preferred that I create the endpoint override within the scope of that single test, or have the functionality added to the initial declaration and add and `afterEach () => {clear mock function}`

EskiMojo14: i think you should create the mock and call enhanceEndpoints within the single test, and at the end of the test block do 
```ts
    api.enhanceEndpoints({
      endpoints: {
        send: {
          onCacheEntryAdded: undefined,
        },
      },
    })
```
to make sure it doesn't continue onto other tests

riqts: That's why you get paid the big bucks, Ben. Recent commit should reflect this change.

EskiMojo14:
thanks!

PR #4084, opened 2024-01-16
FIX(4058): operation parameter should override pathItem parameters 

### LINKED ISSUE #4058: Question/Feature (@rtk-query/codegen-openapi): Support for overriding path-level parameters at operation Level
Hey.
I was trying to set up my query client with the auto generator and noticed that the generated `ApiArg` types are not following the OpenAPI specs for overriding path-level parameters. At least not how I would expect them to.

From the OpenAPI specs on operation object parameters: https://spec.openapis.org/oas/latest.html#operation-object
> A list of parameters that are applicable for this operation. **If a parameter is already defined at the [Path Item](https://spec.openapis.org/oas/latest.html#pathItemParameters), the new definition will override it but can never remove it.** 

An OpenAPI spec example of this from https://swagger.io/docs/specification/describing-parameters/
```yaml
paths:
  /users/{id}:
    parameters:
      - in: path
        name: id
        type: integer
        required: true
        description: The user ID.
        
    # DELETE /users/{id} - uses a single ID.
    # Reuses the {id} parameter definition from the path level.
    delete:
      summary: Deletes the user with the specified ID.
      responses:
        '204':
          description: User was deleted.
    # GET /users/id1,id2,id3 - uses one or more user IDs.
    # Overrides the path-level {id} parameter.
    get:
      summary: Gets one or more users by ID.
      parameters:
        - in: path
          name: id
          required: true
          description: A comma-separated list of user IDs.
          type: array
          items:
            type: integer
          minItems: 1
      responses:
        '200':
          description: OK
```

Which results in the generated `ApiArgs`
```typescript
export type DeleteUsersByIdApiArg = {
  /** The user ID. */
  id: number;
};
export type GetUsersByIdApiArg = {
  /** The user ID. */
  pathId: number;
  /** A comma-separated list of user IDs. */
  _pathId: number[];
};
```

It looks good when not overriding the path-level parameter, but overriding the parameter at operation-level, the generated `id` path-level parameter is now present twice - as both types.
Both of these different parameters then has to be provided in the hooks, which seems odd and makes the querying quite confusing. Is this intended?
image

I would expect the generated `ApiArgs` to match those of the OpenAPI specs and only require the overriding parameter. Like we see in the swagger UI, only requiring operation-level parameter.
image

Any clarification would be appreciated, thanks.

julien-tricent:
This fix #4058


juliengbt:
@EskiMojo14 Can I get you to have a look at that ? :)

markerikson:
Thanks!

juliengbt:
@tricent-lvi

markerikson:

PR #4093, opened 2024-01-18
Fix `composeWithDevTools` spy

aryaemami59:
This PR:

  - [X] Fixes the spy for `composeWithDevTools` using `@internal/` import path.

### DIFF
--- a/packages/toolkit/src/tests/configureStore.test.ts
+++ b/packages/toolkit/src/tests/configureStore.test.ts
@@ -2,10 +2,11 @@ import * as DevTools from '@internal/devtoolsExtension'
 import type { StoreEnhancer } from '@reduxjs/toolkit'
 import { Tuple } from '@reduxjs/toolkit'
 import type * as Redux from 'redux'
+import type { MockInstance } from 'vitest'
 import { vi } from 'vitest'
 
-vi.doMock('redux', async () => {
-  const redux: any = await vi.importActual('redux')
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/tests/configureStore.test.ts
```diff
@@ -15,45 +16,19 @@ vi.doMock('redux', async () => {
...
-  // RTK's internal `composeWithDevtools` function isn't publicly exported,
-  // so we can't mock it. However, it _does_ try to access the global extension method
-  // attached to `window`. So, if we mock _that_, we'll know if the enhancer ran.
-  const mockDevtoolsCompose = vi
-    .fn()
-    .mockImplementation(originalComposeWithDevtools)
-  ;(window as any).__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ = mockDevtoolsCompose
+  const composeWithDevToolsSpy = vi.spyOn(DevTools, 'composeWithDevTools')
+
+  window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ = composeWithDevToolsSpy
```
EskiMojo14, line 31:
```suggestion
```
since we're now spying on composeWithDevTools directly, we don't need to mock the window at all - we can just let it default to the stub and spy on that instead

EskiMojo14: github won't let me make a suggestion removing the declare global lines but those can obviously go too

PR #4102, opened 2024-01-20
Added 'SafePromise' branded Promises for createAsyncThunk

### LINKED ISSUE #4101: Use a branded Promise type for always-resolved Promises
Coming over from https://github.com/typescript-eslint/typescript-eslint/issues/7008#issuecomment-1862478574: there is a known annoyance with the [@typescript-eslint/no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises) lint rule that it flags the Promises created by `createAsyncThunk` as _"floating"_ if not handled, despite those Promises being known "safe" ones that never resolve. There's no way for the lint rule to know that these Promises are safe.

That issue tracks adding an `allow` option to the lint rule that would let users allowlist these Promises. But in order to help with RTK, the known safe Promises would need some kind of specific type that can be added to the rule's allowlist. +1 to @phryneas's suggestion:

> So, it would be nice if we could mark those promises with some kind of branded type to them to signal to eslint that these promises can be handled, but don't need to.

We're up for adding in `allow` on our end if you're up for adding that branded type here! Is that something you're up for? Even if this issue is marked as blocked on us implementing the `allow` option, as long as that's the only blocker we're good to go.

I'd be happy to contribute the actual code change to RTK myself if you think it's doable for a first-timer. 😄

JoshuaKGoldberg:
Fixes #4101.

Build, lint, and test pass for me locally - but I have this sinking feeling that there's got to be more complication to this change?

### DIFF
--- a/packages/toolkit/src/tsHelpers.ts
+++ b/packages/toolkit/src/tsHelpers.ts
@@ -213,5 +213,5 @@ export type UnknownIfNonSpecific<T> = {} extends T ? unknown : T
  * @see https://github.com/reduxjs/redux-toolkit/issues/4101
  */
 export type SafePromise<T> = Promise<T> & {
-  __brand: 'SafePromise'
+  __safetyBrand: 'SafePromise'
 }
--- a/packages/toolkit/src/tsHelpers.ts
+++ b/packages/toolkit/src/tsHelpers.ts
@@ -213,7 +213,7 @@ export type UnknownIfNonSpecific<T> = {} extends T ? unknown : T
  * @see https://github.com/reduxjs/redux-toolkit/issues/4101
  */
 export type SafePromise<T> = Promise<T> & {
-  __safetyBrand: 'SafePromise'
+  __linterBrands: 'SafePromise'
 }
 
 /**
--- a/packages/toolkit/src/tsHelpers.ts
+++ b/packages/toolkit/src/tsHelpers.ts
@@ -207,3 +207,11 @@ export type Tail<T extends any[]> = T extends [any, ...infer Tail]
   : never
 
 export type UnknownIfNonSpecific<T> = {} extends T ? unknown : T
+
+/**
+ * A Promise that will never reject.
+ * @see https://github.com/reduxjs/redux-toolkit/issues/4101
+ */
+export type SafePromise<T> = Promise<T> & {
+  __brand: 'SafePromise'
+}
--- a/packages/toolkit/src/tsHelpers.ts
+++ b/packages/toolkit/src/tsHelpers.ts
@@ -217,7 +217,7 @@ export type SafePromise<T> = Promise<T> & {
 }
 
 /**
- * Properly wraps a Promise as a @link {SafePromise} with .catch(fallback).
+ * Properly wraps a Promise as a {@link SafePromise} with .catch(fallback).
  */
 export function asSafePromise<Resolved, Rejected>(
   promise: Promise<Resolved>,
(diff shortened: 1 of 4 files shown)


packages/toolkit/src/tsHelpers.ts
```diff
@@ -207,3 +207,11 @@ export type Tail<T extends any[]> = T extends [any, ...infer Tail]
   : never
 
 export type UnknownIfNonSpecific<T> = {} extends T ? unknown : T
+
+/**
+ * A Promise that will never reject.
+ * @see https://github.com/reduxjs/redux-toolkit/issues/4101
+ */
+export type SafePromise<T> = Promise<T> & {
+  __brand: 'SafePromise'
```
phryneas, line 216: Is this already set in stone? I fear that `__brand` could get collisions at some point in time - what about `__eslint: 'SafePromise'`?

phryneas:
Very cool :)

Two other places to add this come to mind:

[MutationActionCreatorResult](https://github.com/reduxjs/redux-toolkit/blob/a10e10f7c28eec7594951dbf68a28af2a73d7779/packages/toolkit/src/query/core/buildInitiate.ts#L91)

[QueryActionCreatorResult](https://github.com/reduxjs/redux-toolkit/blob/a10e10f7c28eec7594951dbf68a28af2a73d7779/packages/toolkit/src/query/core/buildInitiate.ts#L61)

packages/toolkit/src/tsHelpers.ts
JoshuaKGoldberg: Oh and - in case it's not clear from the issue, my current proposal is that each library define its own brand. Not that we have a shared built-in in the ts-eslint plugin for folks to use. This means users would have to manually configure the `plugin:@typescript-eslint/no-floating-promises` allowlist for each library with "safe" Promises:

```js
"@typescript-eslint/no-floating-promises": ["error", {
  "allowKnownSafePromises": [
    { "from": "package", "name": "SafePromise", "package": "@reduxjs/toolkit"
  ]
}]
```

JoshuaKGoldberg: Oh and - in case it's not clear from the issue, my current proposal is that each library define its own brand. Not that we have a shared built-in in the ts-eslint plugin for folks to use. This means users would have to manually configure the `plugin:@typescript-eslint/no-floating-promises` allowlist for each library with "safe" Promises:

```js
"@typescript-eslint/no-floating-promises": ["error", {
  "allowKnownSafePromises": [
    { "from": "package", "name": "SafePromise", "package": "@reduxjs/toolkit"
  ]
}]
```

Another strategy would be asking end libraries to take a dependency on some package -likely a `@typescript-eslint/*`- one in order to mark their "safe" Promises. I don't think that's an ask we can put forward. And even if we did, folks will want a way to add an allowlist anyway.

phryneas: > And I'd expect it wouldn't be feasible for most user-facing libs to take a types dependency on any @typescript-eslint/* package.

That's why I added the `@ts-ignore` - if it's installed and available, use it, if not, not.

> Oh and - in case it's not clear from the issue, my current proposal is that each library define its own brand. 

Ah, yeah, that wasn't clear. Yeah, I guess we can explore it that way and we'll see if some pattern will stick out in the end :)

`__safetyBrand` seems fine so far. I'll try one alternative suggestion, though: `__linterBrands`. 
I think this could be defined as a string union, so if any other linter-related brands would come up in the future, they could just add more like `__linterBrands: 'SafePromise' | 'StableValue'`, right?

JoshuaKGoldberg: `__linterBrands` it is! 🚀

EskiMojo14:
cool stuff, thanks!

JoshuaKGoldberg:
🙌 my first contribution! Thanks all!

JoshuaKGoldberg:
🙌 my first code contribution! Thanks all!

JoshuaKGoldberg:
Btw I just edited the OP to make it clear what the changes in a user's ESLint config would be. Only realized after the fact I was never clear on that. Sorry 🙂

PR #4114, opened 2024-01-23
Modernize unit test setup

aryaemami59:
This PR:

  - [X] Renames `vitest.config.ts` to `vitest.config.mts` per [Vitest documentation](https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated)
  - [X] Replaces the deprecated `deps.inline` field inside `vitest.config.mts` with `server.deps.inline`.
  - [X] Renames `vitest.setup.js` to `vitest.setup.ts`.
  - [X] Bumps [Node Fetch](https://github.com/node-fetch/node-fetch) dev dependency.
  - [X] Bumps [MSW](https://mswjs.io/docs) dev dependency.
  - [X] Runs [codemods](https://codemod.com/automations/msw-2-upgrade-recipe) provided by MSW to modernize MSW syntax.
  - [X] Follows [MSW's Migration Guide](https://mswjs.io/docs/migrations/1.x-to-2.x) to convert version 1x syntax to version 2x.


package.json
```diff
@@ -47,7 +47,6 @@
     "@babel/types": "7.19.3",
     "esbuild": "0.19.7",
     "jest-snapshot": "29.3.1",
-    "msw": "patch:msw@npm:0.40.2#.yarn/patches/msw-npm-0.40.2-2107d48752",
```
EskiMojo14, line 50: can you also delete `.yarn/patches/msw-npm-0.40.2-2107d48752`?

packages/toolkit/src/query/tests/buildHooks.test.tsx
```diff
@@ -1977,47 +1976,49 @@ describe('hooks with createApi defaults set', () => {
...
+            return HttpResponse.json(posts)
         }),
       ]
 
       server.use(...handlers)
     })
 
     interface Post {
-      id: number
+      id: string
```
EskiMojo14, line 2021: why make this change if it forces you to add so many `as any`s? The server response id is a number, so I don't understand why it's changed here.

package.json
aryaemami59: Yes sirrrrr

packages/toolkit/src/query/tests/buildHooks.test.tsx
aryaemami59: @EskiMojo14  The reason why I changed the type for `Post` to `string` is because MSW 2 has the `PathParams` type like this:

```ts
type PathParams<KeyType extends keyof any = string> = {
    [ParamName in KeyType]: string | ReadonlyArray<string>;
};
```

so if you have something like:

```ts
{ id: number }
```

it'll yell at you, and I didn't wanna mess with runtime unit tests too much just to make TS happy.


Screenshot 2024-01-23 125845

Issue #4115, opened 2024-01-23
Docs: Missing `reducer` field for `ConfigureStoreOptions`


nickgirardo:
While poking around I noticed that the `reducer` field was missing for the docs for `ConfigureStoreOptions`.  It looks like a few lines were somehow accidentally deleted as there's an unterminated block comment where I'd expect the block comment outlining the purpose of `reducer` to begin.

[Live Docs](https://redux-toolkit.js.org/api/configurestore/)
[Source file defining this page of the docs](https://github.com/reduxjs/redux-toolkit/blob/master/docs/api/configureStore.mdx#L48)
[Source defining `ConfigureStoreOptions`](https://github.com/reduxjs/redux-toolkit/blob/e7a3456eb3c2fa219a8fad462563906746761c47/packages/toolkit/src/configureStore.ts#L49)

Note the few missing lines which are present in the source

This looks like a pretty simple fix.  Assuming I'm not missing anything, I could take care of this if there aren't any contributors who are currently working in the docs who could knock it out easily

markerikson:
Sure, please do!

(We _might_ be able to use the docblock injection syntax we use elsewhere in the docs to keep this always inserted, although not sure if that works in this situation)

nickgirardo:
Closing, was resolved by #4116!

PR #4116, opened 2024-01-24
Add missing reducer field to ConfigureStoreOptions docs

nickgirardo:
The `reducer` field was missing from `ConfigureStoreOptions`

#4115 

I attempted to use docblock to fix these docs, but I wasn't able to have it output the information. I'm not familiar with docblock, but an error message told me the valid references were `summary,remarks,overloadSummary,overloadRemarks,examples,params,,`, and none of those worked.  A manual fix was able to add the missing option.


markerikson:
Thanks!

PR #4121, opened 2024-01-25
Fix `expectExactType` to be compatible with TypeScript 5.4

aryaemami59:
This PR:

  - [X] Fixes `expectExactType` to be compatible with TypeScript 5.4 per reduxjs/react-redux#2123

### DIFF
--- a/packages/toolkit/src/listenerMiddleware/tests/listenerMiddleware.test.ts
+++ b/packages/toolkit/src/listenerMiddleware/tests/listenerMiddleware.test.ts
@@ -96,7 +96,7 @@ type Equals<T, U> = IsAny<
   IsAny<U, never, [T] extends [U] ? ([U] extends [T] ? any : never) : never>
 >
 export function expectExactType<T>(t: T) {
-  return <U extends Equals<T, U>>(u: U) => {}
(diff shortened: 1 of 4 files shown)


EskiMojo14:
Hmm, the removal of the ts-expect-error comments worries me - those *should* be errors.

I've commented over at https://github.com/reduxjs/react-redux/pull/2123.

aryaemami59:
> Hmm, the removal of the ts-expect-error comments worries me - those _should_ be errors.
> 
> I've commented over at [reduxjs/react-redux#2123](https://github.com/reduxjs/react-redux/pull/2123).

Not sure if this counts as a fix but [Here](https://github.com/reduxjs/redux-toolkit/blob/81c1ad741a18c66dca716ef7b65396786aafb926/packages/toolkit/src/query/tests/unionTypes.test-d.ts#L482-L490) I've replaced `expectExactType` with some of Vitest's type test utilities.

EskiMojo14:
yeah, as long as those vitest utilities don't break in 5.4, and we can get the same functionality, i think it would be worth looking into switching our type tests to vitest style ones at some point

PR #4127, opened 2024-01-27
Migrate type tests to Vitest

aryaemami59:
This PR:

  - [X] Converts `typetest.ts` file extensions to `test-d.ts` to be compatible with Vitest.
  - [X] Converts `expectType()` to either `assertType()` or `expectTypeOf().toMatchTypeOf()`.
  - [X] Converts `expectExactType` to `expectTypeOf().toEqualTypeOf()`.
  - [X] Converts lone-blocks to `test` or `describe` blocks.


packages/toolkit/src/tests/createAsyncThunk.test-d.ts
```diff
@@ -0,0 +1,903 @@
...
+        })
+        .catch((error) => {
+          // catch is always any-typed, nothing we can do here
+          expectTypeOf(error).toBeAny()
+        })
+    })()
+  })
+
+  test('More complex usage of thunk args', () => {
+    ;(async function () {
```
EskiMojo14, line 91: this can be un-IIFE'd now it's got its own test block - same for the others in this file

packages/toolkit/src/tests/createAsyncThunk.test-d.ts
```diff
@@ -0,0 +1,903 @@
...
+
+      expectTypeOf(asyncThunk).toBeCallableWith(5)
+
+      expectTypeOf(asyncThunk).parameters.not.toMatchTypeOf<[string]>()
+    })
+
+    test('one argument, specified as any: asyncThunk has required any argument', () => {
+      const asyncThunk = createAsyncThunk('test', (arg: any) => 0)
+
+      expectTypeOf(asyncThunk).parameters.items.toBeAny()
```
EskiMojo14, line 366:
```suggestion
      expectTypeOf(asyncThunk).parameter(0).toBeAny()
```

packages/toolkit/src/tests/createEntityAdapter.test-d.ts
```diff
@@ -0,0 +1,161 @@
...
+  EntityAdapter,
+  EntityId,
+  EntityStateAdapter,
+  Update,
+} from '@reduxjs/toolkit'
+import { createEntityAdapter, createSlice } from '@reduxjs/toolkit'
+
+function extractReducers<T, Id extends EntityId>(
+  adapter: EntityAdapter<T, Id>,
+): Omit<EntityStateAdapter<T, Id>, 'map'> {
```
EskiMojo14, line 13:
```suggestion
): EntityStateAdapter<T, Id> {
```

packages/toolkit/src/tests/createSlice.test-d.ts
```diff
@@ -0,0 +1,959 @@
...
+    test('Reducer', () => {
+      expectTypeOf(slice.reducer).toMatchTypeOf<
+        Reducer<number, PayloadAction>
+      >()
+
+      expectTypeOf(slice.reducer).not.toMatchTypeOf<
+        Reducer<string, PayloadAction>
+      >()
+
+      expectTypeOf().not.toMatchTypeOf<Reducer<string, UnknownAction>>()
```
EskiMojo14, line 91:
```suggestion
```

packages/toolkit/src/tests/createSlice.test-d.ts
```diff
@@ -0,0 +1,959 @@
...
+    expectTypeOf(counter.actions.addTwo).parameters.not.toMatchTypeOf<
+      [number]
+    >()
+
+    expectTypeOf(counter.actions.addTwo).parameters.toEqualTypeOf<
+      [number, number]
+    >()
+  })
+
+  test('Slice action creator types properties are "string"', () => {
```
EskiMojo14, line 164:
```suggestion
  test('Slice action creator types properties are strongly typed', () => {
```

packages/toolkit/src/tests/mapBuilders.test-d.ts
```diff
@@ -0,0 +1,263 @@
...
+          (state, action: ReturnType<typeof increment>) => state,
+        )
+        .addDefaultCase((state, action) => {
+          expectTypeOf(action).toMatchTypeOf<UnknownAction>()
+        })
+
+      test('addMatcher() should prevent further calls to addCase()', () => {
+        const b = builder.addMatcher(increment.match, () => {})
+        // @ts-expect-error
+        b.addCase(increment, () => {})
```
EskiMojo14, line 131: could be switched to `expectTypeOf(b).not.toHaveProperty("addCase")`

and so on

EskiMojo14 (commented):
overall i like this change - the tests read a lot more semantically with the expectTypeOf syntax. this is also something we were discussing back in last April so it's nice to finally get in

it's a shame there's no `expectTypeOf(func).not.toBeCallableWith()`, but `expectTypeOf(func).parameters.not.toMatchTypeOf<Args>()` does the same job

i've left some suggestions for things i think would benefit from tweaks - some of which were carried over from the original type tests, i suspect

thanks for your hard work on this!

packages/toolkit/src/query/tests/createApi.test-d.ts
```diff
@@ -0,0 +1,388 @@
+import type {
+  DefinitionsFromApi,
+  OverrideResultType,
+  TagTypesFromApi,
+} from '@reduxjs/toolkit/dist/query/endpointDefinitions'
```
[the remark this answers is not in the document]
aryaemami59: Do you think we should export it?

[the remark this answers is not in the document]
EskiMojo14: i think we should, yeah

mrazauskas:
> it's a shame there's no `expectTypeOf(func).not.toBeCallableWith()`, but `expectTypeOf(func).parameters.not.toMatchTypeOf<Args>()` does the same job

Just to draw your attention, `.not.toBeAssignable()` is missing as well. This already creates problems in current type tests, consider the following:

https://github.com/reduxjs/redux-toolkit/blob/54b9182926226471174019745e1e2b83f27d9650/packages/toolkit/src/query/tests/retry.test-d.ts#L1-L9

Do you see `RetryOptions` being imported? Here the error caught by `// @ts-expect-error` is: "Cannot find name 'RetryOptions'."

---

For some time I was contributing type tests for Jest repo. Currently they are around 1000 assertions. This scale was hard to handle with current available tooling, so I went on writing my own library. It is called TSTyche and it was released recently after two years of development.

Repo: https://github.com/tstyche/tstyche 
Documentation: https://tstyche.org

---

Perhaps you would consider migrating to TSTyche?

It works just fine with `expectType` assertions, but has it’s own `expect`. The above test would look like this:

```ts
import { describe, expect, test } from 'tstyche'
import type { RetryOptions } from '../retry'

describe('RetryOptions type tests', () => {
  test('RetryOptions only accepts one of maxRetries or retryCondition', () => {
    expect<RetryOptions>().type.toBeAssignable({ maxRetries: 5 })
    expect<RetryOptions>().type.toBeAssignable({ retryCondition: () => false })

    expect<RetryOptions>().type.not.toBeAssignable({
      maxRetries: 5,
      retryCondition: () => false,
    })
  })
})
```

No chance for errors to sneak in.

---

Vitest is simply spawning a process and running `tsc`. Almost the same like to point `tsc` to the test files. The problems with this approach: how to select a single file?, or to use `.skip` / `.only` run mode modifiers? `tsc` does not know nothing about `.skip` and can type check only the whole program. Vitest somewhat improves the situation, but running in this repo `test retry.test` did not work for me. Debugging becomes clumsy in larger code base.

TSTyche is a dedicated tool with a mighty CLI which is build for projects of any scale. And it is packed in only 188kB.

For example, you can run type tests on specified versions of TypeScript: `tstyche --target 4.7,5.0,latest`. Works the same with `rc`, `beta` or `next` tags.

The test helpers support `.only` or `.skip` run mode flags. They can be used with `expect` as well and there is also `expect.fail()` to mark an assertion as failing. That is why I was packing three `expect` in one tests. You can focus or skip any of them at any time.

TSTyche wraps around TypeScrip's `ProjectService`. There is no `ts.createProgram()` or process spawning. Only minimal amount of type information is collected for each test. That is what makes type testing performant.

(By the way, only to type check Vitest globals takes a lot of time. I was playing with TSTyche in the `redux` repo. Running all tests with current setup: 0.6s. If Vitest globals are removed: 0.4s.)

Thanks for reading. Would you be interested to see an example PR?

EskiMojo14:
hey there @mrazauskas, thanks for reaching out! 

It's worth noting the specific example you gave is actually possible with Vitest, it just hadn't been migrated properly at that time (and now has):

https://github.com/reduxjs/redux-toolkit/blob/8bb90a166b0d88c7731c66284cee661e079e6c72/packages/toolkit/src/query/tests/retry.test-d.ts#L1-L11

`toMatchTypeOf` seems to be functionally equivalent to `toBeAssignable`.

I agree that `@ts-expect-error` is a little awkward in some cases, since you don't know which error you're surpressing.

TSTyche looks to be an interesting tool - the fact that you can test against multiple TS versions locally is very nice. I don't think we're likely to change to it for the Redux packages soon as the switch to Vitest was a significant amount of work, and Vitest is a bit more "battle tested".

It's certainly something I'll play around with in my own time though, and I wish you the best for it :)

aryaemami59:
@mrazauskas I will definitely check it out too, it looks very promising, and the documentation is excellent.

mrazauskas:
Thanks for quick response. Are you sure the updated test is correct? Because the following are also passing:

```ts
expectTypeOf<RetryOptions>().not.toMatchTypeOf({ maxRetries: 5 })

expectTypeOf<RetryOptions>().not.toMatchTypeOf({ retryCondition: () => false})
```

---

All is fine with Vitest as a JavaScript test runner. Many people use it and it is indeed battle tested. Altho: what is the test coverage of Vitest library? Top secret (;

Anyway. I speak only about `expectTypeOf`, which is pulled from [`expect-type`](https://www.npmjs.com/package/expect-type) and this library is still experimental (zero release). There are numerous undocumented limitations and issue, but I don’t want to waste your time.

EskiMojo14:
Fair point :)

A more useful test would be to flip the order:
```ts
    // Should not complain if only `maxRetries` exists
    expectTypeOf({ maxRetries: 5 }).toMatchTypeOf<RetryOptions>()

    // Should not complain if only `retryCondition` exists
    expectTypeOf({ retryCondition: () => false }).toMatchTypeOf<RetryOptions>()

    // Should complain if both `maxRetries` and `retryCondition` exist at once
    expectTypeOf({
      maxRetries: 5,
      retryCondition: () => false,
    }).not.toMatchTypeOf<RetryOptions>()
```

mrazauskas:
I was also thinking that flipping the sides is enough and it works, but look at example from Vitest docs:

```ts
expectTypeOf({ a: 1, b: 1 }).toMatchTypeOf({ a: 1 })
```

So it looks like the following should work:

```ts
expectTypeOf<RetryOptions>().toMatchTypeOf({ maxRetries: 5 })
```

But it does not. And the error is: "Expected 0 arguments, but got 1." Puzzling.

mrazauskas:
Took me a while, but I have to admit that flipping the order is right. My example was not good, here is the correct one:

```ts
expect({ b: "foo" }).type.toMatch<{ b: string } | { c: number }>();
```

PR #4135, opened 2024-01-29
Format all files

aryaemami59:


EskiMojo14:
CI failure on this is unrelated - codegen tests have been failing since node-fetch was bumped (ESM vs CJS issues, it seems)

@aryaemami59 is working on converting codegen tests to vitest which should hopefully help with this but in the meantime i think this can be merged

PR #4138, opened 2024-01-29
Centralize tooling configs in monorepo

aryaemami59:
## **This PR**:

  - [X] Creates dedicated repos housing shareable config files for:
  - **ESLint**
  - **Prettier**
  - **TypeScript**
  - **Vitest**
  
## **Part 1 of 3**

This PR precedes #4606 and #4607.

This is the initial step in a three-part effort to centralize tooling configurations so they can be shared across multiple repos.

### **Subsequent PRs**:

- [**Part 2**](https://github.com/reduxjs/redux-toolkit/pull/4606): Adopts the centralized tooling configurations across the codebase.
- [**Part 3**](https://github.com/reduxjs/redux-toolkit/pull/4607): Implements the necessary changes to comply with the new centralized configurations.


packages/toolkit/src/combineSlices.ts
```diff
@@ -121,7 +122,7 @@ export interface CombinedSliceReducer<
     config?: InjectConfig,
   ): CombinedSliceReducer<
     InitialState,
-    Id<DeclaredState & WithSlice<SliceLike<ReducerPath, String>>>
+    Id<DeclaredState & WithSlice<SliceLike<ReducerPath, string>>>
```
EskiMojo14, line 125:
```suggestion
    Id<DeclaredState & WithSlice<SliceLike<ReducerPath, State>>>
```
huh, apparently this is a bug i never noticed - should probably have its own PR

packages/toolkit/src/createAction.ts
```diff
@@ -275,7 +277,7 @@ export function createAction(type: string, prepareAction?: Function): any {
     return { type, payload: args[0] }
   }
 
-  actionCreator.toString = () => `${type}`
+  actionCreator.toString = () => type
```
EskiMojo14, line 280: i think this should be left as is - we don't do any runtime protection against non-strings so we need to make sure this returns a string regardless

packages/toolkit/src/tests/createSlice.test.ts
```diff
@@ -84,7 +83,9 @@ describe('createSlice', () => {
     })
 
     it('should create increment action', () => {
-      expect(actions.hasOwnProperty('increment')).toBe(true)
+      expect(Object.prototype.hasOwnProperty.call(actions, 'increment')).toBe(
+        true,
+      )
```
EskiMojo14, line 88:
```suggestion
      expect(actions).toHaveProperty('increment')
```

EskiMojo14 (approved):

markerikson:
Okay, what I _think_ I'm seeing here is:

- Added the configs
- A lot of lint fixes, mostly around auto-sorting of imports and forcing type imports
- Extracting some reusable TS types like `AnyObject`

Can I get a summary of what lint rules changed?

Also it looks like there's some commented out lint rules in the ESLint config

aryaemami59:
@markerikson

Sure, so mostly what I did was, add 2 configs: 
  - `plugin:@typescript-eslint/recommended`
  - `plugin:@typescript-eslint/stylistic`
  
  Along with `prettier` which turns off ESLint rules that conflict with Prettier.
  Then I fixed all the lint issues, and the ones that were not worth fixing (or were too time consuming) I disabled.
  As for the commented out rules, I will probably put up another PR after this one, with those rules and configs enabled and make adjustments and fixes as needed. I just did not want to do it all in one PR because this PR is already way too large as it is.

packages/configs/eslint/eslint.config.ts
```diff
@@ -0,0 +1,76 @@
...
+    ],
+  },
+  parser: '@typescript-eslint/parser',
+  parserOptions: { project: true, ecmaVersion: 'latest' },
+  plugins: ['@typescript-eslint'],
+  ignorePatterns: ['dist'],
+  root: true,
+  overrides: [
+    { files: ['*.{c,m,}{t,j}s', '*.{t,j}sx'] },
+    { files: ['*{test,spec}.{t,j}s?(x)'] },
```
EskiMojo14, line 72: these overrides don't seem to be doing anything?

aryaemami59: I was gonna see if I needed to override some settings for test files. I'll get rid of them for now.

packages/configs/typescript/package.json
```diff
@@ -0,0 +1,26 @@
+{
```
EskiMojo14, line 1: the other packages don't seem to be using these tsconfigs yet, is that still to do?

EskiMojo14: likewise with vitest

aryaemami59: Yeah, they can be used already, it's just that they don't require that much configuration, so they're pretty much ready to go. They can be used as soon as they are published.

aryaemami59: Same with Vitest, they can be used as soon as they're published, I tested them with Yalc, they seem to work just fine.

aryaemami59:
> could you add a `"prepack": "yarn build"` script to any of the packages with a build step? it means we can use our existing publish action with them

Yeah sure!

packages/configs/typescript/package.json
EskiMojo14: gotcha - i figured through the magic of yarn workspaces they might be able to be used without publishing (though obviously we'll publish them so our other repos can use them)

aryaemami59: They probably can, I just never got around to figuring that out.

timdorr:
Hey @aryaemami59, is ESLint being run in CI anywhere? I just noticed we're not running it on React Redux and it's been bugged for a while in that repo. Our first job is currently `Lint, Test, Build & Pack`, but it appears to only be `Build & Pack` 😬 Luckily we test in the TS matrix, but I don't think ESLint is run anywhere along the way.

aryaemami59:
@timdorr yeah I kinda knew we weren't running ESLint during CI, I think we're only running it for Reselect. Honestly my plan was to add it to CI after we take care of this one. If you want I can take care of it tomorrow.

markerikson:
Can we separate out these changes into multiple PRs?

- One to add all the new tooling configs
- Second to actually switch to _using_ the configs for testing and such
- And a third to apply just any formatting changes

aryaemami59:
I'm gonna split this into multiple PRs soon.

PR #4145, opened 2024-02-02
support default return schemas in openapi codegen

barp:
According to https://swagger.io/docs/specification/describing-responses/#default this is a valid way to describe a possible response.


phryneas:
Could you please add a test that shows a schema with this and the resulting generated code?

barp:
I have added a test, but for some reason I can't run it locally. I get weird errors unrelated to test itself.
```
➤ YN0000:     Jest encountered an unexpected token
➤ YN0000:
➤ YN0000:     Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or wh
en Jest is not configured to support such syntax.
➤ YN0000:
➤ YN0000:     Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configurat
ion.
➤ YN0000:
➤ YN0000:     By default "node_modules" folder is ignored by transformers.
➤ YN0000:
➤ YN0000:     Here's what you can do:
➤ YN0000:      • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
➤ YN0000:      • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
➤ YN0000:      • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config
.
➤ YN0000:      • If you need a custom transformation specify a "transform" option in your config.
➤ YN0000:      • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" co
nfig option.
➤ YN0000:
➤ YN0000:     You'll find more details and examples of these config options in the docs:
➤ YN0000:     https://jestjs.io/docs/configuration
➤ YN0000:     For information about custom transformations, see:
➤ YN0000:     https://jestjs.io/docs/code-transformation
➤ YN0000:
➤ YN0000:     Details:
```

not sure I am doing wrong I just tried to do:
```
yarn install
yarn test
```

EskiMojo14:
codegen tests are currently broken after updating node-fetch (which is now ESM only, it seems) - https://github.com/reduxjs/redux-toolkit/pull/4137 is an attempt to fix it long term by migrating to vitest but we might just need to revert the node-fetch update in the meantime

EskiMojo14:
merged in https://github.com/reduxjs/redux-toolkit/pull/4151 to temporarily revert back to node-fetch v2, try now

barp:
I have fixed the tests.

This adds `| void` to some response types but also fixes some response types for apis that only had default response type (like User apis in petstore).

barp:
The extra void was removed, it was a bug in the filter that removes all "void" from possible types.

barp:
Accidently closed this, reopened in https://github.com/reduxjs/redux-toolkit/pull/4152

PR #4146, opened 2024-02-02
Fix #3970 - Expose types allowing library consumers to build generic utils on top of @reduxjs/toolkit.

### LINKED ISSUE #3970: Error with tsc when attempting to import types from RTK Query using @reduxjs/toolkit 2.0.1
>Sample repo reproducing the issue can be found here: https://github.com/ryerrappa/rtk-query-type-import-issue

- @reduxjs/tookit 2.0.1
- Vite 5.0.0
- Typescript 5.2.2
- tsconfig moduleResolution "bundler"

Receive the following error when attempting to import the `MutationTrigger` type.

```
Cannot find module @reduxjs/toolkit/dist/query/react/buildHooks
```

Using vite with the following tsconfig.

```json
{
  "compilerOptions": {
    "composite": true,
    "skipLibCheck": true,
    "module": "ESNext",
    "moduleResolution": "bundler",
    "allowSyntheticDefaultImports": true
  },
  "include": ["vite.config.ts"]
}
```

lcharlois-neotys:
As discussed on th issue, exposing more types is a good thing as it allows library consumers to build generic piece of code on top of the library.

As I was not sure about which strategy you want to adopt, I choose to expose only the types discussed on the issue.

### DIFF
--- a/packages/toolkit/src/query/react/index.ts
+++ b/packages/toolkit/src/query/react/index.ts
@@ -14,9 +14,14 @@ const createApi = /* @__PURE__ */ buildCreateApi(
 )
 
 export type {
+  LazyQueryTrigger,
+  MutationTrigger,
   TypedUseMutationResult,
   TypedUseQueryHookResult,
   TypedUseQueryStateResult,
   TypedUseQuerySubscriptionResult,
+  UseMutation,
+  UseQuery,
+  UseQueryHookResult,
 } from './buildHooks'
 export { createApi, reactHooksModule, reactHooksModuleName }


EskiMojo14:
My concern is that these types, because they're internal, are not particularly user friendly - they're designed to take the internal endpoint definition that you can't get a hold of easily.

I'd rather a wrapper like we have for TypedUseQueryResult/TypedUseMutationResult etc, which takes friendlier arguments and passes them to the internal types to get the final type.

lcharlois-neotys:
> My concern is that these types, because they're internal, are not particularly user friendly - they're designed to take the internal endpoint definition that you can't get a hold of easily.
> 
> I'd rather a wrapper like we have for TypedUseQueryResult/TypedUseMutationResult etc, which takes friendlier arguments and passes them to the internal types to get the final type.

It looks nice and effectively provide a more user-friendly type.
I will give a try to your porposition to see if it works in my use case.

lcharlois-neotys:
I close this PR since the solution suggested by @EskiMojo14 in #4147 looks to better than this one.

PR #4147, opened 2024-02-02
Create more Typed wrappers for RTKQ hook types

EskiMojo14:
A potential alternate fix to #3970 and #4146


lcharlois-neotys:
Hello,

Just to let you know that I've tested your PR against my use case.
It do the job and better it leads to less complex implementation.
Very nice.

phryneas (approved):

PR #4161, opened 2024-02-06
Incorrect builder method referenced

kantbtrue:
The builder method referred to as `builder.addDefault` is incorrect. Changed to `builder.addDefaultCase`


EskiMojo14:
thanks!

PR #4173, opened 2024-02-08
[Docs, createEntityAdapter API]: add missing setOne and setMany signature

kyselberg:
RTK createEntityAdapter API documentation has minor mistake - it skips `setOne` and `setMany` signatures where other returned CRUD functions are described.
https://redux-toolkit.js.org/api/createEntityAdapter#return-value


EskiMojo14:
Thanks!

PR #4175, opened 2024-02-09
Use vite-tsconfig-paths to make path aliasing easier

aryaemami59:
This PR:

  - [X] Adds [vite-tsconfig-paths](https://github.com/aleclarson/vite-tsconfig-paths) as a dev dependency.
  - [X] Removes `baseUrl` from `tsconfig.base.json`.
  - [X] Removes `dist` paths aliases.
  - [X] Removes `react` path alias.


EskiMojo14:
thanks!

PR #4176, opened 2024-02-09
[Docs] Internal documentation on RTKQ

riqts:


website/sidebars.json
```diff
@@ -143,6 +143,24 @@
               ]
             }
           ]
+        },
```
markerikson, line 146: fwiw I don't think this content needs to be on the website - it's not going to be useful for folks _using_ the library.

riqts: All good, I'll take it off the sidebar and just leave it at docs/internal then!

markerikson:
This material is still useful - going to merge this as-is and you can make any further tweaks you'd like to!

Issue #4149, opened 2024-02-02
[Question] : Generic solution for prefetching paginated queries


julien-tricent:
Hey!

So I have an Api object generated by RTKQ-Codegen. There is a large amount of endpoint and most of them have server-side pagination (i.e `page` and `pageSize` parameters).

I would like to know if it's possible to implement a generic solution (maybe a middleware?) to prefetch the next page of any query that have been fulfilled, if it had a `page` argument.

Something like : 

```typescript
import { Middleware } from '@reduxjs/toolkit';
import { myApi } from 'src/api/enhancedApi';
import { RootState } from './store';

const fetchNextPageMiddleware: Middleware<unknown, RootState> =
  ({ getState, dispatch }) =>
  (next) =>
  (action) => {
    if (action.type === 'api/executeQuery/fulfilled' && typeof action.meta.arg.originalArgs.page === 'number') 
    {
      dispatch(myApi.util.prefetch(action.meta.arg.endpointName,
        {
          ...action.meta.arg.originalArgs,
          page: action.meta.arg.originalArgs.page + 1,
        },
        {},
      ));
    }
    return next(action);
  };
```

However I got some typing issues because action is of type `unknown`. Which means I have to manually check each property.

The `dispatch` also doesn't allow me to actually dispatch the prefetch because the later returns a `ThunkAction`

EskiMojo14:
You can somewhat unsafely skip some steps using type predicates:
```ts
import { Middleware, ThunkDispatch, UnknownAction, isFulfilled } from '@reduxjs/toolkit';
import { myApi } from 'src/api/enhancedApi';
import { RootState } from './store';

const hasEndpointName = (arg: any): arg is { endpointName: string } =>
  typeof arg?.endpointName === "string";

const hasPage = (arg: any): arg is { originalArgs: { page: number } } =>
  typeof arg?.originalArgs?.page === "number";

const fetchNextPageMiddleware: Middleware<
  unknown,
  RootState,
  ThunkDispatch<RootState, unknown, UnknownAction>
> =
  ({ getState, dispatch }) =>
  (next) =>
  (action) => {
    if (
      isFulfilled(action) &&
      hasEndpointName(action.meta.arg) &&
      hasPage(action.meta.arg)
    ) {
      dispatch(
        myApi.util.prefetch(
          action.meta.arg.endpointName as any,
          {
            ...action.meta.arg.originalArgs,
            page: action.meta.arg.originalArgs.page + 1,
          },
[...]

PR #4177, opened 2024-02-09
feat: add isPrefetch property in query action

juliengbt:
I am not so sure where and how to write test for it though.

related to #4149 

@EskiMojo14

### DIFF
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -58,6 +58,7 @@ type StartQueryActionCreator<
 > = (
   arg: QueryArgFrom<D>,
   options?: StartQueryActionCreatorOptions,
+  isPrefetch?: boolean
 ) => ThunkAction<QueryActionCreatorResult<D>, any, any, UnknownAction>
 
 export type QueryActionCreatorResult<
@@ -302,6 +303,7 @@ You must add the middleware for RTK-Query to function correctly!`,
           subscriptionOptions,
           [forceQueryFnSymbol]: forceQueryFn,
         } = {},
+        isPrefetch
       ) =>
       (dispatch, getState) => {
         const queryCacheKey = serializeQueryArgs({
@@ -319,6 +321,7 @@ You must add the middleware for RTK-Query to function correctly!`,
           originalArgs: arg,
           queryCacheKey,
           [forceQueryFnSymbol]: forceQueryFn,
+          isPrefetch
         })
         const selector = (
           api.endpoints[endpointName] as ApiEndpointQuery<any, any>
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -312,6 +312,7 @@ You must add the middleware for RTK-Query to function correctly!`,
         })
 
         const thunk = queryThunk({
+          ...rest,
           type: 'query',
           subscribe,
           forceRefetch: forceRefetch,
@@ -320,7 +321,6 @@ You must add the middleware for RTK-Query to function correctly!`,
           originalArgs: arg,
           queryCacheKey,
           [forceQueryFnSymbol]: forceQueryFn,
-          ...rest
         })
         const selector = (
           api.endpoints[endpointName] as ApiEndpointQuery<any, any>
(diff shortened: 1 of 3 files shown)


EskiMojo14:
my concern with adding a public option is that it allows passing it incorrectly when calling initiate yourself

my thinking would be something a little sneakier, like:
```diff
diff --git a/packages/toolkit/src/query/core/buildInitiate.ts b/packages/toolkit/src/query/core/buildInitiate.ts
index 384d10c5..0d06d8c7 100644
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -301,6 +301,7 @@ You must add the middleware for RTK-Query to function correctly!`,
           forceRefetch,
           subscriptionOptions,
           [forceQueryFnSymbol]: forceQueryFn,
+          ...rest
         } = {},
       ) =>
       (dispatch, getState) => {
@@ -319,6 +320,7 @@ You must add the middleware for RTK-Query to function correctly!`,
           originalArgs: arg,
           queryCacheKey,
           [forceQueryFnSymbol]: forceQueryFn,
+          ...rest,
         })
         const selector = (
           api.endpoints[endpointName] as ApiEndpointQuery<any, any>
diff --git a/packages/toolkit/src/query/core/buildThunks.ts b/packages/toolkit/src/query/core/buildThunks.ts
index 2fe616bb..61acf414 100644
--- a/packages/toolkit/src/query/core/buildThunks.ts
+++ b/packages/toolkit/src/query/core/buildThunks.ts
@@ -609,11 +609,12 @@ In the case of an unhandled error, no tags will be "provided" or "invalidated".`
       const force = hasTheForce(options) && options.force
       const maxAge = hasMaxAge(options) && options.ifOlderThan
 
-      const queryAction = (force: boolean = true) =>
-        (api.endpoints[endpointName] as ApiEndpointQuery<any, any>).initiate(
-          arg,
-          { forceRefetch: force },
-        )
+      const queryAction = (forceRefetch = true) => {
+        const options = { forceRefetch, isPrefetch: true }
+        return (
+          api.endpoints[endpointName] as ApiEndpointQuery<any, any>
+        ).initiate(arg, options)
+      }
       const latestStateValue = (
         api.endpoints[endpointName] as ApiEndpointQuery<any, any>
       ).select(arg)(getState())
```

juliengbt:
> my concern with adding a public option is that it allows passing it incorrectly when calling initiate yourself
> 

@EskiMojo14  Moving `isPrefetch` to `StartQueryActionCreatorOptions` isn't going to make private though.
Or were you thinking of forcing the the type of `options` to be `StartQueryActionCreatorOptions` ? (So `isPrefetch` doesn't show up as an option)

```diff
diff --git a/packages/toolkit/src/query/core/buildThunks.ts b/packages/toolkit/src/query/core/buildThunks.ts
index f07e6700..ac280ada 100644
--- a/packages/toolkit/src/query/core/buildThunks.ts
+++ b/packages/toolkit/src/query/core/buildThunks.ts
@@ -613,7 +613,7 @@ In the case of an unhandled error, no tags will be "provided" or "invalidated".`
       const queryAction = (force: boolean = true) =>
         (api.endpoints[endpointName] as ApiEndpointQuery<any, any>).initiate(
           arg,
-          { forceRefetch: force, isPrefetch: true }
+          { forceRefetch: force, isPrefetch: true } as StartQueryActionCreatorOptions
         )
       const latestStateValue = (
         api.endpoints[endpointName] as ApiEndpointQuery<any, any>
```

EskiMojo14:
no cast necessary - if you see in my diff, saving it to a separate variable first means that structural typing kicks in and doesn't care about the extra key (and the ...rest just think it's an empty object)

though in hindsight a cast would probably be a tiny bit more size efficient codewise, so do feel free to do that

juliengbt:
@EskiMojo14 Yes alright! 
Btw I'm sure you're aware of that, just to make sure. Spreading the rest of the options will also allow the user have their own keys in the `meta.arg` object of the start query action, right ? (when calling `initiate`)

EskiMojo14:
yeah - i think that's okay though

packages/toolkit/src/query/core/buildInitiate.ts
```diff
@@ -319,6 +320,7 @@ You must add the middleware for RTK-Query to function correctly!`,
           originalArgs: arg,
           queryCacheKey,
           [forceQueryFnSymbol]: forceQueryFn,
+          ...rest
```
EskiMojo14, line 323: though it's probably worth moving this spread *before* the other parameters, i think

juliengbt: Sure

EskiMojo14:
thanks!

PR #4184, opened 2024-02-11
allow initialising combined slice reducer with no static slices

EskiMojo14:
```ts
const reducer = combineSlices().withLazyLoadedSlices<LazyLoadedSlices>()
``` 
currently throws an error - this makes it return a "no-op" reducer which creates an empty object and just returns it until any reducers are injected

### DIFF
--- a/packages/toolkit/src/tests/combineSlices.test-d.ts
+++ b/packages/toolkit/src/tests/combineSlices.test-d.ts
@@ -33,7 +33,7 @@ describe('type tests', () => {
     }>()
   })
 
-  test('combineSlices allows for no initial reducers', () => {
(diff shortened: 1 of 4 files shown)


packages/toolkit/src/tests/combineSlices.test-d.ts
```diff
@@ -33,6 +33,19 @@ describe('type tests', () => {
     }>()
   })
 
+  test('combineSlices allows for no initial reducers', () => {
```
EskiMojo14, line 36:
```suggestion
  test('combineSlices allows passing no initial reducers', () => {
```

PR #4186, opened 2024-02-12
proposal fix for axios base query types

smff:
To prevent this type of TS errors in file after @rtk-query/codegen-openapi: 

Снимок экрана 2024-02-12 в 10 22 07

This is acceptable because axios: 
Снимок экрана 2024-02-12 в 10 22 27


EskiMojo14:
thanks!

Issue #4200, opened 2024-02-14
Upgrading from version 2.0.1 to 2.1.0 or 2.2.0 breaks vite build with TS (TS7006) errors


acostaf:
Hi Team

I just upgraded to version 2.2.0 and the change creates those this really weird issue, I am not changing anything on my tsconfig and eslinyrc and suddenly running vite build throws more than 600 implicit any type

Have anyone seen these before?

EskiMojo14:
could you give an example of a piece of code which now errors?

acostaf:
const user = useAppSelector(selectedUser)

user?.uid used to work properly not errors but now the new versions says that user is any but the selector was created from a proper type.

acostaf:
Funny enough the selector is now returning unknown, which force the error on build

EskiMojo14:
any chance you could put together a reproduction in typescript playground, codesandbox or a repo? none of our tests broke 😕

acostaf:
Actually I think the issue is coming from createDrafSafeSelector the function does not detect the original type returning any or unknown

EskiMojo14:
could you show how you're using it?

acostaf:
Sorry about no been able to provide a proper example, I am behind a proxy so cannot copy and paste the real code.

store/Index.ts

const selectOriginalThing = (state: Rootstate) => state.myState.thing; (thing is properly typed as Type? from slice)

export const selectThing = createDraftSafeSelector(selectOriginalThing, (thing) (here is the type error, thing is now “any” instead of inferring type from selectOriginalThing

EskiMojo14:
![image](https://github.com/reduxjs/redux-toolkit/assets/18308300/84d00c3a-4924-4525-8138-5af198dff430)

seems to be working in a playground - what typescript version are you using?

acostaf:
5.3.3

EskiMojo14:
not sure what to suggest at this point then, sorry - without a replication i can't really look into it further

AlbericoD:
I had the same issue, I removed the com package and then deleted node_modules, after which I installed the package again and it worked

rjgotten:
The problem is probably that you have a store slice which use a strongly typed app-specific dispatch.
To create a type for that app-specific dispatch, you need the fully constructed store.
To craete the fully constructed store, you need all the slices.
To make that one slice, you need the app-specific dispatch.

[...]

PR #4190, opened 2024-02-12
Remove trailing commas in `tsconfig.json` files of all CodesandBox examples

aryaemami59:
This PR:

  - [X] Reformats all files to remove trailing commas in `tsconfig.json` files of all CodesandBox examples
  - [X] Resolves #4185

PR #4202, opened 2024-02-14
Bump prettier and format all files

aryaemami59:
## Overview
Due to [this issue](https://github.com/reduxjs/redux-toolkit/issues/4200) which caused prettier to add trailing commas to `tsconfig.json` files which ultimately resulted in codeSandbox examples failing to compile. We had to add `overrides` field in `.prettier.json` in #4190 to resolve the issue. The newest release of Prettier fixes the problem by interpreting `tsconfig.json` files as `json` and not `jsonc`.

**This PR**:

  - [X] Bumps Prettier to 3.2.5.
  - [X] Removes the `overrides` field in `.prettier.json` file as it is no longer needed.
  - [X] Runs Prettier on all files inside the monorepo.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,11 +1,5 @@
 {
   "semi": false,
   "singleQuote": true,
-  "endOfLine": "auto",
-  "overrides": [
-    {
-      "files": ["examples/**/tsconfig*.json"],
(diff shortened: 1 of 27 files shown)


EskiMojo14:
thanks!

PR #4208, opened 2024-02-14
correctly allow extraOptions with required properties, or non-objects

### LINKED ISSUE #1087: Type bug: RTK Query - type for `extraOptions` only works as an object with optional properties
### Description
When defining a `baseQuery` and providing a type for `extraOptions`, trying to use the `baseQuery` with `createApi` has a typescript error unless the `extraOptions` type is an object with *only* optional properties.

e.g. will not work:
```ts
BaseQueryFn<
  string, // Args
  unknown, // Result
  { reason: string }, // Error
  { shout: boolean }, // DefinitionExtraOptions
  { timestamp: number } // Meta
>
```
will work:
```ts
BaseQueryFn<
  string, // Args
  unknown, // Result
  { reason: string }, // Error
  { shout?: boolean }, // DefinitionExtraOptions
  { timestamp: number } // Meta
>
```

### Version
1.6.0-beta.0

### Reproduction example
https://codesandbox.io/s/basequery-extraoptions-type-bug-w4kxt?file=/src/index.ts

```
Type 'BaseQueryFn<string, unknown, { reason: string; }, { shout: boolean; }, { timestamp: number; }>' is not assignable to type 'BaseQueryFn<any, unknown, unknown, {}, {}>'.
  Property 'shout' is missing in type '{}' but required in type '{ shout: boolean; }'.ts(2322)
index.ts(7, 5): 'shout' is declared here.
createApi.d.ts(24, 5): The expected type comes from property 'baseQuery' which is declared here on type 'CreateApiOptions<BaseQueryFn<any, unknown, unknown, {}, {}>, Record<string, EndpointDefinition<any, any, any, any, string>>, "api", never>'
```

EskiMojo14:
fixes #1087

### DIFF
--- a/docs/rtk-query/api/createApi.mdx
+++ b/docs/rtk-query/api/createApi.mdx
@@ -100,7 +100,7 @@ export type BaseQueryFn<
   Args = any,
   Result = unknown,
   Error = unknown,
-  DefinitionExtraOptions = {},
+  DefinitionExtraOptions = any,
   Meta = {},
 > = (
   args: Args,
--- a/docs/rtk-query/usage/customizing-queries.mdx
+++ b/docs/rtk-query/usage/customizing-queries.mdx
@@ -375,7 +375,8 @@ const axiosBaseQuery =
       headers?: AxiosRequestConfig['headers']
     },
     unknown,
-    unknown
+    unknown,
+    {}
   > =>
   async ({ url, method, data, params, headers }) => {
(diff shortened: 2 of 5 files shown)


EskiMojo14:
closing as i can't seem to get this happy, and having extraOptions only support optional properties isn't the end of the world - required config can be returned from `query` instead if needed.

PR #4211, opened 2024-02-15
added bailout to updateQueryData if patches is empty

### LINKED ISSUE #3090: Bail out of `updateQueryData` if there are no patches
Looks like we currently always dispatch inside of `updateQueryData`, even if there are no patches to apply:

```ts
if ('data' in currentState) {
        if (isDraftable(currentState.data)) {
          const [, patches, inversePatches] = produceWithPatches(
            currentState.data,
            updateRecipe
          )
          ret.patches.push(...patches)
          ret.inversePatches.push(...inversePatches)
        } else {
          const value = updateRecipe(currentState.data)
          ret.patches.push({ op: 'replace', path: [], value })
          ret.inversePatches.push({
            op: 'replace',
            path: [],
            value: currentState.data,
          })
        }
      }

      dispatch(api.util.patchQueryData(endpointName, args, ret.patches))
```

We should bail out if `patches.length === 0`.

riqts:
fixes #3090

### DIFF
--- a/packages/toolkit/src/query/core/buildThunks.ts
+++ b/packages/toolkit/src/query/core/buildThunks.ts
@@ -321,6 +321,10 @@ export function buildThunks<
         }
       }
 
+      if (ret.patches.length === 0) {
+        return ret
+      }
+
       dispatch(
         api.util.patchQueryData(
           endpointName,


EskiMojo14:
thanks!

PR #4215, opened 2024-02-15
Throw instances of Error when unwrap, instead of plain objects

pradiev:
Currently, when unwrap any async action, library throws plain objects. This makes such errors hard to debug, especially when using errors collectors like sentry.

I think, using `Error` objects everywhere, the code throws any exceptions, is a must-have practice. Would be even better, if library include separate types inherited from `Error` based on the internal error type we already have as a field.

### DIFF
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -362,7 +362,23 @@ You must add the middleware for RTK-Query to function correctly!`,
               const result = await statePromise
 
               if (result.isError) {
-                throw result.error
+                let errorMessage = '';
+                const errorMap = result.error as any;
+
+                if (errorMap && errorMap.data && errorMap.status) {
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/query/core/buildInitiate.ts
```diff
@@ -362,7 +362,23 @@ You must add the middleware for RTK-Query to function correctly!`,
...
+                }
+
+                const errorToThrow: any = new Error(errorMessage)
+                for (const prop in errorMap) {
+                  if (!errorToThrow[prop]) {
+                    errorToThrow[prop] = errorMap[prop]
+                  }
+                }
+
+                throw errorToThrow
```
EskiMojo14, line 381: i'm generally against trying to inspect the error like this - i would rather make a breaking change for 3.0, and throw something like `new Error("Unwrapped error", { cause: result.error })`

EskiMojo14 (changes requested):

PR #4223, opened 2024-02-18
Entity methods creator

### LINKED ISSUE #3837: Custom reducer creators

EskiMojo14:


packages/toolkit/src/entities/utils.ts
```diff
@@ -64,3 +64,7 @@ export function splitAddedUpdatedEntities<T, Id extends EntityId>(
   }
   return [added, updated, existingIdsArray]
 }
+
+export function capitalize<S extends string>(str: S) {
+  return str && (str.replace(str[0], str[0].toUpperCase()) as Capitalize<S>)
+}
```
aryaemami59, line 70: Can we consolidate with [this](https://github.com/reduxjs/redux-toolkit/blob/11b873c02032f48415f558cab4e89946f9aecd9e/packages/toolkit/src/query/utils/capitalize.ts) instead of duplicating it?

EskiMojo14: RTKQ tends to have its own utils, because it can only import things from the core RTK export that we actually export.

I don't think there's a clean way to consolidate them without exporting it?

PR #4225, opened 2024-02-19
suggest using the `satisfies` operator for `initialState`

k-yle:
This PR recreates #3623, which was reverted because the build system didn't support TypeScript 4.9 syntax.

The build error can be resolved by updating prettier from v2.3 to >v2.8, which I've also done in this commit.

Test plan:
 - run `cd website && yarn netlify-build` to confirm that it does compile this time


EskiMojo14:
thanks! could you resolve the lockfile conflicts?

EskiMojo14:
thanks!

PR #4234, opened 2024-02-21
Update example lockfiles

aryaemami59:
After upgrading to Yarn version 4 I should have also updated the lockfiles for the examples which I did not so this PR:

  - [X] Updates lockfiles for all example projects.


aryaemami59:
@EskiMojo14 Thank you!

PR #4255, opened 2024-03-06
Update TypeScript to 5.4

aryaemami59:
### This PR:

  - [X] Bumps TypeScript version for toolkit and the root `package.json` to 5.4.
  - [X] Fixes TypeScript 5.4 issue related to calling `dispatch` inside `createAsyncThunk`.
  - [X] Fixes TypeScript 5.4 issue related to `builder.addCase` in `createReducer.test-d.ts` type tests.
  - [X] Adds TypeScript 5.4 to TypeScript versions to test against during CI.


packages/toolkit/src/createAsyncThunk.ts
```diff
@@ -617,7 +617,7 @@ export const createAsyncThunk = /* @__PURE__ */ (() => {
               }
               abortController.signal.addEventListener('abort', abortHandler)
             })
-            dispatch(
+            ;(dispatch as ThunkDispatch<unknown, unknown, UnknownAction>)(
```
EskiMojo14, line 620: weird, i wonder what made this break

aryaemami59: It could be related to [this](https://github.com/microsoft/TypeScript/pull/56004). The problem is the second overload for `ThunkDispatch` currently looks like this:

```ts
<Action extends BasicAction>(action: Action): Action;
```

When it should look like this:

```ts
<Action extends BasicAction>(action: Action, ...extraArgs: any[]): Action;
```

The latter matches the actual type of `Dispatch` but the former does not. Obviosuly I don't know the types as well as you 
do but I think that might be the problem.

EskiMojo14: honestly i always found it weird the types for Dispatch allowed for extra arguments when middleware and enhancers only ever pass through one

aryaemami59: I think [this](https://github.com/reduxjs/redux/pull/2560) might be somewhat relevant, if you want I can fix the type for `Dispatch` to not include extra arguments.

PR #4268, opened 2024-03-12
Fixed memory leak in rapid hook arg changing

### LINKED ISSUE #4267: Query entries aren't removed from state if arg changes while query is pending
Hey, I'm encountering a problem where queries never unsubscribe if arg changes while its pending. Here's a codesandbox to reproduce:

[https://codesandbox.io/p/sandbox/distracted-easley-6vnn6x](https://codesandbox.io/p/sandbox/distracted-easley-6vnn6x)

- The test query does nothing but resolve after 3 seconds. `keepUnusedDataFor` is set at zero (i also tested 1 with same results)

- The two buttons fast and slow change the speed at which the arg changes; 1 and 5 seconds respectively

Looking at the redux devtools: 

While the arg changes slowly everything works as intended: the last cached query is removed from state, keeping only one entry active.

However once the query starts changing args faster than its resolve period, a memory leak seems to appear where entries get added indefinietly and never get removed.

As a consequence, it's messing a bit with `onCacheEntryAdded`, since `cacheEntryRemoved` never resolves.

riqts:
Fixes #4267 

Tests failed in github actions referencing delay not being defined but I am not sure what to do to fix that, I did not add the import  for delay, simply referenced it. Any direction for that would be appreciated.

### DIFF
--- a/packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
@@ -69,9 +69,11 @@ export const buildCacheCollectionHandler: InternalHandlerBuilder = ({
   ) => {
     if (unsubscribeQueryResult.match(action) ||
       queryThunk.fulfilled.match(action) ||
-      queryThunk.rejected.match(action)) {
+      queryThunk.rejected.match(action)
+    ) {
       const state = mwApi.getState()[reducerPath]
       let queryCacheKey: QueryCacheKey
+
       if (unsubscribeQueryResult.match(action)) {
         queryCacheKey = action.payload.queryCacheKey
       } else {
--- a/packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
@@ -1,3 +1,4 @@
+import { isAnyOf } from '@reduxjs/toolkit'
 import type { BaseQueryFn } from '../../baseQueryTypes'
 import type { QueryDefinition } from '../../endpointDefinitions'
 import type { ConfigState, QueryCacheKey } from '../apiState'
@@ -55,6 +56,12 @@ export const buildCacheCollectionHandler: InternalHandlerBuilder = ({
 }) => {
   const { removeQueryResult, unsubscribeQueryResult } = api.internalActions
 
+  const canTriggerUnsubscribe = isAnyOf(
+    unsubscribeQueryResult.match,
+    queryThunk.fulfilled,
+    queryThunk.rejected
+  )
+
   function anySubscriptionsRemainingForKey(queryCacheKey: string) {
     const subscriptions = internalState.currentSubscriptions[queryCacheKey]
     return !!subscriptions && !isObjectEmpty(subscriptions)
@@ -67,18 +74,11 @@ export const buildCacheCollectionHandler: InternalHandlerBuilder = ({
     mwApi,
     internalState
   ) => {
-    if (unsubscribeQueryResult.match(action) ||
-      queryThunk.fulfilled.match(action) ||
-      queryThunk.rejected.match(action)
-    ) {
+    if (canTriggerUnsubscribe(action)) {
       const state = mwApi.getState()[reducerPath]
-      let queryCacheKey: QueryCacheKey
-
-      if (unsubscribeQueryResult.match(action)) {
-        queryCacheKey = action.payload.queryCacheKey
-      } else {
-        queryCacheKey = action.meta.arg.queryCacheKey
-      }
+      const { queryCacheKey } = unsubscribeQueryResult.match(action)
+        ? action.payload
+        : action.meta.arg
 
       handleUnsubscribe(
         queryCacheKey,
--- a/packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
@@ -49,6 +49,7 @@ export const THIRTY_TWO_BIT_MAX_TIMER_SECONDS = 2_147_483_647 / 1_000 - 1
 export const buildCacheCollectionHandler: InternalHandlerBuilder = ({
   reducerPath,
   api,
+  queryThunk,
   context,
   internalState,
 }) => {
@@ -99,6 +100,22 @@ export const buildCacheCollectionHandler: InternalHandlerBuilder = ({
           state.config
         )
       }
+    }
+
+    if (
+      queryThunk.fulfilled.match(action) ||
+      queryThunk.rejected.match(action)
+    ) {
+      const state = mwApi.getState()[reducerPath]
+      const queryCacheKey = action.meta.arg.queryCacheKey
+
+      handleUnsubscribe(
+        queryCacheKey,
+        state.queries[queryCacheKey]?.endpointName,
+        mwApi,
+        state.config
+      )
+
     }
   }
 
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
```diff
@@ -100,6 +101,22 @@ export const buildCacheCollectionHandler: InternalHandlerBuilder = ({
         )
       }
     }
+
+    if (
+      queryThunk.fulfilled.match(action) ||
+      queryThunk.rejected.match(action)
+    ) {
+      const state = mwApi.getState()[reducerPath]
+      const queryCacheKey = action.meta.arg.queryCacheKey
```
phryneas, line 110: This block only differs from the above in line 71, so I wonder if those could be consoldiated somehow 🤔

packages/toolkit/src/query/tests/buildHooks.test.tsx
```diff
@@ -730,6 +730,60 @@ describe('hooks tests', () => {
...
+        )
+      }
+
+      render(<User />, { wrapper: storeRef.wrapper })
+      fireEvent.click(screen.getByTestId('fetchButton'))
+      fireEvent.click(screen.getByTestId('fetchButton'))
+      fireEvent.click(screen.getByTestId('fetchButton'))
+
+      await act(async () => {
+        await delay(1000)
```
phryneas, line 780: As that hasn't been defined in this file yet, you'll have to define it somewhere at the top of the file

```ts
  function delay(ms: number) {
    return new Promise((resolve) => setTimeout(resolve, ms))
  }
```

packages/toolkit/src/query/tests/buildHooks.test.tsx
```diff
@@ -730,6 +730,60 @@ describe('hooks tests', () => {
...
+
+      function User() {
+        const [fetchTest, { isFetching, isUninitialized }] =
+          pokemonApi.endpoints.getTest.useLazyQuery()
+
+        return (
+          <div>
+            <div data-testid="isUninitialized">{String(isUninitialized)}</div>
+            <div data-testid="isFetching">{String(isFetching)}</div>
+            <button data-testid="fetchButton" onClick={() => fetchTest(Math.random())}>
```
phryneas, line 767: Maybe better something like `i++` instead of `Math.random` with `i` being a local `let` variable in the test.

packages/toolkit/src/query/tests/buildHooks.test.tsx
aryaemami59: I believe you should also just be able to re-use the `delay` function from `msw`:

```ts
import { delay } from 'msw'

await delay(1000)
```

packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
riqts: My newest commit should reflect these conditionals being turned into one. I am sure there's some TS trick to make the conditional simpler yet again if needed, but I am unsure of what that is off the top of my head.

```ts
 if (unsubscribeQueryResult.match(action) ||
      queryThunk.fulfilled.match(action) ||
      queryThunk.rejected.match(action)
    ) {
      const state = mwApi.getState()[reducerPath]
      let queryCacheKey: QueryCacheKey

      if (unsubscribeQueryResult.match(action)) {
        queryCacheKey = action.payload.queryCacheKey
      } else {
        queryCacheKey = action.meta.arg.queryCacheKey
      }
```

packages/toolkit/src/query/tests/buildHooks.test.tsx
riqts: Good idea, most recent iteration should reflect these changes :)

packages/toolkit/src/query/tests/buildHooks.test.tsx
riqts: My confusion here stems from the fact that it IS imported [here](https://github.com/reduxjs/redux-toolkit/blob/edafc36db93ca3dc6cfd8588f7e21fbb37e7aad2/packages/toolkit/src/query/tests/buildHooks.test.tsx#L39) and the test block directly below it uses it without problem.

Most likely I have just missed something painstakingly obvious, but I added a definition in the test block regardless.

santirodriguezaffonso:
Is this going to be merged someday? 😭

markerikson:
@santirodriguezaffonso : the PR has been up for only 3 days, and we're all busy atm. Please don't beg for things to be merged.

EskiMojo14:
and in the meantime, you can click on the link in the codesandbox message to use the build that's already published.

packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
```diff
@@ -66,9 +67,18 @@ export const buildCacheCollectionHandler: InternalHandlerBuilder = ({
...
   ) => {
-    if (unsubscribeQueryResult.match(action)) {
+    if (unsubscribeQueryResult.match(action) ||
+      queryThunk.fulfilled.match(action) ||
+      queryThunk.rejected.match(action)
+    ) {
       const state = mwApi.getState()[reducerPath]
-      const { queryCacheKey } = action.payload
+      let queryCacheKey: QueryCacheKey
+
```
phryneas, line 76:
```suggestion
  const canTriggerUnsubscribe = isAnyOf(
    unsubscribeQueryResult.match,
    queryThunk.fulfilled,
    queryThunk.rejected
  )
  const handler: ApiMiddlewareInternalHandler = (action, mwApi) => {
    if (canTriggerUnsubscribe(action)) {
      const state = mwApi.getState()[reducerPath]
      const { queryCacheKey } = unsubscribeQueryResult.match(action)
        ? action.payload
        : action.meta.arg
```

Let's get this down in size even a little bit further.

packages/toolkit/src/query/tests/buildHooks.test.tsx
```diff
@@ -730,6 +730,67 @@ describe('hooks tests', () => {
...
+              fetchUser
+            </button>
+          </div>
+        )
+      }
+
+      render(<User />, { wrapper: storeRef.wrapper })
+      fireEvent.click(screen.getByTestId('fetchButton'))
+      fireEvent.click(screen.getByTestId('fetchButton'))
+      fireEvent.click(screen.getByTestId('fetchButton'))
```
phryneas, line 784:
```suggestion
      fireEvent.click(screen.getByTestId('fetchButton'))
      checkNumQueries(3)
```

Let's add a check showing that these are initially 3 and then go down to 1.

phryneas:
Mark already said this, but please be a bit more patient. 

We are all doing this in our free time, and some things can take days, weeks or even months - it all depends on how many resources we have at any given time, given our day jobs and private lifes we have to balance with voluntary Open Source.

packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
riqts: Amazing! This is what I was trying to wrack my brain to come up with originally, but forgot to come back to it.
Newest commit should reflect this change.

packages/toolkit/src/query/tests/buildHooks.test.tsx
riqts: Good idea! Added the additional check

riqts:
> Is this going to be merged someday? 😭

@santirodriguezaffonso 
Sorry buddy, you are being held back by my poor code quality! I require Lenz to do all the thinking for me, I am just his vessel to add code.

phryneas (approved):
No "poor code quality" here - libraries just need different code than applications. 
Your code was good from the start, we just have special requirements to keep bundle size down as a library :)

Thank you for the contribution!

santirodriguezaffonso:
Thanks guys! You rock 🤟🏻

Issue #4290, opened 2024-03-21
How to customize headers (async) on a per endpoint basis?


rwilliams3088:
I am working on implementing security for my web application with AWS Cognito. I've written a little Auth API that will return an access token for a given scope, and I would like to customize my endpoints so that they automatically add the Authorization header to the request with an appropriately scoped access token. 

So, for instance, my endpoint for reading the schedule would need an access token with the scope "schedules.read". 

Since I may need to asynchronously fetch the accessToken if a valid one for that scope isn't cached, I can't use the `headers` property on the endpoint.

There is the `prepareHeaders` function on the baseQuery which looks promising. However, I'd need a good way to inform that function of what scopes to request for the endpoint in question. Is there mechanism to add and retrieve metadata for endpoints that could be used in `prepareHeaders`? Or a way to use an asynchronous function on the endpoint definition to set the headers?

markerikson:
https://github.com/reduxjs/redux-toolkit/releases/tag/v2.3.0 now exposes an `extraOptions` field into `prepareHeaders`, as well as the URL `arg` for the current query.

PR #4291, opened 2024-03-21
fetchBaseQuery: expose extraOptions to prepareHeaders

phryneas:
Triggered by #4290 and the fact that we had similar asks before. Not sure if the extra generic arg is worth it. Thoughts, @EskiMojo14  & @markerikson  ?


EskiMojo14 (approved):
i think some solutions are possible with api.endpoint but this is nicer 👍🏻

phryneas:
Would you prefer it like this or without the generic and just `unknown`? 🤔

EskiMojo14:
definitely the generic

Issue #4285, opened 2024-03-21
RTKQ OpenApi CodeGen Common Parameters


sebws:
Hi there,

Using the RTKQ OpenApi CodeGen, [common parameters](https://swagger.io/docs/specification/describing-parameters/) are tricky to deal with. I have set some common parameters to be put in the headers for every query. The generated hooks require these parameters on every hook use. I would much prefer to be able to define these within the base query or something like that. Is there a way to do something like this?

sebws:
I’ve seen issue #3006 which is also relevant, but has no solution.

Issue #3006, opened 2022-12-14
How to omit certain headers from hook args during codegen?


blwinters:
I'm guessing that this isn't currently possible, but I wanted to ask and start a discussion. I'm using RTK Query and the codegen tool with OpenAPI on our server. Our server requires a custom `Organization-ID` header for almost all endpoints to identify which of the user's organization is active on the client (like different teams on Slack). Because this header is required everywhere, I'm adding it to headers via the `prepareHeaders` function in the RTK Query base query.

This part works great. However, when I run codegen all of the endpoint hooks now require `Organization-ID` as an argument. 

Generated hook arg type:
```
export type VehiclesInspectionCreateApiArg = {
  'Organization-ID': string
  vehicleInspectionReport: VehicleInspectionReport
}
```
which gets called like:
```
const [submitInspectionData] = useVehiclesInspectionCreateMutation()
const onPressSubmit = () => {
    submitInspectionData({
      'Organization-ID': '',
      vehicleInspectionReport: {
        // ...
      },
    })
  }
```

Presumably I can pass it an empty string and that will get overwritten during `prepareHeaders`, but I'd like to clean that up if possible. Is there any way that I can tell codegen to skip that header when generating the arg type for the hooks? If not, does anyone have ideas for how this capability could be added while keeping with the intent and patterns of this project? I'd be interested in contributing if so.

PR #4293, opened 2024-03-22
feat: add filterParameters option for rtkq-codegen-openapi

sebws:
Add an option to filter parameters out for RTK Query codegen for OpenApi schemas.

Both the path parameter and the body parameter aren't included in this, but could possibly be if it is thought that that would be helpful. I wasn't sure they would need to be filtered since they're less likely than other locations to be auto-added by a middleware.

This means it can filter out from the query, header, or cookie as seen in "Parameter Locations" in [the OpenAPI spec](https://swagger.io/specification/#parameter-locations)

Inspired by my issue #4285 and another relevant issue #3006


sebws:
going to open up a new PR with this as an addition to `overrideEndpoints`

PR #4297, opened 2024-03-23
Escape hatch added to defaultSerializeQueryArgs for unserializable data

### LINKED ISSUE #4279: `useStableQueryArgs` chokes on BigInt
When serializing a query object that includes a `BigInt`, rktQuery blows up with:
```
TypeError: Do not know how to serialize a BigInt
```

This seems to be due to this code in `buildQueryHooks`:
```
useStableQueryArgs(skip ? skipToken : arg, defaultSerializeQueryArgs, context.endpointDefinitions[name], name);
```
Which uses `defaultSerializeQueryArgs` instead of user-supplied overrides (i.e. `serializeQueryArgs`) -- where users could override the behavior and make bigints safe to serialize.

riqts:
fixes #4279

There's some discussion there about the expected behaviour and potential other fixes involving additional options or control for a user to override the defaultSerializeQueryArgs implementation.

I figured this is the best initial starting point because if `defaultQueryArgs` does throw an error due to say BigInt being passed, it will just continue the process that involves running the user defined `serializeQueryArgs` anyway.

I am not sure what the handling for the error should be. I presume a warning being sent every single time this happens especially if a user is aware it will happen isn't what we want? But I wanted to put something there for clarity.

### DIFF
--- a/packages/toolkit/src/query/defaultSerializeQueryArgs.ts
+++ b/packages/toolkit/src/query/defaultSerializeQueryArgs.ts
@@ -10,7 +10,7 @@ export const defaultSerializeQueryArgs: SerializeQueryArgs<any> = ({
   endpointName,
   queryArgs,
 }) => {
-  let serialized = ''
+  let serialized
 
   const cached = cache?.get(queryArgs)
 
--- a/packages/toolkit/src/query/defaultSerializeQueryArgs.ts
+++ b/packages/toolkit/src/query/defaultSerializeQueryArgs.ts
@@ -17,7 +17,9 @@ export const defaultSerializeQueryArgs: SerializeQueryArgs<any> = ({
   if (typeof cached === 'string') {
     serialized = cached
   } else {
-    const stringified = JSON.stringify(queryArgs, (key, value) =>
+    let stringified = ''
+    try {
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/query/defaultSerializeQueryArgs.ts
```diff
@@ -27,6 +29,12 @@ export const defaultSerializeQueryArgs: SerializeQueryArgs<any> = ({
             }, {})
         : value,
     )
+    } catch (e) {
```
rkofman, line 32: (I'm relatively new to open source process, so if I should keep my comments to the issue rather than the PR, please let me know)

I suspect that for most use cases, silently swallowing an error is worse than throwing it.

riqts: I think most likely you are correct, that's why I requested guidance on the error handling. Do you have any suggestions?

rkofman: Actually, I have a thought! Let me put together a PR.

rkofman: See: https://github.com/reduxjs/redux-toolkit/pull/4315/files

PR #4305, opened 2024-03-26
feat: extend endpoint overrides for openapi codegen

sebws:
Add an option to filter parameters into endpoint overrides for RTK Query codegen for OpenApi schemas.

Both the path parameter and the body parameter aren't included in this, but could possibly be if it is thought that that would be helpful. I wasn't sure they would need to be filtered since they're less likely than other locations to be auto-added by a middleware.

This means it can filter out from the query, header, or cookie as seen in "Parameter Locations" in [the OpenAPI spec](https://swagger.io/specification/#parameter-locations)

It only runs on the first (by order of endpointOverrides array) filter. This was the existing behaviour for query type, and makes enough sense.

Inspired by my issue https://github.com/reduxjs/redux-toolkit/issues/4285 and another relevant issue https://github.com/reduxjs/redux-toolkit/issues/3006

### DIFF
--- a/docs/rtk-query/usage/code-generation.mdx
+++ b/docs/rtk-query/usage/code-generation.mdx
@@ -64,7 +64,7 @@ npx @rtk-query/codegen-openapi openapi-config.ts
 
 #### Generating tags
 
-If your OpenAPI specification uses [tags](https://swagger.io/docs/specification/grouping-operations-with-tags/), you can specify the `tag` option to the codegen.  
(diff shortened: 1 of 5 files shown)


sebws:
hi there, is there anything I can do to make this more mergeable? same with my other PRs. thanks!

sebws:
hi there, any chance someone could take a look at this?

markerikson:

PR #4307, opened 2024-03-27
Fix codegen-openapi tests

aryaemami59:
## Overview

Previously the codegen tests would create side effects which made them unpredictable and caused them to produce inconsistent results. This PR mainly focuses on making the unit tests reliable both during local development and CI.

## This PR:

  - [X] Replaces the old `cli` command in `cli.test.ts` with a promisified version of Node's `exec` which mimicks the actual cli behavior more closely.
  - [X] Revamps the `tsconfig` setup to have a `tsconfig.build.json` and `tsconfig.json` as opposed to a `test/tsconfig.json` with `baseUrl`.
  - [X] Uses path aliasing to test against the output `lib` folder during CI. `vite-tsconfig-paths` was added to `devDependencies` to manage this easier.
  - [X] Creates a more comprehensive CI workflow for the codegen-openapi that tests against the final build artifact on different operating systems (This change will become active in #4509 since that PR is responsible for fixing the build output).


packages/rtk-query-codegen-openapi/test/cli.test.ts
```diff
@@ -1,76 +1,85 @@
...
+    const fromJson = await fs.readFile(path.resolve(tmpDir, 'example.ts'), 'utf-8');
 
     expect(fromTs).toEqual(fromJs);
     expect(fromJson).toEqual(fromJs);
   }, 120_000);
 
   test("missing parameters doesn't fail", async () => {
-    const out = await cli([`./test/config.invalid-example.json`], __dirname);
-    expect(out.stderr).toContain("Error: path parameter petId does not seem to be defined in '/pet/{petId}'!");
+    await expect(() => cli([`./test/config.invalid-example.json`])).rejects.toThrowError(
```
markerikson, line 81: :hand: looks like this test is failing because an error _isn't_ being thrown

markerikson:
@aryaemami59 fyi the "ts/json/etc should be the same" test seems to be flaky - sometimes it can't find `example.ts`.  Some kind of filesystem race condition, maybe?

aryaemami59:
Yup that's exactly what it is. I'll have to look into it later.

PR #4308, opened 2024-03-27
[Docs]: Replace all instances of `TypedUseSelectorHook` with `useSelector.withTypes<RootState>()`

aryaemami59:
## This PR:

  - [X] Replaces all instances of `TypedUseSelectorHook` with `useSelector.withTypes<RootState>()`.
  - [X] Replaces instances of `const useAppDispatch: () => AppDispatch = useDispatch` with `const useAppDispatch = useDispatch.withTypes<AppDispatch>()`
  - [X] Replaces instances of `const useAppStore: () => AppStore = useStore` with `const useAppStore = useStore.withTypes<AppStore>()`

### DIFF
--- a/docs/tutorials/typescript.md
+++ b/docs/tutorials/typescript.md
@@ -76,13 +76,12 @@ Since these are actual variables, not types, it's important to define them in a 
 
 ```ts title="app/hooks.ts"
 import { useDispatch, useSelector } from 'react-redux'
-import type { TypedUseSelectorHook } from 'react-redux'
 import type { RootState, AppDispatch } from './store'
 
 // highlight-start
 // Use throughout your app instead of plain `useDispatch` and `useSelector`
-export const useAppDispatch: () => AppDispatch = useDispatch
-export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector
+export const useAppDispatch = useDispatch.withTypes<AppDispatch>()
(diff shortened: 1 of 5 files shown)


aryaemami59:
On a side note it might be worth getting rid of some of the packages in the `resolutions` field, seems like they're causing more problems than they solve. If I have some time within the next couple of weeks I'll see if I can start doing that incrementally.

packages/toolkit/src/dynamicMiddleware/react/index.ts
```diff
@@ -79,7 +79,12 @@ export const createDynamicMiddleware = <
       // @ts-ignore
       context === ReactReduxContext
         ? useDefaultDispatch
-        : createDispatchHook(context)
+        : createDispatchHook(
+            context as Context<ReactReduxContextValue<
```
EskiMojo14, line 83: i think we should just fix the parameter to actually expect a nullable context
it'll be a breaking change but technically a bug fix

EskiMojo14: very annoying that the resolutions field meant this was missed 😕

aryaemami59: Agreed!

EskiMojo14:
thanks!

PR #4315, opened 2024-03-30
Handle bigint in useQuerySubscription

rkofman:
fixes https://github.com/reduxjs/redux-toolkit/issues/4279

Handles query arg values that include `bigint`s while retaining behavior of always initiating refetch (even when user-supplied `serializeQueryArgs` doesn't change). This is an alternative proposed solution from https://github.com/reduxjs/redux-toolkit/pull/4297 -- which fixes `bigint` errors but diverges refetching behavior.

Note: this changes behavior for all callers of `defaultSerializeQueryArgs`; which _should_ be safe if the library doesn't try to re-serialize the args anywhere. It _may_ result in slightly surprising behavior for consumers (a potentially unexpected default serialization of bigint values) - but since it currently breaks hard with an exception for those use cases, unlikely to be a breaking change for _anybody_.

### DIFF
--- a/packages/toolkit/src/query/defaultSerializeQueryArgs.ts
+++ b/packages/toolkit/src/query/defaultSerializeQueryArgs.ts
@@ -18,6 +18,7 @@ export const defaultSerializeQueryArgs: SerializeQueryArgs<any> = ({
     serialized = cached
   } else {
     const stringified = JSON.stringify(queryArgs, (key, value) =>
+      // Sort the object keys before stringifying, to prevent useQuery({ a: 1, b: 2 }) having a different cache key than useQuery({ b: 2, a: 1 })
       isPlainObject(value)
         ? Object.keys(value)
             .sort()
@@ -32,7 +33,6 @@ export const defaultSerializeQueryArgs: SerializeQueryArgs<any> = ({
     }
     serialized = stringified
   }
-  // Sort the object keys before stringifying, to prevent useQuery({ a: 1, b: 2 }) having a different cache key than useQuery({ b: 2, a: 1 })
   return `${endpointName}(${serialized})`
 }
 
@@ -48,7 +48,9 @@ export const bigIntSafeSerializeQueryArgs: SerializeQueryArgs<any> = ({
     serialized = cached
   } else {
     const stringified = JSON.stringify(queryArgs, (key, value) => {
+      // Translate bigint fields to a serializable value
       value = typeof value === 'bigint' ? { $bigint: value.toString() } : value
+      // Sort the object keys before stringifying, to prevent useQuery({ a: 1, b: 2 }) having a different cache key than useQuery({ b: 2, a: 1 })
       value = isPlainObject(value)
         ? Object.keys(value)
             .sort()
--- a/packages/toolkit/src/query/defaultSerializeQueryArgs.ts
+++ b/packages/toolkit/src/query/defaultSerializeQueryArgs.ts
@@ -9,36 +9,7 @@ const cache: WeakMap<any, string> | undefined = WeakMap
 export const defaultSerializeQueryArgs: SerializeQueryArgs<any> = ({
   endpointName,
   queryArgs,
-}) => {
-  let serialized = ''
-
-  const cached = cache?.get(queryArgs)
-
-  if (typeof cached === 'string') {
-    serialized = cached
-  } else {
-    const stringified = JSON.stringify(queryArgs, (key, value) =>
-      // Sort the object keys before stringifying, to prevent useQuery({ a: 1, b: 2 }) having a different cache key than useQuery({ b: 2, a: 1 })
-      isPlainObject(value)
-        ? Object.keys(value)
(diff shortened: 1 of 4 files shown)


packages/toolkit/src/query/defaultSerializeQueryArgs.ts
```diff
@@ -17,33 +18,40 @@ export const defaultSerializeQueryArgs: SerializeQueryArgs<any> = ({
...
-  // Sort the object keys before stringifying, to prevent useQuery({ a: 1, b: 2 }) having a different cache key than useQuery({ b: 2, a: 1 })
   return `${endpointName}(${serialized})`
 }
 
 export type SerializeQueryArgs<QueryArgs, ReturnType = string> = (_: {
   queryArgs: QueryArgs
   endpointDefinition: EndpointDefinition<any, any, any, any>
   endpointName: string
+  // Allows for a custom stringify replacer while keeping key-sorting behavior. e.g. for serializing bigint.
+  replacer?: (key: string, value: any) => {}
```
rkofman, line 48: Special reviewer attention is requested here: I'm not sure I fully understand `SerializeQueryArgs` vs. `InternalSerializeQueryArgs` and whether it's proper to update both of them here.

packages/toolkit/src/query/defaultSerializeQueryArgs.ts
rkofman, line 1: I suspect additional tests in `defaultSerializeQueryArgs.test.ts` would be useful before this PR is merge-ready; but I would like overall feedback on this approach before polishing it up.

rkofman:
@riqts thoughts on this PR? Is there anything process-wise I need to do before I get a review? e.g. I'm happy to finish implementing additional unit tests.

riqts:
Haha, I think unfortunately just nothing moves fast in volunteer work. Realistically the primary people who are fully tuned in to this area of the codebase are @phryneas or @markerikson as they wrote the code here and are most familiar with its intentions. 

But unfortunately I think they have been mega busy and there is a decent amount of cognitive load they have to take on for this particular fix. The value vs effort for this particular fix probs just doesn't feel great with the workload.

I will try add some context in the issue of the surrounding code so it's easier for the reviewers to engage with the PRs.

Potentially this PR itself might be preferred if instead of exposing a replacer it just handled potential incoming bigints within the `defaultSerializeQueryArgs` and the implementation of the exposed option can just go in a seperate issue to get it moved along?

markerikson:
Yeah, to be frank, I've been otherwise occupied with personal stuff for much of this year and haven't had mental bandwidth to do meaningful RTK work (either myself or reviewing actual code PRs).

I've seen this PR, I'm very aware of it, but can't give an immediate timeline when I'll have the bandwidth to think through this one.

rkofman:
[non-technical aside]
♥️ to both of you! I have little idea what it takes to maintain a project of this scope / popularity and am very glad that there are people who do. I suspect my "can I be more helpful?" pings may have landed as more demanding  of immediate attention than intended. To be clear: I am not blocked in my project (and tbh, a full half of the places we use bigints today -- we really shouldn't be, but that's another story).

The major pressure I'm feeling is: (1) I just don't know what the expectations are when contributing (how much I should carry water for the change myself, how much to wait for helpful context from others, etc.) and (2) I'm a little afraid that I will drop context over time, and would feel disappointed (at myself) for not helping to finalize and land this change if it pages out of my working memory. That sense of urgency is a matter of personal convenience and takes a back seat to maintainer needs.
[/aside]

@riqts happy to create an alternative PR that handles `bigint` directly in `defaultSerializeQueryArgs`. The reason I shied away from it in the first place is I have a difficult time reasoning through all the places `defaultSerializeQueryArgs` is used; and what impact serializing previously unserializable data would have. For example, are there places in the library that expect to _deserialize_ it? If so, those need to be updated with a mirror implementation. I would probably want to take a few days (up to a week) to poke around and see if I can get mysefl to understand the impact before opening that PR.

rkofman:
One more idea would be to just re-implement `defaultSerializeQueryArgs` in `useQuery`. This would:
1. remove the public interface change
2. maintain correct behavior
3. limit scope of serialization change to avoid regressions elsewhere

The main downside is code duplication -- but that should be easy to cleanup later for somebody with more context (it punts the decision of spreading the change everywhere vs. introducing a configuration parameter).

EskiMojo14:
from memory I don't think we deserialize the results of serializeQueryArgs - it's only used to compare equality of two values (for example, creating a cache key)

we keep hold of the original value for things like refetching so there's no need to deserialize

rkofman:
As suggested; updated to the simplest thing that can work (and added unit test for `defaultSerializeQueryArgs`).

rkofman:
(.. and updated the PR description to match the new code ...)

markerikson:
Okay, finally swinging back to this (had an evening where I had both time and mental energy to chew through a bunch of open PRs).  Thanks for the work here, and for putting up with the delay!

Issue #4304, opened 2024-03-26
flattenArg:true - does NOT work.


krasevych:
I have scheme like this:
```
//......
   "parameters": [
          {
            "name": "keys",
            "in": "query",
            "description": "test",
            "required": false,
            "schema": {
              "type": "array",
              "items": {
                "type": "string"
              }
            }
          }
        ]
//.....
```
take a look on ` "required": false`.

gen result when `flattenArg:true` is `type RequestArg = string[]` - wrong!
gen result when `flattenArg:false` is `type RequestArg ={ keys?: string[]` } - good! because `?`

Expected result when `flattenArg:true`: `type RequestArg = string[] | undefined`

markerikson:
Should be live in https://github.com/reduxjs/redux-toolkit/releases/tag/%40rtk-query%2Fcodegen-openapi%402.0.0-alpha.0 ! Please try it out and let us know if it works.

PR #4317, opened 2024-04-01
fix: flattened optional type not optional

sebws:
when flattening an argument, if the type is optional, it should be a union with undefined.

the PR implements this by adding a union node for the type of the arg. if the type is required, then the union node function is given just the original type and returns it. if the type is not required, then the union node function is given the original type and the undefined keyword, creating an optional type.

see issue #4304 for inspiration

### DIFF
--- a/packages/rtk-query-codegen-openapi/src/generate.ts
+++ b/packages/rtk-query-codegen-openapi/src/generate.ts
@@ -324,7 +324,6 @@ export async function generateApi(
     const queryArgValues = Object.values(queryArg);
 
     const isFlatArg = flattenArg && queryArgValues.length === 1;
-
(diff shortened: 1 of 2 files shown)


krasevych:
when will you merge it?

sebws:
> when will you merge it?

the maintainers aren’t as familiar with code gen so mightn’t be too quick. if you’d like, by clicking on “details” next to the code sandbox merge workflow, there are instructions to use this PR in your package.json

sebws:
hi there, any chance someone can take a look at this?

markerikson:
Looks reasonable, thanks!

markerikson:

PR #4332, opened 2024-04-06
strongly type endpoint name

EskiMojo14:


EskiMojo14:
closing as i don't think it's particularly useful

PR #4348, opened 2024-04-15
Allow providing custom slice creators when calling createSlice

### LINKED ISSUE #3837: Custom reducer creators

EskiMojo14:
```ts
const loaderSlice = createSlice({
  name: "loader",
  initialState: {} as Partial<Record<string, true>>,
  creators: { loader: loaderCreator },
  reducers: (create) => ({
    addLoader: create.loader({}),
  }),
});
```


packages/toolkit/src/tests/createSlice.test.ts
```diff
@@ -1100,6 +1101,43 @@ describe('createSlice', () => {
...
+      })
+      expect(() =>
+        createAppSlice({
+          name: 'loader',
+          initialState: {} as Partial<Record<string, true>>,
+          // @ts-expect-error name overlap
+          creators: { loader: loaderCreator },
+          reducers: (create) => ({
+            addLoader: create.loader({}),
+          }),
```
vahan-sahakyan, line 1135: Could you please explain the differences between this and the traditional builder pattern? Also, would you mind clarifying the purpose of the custom loadCreator?

EskiMojo14: what traditional builder pattern are you referring to?

vahan-sahakyan: My reference was to the action builder pattern in Redux Toolkit, where we use methods like addCase and addMatcher in createSlice to dynamically build actions and reducers. With this PR allowing custom slice creators, I wonder if there are potential issues when multiple creators are provided for the same slice. Could this lead to naming conflicts or unexpected behavior if two creators define overlapping functionality for the same action or state? I might be missing something though haha

EskiMojo14: `extraReducers` has a different builder pattern, because we don't need a return value from it. for `reducers` we need the returned object to infer slice actions and case reducers from, similar to how you're required to return `build.query` when building an RTKQ API.

There's handling for overlaps, this is the third PR in a set of three. See #3837 for the base logic.

vahan-sahakyan: ahh gotcha, thanks Ben!

PR #4352, opened 2024-04-17
Exclude internal docs from the website

aryaemami59:
## __Overview__:

Looks like [This change](https://github.com/reduxjs/redux-toolkit/commit/45cc9ae3350329cabc4ce7244b5bef00a77a35da) is causing the netlify build to fail due to broken links.

## __This PR__:

  - [X] Excludes internal docs from the website.


EskiMojo14:
superceded by #4353 - thank you though!

PR #4356, opened 2024-04-17
fix: pretty-quick ^3.1.0 not working with prettier >3

sebws:
pretty-quick ^3.1.0 doesn't work with prettier >3 (see [here](https://github.com/prettier/pretty-quick/issues/164)).

this is breaking the formatting husky pre-commit hook.

upgrading fixes the issue (see [here](https://github.com/prettier/pretty-quick/pull/182)).


EskiMojo14:
thanks!

PR #4361, opened 2024-04-18
Improve sorted entity adapter sorting performance

markerikson:
This PR:

- Updates the sorting logic for sorted entity adapters to hopefully be more efficient most of the time

Previously we were always getting the list of items via `Object.values(state.entities)`. However, that is always going to give us an array of items in _insertion_ order, not the most recent _sorted_ order.

If we instead go `state.ids.map(id => state.entities[id])`, we at least start with a sorted array.

However, there's a couple tricky cases. If a set of updates replaced an item's ID, that throws things out of whack. It's easier to fall back to the old behavior.  There are also apparently a couple test cases where we can end up with duplicate IDs and have to watch for that.

On my own machine, the perf test case with 100K items gave me numbers like this with the existing code:

```
Upsert: sortComparer called: 3,058,124 times
Duration to upsert an item: 1,177 ms
Update: sortComparer called: 1,529,076 times
Duration to update an item: 1,139 ms
```

After this fix, it drops to:

```
Upsert: sortComparer called: 1,629,402 times
Duration to upsert an item: 533 ms
Update: sortComparer called: 100,032 times
Duration to update an item: 544 ms
```

That's still not _fast_, but it's a 50% improvement.

I _may_ have another idea or two around tracking updated items, removing those, and re-inserting them, but will need to try those later.

### DIFF
--- a/packages/toolkit/src/entities/tests/sorted_state_adapter.test.ts
+++ b/packages/toolkit/src/entities/tests/sorted_state_adapter.test.ts
@@ -1,6 +1,11 @@
 import type { EntityAdapter, EntityState } from '../models'
 import { createEntityAdapter } from '../create_adapter'
-import { createAction, createSlice, configureStore } from '@reduxjs/toolkit'
+import {
+  createAction,
+  createSlice,
+  configureStore,
+  nanoid,
+} from '@reduxjs/toolkit'
 import type { BookModel } from './fixtures/book'
 import {
   TheGreatGatsby,
@@ -247,7 +252,7 @@ describe('Sorted State Adapter', () => {
 
     const { ids, entities } = withUpdated
 
-    expect(ids.length).toBe(2)
+    expect(ids).toEqual(['a', 'c'])
     expect(entities.a).toBeTruthy()
     expect(entities.b).not.toBeTruthy()
     expect(entities.c).toBeTruthy()
@@ -582,6 +587,86 @@ describe('Sorted State Adapter', () => {
 
     expect(withUpdate.ids).toEqual(['b'])
     expect(withUpdate.entities['b']!.title).toBe(book1.title)
+  })
+
+  it('should minimize the amount of sorting work needed', () => {
+    const PARAMETERS = {
+      NUM_ITEMS: 10_000,
+    }
+
+    type Entity = { id: string; name: string; position: number }
+
+    let numSorts = 0
+
+    const adaptor = createEntityAdapter({
+      selectId: (entity: Entity) => entity.id,
+      sortComparer: (a, b) => {
+        numSorts++
+        if (a.position < b.position) return -1
+        else if (a.position > b.position) return 1
+        return 0
+      },
+    })
+
+    const initialState: Entity[] = new Array(PARAMETERS.NUM_ITEMS)
+      .fill(undefined)
+      .map((x, i) => ({
+        name: `${i}`,
+        position: Math.random(),
+        id: nanoid(),
+      }))
+
+    const entitySlice = createSlice({
+      name: 'entity',
+      initialState: adaptor.getInitialState(undefined, initialState),
+      reducers: {
+        updateOne: adaptor.updateOne,
+        upsertOne: adaptor.upsertOne,
+        upsertMany: adaptor.upsertMany,
+      },
+    })
+
+    const store = configureStore({
+      reducer: {
+        entity: entitySlice.reducer,
+      },
+      middleware: (getDefaultMiddleware) => {
+        return getDefaultMiddleware({
+          serializableCheck: false,
+          immutableCheck: false,
+        })
+      },
+    })
+
+    store.dispatch(
+      entitySlice.actions.upsertOne({
+        id: nanoid(),
+        position: Math.random(),
+        name: 'test',
+      }),
+    )
+
+    // These numbers will vary because of the randomness, but generally
+    // with 10K items the old code had 200K+ sort calls, while the new code
+    // is around 130K sort calls.
+    expect(numSorts).toBeLessThan(200_000)
+
+    const { ids } = store.getState().entity
+    const middleItemId = ids[(ids.length / 2) | 0]
+
+    numSorts = 0
+
+    store.dispatch(
+      // Move this middle item near the end
+      entitySlice.actions.updateOne({
+        id: middleItemId,
+        changes: {
+          position: 0.99999,
+        },
+      }),
+    )
+    // The old code was around 120K, the new code is around 10K.
+    expect(numSorts).toBeLessThan(25_000)
   })
 
   describe('can be used mutably when wrapped in createNextState', () => {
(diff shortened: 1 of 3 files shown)


packages/toolkit/src/entities/tests/sorted_state_adapter.test.ts
```diff
@@ -584,6 +589,86 @@ describe('Sorted State Adapter', () => {
...
+        entity: entitySlice.reducer,
+      },
+      middleware: (getDefaultMiddleware) => {
+        return getDefaultMiddleware({
+          serializableCheck: false,
+          immutableCheck: false,
+        })
+      },
+    })
+
```
phryneas, line 640: add a 
```
numSorts = 0
```
here, the important question is how long later sorts take

markerikson:
Revised numbers if we reset the sort counter right before the upsert:

```
// Without fix
Upsert: sortComparer called: 1,529,445 times
Duration to upsert an item: 1,379 ms
Update: sortComparer called: 1,529,444 times
Duration to update an item: 1,482 ms

// With fix
Upsert: sortComparer called: 100,027 times
Duration to upsert an item: 847 ms
Update: sortComparer called: 100,032 times
Duration to update an item: 803 ms
```

phryneas:
Just leaving this here - I suspect that pre-sorting the incoming array and manually merging them might save a bunch of comparisons on top (assuming that we don't only insert at the end):
```js
function merge(models: readonly T[], state: R): void {
    const entities = state.entities as Record<Id, T>
    const oldEntities = Object.values(state.ids).map((id: Id) => entities[id])
    const newEntities = models.slice().sort(sort)

    // Insert/overwrite all new/updated
    models.forEach((model) => {
      entities[selectId(model)] = model
    })

    const newSortedIds: Id[] = []
    let o = 0,
      n = 0
    while (o < oldEntities.length && n < newEntities.length) {
      const oldEntity = oldEntities[o] as T,
        oldId = selectId(oldEntity),
        newEntity = newEntities[n]

      // old entity has been overwritten by new entity, skip comparison
      if (entities[oldId] !== oldEntity) {
        o++
        continue
      }

      const comparison = sort(oldEntity, newEntity)
      if (comparison < 0) {
        newSortedIds.push(oldId)
        o++
        continue
      }
      const newId = selectId(newEntity)
      if (comparison > 0) {
        newSortedIds.push(newId)
        n++
        continue
      }
      newSortedIds.push(oldId)
      o++
      newSortedIds.push(newId)
      n++
    }
    while (o < oldEntities.length) {
      newSortedIds.push(selectId(oldEntities[o]))
      o++
    }
    while (n < newEntities.length) {
      newSortedIds.push(selectId(newEntities[n]))
      n++
    }
    if (!areArraysEqual(state.ids, newSortedIds)) {
      state.ids = newSortedIds
    }
  }
```
This was just a way around `merge` which will only work for insert/upsert, but I believe my approach would also work with updating ids - it just needs the sorted entries from before, the merged object and the inserted/updated objects (also sorted). Could just as well be a method argument like you did.
Feel free to swipe anything if you feel it's a useful approach.

markerikson:
Current benchmark numbers:

**Existing logic:**

```
Original Setup: sortComparer called 1,529,087 times in 839ms
Insert One (random): sortComparer called 1,529,107 times in 1,018ms
Insert One (middle): sortComparer called 1,529,120 times in 1,311ms
Insert One (end): sortComparer called 1,529,137 times in 1,348ms
Add Many: sortComparer called 1,546,947 times in 1,283ms
Update One (end): sortComparer called 1,546,945 times in 1,338ms
Update One (middle): sortComparer called 1,546,948 times in 1,375ms
Update One (replace): sortComparer called 1,547,010 times in 1,403ms
```

**Binary Insertion / this PR:**

```
Original Setup: sortComparer called 1,529,043 times in 642ms
Insert One (random): sortComparer called 17 times in 569ms
Insert One (middle): sortComparer called 16 times in 452ms
Insert One (end): sortComparer called 17 times in 487ms
Add Many: sortComparer called 16,728 times in 519ms
Update One (end): sortComparer called 101,033 times in 590ms
Update One (middle): sortComparer called 101,035 times in 586ms
Update One (replace): sortComparer called 101,055 times in 700ms
```

I haven't tried to do any other perf profiling to measure where the _rest_ of the overhead is coming from. I'd guess that Immer is a large chunk of it.  But still, that's a lot fewer comparisons for those cases.

markerikson:
Latest times with additional Immer-related optimizations:

```
Original Setup: sortComparer called 1,529,339 times in 600ms
Insert One (random): sortComparer called 16 times in 349ms
Insert One (middle): sortComparer called 16 times in 441ms
Insert One (end): sortComparer called 16 times in 377ms
Add Many: sortComparer called 16,710 times in 384ms
Update One (end): sortComparer called 101,034 times in 550ms
Update One (middle): sortComparer called 101,036 times in 492ms
Update One (replace): sortComparer called 101,055 times in 655ms
```

markerikson:
This _should_ be an improvement, let's get it out.

unadlib:
I apologize for the sudden comment on this merged PR.

Here's a performance comparison test between Immer and Mutative that focuses solely on "Insert One (end)." 
The results are as follows:

```
// 10K items
Original Setup: sortComparer called 119,925 times in 40ms
Insert One (end): sortComparer called 13 times in 12ms
Use only Immer: sortComparer called 0 times in 10ms
Use only Mutative: sortComparer called 0 times in 2ms

// 100K items
Original Setup: sortComparer called 1,529,198 times in 349ms
Insert One (end): sortComparer called 17 times in 165ms
Use only Immer: sortComparer called 0 times in 120ms
Use only Mutative: sortComparer called 0 times in 24ms

// 1000K items
Original Setup: sortComparer called 18,604,782 times in 3,559ms
Insert One (end): sortComparer called 19 times in 2,859ms
Use only Immer: sortComparer called 0 times in 1,787ms
Use only Mutative: sortComparer called 0 times in 368ms
```

Here's the test code:

```ts
const initialItems = generateItems(INITIAL_ITEMS)

measureComparisons('Original Setup', () => {
  store.dispatch(entitySlice.actions.upsertMany(initialItems))
})

measureComparisons('Insert One (end)', () => {
  store.dispatch(
    entitySlice.actions.upsertOne({
      id: nanoid(),
      position: 0.9998,
      name: 'test',
    }),
  )
})

measureComparisons('Use only Immer', () => {
  produce(store.getState(), (draft) => {
    const id = nanoid();
    draft.entity.ids.push(id)
    draft.entity.entities[id] = {
      id,
      position: 0.9998,
      name: 'test',
    }
  })
})

measureComparisons('Use only Mutative', () => {
  create(store.getState(), (draft) => {
    const id = nanoid();
    draft.entity.ids.push(id)
    draft.entity.entities[id] = {
      id,
      position: 0.9998,
      name: 'test',
    }
  })
})
```

I know this performance test is quite specific. If you have any further ideas or questions, I’d be more than happy to discuss them.

PR #4363, opened 2024-04-22
Add extended API to ForkedTaskApi

### LINKED ISSUE #3314: Extend ForkedTaskApi to include API
Having considered porting a lot of code from saga to listeners, I've noticed that I am reaching for the same pattern which is a bit verbose in order to have deeply nested forking:

Example:
```js
function doThingEvent(action, api) {
  const forks = [];
  forks.push(api.fork((forkApi) => doAnotherThing({ api, ...forkApi })));
  forks.push(api.fork((forkApi) => doAnotherThing2({ api, ...forkApi })));
  await api.pause(Promise.all(forks.map(f => f.result)));
}
```

When in reality, the `ForkedTaskApi` could just include things from the parent, like `getState()`, `dispatch()`, and `fork()`. Not only would this make the code easier to read `forks.push(api.fork(doAnotherThing))`, but it would also mean the type API for both the fork and the task is the same.

kkirby:
This PR aims to fix #3314.

This is currently a draft as I have yet to fully update all TypeScript types and add tests.

I'm creating this to get any feedback.

### DIFF
--- a/packages/toolkit/src/listenerMiddleware/index.ts
+++ b/packages/toolkit/src/listenerMiddleware/index.ts
@@ -82,52 +82,6 @@ const INTERNAL_NIL_TOKEN = {} as const
 
 const alm = 'listenerMiddleware' as const
 
-const createFork = (
-  parentAbortSignal: AbortSignalWithReason<unknown>,
-  parentBlockingPromises: Promise<any>[],
-) => {
-  const linkControllers = (controller: AbortController) =>
-    addAbortSignalListener(parentAbortSignal, () =>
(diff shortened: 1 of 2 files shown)


markerikson:
Ah, nice, thank you!

Out of curiosity, what prompted the move of `createFork` in the file?

kkirby:
> what prompted the move of `createFork` in the file?

I actually didn't. I moved `createTakePattern` before `createFork` since the updated `createFork` depends on `createTakePattern`. I'll split up the code into more commits to make that more evident.

PR #4364, opened 2024-04-22
fix: ensure hook subscription failures do not reset isLoading state

smacpherson64:
This PR aims to resolve an issue where the isLoading state of a parent component was being reset by a child component if the child component was rendered conditionally.

Relates to: https://github.com/reduxjs/redux-toolkit/discussions/1349#discussioncomment-9175654

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -696,7 +696,7 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
     // isFetching = true any time a request is in flight
     const isFetching = currentState.isLoading
     // isLoading = true only when loading while no data is present yet (initial load with no data in the cache)
-    const isLoading = !hasError && !hasData && isFetching
+    const isLoading = (!lastResult || lastResult.isLoading || lastResult.isUninitialized) && !hasData && isFetching
     // isSuccess = true when data is present
     const isSuccess = currentState.isSuccess || (isFetching && hasData)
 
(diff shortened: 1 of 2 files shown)


phryneas:
That won't prevent a refetch though, just display `isFetching` instead of `isLoading`. Is that really your problem here?

smacpherson64:
@phryneas true! This will still fetch twice. I tried to scope this change to only avoid the infinite loop where `isLoading` reverts back to true for the parent when rendering descendants conditionally.

The `refetchOnError` idea from the discussion would be a much nicer as it would only fetch once.

packages/toolkit/src/query/react/buildHooks.ts
```diff
@@ -687,10 +687,16 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+    let error = currentState.isError ? currentState.error : lastResult?.error
+    if (error === undefined) error = currentState.error
+
+    const hasError = error !== undefined
+
     // isFetching = true any time a request is in flight
     const isFetching = currentState.isLoading
     // isLoading = true only when loading while no data is present yet (initial load with no data in the cache)
-    const isLoading = !hasData && isFetching
+    const isLoading = !hasError && !hasData && isFetching
```
phryneas, line 699:
```suggestion
    const isLoading = (!lastResult || lastResult.isLoading || lastResult.isUninitialized) && !hasData && isFetching
```

Does this fix your bug, too?

smacpherson64: Yes, It does look to pass the test. Will convert to use this.

PR #4379, opened 2024-04-30
Update usage-with-typescript.mdx

kostasx:
- Replaced unused `api` identifier with `usePostsQuery`
  - `api` is imported but never used in this example, whereas `usePostsQuery` is used without explicitly being imported 
- Removed unnecessary else (after return)


EskiMojo14:
thanks!

PR #4386, opened 2024-05-02
try out simplifying matcher code

EskiMojo14:


packages/toolkit/src/matchers.ts
```diff
@@ -404,22 +361,11 @@ export function isAsyncThunkAction<
...
-    const combinedMatcher = isAnyOf(...matchers)
-
-    return combinedMatcher(action)
+  const matchers: Matcher<any>[] = []
+
+  for (const asyncThunk of asyncThunks) {
+    matchers.push(asyncThunk.pending, asyncThunk.rejected, asyncThunk.fulfilled)
   }
+
+  return isAnyOf(...matchers)
```
phryneas, line 370:
```suggestion
  return isAnyOf(...asyncThunks.flatMap(asyncThunk => [asyncThunk.pending, asyncThunk.rejected, asyncThunk.fulfilled]))
```

EskiMojo14: the thought did occur, but i wasn't sure what the support for flatMap was like :)

phryneas: I mean, we already use it in `combineSlices` :)

Also, seems well supported: https://caniuse.com/mdn-javascript_builtins_array_flatmap

EskiMojo14: haha, good point

PR #4389, opened 2024-05-03
Prevent inference of any in create.asyncThunk

EskiMojo14:
fixes https://github.com/reduxjs/redux-toolkit/issues/4060#issuecomment-2091201562

we're also making this change over in #3894 (https://github.com/reduxjs/redux-toolkit/pull/3894/files#diff-ee57feab8b392faa360ea66691a847a2dbed2f82486f5726c3723d6a194f22f0R507) so it's mostly unavoidable once we have creators

### DIFF
--- a/packages/toolkit/src/tests/createSlice.test.ts
+++ b/packages/toolkit/src/tests/createSlice.test.ts
@@ -678,7 +678,7 @@ describe('createSlice', () => {
         initialState: [] as any[],
         reducers: (create) => ({
           thunkReducers: create.asyncThunk(
-            function payloadCreator(arg, api) {
+            function payloadCreator(arg: string, api) {
               return Promise.resolve('resolved payload')
             },
             { pending, fulfilled, rejected, settled },
@@ -722,7 +722,7 @@ describe('createSlice', () => {
         reducers: (create) => ({
           thunkReducers: create.asyncThunk(
             // payloadCreator isn't allowed to return never
-            function payloadCreator(arg, api): any {
+            function payloadCreator(arg: string, api): any {
               throw new Error('')
             },
             { pending, fulfilled, rejected, settled },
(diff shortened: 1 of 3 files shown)


phryneas:
This seems to fail a lot of tests 🤔 

But weird as it is, if you get it working, I'm not opposed to a change like this - seems to fix a real bug.

EskiMojo14:
heh, those tests are relying on broken behaviour - will fix when i can

packages/toolkit/src/tests/createSlice.test.ts
```diff
@@ -678,7 +678,7 @@ describe('createSlice', () => {
         initialState: [] as any[],
         reducers: (create) => ({
           thunkReducers: create.asyncThunk(
-            function payloadCreator(arg, api) {
+            function payloadCreator(arg: string, api) {
```
phryneas, line 681: Hmmm... This brings up the question: will this still be okay in an untyped `.js` file for non-TS-users or will it make it unusable?

EskiMojo14: it's the same behaviour as `createAsyncThunk`. if they're not using type checking on their JS there will be nothing to complain, and if they are then they'll need to annotate:
```js
const getPost = createAsyncThunk(
  'posts/getPost',
  /**
   *
   * @param {number} id
   * @returns {Promise<Post>}
   */
  async (id) => {
    const response = await fetch(
      `https://jsonplaceholder.typicode.com/posts/${id}`,
    )
    const data = await response.json()
    return data
  },
)
```

EskiMojo14: you can even set the ApiConfig:
```js
const getPost = createAsyncThunk(
  'posts/getPost',
  /** @type {import("@reduxjs/toolkit").AsyncThunkPayloadCreator<Post, number, { state: RootState }>}*/
  async (id, api) => {
    const response = await fetch(
      `https://jsonplaceholder.typicode.com/posts/${id}`,
    )
    const data = await response.json()
    return data
  },
)
```

PR #4412, opened 2024-05-16
Fix missed spot where use of `current` may fail if the value is not a draft

markerikson:
I had added a `getCurrent()` util that checks if the value is draftable before calling `current(value)`, but missed a spot.  Turns out that a sequence of `removeAll()` + `upsertMany()` hits the case where now `state.ids` is a plain array, and thus `current(state.ids)` will throw because it's a plain value and not a draft.

### DIFF
--- a/packages/toolkit/package.json
+++ b/packages/toolkit/package.json
@@ -103,7 +103,7 @@
     "format": "prettier --write \"(src|examples)/**/*.{ts,tsx}\" \"**/*.md\"",
     "format:check": "prettier --list-different \"(src|examples)/**/*.{ts,tsx}\" \"docs/*/**.md\"",
     "lint": "eslint src examples",
-    "test": "vitest --run --typecheck",
+    "test": "vitest --typecheck --run ",
     "test:watch": "vitest --watch",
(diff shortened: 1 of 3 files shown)


packages/toolkit/package.json
```diff
@@ -103,7 +103,7 @@
     "format": "prettier --write \"(src|examples)/**/*.{ts,tsx}\" \"**/*.md\"",
     "format:check": "prettier --list-different \"(src|examples)/**/*.{ts,tsx}\" \"docs/*/**.md\"",
     "lint": "eslint src examples",
-    "test": "vitest --run --typecheck",
+    "test": "vitest --typecheck --run ",
```
aryaemami59, line 106: Just out of curiousity, what's the difference?

markerikson: Argument ordering.  If I do `yarn test sorted_state_adapter`, I want that filename to get passed through as a modifier to `--run` so that it filters files.  If you do it with `--typecheck` last, it ignores it, and runs _all_ test files instead.

aryaemami59: ah I see, seems like it's not an issue anymore in the latest version of `vitest`.

PR #4415, opened 2024-05-16
chore: update `size-limit` workflow and configuration

aryaemami59:
## **This PR:**

- [X] Fixes `size-limit` config to monitor bundle size changes in preparation for any potential changes we'll have to make to migrate to React 19.


EskiMojo14:
For what it's worth, I don't think we actually want size info for every single entry point and export possible. We intentionally only measure entry points and exports we're actually concerned about the size for.

It's less of an issue since we filter out unchanged entries, but it still feels a little spammy.

markerikson:
Yeah, this is definitely a lot more info than we need or care about.  (Also, it made the `size` job take 14 minutes to run.)

Is there actually a problem with the current config?

aryaemami59:
I cut down on the `size-limit` config, let me know if this is still not ideal and I will try to look for another solution.

markerikson:
@aryaemami59 what does the new output of this PR look like, and what's the diff vs the current size comments?

packages/toolkit/package.json
```diff
@@ -72,7 +72,7 @@
     "@typescript-eslint/eslint-plugin": "^6",
     "@typescript-eslint/parser": "^6",
     "axios": "^0.19.2",
-    "esbuild": "^0.23.0",
+    "esbuild": "^0.25.0",
```
aryaemami59, line 75: Had to bump `esbuild` to get rid of some TS errors.

EskiMojo14:
it looks like v13 just released, removing the need to use `jiti` (since node can run TS directly now) - could this get updated please? :)

aryaemami59:
> it looks like v13 just released, removing the need to use `jiti` (since node can run TS directly now) - could this get updated please? :)

yes sir.

Issue #4411, opened 2024-05-16
listenerMiddleware/error Error: [Immer] 'current' expects a draft, got: 


zy-jtp:
### Example Repo

https://github.com/zy-jtp/rtk-next14-error

### Context
- Using `pages`, not `app` router.
- After upgrading to next v14.2.3
- This error pops up on certain dispatch calls within the listeners.

The call triggering the error is a dispatch call to modals slice from an `onClick` callback
```
listenerApi.dispatch(
          modalsSlice.actions.openModal({
            dialogId: 'modify',
            closeable: !action.payload?.uncloseableModal,
          })
        )
```
its initial state comes from an `entityAdapter` 
```
export const Modal = createEntityAdapter<ModalConfig, DialogId>({
  selectId: (modal) => modal.dialogId,
});
export const modalsSlice = createSlice({
initialState: Modal.getInitialState()
})
```

### Runtime Error
```javascript
listenerMiddleware/error Error: [Immer] 'current' expects a draft, got: 
    at die (immer.mjs:64:11)
    at current (immer.mjs:704:5)
    at splitAddedUpdatedEntities (redux-toolkit.modern.mjs:1454:74)
    at upsertManyMutably (redux-toolkit.modern.mjs:1577:30)
    at upsertOneMutably (redux-toolkit.modern.mjs:1574:12)
    at runMutator (redux-toolkit.modern.mjs:1423:9)
    at Object.operation [as upsertOne] (redux-toolkit.modern.mjs:1427:7)
    at openModal (modals.slice.ts:18:13)
    at eval (redux-toolkit.modern.mjs:780:20)
    at Immer2.produce (immer.mjs:593:20)
    at eval (redux-toolkit.modern.mjs:779:64)
    at Array.reduce (<anonymous>)
    at reducer (redux-toolkit.modern.mjs:760:25)
    at reducer (redux-toolkit.modern.mjs:1173:14)
    at combination (redux.mjs:327:31)
    at hu (<anonymous>:3:1043)
    at yu (<anonymous>:3:1331)
    at <anonymous>:3:4637
    at Object.dispatch (redux.mjs:172:22)
    at Object.s [as dispatch] (<anonymous>:3:5093)
    at dispatch (redux-toolkit.modern.mjs:598:22)
    at eval (redux-logger.js:1:7667)
    at eval (create-middleware.js:17:22)
    at eval (rtk-query.modern.mjs:2217:17)
[...]

PR #4420, opened 2024-05-21
Fix Immer current usage when calling addManyMutably more than once

JacobShafer:
A getCurrent() util was created that checks if the value is draftable before calling current(value). Asequence of addOne() + addOne() hits the case where now state.ids is a plain array, and thus current(state.ids) will throw because it's a plain value and not a draft.

Resolves: https://github.com/reduxjs/redux-toolkit/issues/4411


EskiMojo14:
thanks!

Issue #2437, opened 2022-06-22
Cannot run codegen CLI command on .cts or .cjs files (CommonJS)


Macludde:
I have a project which uses Vite and thus also uses `type:"module"` in its package.json. 

### Error description
When running `npx @rtk-query/codegen-openapi` I get the following error:

```  
const err = new Error(getErrRequireEsmMessage(filename, parentPath, packageJsonPath))
              ^
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module:/.../src/api/generationConfig.ts
require() of ES modules is not supported.
require() of /.../src/api/generationConfig.ts from /.../node_modules/@rtk-query/codegen-openapi/lib/bin/cli.js is an ES module file as it is a .ts file whose nearest parent package.json contains "type": "module" which defines all .ts files in that package scope as ES modules.
Instead change the requiring code to use import(), or remove "type": "module" from /.../package.json.
```
### Reason for error
The above error is caused by the config file being inside the package.json project whilst also having `type: "module"` specified. One fix to get around this is declaring generationConfig.ts as generationConfig.cts instead to tell it to use CommonJS imports instead of ESM.

However, this doesn't work with `@rtk-query/codegen-openapi` as in [the cli file](https://github.com/reduxjs/redux-toolkit/blob/666f18203db5e62616733fb082ea38178814715c/packages/rtk-query-codegen-openapi/src/bin/cli.ts), on line 46, the regex for checking if the path given is to a correct file is as follows 
`/\.(jsx?|tsx?|jsonc?)?$/`
This doesn't match to files ending in .cjs or .cts

### Possible solution
An easy fix would be to change the regex to
`/\.(c?(jsx?|tsx?)|jsonc?)?$/`

philjones88:
Hit this issue too. Trying to make private NPM packages of our generated files to save having multiple copies all over the place.

Macludde:
Part of [@rtk-query/codegen-openapi@1.0.0-alpha.2](https://github.com/reduxjs/redux-toolkit/releases/tag/%40rtk-query%2Fcodegen-openapi%401.0.0-alpha.2)

philjones88:
@Macludde I've updated to that package but still hit the issue:

```
/Users/philjones/Documents/<redacted>/node_modules/ts-node/dist-raw/node-internal-errors.js:46
  const err = new Error(getErrRequireEsmMessage(filename, parentPath, packageJsonPath))
              ^
[...]

PR #4446, opened 2024-06-06
fix: unclear error when run cts config without ts-node in rtk-query-codegen-openapi

lawvs:
Related to https://github.com/reduxjs/redux-toolkit/issues/2437 https://github.com/reduxjs/redux-toolkit/pull/2438

When you use cts without ts-node in `rtk-query-codegen-openapi`, the error message is unclear and difficult to troubleshoot.

Before
```
(node:56745) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)

/Users/user/project/openapi-config.cts:1
import type { ConfigFile } from '@rtk-query/codegen-openapi'
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

After

```
Encountered a TypeScript configfile, but neither esbuild-runner nor ts-node are installed.
```

### DIFF
--- a/packages/rtk-query-codegen-openapi/src/bin/cli.ts
+++ b/packages/rtk-query-codegen-openapi/src/bin/cli.ts
@@ -38,7 +38,7 @@ const configFile = program.args[0];
 if (program.args.length === 0 || !/\.(c?(jsx?|tsx?)|jsonc?)?$/.test(configFile)) {
   program.help();
 } else {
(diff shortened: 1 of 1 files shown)


lawvs:
> thanks - for future proofing, could you also add detection of .mts?

Done!

markerikson:

Issue #4448, opened 2024-06-07
[TS2742] The inferred type of '...' cannot be named without a reference to ...


ryuujo1573:
> Related to https://github.com/microsoft/TypeScript/issues/58743

It is not working if I try to export reducers/slices from a package.

Is this an actual TS issue?

Here's a possible [reproduction](https://github.com/ryuujo1573/rtk-ts2742-demo)

```
src/logic/client/user.ts:11:7 - error TS2742: The inferred type of 'userSlice' cannot be named without a reference to '../../../node_modules/@reduxjs/toolkit/dist/createSlice'. This is likely not portable. A type annotation is necessary.
```

telephant:
same here

telephant:
emm, don't export `nikkiSlice`, only export `nikkiReducer`

aryaemami59:
Related to #3962.

aryaemami59:
It seems like in your case we have to export the `CaseReducerDefinition` interface from the main `index.d.ts` file.

Side note: It seems like if a type is exported from an internal `.d.ts` file, but not exported from `index.d.ts`, then the error becomes: `The inferred type of 'nikkiSlice' cannot be named without a reference to '../../node_modules/@reduxjs/toolkit/dist/createSlice'. This is likely not portable. A type annotation is necessary.`

But if that same type is not exported at all the error message becomes much clearer like this: `Exported variable 'nikkiSlice' has or is using name 'CaseReducerDefinition' from external module "RTK-issue-3962/node_modules/.pnpm/@reduxjs+toolkit@2.2.5/node_modules/@reduxjs/toolkit/dist/createSlice" but cannot be named.`
which actually tells us which type is the problematic one.

EskiMojo14:
@aryaemami59 that makes sense, thanks. i think we wouldn't be against exporting CaseReducerDefinition - in fact my open PR https://github.com/reduxjs/redux-toolkit/pull/3894/files#diff-5278ccae8e99ae0373490ef372d1953e0f3da303d948485ef7e00bd7de7fffa3 exports more that we can export just in case.

aryaemami59:
@EskiMojo14 Wonderful, I have a list of types that need to be exported, thankfully they're not that many (as of now, might add more to the list if I discover more problems), will _hopefully_ put up a PR soon for a partial fix.

aryaemami59:
Can you run this command and try again to see if it fixes the issue:

```bash
[...]

Issue #4401, opened 2024-05-09
Export DraftableEntityState, IsAny, UseQuery and other types


chenhebing:
https://github.com/microsoft/TypeScript/issues/58474

aryaemami59:
Seems like this is related to #4066

chenhebing:
> Seems like this is related to #4066

Thanks for the reply. It seems to be the same problem, so I will close this issue directly.

aryaemami59:
@chenhebing Thanks, I'm still working on resolving this issue in a way that doesn't involve exporting internal types.

markerikson:
Fixed in https://github.com/reduxjs/redux-toolkit/releases/tag/v2.2.7 !

Issue #4108, opened 2024-01-22
ReducerWithInitialState is not portlable according to TS


krutoo:
Hi, in my project i use toolkit 2.0.0 and i have type check build during emit declaration files only with `isolatedModules: true` option in tsconfig:

```json
{
  "compilerOptions": {
    "target": "ES2021",
    "moduleResolution": "Bundler",
    "baseUrl": ".",
    "jsx": "react-jsx",
    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true,
    "noUncheckedIndexedAccess": true,
    "outDir": "dist/types",
    "declaration": true,
    "sourceMap": true,
    "emitDeclarationOnly": true,
    "isolatedModules": true
  },
  "include": ["src"],
  "exclude": ["node_modules", "*.test.*", "**/__test__/*"]
}
```

i have error:
```
src/internal/core/redux/index.ts:64:14 - error TS2742: The inferred type of 'reducer' cannot be named without a reference to '../../../../node_modules/@reduxjs/toolkit/dist/createReducer'. This is likely not portable. A type annotation is necessary.
```

Can we add export of ReducerWithInitialState from entry point of package?

krutoo:
i have reducer like:

```ts
export const reducer = createReducer({/* ... */}, () => {/* ... */})
```

shermify:
It's a known issue when declaration=true with ES modules
https://github.com/reduxjs/redux-toolkit/issues/4066

aryaemami59:
Can you run this command and try again to see if it fixes the issue:

```bash
npm install https://pkg.csb.dev/reduxjs/redux-toolkit/commit/da50c578/@reduxjs/toolkit
```

markerikson:
Fixed in https://github.com/reduxjs/redux-toolkit/releases/tag/v2.2.7 !

krutoo:
@markerikson looks like release title is 2.2.6 for tag 2.2.7

EskiMojo14:
@krutoo fixed, thanks

Issue #4066, opened 2024-01-12
Typescript errors using Node ESM module resolution and "declaration: true"


shermify:
I noticed myself and a few other people getting an error `The inferred type of X cannot be named without a reference to '../../node_modules/@reduxjs/toolkit/dist/query/react/buildHooks'. This is likely not portable. A type annotation is necessary.`

This happens when you have `declaration: true` or you're using composite projects as can be seen in my very simple reproduction here. https://github.com/shermify/redux-toolkit-ts-repro

The error in slice.ts
![Screenshot from 2024-01-11 20-05-08](https://github.com/reduxjs/redux-toolkit/assets/12567037/39a72b11-56b6-44c0-88a0-38d9e73cc4b6)

I believe the problem is that many of the types are not explicitly exported from redux-toolkit. In this case, it's looking for UseQuery, but there are many others that give me problems. I think this is an issue because ESM modules do not allow you to reach into packages or files that are not declared explicitly in the package.json. Indeed, changing module resolution to "Node" solves it, but "NodeNext", "Node16", and "bundler" have problems.  "Node" resolution is not suggested for new projects.

Exporting the types explicitly from redux-toolkit fixes it, also adding the types to the export section of the package.json of redux-toolkit fixes it.   However, there are a bunch of types that cause this error, so I'm not sure what the best solution would be. I don't think there is a workaround because the user cannot actually import the necessary types even if they wanted to manually due to ECMAScript constraints. 

![Screenshot from 2024-01-11 20-13-29](https://github.com/reduxjs/redux-toolkit/assets/12567037/917633e3-6018-46e8-95dd-338838abd423)

The use case for "declaration: true" would be, for example, creating a library with redux-toolkit actions or having a monorepo with shared packages. Let me know if any of this sounds incorrect!

markerikson:
That sounds plausible.  @eric-crowell was doing some contributions prior to RTK 2.0 related to checking type exports for this sort of thing ( https://github.com/reduxjs/redux-toolkit/pulls?q=is%3Apr+author%3Aeric-crowell ), but it's possible more changes are needed here.

aryaemami59:
@markerikson Do you think it would be worth it to try and see if it would be plausible to bundle the type definitions into one file like we're doing with the other repos?

markerikson:
[...]

PR #4467, opened 2024-06-19
Fix: TypeScript Type Portability Issues

aryaemami59:
# **Overview**

The type portability issues in this repository are primarily caused by three specific TypeScript errors:

1. [**`TS4023`**](https://github.com/microsoft/TypeScript/issues/37888): `Exported variable 'useAppDispatch' has or is using name 'TimeResponse' from external module "/home/runner/work/redux-toolkit/redux-toolkit/examples/publish-ci/vite/src/app/services/times" but cannot be named.`

   - **Solution**: This one can actually be resolved quite often by converting the problematic type from an `interface` to a `type` alias. Other times we might have to export the problematic type.

2. [**`TS2742`**](https://typescript.tv/errors/#ts2742): `The inferred type of 'useGetQuotesQuery' cannot be named without a reference to '../../../node_modules/@reduxjs/toolkit/dist/query/react/buildHooks'. This is likely not portable. A type annotation is necessary.`

   - **Solution**: Bundling the type definitions seems to have resolved this issue.

3. `TS2527`: `The inferred type of 'quotesApiSlice' references an inaccessible 'unique symbol' type. A type annotation is necessary.`
   This one took a while for me to figure out, It turns out in order for a `unique symbol` type to be "accessible", the `unique symbol` has to be imported upon declaration, so we want something that looks like this:

   ```ts
   export declare const coreModuleName: unique symbol
   ```

   But `dts-rollup-plugin` gives us this instead:

   ```ts
   declare const coreModuleName: unique symbol
   export { coreModuleName }
   ```

   But TypeScript treats `declare const coreModuleName: unique symbol` and `export { coreModuleName };` as 2 different `unique symbol`s because each reference to `coreModuleName` is completely unique.

   - **Solution**: While somewhat hacky, I wrote a small script that takes all exported `unique symbol`s and converts them from:

     ```ts
     declare const coreModuleName: unique symbol
     export { coreModuleName }
     ```

     To:

     ```ts
     export declare const coreModuleName: unique symbol
     ```

## **This PR:**

- [x] Resolves:

  - #1806
  - #2314
  - #2485
  - #3202
  - #3568
  - #3962
  - #3970
  - #3983
  - #4066
  - #4108
  - #4401
  - #4448
  - #4492
  - #4498
  - #4518

- [x] Converts the following types from `interface` to `type` alias:

  - `AsyncThunkSliceReducerConfig`
  - `ReducerDefinition`
  - `AsyncThunkSliceReducerDefinition`
  - `MutationTypes`
  - `MutationThunkArg`
  - `MutationCacheLifecycleApi`
  - `MutationLifecycleApi`
  - `BaseEndpointTypes`
  - `CacheLifecyclePromises`
  - `MutationBaseLifecycleApi`
  - `QueryLifecyclePromises`
  - `LifecycleApi`
  - `PromiseWithKnownReason`
  - `QueryTypes`
  - `StartQueryActionCreatorOptions`
  - `QueryThunkArg`
  - `UseQuerySubscriptionOptions`
  - `GetSelectorsOptions`
  - `EndpointDefinitionWithQueryFn`
  - `EndpointDefinitionWithQuery`
  - `BaseActionCreator`
  - `WithMiddleware`
  - `ReactDynamicMiddlewareInstance`
  - `AddListenerOverloads`
  - `TypedActionCreatorWithMatchFunction`
  - `TypedActionCreator`
  - `CaseReducerDefinition`
  - `InjectIntoConfig`
  - `InjectedSlice`
  - `QueryHooks`
  - `MutationHooks`
  - `ThunkWithReturnValue`

- [x] Exports the following types:

  - `StartQueryActionCreatorOptions`
  - `CombinedSliceReducer`
  - `BaseEndpointDefinition`
  - `AddMiddleware`
  - `CreateDispatchWithMiddlewareHook`
  - `DynamicDispatch`

- [x] Exports the following values:

  - `UNINITIALIZED_VALUE`
  - `_NEVER`

- [x] Removes relative `declare module` statements.
- [x] Fixes issues related to name collision of types. Sometimes `tsup` has a hard time with aliasing type imports so if we have something like this:

  ```ts
  import type { Dispatch as ReduxDispatch } from 'redux'
  ```

  `tsup` somehow renames `ReduxDispatch` back to `Dispatch` which causes issues when you have a generic type parameter with the same name so this:

  ```ts
  export type AddMiddleware<
    State = any,
    Dispatch extends ReduxDispatch<UnknownAction> = ReduxDispatch<UnknownAction>
  > = {
    (...middlewares: Middleware<any, State, Dispatch>[]): void
    withTypes<MiddlewareConfig extends MiddlewareApiConfig>(): AddMiddleware<
      GetState<MiddlewareConfig>,
      GetDispatch<MiddlewareConfig>
    >
  }
  ```

  Becomes:

  ```ts
  export type AddMiddleware<
    State = any,
    Dispatch extends Dispatch<UnknownAction> = Dispatch<UnknownAction>
  > = {
    (...middlewares: Middleware<any, State, Dispatch>[]): void
    withTypes<MiddlewareConfig extends MiddlewareApiConfig>(): AddMiddleware<
      GetState<MiddlewareConfig>,
      GetDispatch<MiddlewareConfig>
    >
  }
  ```

  Which can cause issues as well as make things confusing. This issue is fixed by doing 2 things:

  1. Renaming some generic type parameters to prevent name collision with imported external types:

     ```diff
     export type AddMiddleware<
       State = any,
     - Dispatch extends ReduxDispatch<UnknownAction> = ReduxDispatch<UnknownAction>,
     + DispatchType extends ReduxDispatch<UnknownAction> = ReduxDispatch<UnknownAction>,
       > = {
      (...middlewares: Middleware<any, State, DispatchType>[]): void
      withTypes<MiddlewareConfig extends MiddlewareApiConfig>(): AddMiddleware<
       GetState<MiddlewareConfig>,
           GetDispatch<MiddlewareConfig>
         >
      }
     ```

  2. Removing some import aliases:

     ```diff
     - import type { Dispatch as ReduxDispatch } from 'redux'
     + import type { Dispatch } from 'redux'
     ```

- [x] Patches `console-testing-library` so we can run the type tests with `"moduleResolution": "Bundler"`. Previously it would fail due to `console-testing-library` not having a `types` field in it's subpath exports.
- [x] Adds 3 new example workspaces:

1. **`@examples-type-portability/bundler`** to test for `"moduleResolution": "Bundler"`.
2. **`@examples-type-portability/nodenext-cjs`** to test for `"moduleResolution": "NodeNext"` with TypeScript's CJS syntax and `"type": "commonjs"` in `package.json`.
3. **`@examples-type-portability/nodenext-esm`** to test for `"moduleResolution": "NodeNext"` with ESM syntax and `"type": "module"` in `package.json`.

## **Things to keep in mind**

- Although I prefer using `interfaces`, it might be worthwhile to switch to `type` aliases going forward. `interfaces` are more prone to type portability issues compared to `type` aliases, which often dissolve into their constituents. I highly recommend using `type` aliases for all new types unless `interfaces` are necessary (e.g., for declaration merging).

- I've noticed that using an `interface` as the return type of an external function often leads to type portability issues unless the `interface` itself is also exported.


EskiMojo14:
this is going to cause a painful amount of conflicts with other PRs 😅

aryaemami59:
I know, I'm sorry :( but I really think this will be worth it.

aryaemami59:
Just out of curiosity what are some of the other PRs that are gonna have conflicts with this one? And is there anything I can do to minimize said conflicts?

EskiMojo14:
the changes in this PR are very wide reaching (including some renames that seem unrelated to the issue at hand?), so i suspect quite a few will encounter issues, but in particular the creator PR #3837 is quite significant in how much it reworks, and will certainly encounter issues when trying to reconcile with this.

aryaemami59:
The renames were necessary, they were causing issues with name collisions, I'll see if I can work around it to minimize conflicts with other PRs.

aryaemami59:
Alright this one should be good to go. Let me know if this needs anything.

packages/toolkit/src/createAsyncThunk.ts
```diff
@@ -125,15 +118,10 @@ export type AsyncThunkConfig = {
...
-type GetState<ThunkApiConfig> = ThunkApiConfig extends {
-  state: infer State
-}
-  ? State
-  : unknown
 type GetExtra<ThunkApiConfig> = ThunkApiConfig extends { extra: infer Extra }
   ? Extra
   : unknown
-type GetDispatch<ThunkApiConfig> = ThunkApiConfig extends {
+type GetDispatchType<ThunkApiConfig> = ThunkApiConfig extends {
```
markerikson, line 124: :question: what required the renames?

packages/toolkit/src/dynamicMiddleware/index.ts
```diff
@@ -1,27 +1,29 @@
...
+  DynamicMiddlewareInstance,
+  GetDispatch,
+  GetState,
+  MiddlewareApiConfig,
 } from './types'
 
 const createMiddlewareEntry = <
   State = any,
-  Dispatch extends ReduxDispatch<UnknownAction> = ReduxDispatch<UnknownAction>,
+  DispatchType extends Dispatch<UnknownAction> = Dispatch<UnknownAction>,
```
markerikson, line 23: :question: This seems cosmetic - why was this needed?

packages/toolkit/src/query/apiTypes.ts
```diff
@@ -1,30 +1,358 @@
...
   BaseQuery extends BaseQueryFn,
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   Definitions extends EndpointDefinitions,
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   ReducerPath extends string,
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   TagTypes extends string,
-> {}
+> {
+  [coreModuleName]: {
```
markerikson, line 53: :question: I assume this was added because we're no longer doing interface merging, or something along those lines?

markerikson (approved):
Probably have more questions later, but these are the first things that come to mind.

packages/toolkit/src/createAsyncThunk.ts
```diff
@@ -1,24 +1,17 @@
 import type { Dispatch, UnknownAction } from 'redux'
-import type {
-  PayloadAction,
-  ActionCreatorWithPreparedPayload,
-} from './createAction'
-import { createAction } from './createAction'
 import type { ThunkDispatch } from 'redux-thunk'
+import type { ActionCreatorWithPreparedPayload } from './createAction'
+import { createAction } from './createAction'
+import type { GetState } from './dynamicMiddleware/types'
```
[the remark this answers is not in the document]
aryaemami59: I did it to de-duplicate the `GetState` type to prevent having `GetState$1` in the type declaration file.

packages/toolkit/src/createAsyncThunk.ts
aryaemami59: This one was done to avoid name collision with `GetDispatch` type in `src/dynamicMiddleware/types.ts`.

packages/toolkit/src/dynamicMiddleware/index.ts
aryaemami59: The renames were done because `tsup` sometimes has a hard time with aliasing type imports so something like:

```ts
import type { Dispatch as ReduxDispatch } from 'redux'
```

It somehow renames `ReduxDispatch` back to `Dispatch` which causes issues when you have a generic type parameter with the same name:

```ts
export type AddMiddleware<
  State = any,
  Dispatch extends ReduxDispatch<UnknownAction> = ReduxDispatch<UnknownAction>,
> = {
  (...middlewares: Middleware<any, State, Dispatch>[]): void
  withTypes<MiddlewareConfig extends MiddlewareApiConfig>(): AddMiddleware<
    GetState<MiddlewareConfig>,
    GetDispatch<MiddlewareConfig>
  >
}
```

which can turn into:

```ts
export type AddMiddleware<
  State = any,
  Dispatch extends Dispatch<UnknownAction> = Dispatch<UnknownAction>,
> = {
  (...middlewares: Middleware<any, State, Dispatch>[]): void
  withTypes<MiddlewareConfig extends MiddlewareApiConfig>(): AddMiddleware<
    GetState<MiddlewareConfig>,
    GetDispatch<MiddlewareConfig>
  >
}
```

Which can cause issues.

packages/toolkit/src/query/apiTypes.ts
aryaemami59: We can still do `interface` merging, I only removed `declare module`s that were using relative paths. Since now we are bundling the type defs, and relative paths will no longer be applicable.

packages/toolkit/src/createAsyncThunk.ts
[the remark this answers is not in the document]
markerikson: Could we either rename the type from `dynamicMiddleware` instead, or consolidate into one type somehow?

packages/toolkit/src/createAsyncThunk.ts
markerikson: Same thing. I'd rather keep the `cAT` version as the "primary" - seems like there'd be less churn overall, and `dynamicMiddleware` is both a newer file and less important.

aryaemami59: Sure thing

aryaemami59: The only issue is the `GetDispatch` type from `src/dynamicMiddleware/types.ts` is the one that is exported through the main entry point, so if I rename that one, it'll be a breaking change.

markerikson:
Okay, changes here make sense based on the descriptions, and I think Arya's got it cleaned up to where this is basically "just" the `interface -> type` changes (rather than moving type definitions or contents between files).

Let's get this in.

PR #4475, opened 2024-06-21
Fix CI tests

aryaemami59:
## **Overview**

It seems like the new version of Vite much like Node.js allows packages to self-reference themselves, which became a problem considering how we were running the tests against the build artifact during CI. The main issue was that when running the tests against the build artifact, Vitest was prioritizing the `@reduxjs/toolkit` source files instead of the external `@reduxjs/toolkit` inside the `node_modules` folder.


EskiMojo14:
thanks!

PR #4509, opened 2024-07-12
Fix codegen-openapi build output

aryaemami59:
#4307 Will need to be merged first.

## **This PR**:

- [X] Fixes codegen-openapi build output by adding `tsup`.
- [X] Fixes the CI tests for codegen-openapi.
- [X] Modernizes build output by adding the `exports` field to provide ESM/CJS dual compatibility.
- [X] Uses the compact `lodash.camelcase` package instead of the entire `lodash` library in order to minimize package bundle size.
- [X] Updates TypeScript and Vitest to make sure the library is compatible with the latest versions of those libraries.
- [X] Fixes the unit tests by copying over the changes made in #4307.
- [X] Adds `vite-tsconfig-paths` to get a similar setup for unit tests as `@reduxjs/toolkit`.


markerikson:
This does a _lot_ more than just "add `tsup`".  Could you summarize the rest of the changes and why they're necessary?

Also, note that adding `exports` will require a major version bump.

markerikson:
Okay, merged the test changes over and this is clean.

@aryaemami59 just to check: could you remind me _why_ we need any of the build setup changes in the first place?

aryaemami59:
Because the current build output is broken. The reason why it's broken is because we build with `moduleResolution` set to `Bundler` but we don't use one. So our build output is ESM but the library is CJS. Hence why we get the "cannot use import statement outside of a module" error.

markerikson:
Okay. Can we do this PR _without_ the `package.exports` change at all? That will require a major version bump. We can do that if necessary, I just don't want to do it casually.

aryaemami59:
1 - I don't think it can be done without.
2 - Since we're changing the entire build step, I think this could be considered a breaking change even without the exports field added.

markerikson:
Okay. Any idea what other codegen-related changes (PRs, issues, etc) ought to go in as well?

aryaemami59:
I'm not sure tbh. #4568 looks good.

PR #4510, opened 2024-07-12
Disable transparent workspaces

aryaemami59:
## **This PR**:

- [X] Sets `enableTransparentWorkspaces` to `false` in `.yarnrc.yml` file. By default Yarn will link workspaces if their versions match a semver range. We disable this behavior as to not accidentally require a workspace instead of an actual published node module.


EskiMojo14:
hmm, why would it be a bad thing to use a workspace instead of the node module?

from my understanding this would break things like being able to make changes to both the base package and the codegen, for example

markerikson:
Yeah, can you clarify why this change is needed overall? Is this an actual problem we already have, or trying to prevent a future problem?

aryaemami59:
@markerikson Mostly trying to prevent a future problem. If we have something like this in any of the workspaces:

```json
"devDependencies": {
  "@reduxjs/toolkit": "^2.0.0"
}
```

Yarn will just use the local workspace version of `@reduxjs/toolkit` when we might not want to, setting `enableTransparentWorkspaces` to `false` forces us to be explicit about whether we want to use the local workspace or not. If we do want to use the local workspace version then we have to use the "workspace protocol":

```json
"devDependencies": {
  "@reduxjs/toolkit": "workspace:^"
}
```

PR #4519, opened 2024-07-18
Export UpdateDefinitions from toolkit/query

### LINKED ISSUE #4492: `UpdateDefinitions` type not exported from `@reduxjs/toolkit/dist/query/endpointDefinitions`
I'm trying to upgrade to Redux Toolkit v2 but running into a problem that I'm pretty sure is caused by the `UpdateDefinitions` type not being exported from `@reduxjs/toolkit/query`.

I have a monorepo with a separate package for my API endpoint, which is further split up into smaller chunks. `@reduxjs/toolkit` is a peer dependency in that package.

I have a base API defined like this:

```ts
export const baseApi = createApi({
  reducerPath: "api",
  endpoints: () => ({}),
});
```

And the additional API chunks are handled similar to this:

```ts
export const authApi = baseApi
  .enhanceEndpoints({
    addTagTypes: ["Permissions"],
  })
  .injectEndpoints({
    endpoints: (build) => ({
      permissions: build.query<PermissionsResponse, void>({
        query: () => "user/permissions",
        providesTags: () => [{ type: "Permissions", id: "CURRENT" }],
      }),
    }),
  });
```

This was working fine with v1, but after trying to upgrade to v2 I'm getting the error (at the `authApi` definition):

> The inferred type of 'authApi' cannot be named without a reference to '@reduxjs/toolkit/dist/query/endpointDefinitions'. This is likely not portable. A type annotation is necessary.ts(2742)

If I remove the `enhanceEndpoints` call (and leave just the `injectEndpoints` call) the issues goes away, but then I can no longer use the "Permissions" tag in `provideTags`. I'm fairly certain I've narrowed down the problem to the fact that the `UpdateDefinitions` type used by `enhanceEndpoints` is not exported by `@reduxjs/toolkit/query`. All the other referenced types seem to be exported except for this one. Is there a specific reason for omitting it? Or was it an oversight? Would it be worth me putting up a PR to export it?

joekrill:
When using `enhanceEndpoints` in a library, the return type cannot be inferred because the `UpdateDefinitions` type is not exported from `@reduxjs/toolkit/query`.  This exports that type. Fixes #4492. 

The CONTRIBUTING.md file mentions needing to update the `etc/redux-toolkit.api.md` file "If you changed external-facing types". I'm not sure if that's required in this case? I wasn't able to figure out how to regenerate that file. The instructions make it sounds like it gets generated during the build process - but when I deleted the file and re-ran the build script it did not get re-generated. None of the other scripts seem to regenerate this file, either. I tried running `api-extractor` directly but get an error. Is there some other manually step that needs to be taken to generate that file?

### DIFF
--- a/packages/toolkit/src/query/index.ts
+++ b/packages/toolkit/src/query/index.ts
@@ -31,6 +31,7 @@ export type {
   DefinitionsFromApi,
   OverrideResultType,
   TagTypesFromApi,
+  UpdateDefinitions,
 } from './endpointDefinitions'
 export { fetchBaseQuery } from './fetchBaseQuery'
 export type {


markerikson:
Yeah, we stopped using API Extractor for anything a while back and never happened to update `contributing.md`.

PR #4523, opened 2024-07-21
Improve doc clarity around TypeScript and `createAsyncThunk` usage

EvHaus:
I had a difficult time figuring out how to get proper TypeScript responses for dispatched thunk actions. I read the docs several times, and even though the answer was there (use type narrowing via match), I missed it each time.

I think I skipped it because the relevant information was nested under the "Manually defining thunkAPI types" category, which I wasn't doing in my case.

Since this type narrowing is very useful and relevant even when you're not manually typing `thunkAPI`, I'm making a proposal in this PR to move this information slightly higher in the doc and give it a dedicated section.

For more context, see my question [on Discord](https://discord.com/channels/102860784329052160/103538784460615680/1264479048524890224).

I hope you'll consider making this small change. Thank you!


markerikson:
Not sure about the "in the component" part. The preferred approach would be to use `await dispatch(myAsyncThunk()).unwrap()`, which would actually throw an error:

- https://redux-toolkit.js.org/api/createAsyncThunk#unwrapping-result-actions

EvHaus:
The content, including "in the component" is a copy/paste of the existing doc. Just moved.

EvHaus:
Pushed a new update that instead promotes the use of `.wrap()` rather than modifying the previous docs.

PR #4549, opened 2024-08-07
Add "createAsyncThunkCreator" with option for customizing default error serializer

### LINKED ISSUE #4548: Ability to configure createAsyncThunk for a custom serializeError
Instead of having to provide `serializeError` to the options of every `createAsyncThunk` call it would be nice to be able to provide your own `serializeError` default implementation for all thunks.

Suggestions:
* Add an `createAsyncThunkCreator`

```ts
const createAppAsyncThunk = createAsyncThunkCreator({
  serializeError(error) {
    return mySerializer(error)
  }
})
```

In my case I would like to add a new property `isAxiosError` to all the serialized errors.

jesperjohansson:
Adds a new `createAsyncThunkCreator` method which can be used to set the custom error serializer.

```ts
import { createAsyncThunkCreator, SerializedError } from '@reduxjs/toolkit'

export interface AppSerializedError extends SerializedError {
  isAxiosError?: boolean
}

type ThunkApiConfig = {
  state: RootState
  serializedErrorType: AppSerializedError
}

const createAppAsyncThunkCreator = createAsyncThunkCreator<ThunkApiConfig>({
  serializeError(error, defaultSerializer) {
    const serializedError = defaultSerializer(error)
    serializedError.isAxiosError = error.isAxiosError
    return serializedError
  },
})

function createAppAsyncThunk<
  Returned,
  ThunkArg = void,
  T extends ThunkApiConfig = ThunkApiConfig,
>(
  typePrefix: string,
  payloadCreator: AsyncThunkPayloadCreator<Returned, ThunkArg, T>,
  options?: AsyncThunkOptions<ThunkArg, T>,
): AsyncThunk<Returned, ThunkArg, T> {
  return createAppAsyncThunkCreator<Returned, ThunkArg, T>(
    typePrefix,
    payloadCreator,
    options,
  )
}
```

Closes #4548

### DIFF
--- a/packages/toolkit/src/createAsyncThunk.ts
+++ b/packages/toolkit/src/createAsyncThunk.ts
@@ -487,34 +487,61 @@ type CreateAsyncThunk<CurriedThunkApiConfig extends AsyncThunkConfig> = {
   >
 }
 
-type InternalCreateAsyncThunkCreatorOptions<
+/**
+ * @public
+ */
+export type CreateAsyncThunkCreatorOptions<
   ThunkApiConfig extends AsyncThunkConfig,
-> = {
-  serializeError?: ErrorSerializer<ThunkApiConfig>
-}
+> = Pick<
+  AsyncThunkOptions<unknown, ThunkApiConfig>,
+  'serializeError' | 'idGenerator'
+>
 
-function internalCreateAsyncThunkCreator<
+export function createAsyncThunkCreator<
   CreatorThunkApiConfig extends AsyncThunkConfig = {},
 >(
-  creatorOptions?: InternalCreateAsyncThunkCreatorOptions<CreatorThunkApiConfig>,
+  creatorOptions?: CreateAsyncThunkCreatorOptions<CreatorThunkApiConfig>,
 ): CreateAsyncThunk<CreatorThunkApiConfig> {
   function createAsyncThunk<
     Returned,
     ThunkArg,
-    ThunkApiConfig extends CreatorThunkApiConfig,
+    CallThunkApiConfig extends AsyncThunkConfig,
   >(
     typePrefix: string,
     payloadCreator: AsyncThunkPayloadCreator<
       Returned,
       ThunkArg,
-      ThunkApiConfig
+      OverrideThunkApiConfigs<CreatorThunkApiConfig, CallThunkApiConfig>
     >,
-    options?: AsyncThunkOptions<ThunkArg, ThunkApiConfig>,
-  ): AsyncThunk<Returned, ThunkArg, ThunkApiConfig> {
+    options?: AsyncThunkOptions<
+      ThunkArg,
+      OverrideThunkApiConfigs<CreatorThunkApiConfig, CallThunkApiConfig>
+    >,
+  ): AsyncThunk<
+    Returned,
+    ThunkArg,
+    OverrideThunkApiConfigs<CreatorThunkApiConfig, CallThunkApiConfig>
+  > {
+    type ThunkApiConfig = OverrideThunkApiConfigs<
+      CreatorThunkApiConfig,
+      CallThunkApiConfig
+    >
     type RejectedValue = GetRejectValue<ThunkApiConfig>
     type PendingMeta = GetPendingMeta<ThunkApiConfig>
     type FulfilledMeta = GetFulfilledMeta<ThunkApiConfig>
     type RejectedMeta = GetRejectedMeta<ThunkApiConfig>
+
+    const {
+      serializeError = miniSerializeError,
+      // nanoid needs to be wrapped because it accepts a size argument
+      idGenerator = () => nanoid(),
+      getPendingMeta,
+      condition,
+      dispatchConditionRejection,
+    } = {
+      ...creatorOptions,
+      ...options,
+    }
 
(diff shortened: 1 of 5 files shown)


EskiMojo14:
the type changes to miniSerializeError are not safe - the function will only ever return a SerializedError shape, regardless of what your config says.

packages/toolkit/src/createAsyncThunk.ts
```diff
@@ -542,6 +552,18 @@ export const createAsyncThunk = /* @__PURE__ */ (() => {
         }),
       )
 
+    function getError(x: unknown): GetSerializedErrorType<ThunkApiConfig> {
+      if (options && options.serializeError) {
+        return options.serializeError(x)
+      }
+
+      if (creatorOptions && creatorOptions.serializeError) {
+        return creatorOptions.serializeError(x, miniSerializeError)
```
EskiMojo14, line 561: because we export miniSerializeError, i don't think we need to pass it here - the user can just import it if they want it

EskiMojo14: i'd also prefer something like ```ts
const serializeError = options?.serializeError || creatorOptions?.serializeError || miniSerializeError
``` rather than this long function

File: packages/toolkit/src/createAsyncThunk.ts [hunk 2]
```diff
@@ -93,7 +93,7 @@ class FulfillWithMeta<Payload, FulfilledMeta> {
  */
 export const miniSerializeError = (value: any): SerializedError => {
   if (typeof value === 'object' && value !== null) {
-    const simpleError: SerializedError = {}
+    const simpleError = {} as Record<string, string>
```
EskiMojo14, line 96: why change this?

packages/toolkit/src/createAsyncThunk.ts
```diff
@@ -702,9 +725,14 @@ export const createAsyncThunk = /* @__PURE__ */ (() => {
...
   }
+
   createAsyncThunk.withTypes = () => createAsyncThunk
 
   return createAsyncThunk as CreateAsyncThunk<AsyncThunkConfig>
+}
+
+export const createAsyncThunk = /* @__PURE__ */ (() => {
+  return internalCreateAsyncThunkCreator() as CreateAsyncThunk<AsyncThunkConfig>
 })()
```
EskiMojo14, line 736:
```suggestion
export const createAsyncThunk = /* @__PURE__ */ internalCreateAsyncThunkCreator()
```

packages/toolkit/src/createAsyncThunk.ts
```diff
@@ -744,3 +772,31 @@ function isThenable(value: any): value is PromiseLike<any> {
...
+) => GetSerializedErrorType<ThunkApiConfig>
+
+/**
+ * @public
+ */
+type CreateAsyncThunkCreatorOptions<ThunkApiConfig extends AsyncThunkConfig> = {
+  serializeError?: ErrorSerializer<ThunkApiConfig>
+}
+
+export const createAsyncThunkCreator = /* @__PURE__ */ (() => {
```
EskiMojo14, line 796: why have this instead of just exporting (and renaming) internalCreateAsyncThunkCreator?

phryneas:
I have to admit I'm a bit hesitant to add additional runtime code for this, especially with a name like `createAsyncThunkCreator`, which would probably be picked up by autocomplete a lot.
Also, the whole thing needs a lot of docs and explanation and might confuse people.

Alternative suggestion - we could export something like a `CreateAsyncThunkWithoutWithTypes` type, which would only be the function signature of the `CreateAsyncThunk` type without the `withTypes` callback.

That would make it possible to define something like this in userland:

```ts
export const createAppAsyncThunk: CreateAsyncThunkWithoutWithTypes<{
  serializedErrorType: number;
}> = (typePrefix: string, payloadCreator: any, options: any) => {
  return createAsyncThunk(typePrefix, payloadCreator, {
    ...options,
    serializeError: mySerializeError,
  }) as any;
};
```

What do you think?

jesperjohansson:
> I have to admit I'm a bit hesitant to add additional runtime code for this, especially with a name like `createAsyncThunkCreator`, which would probably be picked up by autocomplete a lot. Also, the whole thing needs a lot of docs and explanation and might confuse people.
> 
> Alternative suggestion - we could export something like a `CreateAsyncThunkWithoutWithTypes` type, which would only be the function signature of the `CreateAsyncThunk` type without the `withTypes` callback.
> 
> That would make it possible to define something like this in userland:
> 
> ```ts
> export const createAppAsyncThunk: CreateAsyncThunkWithoutWithTypes<{
>   serializedErrorType: number;
> }> = (typePrefix: string, payloadCreator: any, options: any) => {
>   return createAsyncThunk(typePrefix, payloadCreator, {
>     ...options,
>     serializeError: mySerializeError,
>   }) as any;
> };
> ```
> 
> What do you think?

Yeah, this looks like a good solution to me.

EskiMojo14:
@jesperjohansson raised #4667 with the solution posed - that should have a build you'll be able to install and try out in the meantime of us getting this in.

happy for this PR to be closed in favour of the other?

markerikson:
Out in https://github.com/reduxjs/redux-toolkit/releases/tag/v2.4.0 !

PR #4555, opened 2024-08-10
[DOCS] invalidationBehavior documentation

thoughtworks-tcaceres:
- update incorrect default option of invalidationBehavior default property in documentation
- correct spelling of property reference of apiSlice


EskiMojo14:
thanks!

although as a Brit i will maintain that you're all spelling behaviour wrong 😄

thoughtworks-tcaceres:
> thanks!
> 
> although as a Brit i will maintain that you're all spelling behaviour wrong 😄

As a Canadian, I agree.  😅

Issue #4106, opened 2024-01-21
rtk-query normalization and manual insert into cache


sebastian-dor:
API has an endpoint that provides a list of items that we need. E.g. we have a map application that shows a lot of structures for a specific scenario id. We then want to use deletion, updating, selecting the items by their respective ID. The way RTK Query builds the cache (key) is by the function name in combination with the (scenario) ID. If one of the items would change, it would invalidate the whole list.

## possible solutions:

1. A normalized cache, but https://redux-toolkit.js.org/rtk-query/usage/cache-behavior#no-normalized-or-de-duplicated-cache
2. onQueryStarted and using dispatch(api.util.upsertQueryData(...), but dispatch has quite the overhead and using it for more than a handful items slows down UI considerably
3. using the concept of a top-level-reducer that wraps the api reducer

## the makeshift solution for 3.
Creating a so-called top-level reducer in the root reducer that wraps the actual call to the list endpoint and creates fake actions to insert the list items one by one into the cache.

```typescript
export const rootReducer = {
[...]
  // [structuredataApi.reducerPath]: structuredataApi.reducer,
  [structuredataApi.reducerPath]: structuretlr,
 [...]
};

function structuretlr(state: any, action: any) {
  let newState: any = undefined;
  let intermediateState = structuredataApi.reducer(state, action);
  switch (action.type) {
    //query is done
    case 'structuredataApi/executeQuery/fulfilled': {
      //list endpoint (get all structures by scenario)
      if ('getStructuredataByScenarioId' == action.meta.arg.endpointName) {
        //const startTime = performance.now();

        let previousState: any = undefined;
        action.payload.forEach((element: { id: any }) => {
          const entryToAdd = [element];
          const idToAdd = element.id;
          //this is necessary as otherwise the fulfilled action won't insert
          let fakePendingAction = {
            type: 'structuredataApi/executeQuery/pending',
            payload: undefined,
            meta: {
              ...action.meta,
              requestStatus: 'pending',
              //not sure if this is necessary
              requestId: action.meta.requestId + '_1',
              arg: {
                ...action.meta.arg,
                requestStatus: 'pending',
                endpointName: `getStructuredataById`,
                queryCacheKey: `getStructuredataById(${idToAdd})`,
                originalArgs: idToAdd
              }
            }
          };
          //verbosity for debugging
          let stateToSet = previousState ?? intermediateState;
          previousState = structuredataApi.reducer(stateToSet, fakePendingAction);

[...]

Issue #4073, opened 2024-01-12
Automatic normalization addon to rtk-query


klis87:
Hi, as discussed in https://github.com/reduxjs/redux-toolkit/issues/3248 , I integrated rtk-query with normy, a library, which allows automatic normalization and data updates for fetching libraries. There is already react-query and swr integration, now also rtk-query is supported. You can find it here - https://github.com/klis87/normy/tree/master/packages/normy-rtk-query#normyrtk-query

This is to solve problem mentioned here - https://redux-toolkit.js.org/rtk-query/comparison#no-normalized-or-deduplicated-cache and gives you apollo graphql like experience - relevant queries are automatically updated based on mutations results with no additional code.

I raised this issue not only as information, but also to mention several potential improvements, which could require some rtk-query changes
1) the integration is implemented as middleware, which also listens to rtk-query actions. is it possible to get those action types somehow? for now I wrote them by hand - https://github.com/klis87/normy/blob/master/packages/normy-rtk-query/src/index.ts#L22 , but ofc it would be much better to reuse rtk-query types
2) I could not manage to find a way to pass some meta attribute from query definition - like this - https://github.com/klis87/normy/blob/master/examples/rtk-query/src/api.js#L9 to the action here https://github.com/klis87/normy/blob/master/packages/normy-rtk-query/src/index.ts#L135 . This is problematic because without it I cannot allow an easy way to disable normalizations per specific queries and mutations. I added only global way as `normalizeQuery` and `normalizeMutation` options, but it is global, collocation style inside specific queries and mutations would be much better, like implemented in react-query addon - https://github.com/klis87/normy/tree/master/packages/normy-react-query#disabling-of-normalization-per-query-and-mutation-arrow_up
3) This library uses structural sharing, so if it works like I think, after each refetch it is checked that `data` is different. it would be cool to have this information in actions, so that middleware could give up normalization for a refetch which did not update data, which is a waste as it will end up with normalization store exactly the same as it was. This would be very nice optimization, especially that rtk-query has features like refetch on refocus, which potentially could cause multiple refetches at the same time.

markerikson:
Finally have a chance to look at this topic and think about it a bit.

I don't think we expose the `executeQuery/executeMutation` action types anywhere. Internally, those are part of the `queryThunk` and `mutationThunk` variables, and it doesn't look like those are included.  I suppose in theory we could add those to `api.internalActions`, but also not sure that's something we _ought_ to do.

[...]

PR #4561, opened 2024-08-14
Implement a util function to batch-upsert cache entries

markerikson:
This PR:

- Extracts the current reducer logic for writing `pending` and `fulfilled` cache entries into reusable helper functions
- Implements a new `util.upsertEntries` action creator + reducer that accepts an array of `{endpointName, arg, value}` entries, and upserts those into the cache, overwriting any prior entries

### Use Cases

One use case might be just preloading some data into the cache programmatically.

The other main use case I'm thinking of is a pseudo-normalization approach, ala #4106 and #4073.  

Say you have a `getPosts` endpoint and it returns the usual array.  RTKQ won't deduplicate individual objects across multiple endpoints, so if you then try to access `getPost("2")`, we have to make that request separately.  

It seems very reasonable to want to take the response contents from the `getPosts` query, and prepopulate all of the corresponding individual `getPost` cache entries.

We have `upsertQueryData`, which relies on the entire async thunk lifecycle, and like the rest of our utils it only does a single entry at a time.  That's always felt too limiting, both in terms of possibly wanting to update many cache entries at once, and performance due to the number of actions that would have to get dispatched.

### Status

This is the first seemingly-working implementation.  Thanks to @EskiMojo14 the external types seem to work (ie, `{endpointName: 'getPost'}` correctly expects to go along with `{arg: string, value: Post}`).  I threw in one POC unit test 

### Perf

Per #4106, doing this by calling the `apiSlice.reducer()` with a separate faked action for each item was still relatively slow - 1500ms for 1000 items (as compared to the abominably slow 6000ms for 1000 items using individual `upsertQueryData` calls).

In a quick local test, a single `dispatch(api.util.upsertEntries())` call with 1000 separate `getPost` entries was only 31ms.  That's because we only called the reducer once, so we did all that work inside a single Immer call.  That should reduce the perf overhead significantly.

### Plans

The starting point would be to just ship this util action creator as-is, and users would manually dispatch that in `onQueryStarted` after the response resolves.

I can also imagine a new `upsertCacheEntries` option to make this more of a first-class citizen, which would look roughly like `(data: T) => NormalizedQueryUpsertEntry[]`.

I'm sure there's edge cases here. For example, `merge` expects `baseQueryMeta`, and since we aren't even using a real request here, we don't have that. Right now I'm faking it with an empty object. I could imagine a `merge` function actually expecting a real value instead.

### DIFF
--- a/packages/toolkit/src/query/core/buildSlice.ts
+++ b/packages/toolkit/src/query/core/buildSlice.ts
@@ -112,6 +112,100 @@ export function buildSlice({
   >
 }) {
   const resetApiState = createAction(`${reducerPath}/resetApiState`)
+
+  function writePendingCacheEntry(
+    draft: QueryState<any>,
+    arg: QueryThunkArg,
+    upserting: boolean,
+    meta: {
+      arg: QueryThunkArg
+      requestId: string
+      // requestStatus: 'pending'
+    } & { startedTimeStamp: number },
+  ) {
+    draft[arg.queryCacheKey] ??= {
+      status: QueryStatus.uninitialized,
+      endpointName: arg.endpointName,
+    }
+
+    updateQuerySubstateIfExists(draft, arg.queryCacheKey, (substate) => {
+      substate.status = QueryStatus.pending
+
+      substate.requestId =
+        upserting && substate.requestId
+          ? // for `upsertQuery` **updates**, keep the current `requestId`
+            substate.requestId
+          : // for normal queries or `upsertQuery` **inserts** always update the `requestId`
+            meta.requestId
+      if (arg.originalArgs !== undefined) {
+        substate.originalArgs = arg.originalArgs
+      }
+      substate.startedTimeStamp = meta.startedTimeStamp
+    })
+  }
+
+  function writeFulfilledCacheEntry(
+    draft: QueryState<any>,
+    meta: {
+      arg: QueryThunkArg
+      requestId: string
+      // requestStatus: 'fulfilled'
+    } & {
+      fulfilledTimeStamp: number
+      baseQueryMeta: unknown
+      // RTK_autoBatch: true
+    },
+    payload: unknown,
+  ) {
+    updateQuerySubstateIfExists(draft, meta.arg.queryCacheKey, (substate) => {
+      if (substate.requestId !== meta.requestId && !isUpsertQuery(meta.arg))
+        return
+      const { merge } = definitions[meta.arg.endpointName] as QueryDefinition<
+        any,
+        any,
+        any,
+        any
+      >
+      substate.status = QueryStatus.fulfilled
+
+      if (merge) {
+        if (substate.data !== undefined) {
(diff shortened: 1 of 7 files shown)


packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -42,6 +53,44 @@ import {
...
+>(
+  entries: [
+    ...{
+      [I in keyof EndpointNames]: NormalizedQueryUpsertEntry<
+        Definitions,
+        EndpointNames[I]
+      >
+    },
+  ],
+) => PayloadAction<NormalizedQueryUpsertEntryPayload>
```
EskiMojo14, line 93:
```suggestion
) => PayloadAction<NormalizedQueryUpsertEntryPayload[]>
```

phryneas:
My main critique point here is conceptual - if you do normalization, oftentimes you have a list with 50 items, and a request might only update a few of those (and also not return the unchanged ones).

So to update three items in a list of 50, userland code would have to retrieve the current cache entry with 50 items, manually update those three items and write them back. Essentially, moving business logic out of Redux.

That's why we have the callback notation in `updateQueryData` (which moves the business logic kinda into Redux), and in my eyes `upsertQueryData` has only very limited use case.

Could I suggest we modify this a bit so something like this would become possible?

```js
upsertEntries([
  {
    endpointName: "item",
    arg: 1,
    value: item1,
    // this is the default
    merge: (_, value) => value,
  },
  {
    endpointName: "item",
    arg: 20,
    value: item20,
    // if someone wanted to opt out of overriding existing values, they could do something like this
    merge: (existing) => existing,
  },
  {
    endpointName: "items",
    arg: undefined,
    value: { 1: item1, 20: item20 },
    // and here someone could have custom merging logic
    merge(existing, upserted) {
      return { ...existing, ...upserted }
    },
  },
]);
```

The example above shows one weakness regarding usage with entityAdapter, though. The following would not be possible:
```ts
{
    endpointName: "items",
    arg: undefined,
    value: [ item1, item20 ],
    merge(existing, upserted) {
      return itemsAdapter.setAll(existing, upserted);
    },
  },
```

The problem here being the case of "nothing exists yet, so we don't call merge". In that case, `[ item1, item20 ]` would be upserted without calling `merge`, which would result in the wrong cache entry shape for future cache updates.
But if a user worked around that by doing `value: itemsAdapter.setAll(itemsAdapter.initialState, upserted)`, the `merge` function itself wouldn't work out, since `setAll` doesn't accept a full entity-adapter-state as second argument.

I'm open for ideas here :)

markerikson:
Lenz and I discussed this internally, and we noted that we already have a `merge` endpoint option. With the way this new util is constructed, all the normal `merge` logic applies. That said, we _don't_ get `transformResponse`, because A) that runs inside query thunks, and B) it can be async.  So, you'd need to pass in the values as `transformResponse` would return them.

Issue #3822, opened 2023-10-25
Add an option to encode query arguments in openApi codegen


ifozest:
Hi all,

Currently, there is no way to specify that query arguments should be encoded when generating OpenAPI code using rtk query codegen. This can be a problem if you are using an argument with special characters, such as #.

For example, the following query function would result in an OpenAPI endpoint:
```
    doFoo: build.mutation<void, { foo: string, ... }>({
      query: (queryArg) => ({
        url: `/api/${queryArg.foo}`,
        method: "POST",
        body: queryArg.requestObject,
      }),
    }),
```

To fix this, I would like to request an option to enable query argument encoding in OpenAPI codegen. This would generate the following query function and generate an encoded endpoint:  **`/api/${encodeURIComponent(queryArg.foo)}`** or similar

I believe that this would be a valuable addition to rtk query openAPI codegen, as it would make it easier to use query args with special characters.

Thank you for your consideration.

Meantime I'm using `injectEndpoints` and overwriting them which is not ideal.

rjgotten:
Technically this is a correctness problem, isn't it?
Anything going into an URI path _should_ be encoded as an URI component, period.

shivam178:
I faced the same issue. Used `encodeURIComponent` in the same way to overcome it.
I'm facing this issue only in Safari and Firefox and it's working fine in Chromium browsers as they seem auto encode URLs before requesting them.

Richard87:
I was just bit by this issue in Chrome and absolutley agree this should be handled in the API Client library... Is there any fix available? 

Or would RTK team accept a PR to add this?

bayasdev:
I'm facing the same issue, would a fix be available soon?

PR #4568, opened 2024-08-17
Add "encodeParams" option to rtkq-codegen-openapi

bayasdev:
## Overview

This PR introduces a new `encodeParams` option to the Redux Toolkit Query (RTKQ) OpenAPI code generator. When enabled, this option will automatically apply `encodeURIComponent` to the generated query parameters, ensuring that special characters in query strings are correctly encoded.

This feature will be useful for users who need to ensure their query parameters are URL safe, particularly in cases where special characters are commonly used.

Fixes https://github.com/reduxjs/redux-toolkit/issues/3822

## This PR:

- [x] Adds `encodeParams` option to automatically perform URL safe encoding to the generated query and path parameters
- [X] Adds test cases for `encodeParams` option
- [X] Fixes `flattenArg` test case which was not passing due to formatting

### DIFF
--- a/packages/rtk-query-codegen-openapi/src/generate.ts
+++ b/packages/rtk-query-codegen-openapi/src/generate.ts
@@ -396,7 +396,9 @@ export async function generateApi(
         return createPropertyAssignment(
           param.originalName,
           encodeParams && param.param?.in === 'query'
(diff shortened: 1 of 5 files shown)


markerikson:
Seems reasonable, thanks!

PR #4585, opened 2024-08-28
Fix `noImplicitOverrides`-related issues

aryaemami59:
## **This PR**:

- [X] Fixes `noImplicitOverrides`-related issues (part of #4138).


EskiMojo14:
hmm, what's the benefit of enabling this flag?

aryaemami59:
It forces you to be slightly more explicit when overloading functions that are inherited from the base class.

- [Docs](https://www.typescriptlang.org/tsconfig/#noImplicitOverride)

EskiMojo14:
honestly i don't see the point, but ultimately it's five words that don't affect the final bundle size at all, so doesn't really hurt.

PR #4591, opened 2024-08-31
Add "includeDefault" option to rtkq-codegen-openapi

bayasdev:
## Overview

This PR adds a new `includeDefault` option to the Redux Toolkit Query (RTKQ) OpenAPI code generator. When enabled, this option includes the default response type in the generated code for all endpoints, as defined in the OpenAPI schema.

The `includeDefault` option is disabled by default because the default response type typically represents an error response (see [OpenAPI documentation on default responses](https://swagger.io/docs/specification/describing-responses/#default)). Enabling it by default could break existing code if types are not narrowed correctly.

### DIFF
--- a/packages/rtk-query-codegen-openapi/src/generate.ts
+++ b/packages/rtk-query-codegen-openapi/src/generate.ts
@@ -30,8 +30,8 @@ import { factory } from './utils/factory';
 const generatedApiName = 'injectedRtkApi';
 const v3DocCache: Record<string, OpenAPIV3.Document> = {};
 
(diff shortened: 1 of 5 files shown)


markerikson:
Okay, published as https://github.com/reduxjs/redux-toolkit/releases/tag/%40rtk-query%2Fcodegen-openapi%402.0.0-alpha.1

bayasdev:
> Okay, published as https://github.com/reduxjs/redux-toolkit/releases/tag/%40rtk-query%2Fcodegen-openapi%402.0.0-alpha.1

Thanks!

PR #4594, opened 2024-09-02
Update immer-reducers.md

dominikbulaj:
Used proper `immer` import (see immer docs @ https://immerjs.github.io/immer/#with-immer)


EskiMojo14:
thanks! the default import was removed in immer's last major release, looks like this didn't get updated to match :)

PR #4409, opened 2024-05-14
Migrate to React 19 

aryaemami59:
## **This PR:**

- [X] Updates `react` and `react-dom` to v19.
- [X] Sets the `@types/react` and `@types/react-dom` to v19 according to the [React 19 migration guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#installing).
- [X] Runs tests against different versions of React during CI. Not sure if we want to keep this addition but for now it serves as a way for us to make sure the changes we make are non-breaking and can work with both React 18 and 19.
- [X] Updates `@testing-library/react` to latest version.
- [X] Updates `@testing-library/user-event` to latest version.
- [X] Adds latest version of `@testing-library/dom`.
- [X] Replaces the now removed `ReactDOM.render` calls with the `render` function from `@testing-library/react` as recommended in the [React 19 migration guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-reactdom-render).
- [X] Replaces the now removed `ReactDOM.unmountComponentAtNode` calls with the `unmount` function from `@testing-library/react` as recommended in the [React 19 migration guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-unmountcomponentatnode).
- [X] Removes redundant `act` wrappers around `fireEvent` method calls since they are already wrapped in `act` calls.
- [X] Expands version ranges of `react` in `peerDependencies` to include version 19.
- [X] Updates React-related code snippets in docs.
- [X] Fixes some issues related to `console` spies.
- [X] Fixes some issues related to stubbing environments.
- [X] Fixes some issues related to `act` warnings emitted by `@testing-library/react`.
- [X] Fixes some issues related to `sinon` fake timers.

PR #4603, opened 2024-09-04
Remove `console-testing-library`

aryaemami59:
## **This PR**:

- [X] Removes `console-testing-library` (part of #4138).
- [X] Removes `jest-snapshot` from `resolutions` field of root `package.json`.

### DIFF
--- a/packages/toolkit/src/query/tests/queryFn.test.tsx
+++ b/packages/toolkit/src/query/tests/queryFn.test.tsx
@@ -244,10 +244,10 @@ describe('queryFn base implementation tests', () => {
 
     const thunk = endpoint.initiate(endpointName)
 
-    const result:
+    const result = (await store.dispatch(thunk)) satisfies
       | undefined
       | { data: string }
-      | { error: string | SerializedError } = await store.dispatch(thunk)
+      | { error: string | SerializedError }
 
     if (endpointName.includes('Throw')) {
       expect(consoleErrorSpy).toHaveBeenCalledOnce()
@@ -291,7 +291,7 @@ describe('queryFn base implementation tests', () => {
     {
       const thunk = withNeither.initiate('withNeither')
 
-      const result: QuerySubState<any> = await store.dispatch(thunk)
+      const result = (await store.dispatch(thunk)) satisfies QuerySubState<any>
 
       expect(consoleErrorSpy).toHaveBeenCalledOnce()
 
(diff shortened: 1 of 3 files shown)


EskiMojo14:
hmm, what's the benefit of switching away from console-testing-library?

phryneas:
It seemed to be a good idea back then but I think at this point it's more tech debt. I do have a request here though, will add something.

packages/toolkit/src/listenerMiddleware/tests/effectScenarios.test.ts
```diff
@@ -57,10 +54,7 @@ describe('Saga-style Effects Scenarios', () => {
     return new Promise((resolve) => setTimeout(resolve, ms))
   }
 
-  beforeAll(() => {
-    const noop = () => {}
-    vi.spyOn(console, 'error').mockImplementation(noop)
-  })
+  const consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(noop)
```
phryneas, line 57: I just wanted to suggest that we do 
```suggestion
  using consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(noop)
```
inside the tests that actually need this, and skip the `mockClear` and `mockRestore` calls, but I'm not sure if `vitest` actually returns a `Disposable` here. I know `jest` does (because I added the functionality there) - could you check if `vitest` has that now, too?

aryaemami59: Sure I can try it. The type tests will probably fail, so I'll probably exclude the runtime tests from the type tests.

aryaemami59:
@EskiMojo14 There is multiple reasons:

1. like @phryneas mentioned, it's tech debt.
2. The inline snapshots caused some issues with Prettier and ESLint in #4138.
3. It caused some issues in #4409 with React 19.
4. It's also somewhat of an old library that hasn't been maintained in years which is why we had to patch it several times.
5. We also had to override the version of `jest-snapshot` in the root `package.json` because of it which in turn causes the tests for all the example apps to fail locally.
6. We can also replace what it does by spying on the `console` object itself easily.

packages/toolkit/src/listenerMiddleware/tests/effectScenarios.test.ts
aryaemami59: Looks like `vitest` doesn't return  a `Disposable`, though I might try to submit a PR to add that functionality for `vitest`.

phryneas: Will definitely be a win in the long run, even if it's not applicable to this PR in the end :)

PR #4628, opened 2024-09-19
Fix `AbortError` being triggered incorrectly on `createApi` endpoint timeout

### LINKED ISSUE #2915: Timeout not throwing error with status TIMEOUT_ERROR
Hey, PR #2401 shows that `fetchBaseQuery` should throw a `TIMEOUT_ERROR` when time out delay is reached. However, I got an error without status `{"message": "Aborted", "name": "AbortError"}`. I'm using 
`"@reduxjs/toolkit": "^1.9.0"` with` "react-native": "0.70.3"`,  `"react": "18.1.0"`. Here is my base API:

```
export const BaseApi = createApi({
  reducerPath: 'baseApi',
  baseQuery: fetchBaseQuery({
    baseUrl: Config.BASE_URL,
    prepareHeaders: headers => {
      const token = Preferences.getToken();
      headers.set('authorization', `Bearer ${token}`);

      return headers;
    },
    timeout: 1,
  }),
  tagTypes: ['Auth', 'Notification', 'Restaurant'],
  endpoints: () => ({}),
});
```

andrejpavlovic:
## **This PR**:
* [x]  Fixes bug where a timeout triggered on `createApi` endpoint would return a generic `AbortError` instead of  `TIMEOUT_ERROR`.
* [x] Adds a testcase for that validates this behavior.
* [x]  Resolves #2915

### DIFF
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -213,7 +213,7 @@ export function fetchBaseQuery({
     )
   }
   return async (arg, api) => {
-    const { signal, getState, extra, endpoint, forced, type } = api
+    const { getState, extra, endpoint, forced, type } = api
     let meta: FetchBaseQueryMeta | undefined
     let {
       url,
@@ -224,6 +224,14 @@ export function fetchBaseQuery({
       timeout = defaultTimeout,
       ...rest
     } = typeof arg == 'string' ? { url: arg } : arg
+
+    let abortController: AbortController | undefined, signal = api.signal
+    if (timeout) {
+      abortController = new AbortController()
+      api.signal.addEventListener('abort', abortController.abort)
+      signal = abortController.signal
+    }
+
     let config: RequestInit = {
       ...baseFetchOptions,
       signal,
@@ -272,10 +280,10 @@ export function fetchBaseQuery({
     let response,
       timedOut = false,
       timeoutId =
-        timeout &&
+        abortController &&
         setTimeout(() => {
           timedOut = true
-          api.abort()
+          abortController!.abort()
         }, timeout)
     try {
       response = await fetchFn(request)
@@ -289,6 +297,7 @@ export function fetchBaseQuery({
       }
     } finally {
       if (timeoutId) clearTimeout(timeoutId)
+      abortController?.signal.removeEventListener('abort', abortController.abort)
     }
     const responseClone = response.clone()
 
--- a/packages/toolkit/src/query/tests/createApi.test.ts
+++ b/packages/toolkit/src/query/tests/createApi.test.ts
@@ -1128,3 +1128,38 @@ describe('custom serializeQueryArgs per endpoint', () => {
     })
   })
 })
+
+describe('timeout behavior', () => {
+  test('triggers TIMEOUT_ERROR', async () => {
+    const api = createApi({
+      baseQuery: fetchBaseQuery({ baseUrl: 'https://example.com', timeout: 5 }),
+      endpoints: (build) => ({
+        query: build.query<unknown, void>({
+          query: () => '/success',
+        }),
+      }),
+    })
+  
+    const storeRef = setupApiStore(api, undefined, {
+      withoutTestLifecycles: true,
+    })
+    
+    server.use(
+      http.get(
+        'https://example.com/success',
+        async () => {
+          await delay(10)
+          return HttpResponse.json({ value: 'failed' }, { status: 500 })
+        },
+        { once: true },
+      ),
+    )
+
+    const result = await storeRef.store.dispatch(api.endpoints.query.initiate())
+
+    expect(result?.error).toEqual({
+      status: 'TIMEOUT_ERROR',
+      error: expect.stringMatching(/^AbortError:/),
+    })
+  })
+})


phryneas:
Could you please explain the reasoning here a bit? It's not immediately obvious to me.

andrejpavlovic:
> Could you please explain the reasoning here a bit? It's not immediately obvious to me.

Sure thing! As I alluded to in #2915, the issue exists because of a bit of a race condition happening.
1. When `fetchBaseQuery` times out, it triggers `api.abort()` on global abort controller (ie. on the one that is created inside `createAsyncTrunk`)
https://github.com/reduxjs/redux-toolkit/blob/bc0bf0ffced8d1e9aa927faf81e37f42e681058c/packages/toolkit/src/query/fetchBaseQuery.ts#L276-L279
2. That in turns rejects the `abortPromise` in `createAsyncTrunk` with the generic `AbortError`.
https://github.com/reduxjs/redux-toolkit/blob/bc0bf0ffced8d1e9aa927faf81e37f42e681058c/packages/toolkit/src/createAsyncThunk.ts#L604-L610
3. Which then rejects the `Promise.race` promise here:
https://github.com/reduxjs/redux-toolkit/blob/bc0bf0ffced8d1e9aa927faf81e37f42e681058c/packages/toolkit/src/createAsyncThunk.ts#L623-L626
4. And this rejection (the generic `AbortError`) gets returned by the `createAsyncTrunk` here:
https://github.com/reduxjs/redux-toolkit/blob/bc0bf0ffced8d1e9aa927faf81e37f42e681058c/packages/toolkit/src/createAsyncThunk.ts#L654-L657

So even though `fetchBaseQuery` would resolve to `TIMEOUT_ERROR` (see below), it wouldn't matter, because the global `abortPromise` would always resolve first.
https://github.com/reduxjs/redux-toolkit/blob/bc0bf0ffced8d1e9aa927faf81e37f42e681058c/packages/toolkit/src/query/fetchBaseQuery.ts#L282-L286

Since `fetchBaseQuery` is quite capable of resolving the timeout locally, there is no need to trigger the global abort inside of it. 

The solution I applied was to create an `AbortController` inside of `fetchBaseQuery` that would handle the timeout locally, but would also abort if `abort` is triggered on the global `signal`:
https://github.com/reduxjs/redux-toolkit/blob/d9a9106c204abd60ffa4969a638d3afa33050f25/packages/toolkit/src/query/fetchBaseQuery.ts#L228-L233

To sum it up, in this PR, we avoid the race condition in `createAsyncThunk` between 1. triggering global abort and 2. handling the timeout in `fetchBaseQuery`, by never triggering the global abort on timeout in the first place.

**Of note**: As far as I understand best practices for using `AbortController`, `abort` callback should not be passed to `fetchBaseQuery` in the first place, just the `signal`. I would go as far as suggest to remove passing this parameter:
https://github.com/reduxjs/redux-toolkit/blob/bc0bf0ffced8d1e9aa927faf81e37f42e681058c/packages/toolkit/src/createAsyncThunk.ts#L632

phryneas:
I see, thank you for the explanation.

It's a shame that `AbortSignal.any` is so new that we probably can't use it, or otherwise we could have done something much shorter like 

```diff
diff --git a/packages/toolkit/src/query/fetchBaseQuery.ts b/packages/toolkit/src/query/fetchBaseQuery.ts
index 9dc14d97..f2a3b732 100644
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -225,12 +225,8 @@ export function fetchBaseQuery({
       ...rest
     } = typeof arg == 'string' ? { url: arg } : arg

-    let abortController: AbortController | undefined, signal = api.signal
-    if (timeout) {
-      abortController = new AbortController()
-      api.signal.addEventListener('abort', abortController.abort)
-      signal = abortController.signal
-    }
+    const timeoutSignal = timeout ? AbortSignal.timeout(timeout) : undefined
+    const signal = timeoutSignal ? AbortSignal.any(api.signal, timeoutSignal) : api.signal

     let config: RequestInit = {
       ...baseFetchOptions,
@@ -277,27 +273,17 @@ export function fetchBaseQuery({
     const requestClone = new Request(url, config)
     meta = { request: requestClone }

-    let response,
-      timedOut = false,
-      timeoutId =
-        abortController &&
-        setTimeout(() => {
-          timedOut = true
-          abortController!.abort()
-        }, timeout)
+    let response
     try {
       response = await fetchFn(request)
     } catch (e) {
       return {
         error: {
-          status: timedOut ? 'TIMEOUT_ERROR' : 'FETCH_ERROR',
+          status: timeoutSignal?.aborted ? 'TIMEOUT_ERROR' : 'FETCH_ERROR',
           error: String(e),
         },
         meta,
       }
-    } finally {
-      if (timeoutId) clearTimeout(timeoutId)
-      abortController?.signal.removeEventListener('abort', abortController.abort)
     }
     const responseClone = response.clone()
```

phryneas (approved):
Looks good to me :)

andrejpavlovic:
> I see, thank you for the explanation.
> 
> It's a shame that `AbortSignal.any` is so new that we probably can't use it, or otherwise we could have done something much shorter like
> 
> ```diff
> diff --git a/packages/toolkit/src/query/fetchBaseQuery.ts b/packages/toolkit/src/query/fetchBaseQuery.ts
> index 9dc14d97..f2a3b732 100644
> --- a/packages/toolkit/src/query/fetchBaseQuery.ts
> +++ b/packages/toolkit/src/query/fetchBaseQuery.ts
> @@ -225,12 +225,8 @@ export function fetchBaseQuery({
>        ...rest
>      } = typeof arg == 'string' ? { url: arg } : arg
> 
> -    let abortController: AbortController | undefined, signal = api.signal
> -    if (timeout) {
> -      abortController = new AbortController()
> -      api.signal.addEventListener('abort', abortController.abort)
> -      signal = abortController.signal
> -    }
> +    const timeoutSignal = timeout ? AbortSignal.timeout(timeout) : undefined
> +    const signal = timeoutSignal ? AbortSignal.any(api.signal, timeoutSignal) : api.signal
> 
>      let config: RequestInit = {
>        ...baseFetchOptions,
> @@ -277,27 +273,17 @@ export function fetchBaseQuery({
>      const requestClone = new Request(url, config)
>      meta = { request: requestClone }
> 
> -    let response,
> -      timedOut = false,
> -      timeoutId =
> -        abortController &&
> -        setTimeout(() => {
> -          timedOut = true
> -          abortController!.abort()
> -        }, timeout)
> +    let response
>      try {
>        response = await fetchFn(request)
>      } catch (e) {
>        return {
>          error: {
> -          status: timedOut ? 'TIMEOUT_ERROR' : 'FETCH_ERROR',
> +          status: timeoutSignal?.aborted ? 'TIMEOUT_ERROR' : 'FETCH_ERROR',
>            error: String(e),
>          },
>          meta,
>        }
> -    } finally {
> -      if (timeoutId) clearTimeout(timeoutId)
> -      abortController?.signal.removeEventListener('abort', abortController.abort)
>      }
>      const responseClone = response.clone()
> ```

Ah, yes, that's a thing of beauty!

EskiMojo14:
thanks!

PR #4636, opened 2024-09-20
feat(baseQuery): expose queryCacheKey in baseQuery

### LINKED ISSUE #4615: Expose `queryCacheKey` in api `baseQuery`
## As discussed in [this Discord thread](https://discord.com/channels/102860784329052160/1281717350793478166)
In very broad strokes, a user-facing overview of the different layers of RTK Query looks a little bit like this:

Screenshot 2024-09-09 at 09 21 29

The availability of _"query-stable"_ identifiers is currently a little bit lacking across these layers. As discussed in the Discord thread, it would be helpful to expose `queryCacheKey` in the arguments of `baseQuery` in the api definition since this would allow us to create stable identifiers that reference a specific query that we can compare against in other layers as well.

### Context
Here is a process diagram showing what I want to achieve on a higher level:
Screenshot 2024-09-09 at 09 18 46

Code flow looks a little bit like this:
Screenshot 2024-09-09 at 09 24 09

In essence, I need to orchestrate the downloading of a set of server endpoints into local device storage so that users can work with my app in offline mode. By moving choice of querying against a remote server vs my local storage futher into the fetching / network layer of my codebase, I can use RTK Query agnostic to _how_ the data is fetched. This has a lot of benefits from an architectural point of view.

In my async thunk download orchestrator, I am dispatching endpoints and storing their resulting data in my local storage. I need a key / id for each of these endpoints that I can store the response data under. In the dispatching layer, the `queryCacheKey` is available to me, which together with the reducer name should allow me to uniquely set a key for each endpoint in storage. Since this key is not available in the fetching / networking layer however, it is currently impossible for me to query against my local storage since there is no way to retrieve this key based on the contextual arguments that `baseQuery` is served.

Ideally, I'd want to store the response data by the URL of the endpoint I downloaded the data for since this is expected to be more stable than reducer names and query names (+ arg names). This would require the `meta` of a query to be exposed in the return value of an awaited query dispatch, which I understand from the Discord thread would be hard to achieve.

HaakonSvane:
- Add `queryCacheKey` to `BaseQueryApi` typings
- Expose query cache key to the `baseQuery` api context
- Write tests to ensure correct typings and for ensuring matching keys between thunk dispatches and baseQuery 

closes #4615

### DIFF
--- a/docs/rtk-query/api/createApi.mdx
+++ b/docs/rtk-query/api/createApi.mdx
@@ -91,6 +91,7 @@ export const { useGetPokemonByNameQuery } = pokemonApi
   - `endpoint` - The name of the endpoint.
   - `type` - Type of request (`query` or `mutation`).
   - `forced` - Indicates if a query has been forced.
+  - `queryCacheKey`- The computed query cache key
 - `extraOptions` - The value of the optional `extraOptions` property provided for a given endpoint
 
 #### baseQuery function signature
--- a/docs/rtk-query/api/createApi.mdx
+++ b/docs/rtk-query/api/createApi.mdx
@@ -91,7 +91,7 @@ export const { useGetPokemonByNameQuery } = pokemonApi
   - `endpoint` - The name of the endpoint.
   - `type` - Type of request (`query` or `mutation`).
   - `forced` - Indicates if a query has been forced.
-  - `queryCacheKey`- The computed query cache key
+  - `queryCacheKey`- The computed query cache key.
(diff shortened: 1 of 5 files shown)


HaakonSvane:
Seems someone has exposed the `forced` field in the api previously. I'd ideally want to ensure this and the added queryCacheKey to be typed with a discriminated union on the endpoint `type`, but changing that would be a breaking change, so I added this field in a similar fashion

phryneas (approved):

EskiMojo14:
thanks!

PR #4638, opened 2024-09-23
Pass query args to prepareHeaders function

### LINKED ISSUE #4238: Passing query args to `prepareHeaders`?
I have a use case where some header values are dependent on the args passed to my query.  Could we pass `arg` into the `prepareHeaders` function [here](https://github.com/reduxjs/redux-toolkit/blob/ed9fc19027a31afee2234dea54fb36c8beef46ad/packages/toolkit/src/query/fetchBaseQuery.ts#L235)?

Happy to work on this if this feature is considered.

kyletsang:
Closes #4238

Passes args in the query function to the `prepareHeaders` call as the third argument.  

In the base fetch function, I renamed the first argument from `arg` to `args` to make it consistent with the name in the types and docs.

### DIFF
--- a/docs/rtk-query/api/fetchBaseQuery.mdx
+++ b/docs/rtk-query/api/fetchBaseQuery.mdx
@@ -64,8 +64,7 @@ type FetchBaseQueryArgs = {
     api: Pick<
       BaseQueryApi,
       'getState' | 'extra' | 'endpoint' | 'type' | 'forced'
-    >,
-    args: string | FetchArgs
+    > & { args: string | FetchArgs },
   ) => MaybePromise<Headers | void>
   fetchFn?: (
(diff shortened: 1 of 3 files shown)


markerikson:
I think I'd probably prefer adding `args` to the object second argument, rather than passing it as a third argument.

kyletsang:
> I think I'd probably prefer adding `args` to the object second argument, rather than passing it as a third argument.

Thanks for the feedback Mark.  I've modified the code to pass into the second args object.

markerikson:
Looks like we have uses of both `arg` and `args` in the codebase, but more `arg` usage - `args` is mostly the name given to the parameter in some thunks.

Given that, I'm going to switch this back to `arg`.

Issue #4554, opened 2024-08-09
Previously accessible types are no longer accessible from 2.2.6 -> 2.2.7


JP250552:
With the consolidated types change in 2.2.7, certain types are no longer available. 
While I understand these may be considered "internal" types, they are really helpful when utilizing a custom base query. 

I would ask for your consideration in exposing at minimum the following:

`@reduxjs/toolkit/dist/query/baseQueryTypes`
- `BaseQueryError`
- `BaseQueryMeta`

`@reduxjs/toolkit/dist/query/endpointDefinitions`
- `MutationExtraOptions`
- `QueryExtraOptions`
- `ResultDescription`

As a workaround, it looks like these are available in the `src` directory.

markerikson:
Can you give some examples of where / how you're trying to use these types?

JP250552:
Yes certainly. 

We've created a base query that handles API functions generated from our open api schema. We have some utilities created to make creating fully typed endpoints based on those functions. 

Here is an example of where we create a custom `EndpointBuilder` and `QueryExtraOptions` to handle queries. 

```ts
// Extended types

export type CustomEndpointBuilder<T = unknown> = EndpointBuilder<
  BaseQueryFn<QueryCallback<T>>,
  string,
  'api'
>;

export interface CustomQueryExtraOptions<T extends (...args: any) => any>
  extends Omit<
    QueryExtraOptions<
      any,
      BaseQueryReturnType<T>,
      QueryArgs<T>,
      BaseQueryFn<QueryCallback<T>>,
      'api'
    >,
    'type'
  > {
  providesTags?: ResultDescription<
    any,
    BaseQueryReturnType<T>,
    QueryArgs<T>,
    BaseQueryError<BaseQueryFn<QueryCallback<T>>>,
    BaseQueryMeta<BaseQueryFn<QueryCallback<T>>>
  >;
}
```

```ts
// Utility functions for working with custom API functions

export function createQuery(builder: CustomEndpointBuilder) {
  return <T extends (...args: any[]) => QueryCallback>(
    apiFn: T,
    options?: CustomQueryExtraOptions<T>,
  ) =>
    builder.query<BaseQueryReturnType<T>, QueryArgs<T>>({
      ...options,
      query: apiFn,
    });
}
```

AndreasNH:
I have a usecase where we use the QueryReturnValue type to create a reposehandler to enrich our logging middleware with metadata. 

```ts
/**
[...]

PR #4640, opened 2024-09-24
type: export QueryReturnValue

hornta:
This PR exports the type `QueryReturnValue` as described in https://github.com/reduxjs/redux-toolkit/issues/4554.


EskiMojo14:
thanks!

PR #4689, opened 2024-11-03
Add ability to reset lazy query hooks

alexmotoc:
Having another go at addressing #2055 since #2133 has been stale for quite a while now.
Implements the suggestion from https://github.com/reduxjs/redux-toolkit/pull/2133#issuecomment-1173164762

I have tried to follow a similar approach to the way the `reset` method has been implemented for the `useMutation` hook. I am new to this codebase so please let me know if something does not makes sense!


markerikson:
Out in https://github.com/reduxjs/redux-toolkit/releases/tag/v2.4.0 !

Issue #4693, opened 2024-11-05
React component using `useQuery` fails to render data when using `jest.useFakeTimers`


ensconced:
See [here](https://github.com/ensconced/rtk-fake-timers-bug/blob/master/test.spec.js) for details on how to reproduce this issue.

You should be able to reproduce the issue just by cloning that repo, running `npm install` and then `npm run test`. 
There are two tests, each pretty much the same except that one uses fake timers, which seems to cause it to fail.

From what I can tell, the store is always updated with the data from the (stubbed) API response, but when using fake timers this does not result in the react component actually being rendered with the data.

ensconced:
It seems that passing `{ doNotFake: ['clearTimeout'] }` into `jest.useFakeTimers` fixes this. (This isn't viable as a workaround for me - just mentioning in case it helps the investigation).

phryneas:
I mean, you're configuring your runtime to not run per the specification. That *will* break every tool using `setTimeout` and `clearTimeout` - and RTK Query uses those.

So honestly I don't know what we could do to fix that - RTKQ needs them to work :/

ensconced:
Sorry if this wasn't clear, but the reason I expect this to work is that the `findBy` functions provided by react testing library internally advance fake timers if they detect that jest is using fake timers. 

I've done some more digging and I'm pretty sure this is the problem: https://github.com/reduxjs/redux-toolkit/blob/24286f1745e42217448de1f3486886fa8558c1b5/packages/toolkit/src/autoBatchEnhancer.ts#L20

`requestAnimationFrame` is read into `rAF` at the top level of that module, meaning it will always use the real version of `window.requestAnimationFrame` and never the fake version inserted by jest.

ensconced:
If I move the assignment of `rAF` into the `autoBatchEnhancer` function, just before it's used, it fixes my test.

ensconced:
Let me know if there's anything else I can do to help this get merged 🙂

markerikson:
Out in https://github.com/reduxjs/redux-toolkit/releases/tag/v2.4.0 !

PR #4701, opened 2024-11-10
Read `window.rAF` later to allow fake timers to work correctly

ensconced:
This fixes issue https://github.com/reduxjs/redux-toolkit/issues/4693 by reading window.requestAnimationFrame within the autoBatchEnhancer function instead of in the top-level module scope. This ensures that we use the fake version of requestAnimationFrame if jest is using fake timers.

### DIFF
--- a/packages/toolkit/src/autoBatchEnhancer.ts
+++ b/packages/toolkit/src/autoBatchEnhancer.ts
@@ -15,13 +15,6 @@ const createQueueWithTimer = (timeout: number) => {
   }
 }
 
-// requestAnimationFrame won't exist in SSR environments.
-// Fall back to a vague approximation just to keep from erroring.
-const rAF =
-  typeof window !== 'undefined' && window.requestAnimationFrame
-    ? window.requestAnimationFrame
-    : createQueueWithTimer(10)
-
 export type AutoBatchOptions =
   | { type: 'tick' }
   | { type: 'timer'; timeout: number }
@@ -66,7 +59,10 @@ export const autoBatchEnhancer =
       options.type === 'tick'
         ? queueMicrotask
         : options.type === 'raf'
-          ? rAF
+          ? // requestAnimationFrame won't exist in SSR environments. Fall back to a vague approximation just to keep from erroring.
+            typeof window !== 'undefined' && window.requestAnimationFrame
+            ? window.requestAnimationFrame
+            : createQueueWithTimer(10)
           : options.type === 'callback'
             ? options.queueNotification
             : createQueueWithTimer(options.timeout)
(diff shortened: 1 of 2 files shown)


ensconced:
I'm just trying to figure out how I can write a test for this

packages/toolkit/src/autoBatchEnhancer.ts
```diff
@@ -62,6 +55,13 @@ export const autoBatchEnhancer =
 
     const listeners = new Set<() => void>()
 
+    // requestAnimationFrame won't exist in SSR environments.
+    // Fall back to a vague approximation just to keep from erroring.
+    const rAF =
+      typeof window !== 'undefined' && window.requestAnimationFrame
+        ? window.requestAnimationFrame
+        : createQueueWithTimer(10)
+
```
phryneas, line 64: At this point, this could probably be inlined in line 69.

ensconced: Makes sense, I have pushed this change.

packages/toolkit/src/tests/autoBatchEnhancer.test.ts
```diff
@@ -125,3 +125,84 @@ describe.each(cases)('autoBatchEnhancer: %j', (autoBatchOptions) => {
     expect(subscriptionNotifications).toBe(3)
   })
 })
+
+describe.each(cases)(
```
ensconced, line 129: Two of these tests were failing before the fix was applied

packages/toolkit/src/autoBatchEnhancer.ts
```diff
@@ -66,7 +59,8 @@ export const autoBatchEnhancer =
       options.type === 'tick'
         ? queueMicrotask
         : options.type === 'raf'
-          ? rAF
+          ? // requestAnimationFrame won't exist in SSR environments. Fall back to a vague approximation just to keep from erroring.
+            window?.requestAnimationFrame ?? createQueueWithTimer(10)
```
phryneas, line 63: `window?` doesn't cover all cases that `typeof window !== 'undefined'` does - optional chaining still can cause a reference error if a variable with that name never was defined:
grafik

ensconced: ah, of course. I have just pushed a fix to revert it back to a ternary like before

ensconced:
Let me know if there's anything else I can do to help this get merged 🙂

phryneas:
Mostly, "make the day have 48 hours" :)

Frankly, we have a ton to do (here and in our day-to-day lifes), so PRs can sit around for quite a while until one of us finds the time to thoroughly go over a few of them.

At first glance, this looks good, but please don't expect this to be merged/released soon. You can use the CodeSandbox release from the CSB comment in the meantime!

PR #4721, opened 2024-11-21
Preventing retries should result in the same result structure as if the retry wasn't used. 

### LINKED ISSUE #3789: Return `meta` from `retry.fail()`
Currently, if you use `retry` with `fetchBaseQuery()`, and you bail out early (eg, because of a 4xx status), then the result is missing `meta` data.

The code for `fail` is:

```typescript
function fail(e: any): never {
  throw Object.assign(new HandledError({ error: e }), {
    throwImmediately: true,
  })
}
```

So `retry.fail()` is only returning / throwing the error, and not returning the meta, although `HandledError` is capable of handling meta:

```typescript
export class HandledError {
  constructor(
    public readonly value: any,
    public readonly meta: any = undefined
  ) {}
}
```

LordZardeck:
This PR fixes #3789.  Originally, by forcing the value of `HandledError` to be of type `{error: any}` instead of the full result value, the meta field would be lost. This means there would be no way to fail out of retrying but keep the result consistent as if no retry had happened.

This PR fixes what I believe to be a bug, but could be a breaking change. Alternatives could be to add an additional fail option such as `failResult`, or by exporting the `HandledError` class to be manually thrown in whatever fashion the user desires.

### DIFF
--- a/packages/toolkit/src/query/retry.ts
+++ b/packages/toolkit/src/query/retry.ts
@@ -65,7 +65,7 @@ export type RetryOptions = {
 )
 
 function fail(e: any): never {
-  throw Object.assign(new HandledError({ error: e }), {
+  throw Object.assign(new HandledError(e), {
     throwImmediately: true,
   })
 }


EskiMojo14:
This would be a breaking change, yes - as the docs show, the fail utility only expects to be passed the error itself, not the entire result shape.

You could make it non-breaking by changing it back and moving meta to a second parameter: `retry.fail(result.error, result.meta)`

markerikson:
Opted for #4723 instead, thanks!

PR #4723, opened 2024-11-22
Allow passing meta to retry.fail, and passing baseQuery to ensure types match

EskiMojo14:
Alternative fix to #3789, compared to #4721 which would be a breaking change.

This makes the change non-breaking, and also allows providing the baseQuery as a type parameter to retry.fail for correct error and meta type checking.

Meta parameter being optional is fine as the meta property in QueryReturnValue is also optional.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -111,7 +111,7 @@ jobs:
       fail-fast: false
       matrix:
         node: ['20.x']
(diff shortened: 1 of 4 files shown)


markerikson:
Given that the only test failure was the TS 4.7 typetests, and that DefinitelyTyped itself only tests against "last 2 years of TS releases", now seems like a good time to update our own TS support matrix to drop support for TS <5.0.

PR #4758, opened 2024-12-09
fix(rtk-query): `useQuery` hook does not refetch after `resetApiState`

### LINKED ISSUE #3778: [RTK-Query]: `useQuery` hook does not refetch after `resetApiState` (2)
### Description:
This issue appears to be a bug in RTK-Query, which should have been resolved by #1735 and #3333. However, it appears that the problem persists.

### Reproduction Steps:
1. Access the following example showcasing the issue: [CodeSandbox](https://codesandbox.io/s/rtkq-resetapistate-query-2mxs7f)
2. Click on the "Rerender Component" button.
3. Click on the "Reset Api State" button.

### Expected Behavior:
After clicking "Reset API State," the request should not remain stuck in a pending status. Instead, it should automatically re-fetch the query. Currently, the request remains pending until the component is re-rendered. Even if we attempt to manually refetch, the query is executed, but the result remains pending.

juniusfree:
This PR fixes #3778 

Problem:
- The current implementation uses a `lastRenderHadSubscription` ref to track if the last render had a subscription. However, this ref can become out of sync with the actual subscription state, causing failed refetches, which lead to missing data and eventually stuck loading states.

Solution:
- We now use `promiseRef.current` to check if there was a previous subscription, as it is sufficient to track the subscription state. This removes redundant logic and ensures correct behavior.

### DIFF
--- a/packages/toolkit/src/query/tests/buildHooks.test.tsx
+++ b/packages/toolkit/src/query/tests/buildHooks.test.tsx
@@ -903,7 +903,7 @@ describe('hooks tests', () => {
         })
       })
 
-      test('reset after unmount/remount', async () => {
+      test('hook should not be stuck loading post resetApiState after re-render', async () => {
         const user = userEvent.setup()
 
         function QueryComponent() {
--- a/packages/toolkit/src/query/tests/buildHooks.test.tsx
+++ b/packages/toolkit/src/query/tests/buildHooks.test.tsx
@@ -902,6 +902,71 @@ describe('hooks tests', () => {
           status: 'uninitialized',
         })
       })
+
+      test('reset after unmount/remount', async () => {
+        const user = userEvent.setup()
+
+        function QueryComponent() {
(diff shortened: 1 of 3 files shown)


packages/toolkit/src/query/tests/buildHooks.test.tsx
```diff
@@ -895,6 +895,71 @@ describe('hooks tests', () => {
           status: 'uninitialized',
         })
       })
+
+      test('reset after unmount/remount', async () => {
```
riqts, line 899:
```suggestion
test('hook should complete post resetApiState after rerender', async () => {
```

onmotion:
2.4.0 works fine for me, however on 2.5.0 and 2.5.1 `resetApiState` doesn't trigger the hook state updating

andrejpavlovic:
@markerikson Any chance we can get this merged? Is there more work that needs to be done?

PR #4768, opened 2024-12-14
Ensure upserted cache entries always get written

### LINKED ISSUE #4749: UpsertQueryEntries breaks query invalidation when the invalidationBehavior is set to delayed
```ts
// api configuration
invalidationBehavior: "delayed"

// queries
getQuery: build.query<>({
      query:...,
      providesTags: ["QueryTag"]
}),
editQuery: build.mutation<>({
     invalidatesTags: ["QueryTag"]
}),

getSomeUnrelatedListData: buld.query<>({
     query:...
     async onQueryStarted(_, { dispatch, queryFulfilled }) {
        try {
          const res = await queryFulfilled;
          const results = res.data;
          dispatch(
           apiI.util.upsertQueryEntries(
              results.map((result) => ({
                endpointName: "getSomeUnrelatedSingleItemData",
                arg: { id: result.id },
                value: result,
              })),
            ),
          );
        } catch (err) {
          void err;
        }
      },
    }),
}),

getSomeUnrelatedSingleItemData: buld.query<unknown, {id: number}>({
     
})
```

Steps:

1. Trigger fetching the data by calling the getQuery (useGetQueryQuery())
2. Trigger fetching the data by calling the getSomeUnrelatedListData (useGetSomeUnrelatedListDataQuery())
     *  this populates entries for getSomeUnrelatedSingleItemData
3. Trigger getQuery data invalidation by calling the editQuery mutation

Expected result:
data returned by the useGetQueryQuery() is re-fetched

Actual Result:
the data is not re-fetched

```
"@reduxjs/toolkit": "2.4.0",
```

markerikson:
This PR:

- Fixes a bug with `upsertQueryEntries` where updates to an existing cache entry would not actually get completed, leaving the entry in an incorrect `pending` state indefinitely

Fixes #4749 

## Investigation

Made a Replay that shows the repro in action:

- https://app.replay.io/recording/rtk-query-4749-repro-upsertqueryentries--08eb2b78-7fc9-4b3a-b9fb-0a944f587fd7

Looks like there was a small flaw in the logic.  We already had a check for the `fulfilled` case that bails out of writing the result if `substate.requestId !== meta.requestId`, so that we don't end up with conflicting request results being written.  However, it also had a check for the older forced upsert logic as well, and it wasn't accounting for this newer approach.  

Meanwhile, the `pending` logic actually keeps around the existing request ID:

```ts
      substate.requestId =
        upserting && substate.requestId
          ? // for `upsertQuery` **updates**, keep the current `requestId`
            substate.requestId
          : // for normal queries or `upsertQuery` **inserts** always update the `requestId`
            meta.requestId
```

So, what was happening was roughly:

- Actual request for ID 2 resolves
- `fulfilled` action for ID 2 is written, with its request ID
- `upsertQueryEntries` for ID 2 is dispatched
- Inside the reducer:
  - `pending` write logic for the upsert sees there's an existing entry, and keeps the _old_ request ID from the actual API response, instead of overwriting it with the fake request ID from the upsert
  - Immediately after that, the `fulfilled` write logic for the upsert tries to update the entry... but the existing and incoming request IDs don't match, _and_ it wasn't being detected as a forced upsert because that logic only was checking for the older `upsertQueryData` approach

and so we're left with the existing entry, updated to be in a `pending` state, and without the upserted data written and marked as fulfilled.

### DIFF
--- a/packages/toolkit/src/query/core/buildSlice.ts
+++ b/packages/toolkit/src/query/core/buildSlice.ts
@@ -84,7 +84,7 @@ export type ProcessedQueryUpsertEntry = {
 /**
  * A typesafe representation of a util action creator that accepts cache entry descriptions to upsert
  */
-export type UpsertEntries<Definitions extends EndpointDefinitions> = <
+export type UpsertEntries<Definitions extends EndpointDefinitions> = (<
   EndpointNames extends Array<QueryKeys<Definitions>>,
 >(
   entries: [
@@ -95,7 +95,11 @@ export type UpsertEntries<Definitions extends EndpointDefinitions> = <
       >
     },
   ],
-) => PayloadAction<NormalizedQueryUpsertEntryPayload[]>
+) => PayloadAction<NormalizedQueryUpsertEntryPayload[]>) & {
+  match: (
+    action: unknown,
+  ) => action is PayloadAction<NormalizedQueryUpsertEntryPayload[]>
+}
 
 function updateQuerySubstateIfExists(
   state: QueryState<any>,
@@ -212,10 +216,10 @@ export function buildSlice({
       // RTK_autoBatch: true
     },
     payload: unknown,
+    upserting: boolean,
   ) {
     updateQuerySubstateIfExists(draft, meta.arg.queryCacheKey, (substate) => {
-      if (substate.requestId !== meta.requestId && !isUpsertQuery(meta.arg))
-        return
+      if (substate.requestId !== meta.requestId && !upserting) return
       const { merge } = definitions[meta.arg.endpointName] as QueryDefinition<
         any,
         any,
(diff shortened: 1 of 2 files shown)


ceisele-r:
@markerikson it's been a while since you fixed this without a new version being released since then - would it be possible to craft a release that includes this? That would be great!

markerikson:
Ah, yeah, this didn't get released. I'll try to find time to put out a new build.

ceisele-r:
Great, thank you for the release @markerikson !

markerikson:
Oh yeah, forgot to comment back here :) Out in https://github.com/reduxjs/redux-toolkit/releases/tag/v2.5.1 !

PR #4774, opened 2024-12-15
Update Docusaurus to v3

### LINKED ISSUE #4766: theme button and javascript code button
when cursor goes to the theme button it change into not allowed and javascript button not working it is not showing the javascript redux code

### LINKED ISSUE #4767: Search not working on a website
### Issue:

When visiting https://redux-toolkit.js.org/ and clicking the **Search** button nothing happens.

Image

### Expected behavior:

A popup would open after clicking the **Search** button, prompting the user to type and search the documentation.

markerikson:
This PR:

- [x] Upgrades Docusaurus from v2 to v3
- [x] Switches to use RSPack for builds

Fixes #4766 
Fixes #4767 

## Status

This was a rather painful last 8 hours. Recapping:

- the DS v3 upgrade was trivial. So was switching to use their new RSPack config option.
- upgrading to DS v3 seems to have fixed all the "things aren't working on the client side" issues (search, snippet copying), even with us using React 19 everywhere
- Webpack builds were 10min cold, 20s cached. RSPack was 5 min.
- the vast majority of our slow RTK docs build time is the `remark-ts-tools` plugin compiling our snippets
- most of _that_ time, in turn, appears to be scanning the file system (including all of the `/@types/` folders in the root `node_modules_`, _and_ loading source file contents. over and over and over.
- disabling the code snippet compilation dropped build time to 20s with RSPack
- given that, I've changed the DS config to only run the snippet compilation in CI
- I added a couple layers of caching inside `remark-ts-tools`, for module resolution and file contents. Once I finally got it cooperating, RSPack builds with TS compilation are down to ~90s
- I couldn't get the Yalc-installed local build to work in CI, so I'll put up a PR to `remark-ts-tools` and hopefully we can get that merged and published shortly


rcherny:
I am wondering if it is worth merging / deploying this even if the TS/JS toggles aren't working, simply because this does resolve a multitude of other issues.

Obviously up to the team, though 🤷‍♂️

markerikson:
@rcherny unfortunately it turns out that the current changes actually _delete_ all of the code snippets entirely. Which is... bad :)

It's something about the Markdown + TS compilation plugin, but I don't know what yet.

markerikson:
After a lot more investigation, I found that:

- `remark-typescript-tools` already had a PR merged to support MDX v3: https://github.com/phryneas/remark-typescript-tools/pull/15 . That had been released as `2.0.0-alpha.0`
- That same PR also upgraded Prettier to v3
- Prettier v3's API is now entirely async, so the PR converted the internals of the conversion to return Markdown nodes async
- _However_... while the plugin's `transform` is allowed to be async, it wants you to return the actual Markdown structure as real sync nodes and not promises. _That_ is why the code blocks were getting removed - the Markdown processing was looking for `Node` and getting `Promise<Node>` instead, so those got ignored.

I briefly tried converting the transformation logic to await those nested promises and struggled. So, I instead converted the Prettier logic to use `@prettier/sync` instead:

- https://github.com/phryneas/remark-typescript-tools/pull/17

We've published that as `remark-typescript-tools@2.0.0-alpha.1`, and _that_ actually fixes the snippet problems here:

image

PR #4775, opened 2024-12-16
Add infinite query examples

remus-selea:
- Add a cursor-based infinite scroll example 
- Add limit and offset infinite scroll example
- Add page and size infinite scroll example
- Add an intersection observer callback ref hook
- Update routing to easily go back to the menu from the top section
- Update handlers.ts with new handlers for the examples
- Update redux-toolkit version for the infinite scroll examples


remus-selea:
I got inspired by how discord does bidirectional pagination when jumping to a message from search. That's where I got "after", "before", "around"  and "limit" from.

PR #4782, opened 2024-12-19
Fix incorrect handling of query parameters with values 0 or empty str…

### LINKED ISSUE #4781: Incorrect handling of query parameters with values 0 or empty strings in CodeGeneration output
In RTK Query's CodeGeneration, there is an issue with the generated code when `encodeQueryParams` is set to `true`. The current generated code outputs the following logic:

```ts
queryArg.status
  ? encodeURIComponent(String(queryArg.status))
  : undefined,
```

With this logic, if `queryArg.status` is `''` or `0`, it returns `undefined`, which results in the `status` query parameter being omitted from the request URL. However, I believe it is more natural for the values `''` and `0` to be represented as `?status=` and `?status=0` in query parameters. In fact, the behavior of [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) demonstrates this as follows:

```ts
const paramsObj = { foo: "", bar: "0" };
const searchParams = new URLSearchParams(paramsObj);
console.log(searchParams.toString()); // "foo=&bar=0"
```

Therefore, in the generated code from the CodeGenerator, the logic should be correctly implemented as follows:

```ts
queryArg.status != null
  ? encodeURIComponent(String(queryArg.status))
  : undefined,
```

This code ensures that only `undefined` and `null` result in `queryArg.status` being undefined.

yukiyokotani:
fixes #4781

The generated code in RTK Query's CodeGeneration is modified as follows when `encodeQueryParams` is set to `true`:

Before:
```ts
queryArg.status
  ? encodeURIComponent(String(queryArg.status))
  : undefined,
```

After:
```ts
queryArg.status != null
  ? encodeURIComponent(String(queryArg.status))
  : undefined,
```

### DIFF
--- a/packages/rtk-query-codegen-openapi/src/generate.ts
+++ b/packages/rtk-query-codegen-openapi/src/generate.ts
@@ -448,7 +448,11 @@ export async function generateApi(
         const encodedValue =
           encodeQueryParams && param.param?.in === 'query'
             ? factory.createConditionalExpression(
-                value,
+                factory.createBinaryExpression(
+                  value,
+                  ts.SyntaxKind.ExclamationEqualsToken,
+                  factory.createNull()
+                ),
                 undefined,
                 factory.createCallExpression(factory.createIdentifier('encodeURIComponent'), undefined, [
                   factory.createCallExpression(factory.createIdentifier('String'), undefined, [value]),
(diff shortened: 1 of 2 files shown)


markerikson:
Out in https://github.com/reduxjs/redux-toolkit/releases/tag/%40rtk-query%2Fcodegen-openapi%402.1.0 !

PR #4813, opened 2025-01-13
Add skip condition on endpoint definition

### LINKED ISSUE #4810: RTK Query: Conditionally skip the query from the endpoints definition for hooks
Is there a way to set a condition or even a callback on a endpoint to conditionally skip it, instead of at the hook using skip or skipToken?

I have a few queries in our app that we should only make if the user has a specific permission. These queries are all around our app, and currently we are having to add logic to ever screen in order to set the hook to skip the query. This is some unnecessary boilerplate code for most queries, and we have thought about making a wrapper hook to extend the skip logic to include the permissions. Is there a better way?

SYoder1:
Adds a skip condition function to the endpoint definition to "skip" making the network request when returned `true`. Instead of needing to set the skip on every query hook it can be applied at a more global layer.

In order to keep `isUninitialized` set to `true`, the same way it works with `skip` on the hook, I needed to add a check for it in the useQuery hook.

Closes #4810

### DIFF
--- a/packages/toolkit/src/query/core/buildThunks.ts
+++ b/packages/toolkit/src/query/core/buildThunks.ts
@@ -955,6 +955,14 @@ In the case of an unhandled error, no tags will be "provided" or "invalidated".`
 
         if (
           isQueryDefinition(endpointDefinition) &&
+          endpointDefinition?.skipCondition &&
+          endpointDefinition?.skipCondition(state)
+        ) {
+          return false
+        }
+
+        if (
+          isQueryDefinition(endpointDefinition) &&
           endpointDefinition?.forceRefetch?.({
             currentArg,
             previousArg,
--- a/packages/toolkit/src/query/endpointDefinitions.ts
+++ b/packages/toolkit/src/query/endpointDefinitions.ts
@@ -686,6 +686,18 @@ export interface QueryExtraOptions<
   invalidatesTags?: never
 
   /**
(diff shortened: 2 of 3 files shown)


markerikson:
From some internal discussion:

> If we ever had suspense hooks, that could be a serious blockers for correct types for those, just to bring that up
>
> ```ts
>   const data = useSuspenseQuery({ ... })
>   // ^? MyDataType
>   const data = useSuspenseQuery(skipToken)
>   // ^? undefined
>   const data = useSuspenseQuery(foo ? skipToken : { ... })
>   // ^? MyDataType | undefined
> ```
>
> now assume `skip` could be set somewhere else, the first case would have to be `| undefined`

which is a valid concern.

I like the idea of this option, but it also feels like it will need some deeper consideration to figure out ramifications.

SYoder1:
Hey @markerikson, just circling back to this, is there anything I can do on my end for this or does this need to be solved at a project level since it effects other things already?

markerikson:
Unfortunately it's not something I'm going to have time to look at any time soon.  My current focus is wrapping up the Immer perf work, and then I'll be turning my attention to React-Redux.  

I think the last comment is still where things stand - I like the idea conceptually, it's reasonable, but I'm pretty sure this will end up having some unexpected impacts and we'd need time to try to figure out what those _might_ be.

I realize that's a very unsatisfying answer, and I apologize that I don't have more concrete feedback at this point :(

PR #4836, opened 2025-01-29
Docs: Add simpler Basic Usage example for `createListener` Middleware (suggestion)

kostasx:
I find the `Basic Usage` example way too convoluted and definitely not Basic. I suggest this is split into a very simple version  for setting up and dispatching an action (I tried to keep it as simple as possible) and an extended version with some of the other options available.

I've also included another listener on the extended version.

P.S. Up to the team and the community to decide, but I find myself in trouble whenever I need to reach out for a very basic setup in order to implement a listener in my code.

### DIFF
--- a/docs/api/createListenerMiddleware.mdx
+++ b/docs/api/createListenerMiddleware.mdx
@@ -25,14 +25,9 @@ Listeners can be defined statically by calling `listenerMiddleware.startListenin
 
 ```js
 import { configureStore, createListenerMiddleware } from '@reduxjs/toolkit'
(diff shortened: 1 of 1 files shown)


markerikson:
Is your concern around the _setup_ part (`createListenerMiddleware()`, `getDefaultMiddleware.concat(listenerMiddleware.middleware)`?  Or the "things you can do inside the `effect` callback" part?

kostasx:
Mostly the effect callback contents. The setup part, is mandatory for a basic example, I guess.

kostasx:
For a basic example, I think, all the things inside the callback are a bit too much.

PR #4844, opened 2025-02-05
fix trackProperties recursion

paztis:
passdown the checkedObjects to detect circularity

### DIFF
--- a/packages/toolkit/src/immutableStateInvariantMiddleware.ts
+++ b/packages/toolkit/src/immutableStateInvariantMiddleware.ts
@@ -55,6 +55,7 @@ function trackProperties(
         ignorePaths,
         obj[key],
         childPath,
+        checkedObjects
       )
     }
   }


soyjuanmacias:
Same issue here that can be solved with this PR.

markerikson:
I can reproduce the brokenness from this change locally, although the line numbers are different:

```
TypeError: Cannot read properties of undefined (reading 'value')
 ❯ detectMutations src/immutableStateInvariantMiddleware.ts:117:23
    115|       isImmutable,
    116|       ignoredPaths,
    117|       trackedProperty.children[key],
       |                       ^
    118|       obj[key],
    119|       sameRef,
 ❯ detectMutations src/immutableStateInvariantMiddleware.ts:114:20
 ❯ detectMutations src/immutableStateInvariantMiddleware.ts:114:20
 ❯ detectMutations src/immutableStateInvariantMiddleware.ts:114:20
 ❯ Object.detectMutations src/immutableStateInvariantMiddleware.ts:24:14
 ❯ src/immutableStateInvariantMiddleware.ts:228:28
 ❯ Object.measureTime src/utils.ts:9:16
 ❯ src/immutableStateInvariantMiddleware.ts:225:22
 ❯ src/tests/immutableStateInvariantMiddleware.test.ts:148:7
```

markerikson:
It looks like the "`.value` of undefined" bit is related to the mock data being generated for the test: 

```ts
state.foo.bar = new Array(10).fill({ value: 'more' })
```

rather than the `.value` field from the `TrackedProperty` objects.

Honestly I'm not sure what's going on here, and this is low enough priority that I don't want to get bogged down debugging it atm.

soyjuanmacias:
Just solve it @paztis. Can you try it and modify #4844 ?

**Problem:**

I was encountering false-positive mutation detections because my `trackProperties` function wasn't consistently memoizing object references. When the same object was encountered multiple times (for example, using `Array.fill({ value: 'more' })` or updating `isPrefetch`), a new tracker with an empty `children` property was created instead of reusing the existing one. This inconsistency caused the mutation detector to incorrectly report changes.

**Solution:**

I fixed the issue by introducing memoization in `trackProperties` using a `WeakMap`. Now, when processing an object:
- If the object is a primitive or already considered immutable, I simply return a basic tracker.
- If I’ve already seen the object, I retrieve its tracker from the `WeakMap` to ensure consistency.
- Otherwise, I create a new tracker, store it in the `WeakMap`, and use it for all future detections.

This change ensures that each object reference is tracked only once, completely eliminating the false-positive mutation detections.

This is the complete code for trackProperties

```ts
function trackProperties(
  isImmutable: IsImmutableFunc,
  ignorePaths: IgnorePaths = [],
  obj: Record<string, any>,
  path: string = '',
  checkedObjects: WeakMap<any, TrackedProperty> = new WeakMap(),
) {
  if (typeof obj !== 'object' || obj === null || isImmutable(obj)) {
    return { value: obj, children: {} }
  }

  if (checkedObjects.has(obj)) {
    return checkedObjects.get(obj)!
  }

  const tracked: TrackedProperty = { value: obj, children: {} }
  checkedObjects.set(obj, tracked)

  for (const key in obj) {
    const childPath = path ? path + '.' + key : key
    if (ignorePaths.length && ignorePaths.indexOf(childPath) !== -1) {
      continue
      }

    tracked.children[key] = trackProperties(
      isImmutable,
      ignorePaths,
      obj[key],
      childPath,
      checkedObjects,
    )
  }
  return tracked as TrackedProperty
}
``` 

And here, tests passed locally:
image

image

markerikson:
@soyjuanmacias  :

I just tried out your suggested changes, and I don't think it actually fixes the issue.

I think the first issue is that our existing tests don't correctly capture this behavior.  So, I tried changing one of our existing tests to try to force a failure:

```ts
  it('works correctly with circular references', () => {
    const next: MWNext = (action) => action

    const dispatch = middleware()(next)

    let x: any = {}
    let y: any = {}
    x.y = y
    y.x = x

    y.items = new Array(5).fill({ data: 'testing' })

    expect(() => {
      dispatch({ type: 'SOME_ACTION', x })
    }).not.toThrow()

    // ADDED
    const tracker = trackForMutations(isImmutableDefault, [], x)
    expect(() => tracker.detectMutations()).not.toThrow()
  })
```

That does induce an error:

```
RangeError: Maximum call stack size exceeded
 ❯ trackProperties src/immutableStateInvariantMiddleware.ts:39:46
     37|   obj: Record<string, any>,
     38|   path: string = '',
     39|   checkedObjects: Set<Record<string, any>> = new Set(),
       |                                              ^
     40| ) {
     41|   const tracked: Partial<TrackedProperty> = { value: obj }
 ❯ trackProperties src/immutableStateInvariantMiddleware.ts:53:31
 ```

However, when I tried using your suggested `trackProperties` implementation, I get a similar error.

I'd been hoping to include this in 2.7.0, which I'm going to publish shortly.  But, given that, I'm going to set this aside again.

Could you please file a new PR that includes both the code changes _and_ a test that fails with the existing implementation and passes with the new one?

soyjuanmacias:
> Could you please file a new PR that includes both the code changes _and_ a test that fails with the existing implementation and passes with the new one?

Of course @markerikson.

I'm going to start working on a new PR that addresses these issues and will create it as soon as possible so it can be incorporated into the next release.

Maybe you can close this one, and I will reference this in the new PR.

PR #4845, opened 2025-02-05
Re-export `setGlobalDevModeChecks` from `reselect`

### LINKED ISSUE #4681: Re-export setGlobalDevModeChecks from reselect
Hi, we'd love the ability to directly import setGlobalDevModeChecks from @reduxjs/toolkit rather than importing the values from reselect

We are importing it currently using this so that we can run our test files without the stability checks for performance reasons. We have these turned back on for our tests in CI
```ts
const { setGlobalDevModeChecks } = require(
    require.resolve("reselect", {
        paths: [require.resolve("@reduxjs/toolkit")],
    }),
);
```

Your current exports for re-select
```ts
export type { Selector, OutputSelector } from 'reselect'
```

Happy to raise a pr to change to this, as long as you are happy with the change
```ts
export type { Selector, OutputSelector, setGlobalDevModeChecks } from 'reselect'
```

lg-kialo:
Resolves #4681.

I tried to update the API report file `etc/redux-toolkit.api.md` as suggested by the contribution guide but it didn't work, is that maybe broken? Seems like that report hasn't been updated in a while.

EDIT: Ah yes it was acknowledged [here](https://github.com/reduxjs/redux-toolkit/pull/4519#issuecomment-2249140044).

### DIFF
--- a/packages/toolkit/src/index.ts
+++ b/packages/toolkit/src/index.ts
@@ -15,6 +15,7 @@ export {
   createSelector,
   createSelectorCreator,
   lruMemoize,
+  setGlobalDevModeChecks,
   weakMapMemoize,
 } from 'reselect'
 export type { Selector, OutputSelector } from 'reselect'


markerikson:
Appreciate the PR, but per #4681 this isn't something we want to re-export directly.

Issue #4834, opened 2025-01-29
[RTK Query] Reset state for lazy query, similar to mutation


Tasin5541:
Having a `reset` option for query would definitely help. Take for example, we have a derived variable that reads from the result of 2 queries (whichever is present or the latest one fetched).

```
const { trigger1, { data: data1 } } = useData1LazyQuery();
const { trigger2, { data: data2 } } = useData2LazyQuery();
const derVal = data1 || data2 || null
```

We need the `derVal` to always point to the latest data. If `data2` is fetched after fetcing `data1`, `derVal` should be `data2`.

Having an option to reset the other query's state during triggering one query would definitely be the best approach here.

```
const { trigger1, { data: data1, reset: reset1 } } = useData1LazyQuery();

// trigger data2 handler
reset1(); 
trigger2();
```

alexmotoc:
Hey there! The ability to reset lazy queries has been added as part of [v2.4.0](https://github.com/reduxjs/redux-toolkit/releases/tag/v2.4.0)

Tasin5541:
> Hey there! The ability to reset lazy queries has been added as part of [v2.4.0](https://github.com/reduxjs/redux-toolkit/releases/tag/v2.4.0)

Can we please add it in the docs? 
https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#uselazyquery does not have a definition for `reset`

EskiMojo14:
PRs welcome :)

Tasin5541:
> PRs welcome :)

PR created :)

PR #4847, opened 2025-02-08
doc: update type definitions for lazy query hooks to include reset method

Tasin5541:
Updates the return type of `UseLazyQuery`:
`UseQueryStateResult` -> `UseLazyQueryStateResult`

Adds `reset` method in the `UseLazyQueryStateResult` definition

Addresses #4834


EskiMojo14:
thanks!

PR #4864, opened 2025-02-26
Allow standard schemas to validate endpoint values

EskiMojo14:
Type-wise, schemas are required to match the type of the value they're validating - they can't change a string to a number for example. Transformations that preserve the correct type are allowed, however.

Schemas can also be used as a source of inference, for example 
```ts
build.query({
  query: ({ id }) => `posts/${id}`,
  argSchema: v.object({ id: v.number() }),
  responseSchema: postSchema
})
// or more likely in a TS app, since arg will be checked statically
build.query({
  query: ({ id }: { id: number }) => `posts/${id}`,
  responseSchema: postSchema
})
```

Schema parse failures are treated as unhandled errors, i.e. they're thrown and serialized rather than being returned. This is mainly because we have no idea what the base query error shape should look like, so we can't match it.

For transformable values, `raw<value>Schema` checks the value before transformation, and the `<value>Schema` checks the possibly transformed value.

```ts
build.query({
  query: (_arg: void) => `posts`,
  rawResponseSchema: v.array(postSchema),
  // raw response is inferred from schema
  transformResponse: (posts) =>
    postAdapter.getInitialState(undefined, posts),
  responseSchema: v.object({
    ids: v.array(v.number()),
    entities: v.record(
      v.pipe(v.string(), v.transform(Number), v.number()),
      postSchema,
    ),
  }),
})
```


agusterodin:
FWIW, I really like where your head is at in the second example you proposed.

Input/request validation is probably out of scope for a data fetching solution anyway.

I definitely prefer the terminology `responseSchema` btw, as it feels more intuitive, but I understand that RTKQ can be applied to async state management use-cases beyond just simple HTTP requests.

build.query({
  query: ({ id }: { id: number }) => `posts/${id}`,
  resultSchema: postSchema
})

agusterodin:
Here is link to my initial idea pitch just so there is a paper trail:
https://x.com/agusterodin/status/1894623542471528888

Will respond directly to this PR for now on for brainstorming.

agusterodin:
It might be overkill, but i'm really enticed by the idea of performing schema validation in a web worker (off main thread). I have seen brief main thread lag when validating massive payloads from server (10MB+). One of the projects I work on professionally is a Tableau-esque data viewer where user can view and interact with up to 500k geospatial results.

One of my `transformResponse` functions was bogging the main thread down, so I experimented moving that transformation into a web worker. I ultimately decided not to go through with this, particularly because the type-safety situation was less than ideal.

Attached below is my rough attempt. The web worker was usable via promise.

Usage inside API slice
```
transformResponse: async (response: DashboardDataTableResult[]) => {
        return await transformDataTableResponseOffMainThread(response)
}
```

workerPromise.ts
```
import { DashboardDataTableResult } from 'datamodels'

export type TableRowByIdDictionary = Record<string, DashboardDataTableResult>

export function transformDataTableResponseOffMainThread(tableResults: DashboardDataTableResult[]): Promise<TableRowByIdDictionary> {
  return new Promise((resolve, reject) => {
    const worker = new Worker(new URL('./transformWorker.ts', import.meta.url))

    worker.onmessage = (event: MessageEvent<TableRowByIdDictionary>) => {
      resolve(event.data)
      worker.terminate()
    }

    worker.onerror = error => {
      reject(error)
      worker.terminate()
    }

    worker.postMessage(tableResults)
  })
}
```

workerLogic.ts
```
import { keyBy } from 'lodash'
import { DashboardDataTableResult } from 'datamodels'

self.onmessage = function (event: MessageEvent<DashboardDataTableResult[]>) {
  const data = event.data
  const transformedData = transformData(data)
  self.postMessage(transformedData)
}

function transformData(data: DashboardDataTableResult[]) {
  const chan: Partial<DashboardDataTableResult>[] = new Array(5000000).fill(0).map((_, i) => ({
    id: i,
    ipv4: 'zcxv',
    ipv6: '342342'
  }))
  return keyBy(chan, 'id')
}
```

Getting web worker to work with a variety of build tooling could be a pain though, particularly this kind of thing.
`const worker = new Worker(new URL('./transformWorker.ts', import.meta.url))`. It works well in Next.js, but I have no idea if other bundlers handle it the same way.

My example above is obviously for a different use case (transforming a response instead of validating a payload), but I figure is worth sharing.

agusterodin:
As for question about whether validation would occur before transformResponse or after:

"I would imagine responseSchema validating what data came back from the server pre-transformResponse. 

If a transformResponse function is provided, perhaps the RTKQ endpoint will use the return type of their provided transformResponse function. 

If no transformResponse function is provided, the RTKQ endpoint return type will be whatever was provided as responseSchema."

EskiMojo14:
> If rawReponseSchema for pre-tranformResponse is the route you end up going, it would be cool if responseSchema would be optional in if rawResponseSchema is provided.

All schemas are optional - the difficult thing with `responseSchema` vs `rawResponseSchema` is which one should be used for inference/checked against the provided type parameter 🤔 

Currently I have it as `rawResponseSchema` is always `StandardSchemaV1<BaseQueryResult<BaseQuery>>`, and `responseSchema` is always `StandardSchemaV1<ResultType>`, which matches `transformResponse`'s `(res: BaseQueryResult<BaseQuery) => ResultType`.

The difficulty with that is that `rawResponseSchema` cannot be used as a source of inference, so the below would not infer properly: 
```ts
build.query({
  query: ({ id }: { id: number }) => `posts/${id}`,
  rawResponseSchema: postSchema
})
```
*however*, this doesn't seem like a huge issue to me, as `responseSchema` will be used even if transformResponse isn't provided (it still happens, it just defaults to an identity function).

an *ideal* state of affairs would be 
```ts
build.query({
  query: ({ id }: { id: number }) => `posts/${id}`,
  rawResponseSchema: postSchema,
  transformResponse: (res /* inferred from rawResponseSchema */) => transformed // provides final data type
})
// or no transformResponse
build.query({
  query: ({ id }: { id: number }) => `posts/${id}`,
  responseSchema: postSchema,
})
```
but this isn't currently possible (the inference part at least). The closest you'd get is manually annotating
```ts
build.query({
  query: ({ id }: { id: number }) => `posts/${id}`,
  rawResponseSchema: postSchema,
  transformResponse: (res: Infer<typeof postSchema>) => transformed // provides final data type
})
```

EskiMojo14:
cool, that wasn't actually too bad 😄 
so basically:
```ts
// without transformResponse
build.query({
  query,
  responseSchema
})
// with transformResponse
build.query({
  query,
  rawResponseSchema,
  transformResponse
})
```
We *could* use a union to enforce this, but honestly the types are complicated enough and i think some people may still want to use both.

agusterodin:
> > If rawReponseSchema for pre-tranformResponse is the route you end up going, it would be cool if responseSchema would be optional in if rawResponseSchema is provided.
> 
> All schemas are optional - the difficult thing with `responseSchema` vs `rawResponseSchema` is which one should be used for inference/checked against the provided type parameter 🤔
> 
> Currently I have it as `rawResponseSchema` is always `StandardSchemaV1<BaseQueryResult<BaseQuery>>`, and `responseSchema` is always `StandardSchemaV1<ResultType>`, which matches `transformResponse`'s `(res: BaseQueryResult<BaseQuery) => ResultType`.
> 
> The difficulty with that is that `rawResponseSchema` cannot be used as a source of inference, so the below would not infer properly:
> 
> ```ts
> build.query({
>   query: ({ id }: { id: number }) => `posts/${id}`,
>   rawResponseSchema: postSchema
> })
> ```
> 
> _however_, this doesn't seem like a huge issue to me, as `responseSchema` will be used even if transformResponse isn't provided (it still happens, it just defaults to an identity function).
> 
> an _ideal_ state of affairs would be
> 
> ```ts
> build.query({
>   query: ({ id }: { id: number }) => `posts/${id}`,
>   rawResponseSchema: postSchema,
>   transformResponse: (res /* inferred from rawResponseSchema */) => transformed // provides final data type
> })
> // or no transformResponse
> build.query({
>   query: ({ id }: { id: number }) => `posts/${id}`,
>   responseSchema: postSchema,
> })
> ```
> 
> but this isn't currently possible (the inference part at least). The closest you'd get is manually annotating
> 
> ```ts
> build.query({
>   query: ({ id }: { id: number }) => `posts/${id}`,
>   rawResponseSchema: postSchema,
>   transformResponse: (res: Infer<typeof postSchema>) => transformed // provides final data type
> })
> ```

Is it not possible due to limitations in standard schema specification, the way RTKQ is currently structured, or due to limitation in TypeScript language itself?

EskiMojo14:
> Is it not possible due to limitations in standard schema specification, the way RTKQ is currently structured, or due to limitation in TypeScript language itself?
Due to how it's currently structured - I've just pushed a change adding the inference :)

agusterodin:
Ability to provide a global callback for situation where schema validation fails would be super valuable btw. Would allow for warning toast, logging to external service, etc.

Callback would contain information about the request such as endpoint name, request payload, etc.

agusterodin:
Not sure if useful, but noticed some data fetching libraries (prior art) listed in standard schema readme. These appear to be very lightweight / way less fully-fledged than RTKQ and TanStack.

https://github.com/bekacru/better-fetch
https://github.com/zayne-labs/callapi

EskiMojo14:
> Ability to provide a global callback for situation where schema validation fails would be super valuable btw. Would allow for warning toast, logging to external service, etc.

> Callback would contain information about the request such as endpoint name, request payload, etc.

Added - callback parameters are `onSchemaFailure(error, info)` where error is an extended `SchemaError` with `.value` (the original value before parsing) and `.schemaName` (e.g. `"argSchema"`) and info is an object like `{ endpoint, arg, type, queryCacheKey? }`

markerikson:
@agusterodin can you give an example of when/why an option to skip schema validation would actually be necessary?  Like, in that geospatial results blog example: do you really need to validate the entire response?  when would you want to supply a schema for that endpoint and then _not_ run it?  not sure I follow the intended use case for such an option.

agusterodin:
> @agusterodin can you give an example of when/why an option to skip schema validation would actually be necessary? Like, in that geospatial results blog example: do you really need to validate the entire response? when would you want to supply a schema for that endpoint and then _not_ run it? not sure I follow the intended use case for such an option.
> 
> is it just to get the type inference? "trust me, the data will look like this, just don't bother checking for me"?

The rationale would be to still take advantage of type inference and so that endpoint definitions can be consistent. Mixing-and-matching providing standard schema vs. providing generics (the current way of defining request/response types) based solely on whether you want to validate payload or not would likely be clunky.

***

More context on why we skip schema validation for certain endpoints: mostly for performance reasons. There are a few endpoints we use that return massive amounts of data (10MB+) and cause the browser to dramatically slow down when trying to validate the schema. 

For these rare cases, we leave the Zod schema in the code but comment it out.

image

There are probably other ways to mitigate:
- Paginating the data (would require extensive changes to the way UI works).
- Performing schema validation in a web worker (briefly mentioned earlier in thread, not sure if overhead of thread communication is worth it and if type-safety situation is viable).

markerikson:
@agusterodin gotcha, thanks for the response.

I'm planning to ship this in 2.7.0.  Was _going_ to release that today :)  but realized we still need docs for this feature, plus a couple other tweaks.  Could you give the current PR build a shot and give me any last-minute feedback before this goes live?

agusterodin:
> @agusterodin gotcha, thanks for the response.
> 
> I'm planning to ship this in 2.7.0. Was _going_ to release that today :) but realized we still need docs for this feature, plus a couple other tweaks. Could you give the current PR build a shot and give me any last-minute feedback before this goes live?

Hell yeah! Will test and provide feedback over next couple days

EskiMojo14:
The root issue here is that with the way RTKQ is architected, there are only two types of error:

- "expected" errors 
  - type is determined by base query (in this case `FetchBaseQueryError`), and `queryFn` errors are required to match this
  - strangely `transformErrorResponse` breaks this contract - you can return anything and the rest of your app will act like it matches the base query errors
- "unexpected" errors
  - always `SerializedError`, prevents tag invalidation

There's no such thing as a unique error type for an endpoint, for example, and adding this would be a major restructure of RTKQ's types.

With regards to schemas, we have absolutely no way of knowing how to turn our schema errors into a shape that would match the base query's errors. We could possibly leave this up to the user:
```ts
const api = createApi({
  baseQuery: fetchBaseQuery(),
  catchSchemaFailure: (error, info) => ({ status: "CUSTOM_ERROR", data: error.issues, error: `${error.schemaName} failed validation` })
  endpoints: () => ({})
})
```

This still raises questions - should the schema error be passed to `transformErrorResponse`? should it also be passed to the error schemas?

A more drastic approach would be adding a whole new type of error to the system specifically for schema failures, but this would be potentially breaking for code like the above that's only expecting `BaseQueryError<BaseQuery> | SerializedError`.

All the reasons above is also why your errorResponseSchema needs to match the whole base query error, because we'd have *no idea* how to extract only the server response from the shape returned from the base query. None of our code can be written assuming that errors will only ever be `FetchBaseQueryError`, because other base queries may use errors that use look nothing like it.

EskiMojo14:
@agusterodin i've raised #4934 to add `catchSchemaFailure` - would you be able to try out the build from that and see if that helps?

PR #4872, opened 2025-03-03
Add providesTags handling for upsertQueryEntries

### LINKED ISSUE #4870: `upsertQueryEntries` does not create subscriptions and provided tags
Hi! 

I have tried to utilize the `upsertQueryEntries` util to populate individual cache entries, similar to the `getPosts`/`getPost` example in [the docs](https://redux-toolkit.js.org/rtk-query/api/created-api/api-slice-utils#upsertqueryentries) and [v2.3.0 release notes](https://github.com/reduxjs/redux-toolkit/releases/tag/v2.3.0).
This works fine, until I want to perform some optimistic updates on the different entries, as `selectInvalidatedBy` does not contain the individual entity queries.

I tested with replacing `upsertQueryEntries` with `upsertQueryData`, and this seems to work. Inspecting the difference in api state from the two utils, it looks like they both add all the `queries` to the api state, but `upsertQueryData` adds subscriptions and `provided` tags for all the individual entries, while `upsertQueryEntries` does not. 

I have created a codesandbox where you can easily toggle between the to different utils. 
Codesandbox: https://codesandbox.io/p/sandbox/agitated-surf-zxvyw5

So now I'm wondering if this is a bug, or intentional behavior and that I will have to resort to `upsertQueryData` if i want to be able to do optimistic updates on the individual cache entries.

markerikson:
Fixes #4870 .

### DIFF
--- a/packages/toolkit/src/query/core/buildSlice.ts
+++ b/packages/toolkit/src/query/core/buildSlice.ts
@@ -1,4 +1,4 @@
-import type { Action, PayloadAction, UnknownAction } from '@reduxjs/toolkit'
+import type { PayloadAction } from '@reduxjs/toolkit'
 import {
   combineReducers,
   createAction,
@@ -23,7 +23,6 @@ import type {
   QueryCacheKey,
   SubscriptionState,
   ConfigState,
-  QueryKeys,
   InfiniteQuerySubState,
   InfiniteQueryDirection,
 } from './apiState'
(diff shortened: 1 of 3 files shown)


sondreluc:
Swiftly handled @markerikson! 💪 
Tested this in my project,  and the codesanbox mentioned in #4870, and it does fixes the issues with optimistic updates not working because tags not being created by `upsertQueryEntries`! 🙌

PR #4882, opened 2025-03-07
chore: update TypeScript version to 5.8

aryaemami59:
## **This PR**:

- [X] Bumps TypeScript to version 5.8.
- [X] Adds TypeScript 5.8 to TS versions to test against during CI.
- [X] Remove TypeScript versions lower than 5.0 from the CI matrix.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -131,7 +131,7 @@ jobs:
       fail-fast: false
       matrix:
         node: ['22.x']
-        ts: ['5.0', '5.1', '5.2', '5.3', '5.4', '5.5']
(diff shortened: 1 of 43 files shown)


timdorr:
On our test suite, how far back do we want to go on the Typescript versions? 5.1 is coming up on 2 years old, so not the youngest version but not _that_ far behind. More of a question than anything else.

aryaemami59:
@timdorr Not sure tbh but we just got rid of <5.0 I'd say let's stick with this for at least a little while then I can get rid of 5.0 and 5.1 if everyone else is okay with it.

PR #4887, opened 2025-03-11
fix: React-Native package exports

### LINKED ISSUE #4524: Warning with @reduxjs/toolkit/query/react Is this an Expo or RTK issue?
### Describe the bug
When using Expo with `@reduxjs/toolkit/query/react`, I encounter the following warning:

warning: The package /node_modules/@reduxjs/toolkit/query/react contains an invalid package.json configuration. 
Consider raising this issue with the package maintainer(s).
Reason: The target for "." defined in "exports" is "./../../dist/query/react/cjs/index.js", however this value is an invalid subpath or subpath pattern because it includes "..". Falling back to file-based resolution.

### Questions
1. **Should I be concerned about this warning?**
2.  **Is this issue related to the configuration of Redux Toolkit (RTK) or something specific to Expo?**

aryaemami59:
## **This PR**:

- [X] Resolves #4524.


packages/toolkit/query/package.json
```diff
@@ -4,16 +4,9 @@
   "description": "",
   "type": "module",
   "module": "../dist/query/rtk-query.legacy-esm.js",
-  "main": "../dist/query/cjs/index.js",
+  "main": "./../dist/query/rtk-query.modern.mjs",
   "types": "./../dist/query/index.d.ts",
-  "exports": {
```
phryneas, line 9: Dropping the `exports` field should be okay - a bundler that supports `exports` should look at the root-level `package.json`.

packages/toolkit/query/package.json
```diff
@@ -4,16 +4,9 @@
   "description": "",
   "type": "module",
   "module": "../dist/query/rtk-query.legacy-esm.js",
-  "main": "../dist/query/cjs/index.js",
+  "main": "./../dist/query/rtk-query.modern.mjs",
```
phryneas, line 7: Here be dragons - this switches from CJS to ESM. Probably be fine, since only outdated bundlers would look at this file at all, but pointing it out.

phryneas (commented):
Not sure about the tsup changes (was this just a port to TS or bigger changes?) but the `exports` field changes generally seem reasonable. This should at least be a minor bump, though.

examples/publish-ci/node-esm/package.json
```diff
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "build": "echo Done",
-    "test": "node test-cjs.cjs && node test-esm.mjs"
+    "test": "node --no-experimental-require-module test-cjs.cjs && node test-esm.mjs"
```
aryaemami59, line 8: Needed to avoid the **`module-sync`** condition.

examples/publish-ci/node-standard/package.json
```diff
@@ -4,7 +4,7 @@
   "license": "MIT",
   "scripts": {
     "build": "echo Done",
-    "test": "node test-cjs.js && node test-esm.mjs"
+    "test": "node --no-experimental-require-module test-cjs.js && node test-esm.mjs"
```
aryaemami59, line 7: Needed to avoid the **`module-sync`** condition.

packages/toolkit/query/package.json
aryaemami59: Without this `vitest` will fail unless we explicitly specify `resolve.mainFields` in `vitest` config files. Still not entirely sure why this happens, I know it's `vite` related and with this change added everything seems to work okay.

markerikson: On the flip side, this was here to keep Webpack 4 working.  

Obviously at this point we shouldn't want to care about Webpack 4 much, but that's why it's here.  

Why is Vitest looking at this field at all anyway?  Shouldn't it be following `exports` instead and not even looking at this file?

markerikson (approved):

markerikson:
Spent a couple hours poking at this and concluded that Vitest does _not_ seem to be looking at `exports` for those nested entry points :(

I honestly still don't know why.  This is the closest I could find:

- https://github.com/vitest-dev/vitest/discussions/4233

but that just says "it uses the Node resolution algorithm", which _ought_ to still involve `exports` I'd think.

As a result, we were seeing a module mismatch - it was importing `redux-toolkit.modern.mjs`, but `rtk-query.development.cjs`, and thus Bad Things happened.

What we've ended up on is specifying `resolve: {mainFields: ["module"]}`, which forces it to stop looking at the `main` field and land on `module`, which currently points to our `legacy-esm.js` artifacts.  That's not _good_ and will almost definitely bite us at some point, but it lets the "test against dist" test pass for now.

markerikson:
Further updates:

After some discussion on Bluesky, I wondered if this was specific to either Yarn or NPM in general, due to `package.json` also having the `name: "@reduxjs/toolkit"` value, and thus maybe preventing Vitest from going into `node_modules` to find the installed version.

After further ~~random flailing and changing settings~~ investigation, I concluded that this is indeed the case.  So, instead of adding a `TEST_DIST`-specific alias, what we really need is to temporarily alter the package `name` field so that Vitest is forced to look inside of `node_modules` instead.  Did that in the `tests.yml` workflow.

I've confirmed (locally at least) that this does indeed result in Vitest loading the `rtk-query.modern.mjs` artifact as we would want.

markerikson:
Worth extracting @phryneas 's explanations for visibility:

> Going over one entry point's `exports` in order here:
> 
> * `module-sync` is used to enable `require(ESM)` if supported by the runtime
> * `module` aims at bundlers
> * `react-native`, with ESM for `import` and CJS for `require`
> * `browser` with the ESM browser build
> * `import` with the ESM build
> * fallback to CJS, will probably only be hit in outdated runtimes that don't support `require(ESM)`
>
> On top of that, for RN versions that support ESM, but not exports, pointing at the legacy ESM build [with a top-level `react-native` condition]

Issue #3429, opened 2023-05-09
upsertMany with multiple updates with the same id throws away updates


johannaljunggren:
I'm using upsertMany to push multiple updates with the same id. If the id previously didn't exist only the last update in the array of updates is added to the store. Example using the books example from the createEntityAdapter documentation https://redux-toolkit.js.org/api/createEntityAdapter:
```ts
const booksAdapter = createEntityAdapter({
  selectId: (book) => book.customId,
  sortComparer: (a, b) => (a.customId > b.customId ? 1 : -1)
});

const booksSlice = createSlice({
  name: "books",
  initialState: booksAdapter.getInitialState(),
  reducers: {
    booksUpserted: booksAdapter.upsertMany,
  }
});

const { booksUpserted } = booksSlice.actions;

const store = configureStore({
  reducer: {
    books: booksSlice.reducer
  }
});

store.dispatch(booksUpserted([
    { customId: "a", title: "First" },
    { customId: "a", author: "Author" },
    { customId: "a", description: "Description" }
  ])
);
```
After dispatching dispatching I expected the state to look like this:
`{"ids":["a"],"entities":{"a":{"customId":"a", title: "First", author: "Author", "description":"Description"}}}`
Instead it looks like this: 
`{"ids":["a"],"entities":{"a":{"customId":"a","description":"Description"}}}`
I created a codesandbox example to demonstrate the issue https://codesandbox.io/s/upsertmany-test-lwmfy4?file=/src/App.js

markerikson:
Hmm. Currently the logic is:

```ts
  function upsertManyMutably(
    newEntities: readonly T[] | Record<EntityId, T>,
    state: R
  ): void {
    const [added, updated] = splitAddedUpdatedEntities<T>(
      newEntities,
      selectId,
      state
    )

    updateManyMutably(updated, state)
    addManyMutably(added, state)
  }
```

I wonder what would happen if we just flip the `update` and `add` steps?

dlarroder:
@markerikson any updates on this?

markerikson:
@dlarroder there's been no comments since I wrote one a year ago, and the issue is still open, so no :)

PR #4890, opened 2025-03-12
Process multiple update for the same new entity

demyanm:
Fixes https://github.com/reduxjs/redux-toolkit/issues/3429


packages/toolkit/src/tests/createEntityAdapter.test.ts
```diff
@@ -0,0 +1,53 @@
+import { createAction, createEntityAdapter, type EntityId } from '@reduxjs/toolkit'
+
+describe('createEntityAdapter', () => {
```
EskiMojo14, line 3: tests for createEntityAdapter are in `packages/toolkit/src/entities/tests`

EskiMojo14 (approved):
makes sense, thanks!

Issue #4009, opened 2023-12-22
resetApiState is not resetting the local state of the hook


saibs:
On redux toolkit 1.9.2, resetApiState resets the hooks local state.
After upgrading to toolkit to 2.0.1 we found an issue in our app relating to resetApiState call is not resetting the local state of the hook

[Sandbox using toolkit 1.9.2](https://codesandbox.io/p/sandbox/priceless-poincare-cnctxs?file=%2Fpackage.json%3A12%2C26)
[Sandbox using  toolkit 2.0.1](https://codesandbox.io/p/sandbox/unruffled-platform-6tjjht?file=%2Fpackage.json%3A7%2C29)

Reproduction steps:
- Click on login button
- In the next screen you should see 
<img width="959" alt="image" src="https://github.com/reduxjs/redux-toolkit/assets/1705639/2b996d5a-4187-4a61-946b-14611b00a021">
- Click logout

Result in 1.9.2
<img width="590" alt="image" src="https://github.com/reduxjs/redux-toolkit/assets/1705639/2deadfa0-190e-4795-afd3-447fbc6ac4d0">

Result in 2.0.1 - this is a bug
<img width="972" alt="image" src="https://github.com/reduxjs/redux-toolkit/assets/1705639/0d315a77-431b-46f4-b2ac-39eebd5f6f6b">

Luk-z:
Hi guys, me too. This is the code to reproduce the bug. If you press on "reset api" button restApiState is called but data still remain populated. isUnitialized is setted to true correctly.

https://playcode.io/2255811

Luk-z:
@markerikson any update on this issue? currently we wrap each api in a custom hook checking `isUninitialized` and forcing `data` undefined if `isUninitialized=true`...

Luk-z:
maybe this fix the problem https://github.com/reduxjs/redux-toolkit/pull/4896

markerikson:
Looks like we already had an issue for this, #3778 .  Working on merging the PR from that issue, #4758 .

PR #4896, opened 2025-03-17
try to solve #4009

Luk-z:


markerikson:
Should be handled by #4758 .

Issue #4909, opened 2025-03-27
Performance regression in `2.6.1` (after updating from `2.6.0`) when using `upsertQueryEntries`


ceisele-r:
When using `upsertQueryEntries` to upsert ~5000 entries, the performance massively degraded after the update from `2.6.0` to `2.6.1` making the browser hang completely for a few seconds where it was previously performing well.

![Image](https://github.com/user-attachments/assets/356f9f31-f465-4a0f-8591-67187b46c866)

I suspect #4872 to be the cause of the issue, though I have not traced it down completely to this specific commit.

But I can confirm the `upsertQueryEntries` is performing well in `2.6.0` while it is making the browser hang for ~5s after the update to `2.6.1` using the same code.

ceisele-r:
@markerikson yeah, please see the following:
**v2.6.0 with good performance:**
https://codesandbox.io/p/devbox/rtk-query-4909-forked-wfr6yd?workspaceId=ws_AYiY7CpdS8ZRZrgT98ctux

**v2.6.1 with bad performance:**
https://codesandbox.io/p/devbox/redux-with-redux-toolkit-2-forked-z7ng7l?workspaceId=ws_AYiY7CpdS8ZRZrgT98ctux

After clicking the button "Test", the query is started with upserting entries.
Once completed, it displays the loaded data.

As you can see in the second (v2.6.1) demo, the preview hangs and the upserts take way longer.

markerikson:
Ouch. Yeah.  Confirmed:

![Image](https://github.com/user-attachments/assets/d23b66e4-c678-42ee-9f2e-d55c80ec1915)

I think the issue is that internally this is reusing the existing logic for updating the `api.provided` slice of state, which maps `Tag > ID > QueryCacheKey[]`:

![Image](https://github.com/user-attachments/assets/b353576e-5515-4813-9cdb-b42747a6ee63)

That logic looks like:

```ts
updateProvidedBy: {
        reducer(
          draft,
          action: PayloadAction<{
            queryCacheKey: QueryCacheKey
            providedTags: readonly FullTagDescription<string>[]
          }>,
        ) {
          const { queryCacheKey, providedTags } = action.payload

          for (const tagTypeSubscriptions of Object.values(draft)) {
            for (const idSubscriptions of Object.values(tagTypeSubscriptions)) {
              const foundAt = idSubscriptions.indexOf(queryCacheKey)
              if (foundAt !== -1) {
[...]

PR #4910, opened 2025-03-27
Rewrite providedTags handling for better perf

markerikson:
Per #4909 , our recent update to add the missing `providesTags` handling for `upsertQueryEntries` in #4872 has resulted in a significant perf regression for that use case.

The current `api.provides` state structure looks like:

427705022-b353576e-5515-4813-9cdb-b42747a6ee63

When we update tags for a cache key, we first need to go in and remove that cache key from all tag entries that referenced it.  Unfortunately, the existing logic is doing that via a brute-force search over _all_ existing tag > cache key arrays.  Not only is this O(n), but it's doing so by going through an Immer draft proxy, which is slower:

```ts
      updateProvidedBy: {
        reducer(
          draft,
          action: PayloadAction<{
            queryCacheKey: QueryCacheKey
            providedTags: readonly FullTagDescription<string>[]
          }>,
        ) {
          const { queryCacheKey, providedTags } = action.payload

          for (const tagTypeSubscriptions of Object.values(draft)) {
            for (const idSubscriptions of Object.values(tagTypeSubscriptions)) {
              const foundAt = idSubscriptions.indexOf(queryCacheKey)
              if (foundAt !== -1) {
                idSubscriptions.splice(foundAt, 1)
              }
            }
          }

          for (const { type, id } of providedTags) {
            const subscribedQueries = ((draft[type] ??= {})[
              id || '__internal_without_id'
            ] ??= [])
            const alreadySubscribed = subscribedQueries.includes(queryCacheKey)
            if (!alreadySubscribed) {
              subscribedQueries.push(queryCacheKey)
            }
          }
        },
```

To make things worse, the `upsertQueryEntries` handling was doing this one cache key at a time. So, when we try to insert 5000 cache entries, we do the whole "painter's algorithm" anti-pattern and iterate through 1, 2, .... 4998, 4999 different tag > cache key arrays just to _try_ to clear out any existing uses of each cache key.

This PR restructures the invalidation state to be:

```ts
export type InvalidationState<TagTypes extends string> = {
  tags: {
    [_ in TagTypes]: {
      [id: string]: Array<QueryCacheKey>
      [id: number]: Array<QueryCacheKey>
    }
  }
  keys: Record<QueryCacheKey, Array<FullTagDescription<any>>>
}
```

From there, we just save the per-cache-key array of tags in a second lookup table, which then means that deletion simplifies down to grabbing that array of tags and removing the cache key from the corresponding per-tag cache key arrays.  So, instead of iterating over the entire list of all tags, we can just grab the few specific tags and target those directly.  (To toss _two_ CS-like bits of knowledge in here, we're "trading memory for speed".)

I also updated the `upsertQueryEntries` handling for tags to try to batch together all of the entries into one nested reducer call. Didn't appear to make much of a difference, but shouldn't hurt.

In local testing, this knocked the runtime for inserting 5000 entries from 23 _seconds_ down to 90 _milliseconds_.  (Yeah.  It was that bad.)

The downside is that this _does_ change the structure of `state.api.provided`.  We've always considered that internal, so this is not a _public_ breaking change and can thus be a patch release.

Unfortunately, the Redux DevTools RTKQ display _does_ expect the existing state structure, and I've confirmed that trying to view a cache entry with this build causes the entire DevTools to crash when it accesses an undefined field due to the structure change.

So, we're going to need to update the DevTools to handle both cases and get that release out, _then_ publish this fix.

The PR preview build here can be used as a workaround in the meantime.

### DIFF
--- a/packages/toolkit/src/query/core/apiState.ts
+++ b/packages/toolkit/src/query/core/apiState.ts
@@ -3,6 +3,7 @@ import type { BaseQueryError } from '../baseQueryTypes'
 import type {
   BaseEndpointDefinition,
   EndpointDefinitions,
+  FullTagDescription,
   InfiniteQueryDefinition,
   MutationDefinition,
   PageParamFrom,
@@ -15,14 +16,8 @@ import type { Id, WithRequiredProp } from '../tsHelpers'
 export type QueryCacheKey = string & { _type: 'queryCacheKey' }
 export type QuerySubstateIdentifier = { queryCacheKey: QueryCacheKey }
 export type MutationSubstateIdentifier =
-  | {
-      requestId: string
-      fixedCacheKey?: string
-    }
-  | {
-      requestId?: string
-      fixedCacheKey: string
-    }
+  | { requestId: string; fixedCacheKey?: string }
+  | { requestId?: string; fixedCacheKey: string }
 
 export type RefetchConfigOptions = {
   refetchOnMountOrArgChange: boolean | number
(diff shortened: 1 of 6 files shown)


markerikson:
https://github.com/reduxjs/redux-devtools/releases/tag/remotedev-redux-devtools-extensions%403.2.10 is out now, so we can land this.

PR #4926, opened 2025-04-06
add docs for RTKQ schema features

EskiMojo14:

PR #4934, opened 2025-04-08
Add `catchSchemaFailure`, and docs for RTKQ schema features

EskiMojo14:
replaces #4926

### DIFF
--- a/docs/rtk-query/api/createApi.mdx
+++ b/docs/rtk-query/api/createApi.mdx
@@ -221,6 +221,20 @@ export type QueryDefinition<
       updateCachedData, // available for query endpoints only
     }: QueryCacheLifecycleApi,
   ): Promise<void>
+
+  argSchema?: StandardSchemaV1<QueryArg>
+
+  /* only available with `query`, not `queryFn` */
+  rawResponseSchema?: StandardSchemaV1<BaseQueryResult<BaseQuery>>
+
+  responseSchema?: StandardSchemaV1<ResultType>
+
+  /* only available with `query`, not `queryFn` */
+  rawErrorResponseSchema?: StandardSchemaV1<BaseQueryError<BaseQuery>>
+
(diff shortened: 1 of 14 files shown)


agusterodin:
This addresses the [concerns I mentioned in the original standard schema MR](https://github.com/reduxjs/redux-toolkit/pull/4864#issuecomment-2785161278) for catching errors and providing the ability to log the "parse issue" errors to the console (I didn't realize `error.issues` was a thing) 

[I looked in Zod's source code](https://github.com/copilot/share/0843500e-0a80-88e3-a140-b20124894034) to see what formatting of `ZodError.toString()` is. `util.jsonStringifyReplacer` [appears to just format bigint values to string when present](https://github.com/colinhacks/zod/blob/859ad2495e3d24559d3c8e33a00cf56e2dbaa16a/src/helpers/util.ts#L91). I don't use bigint values, so I left the `replacer` function as `null` instead.

With the code snippet below, I no longer see "uncaught error" in console and am able to see all "issues" formatted verbatim the way i'm used to when directly using Zod 👍 

```
catchSchemaFailure: error => {
    console.error('ZodError:', JSON.stringify(error.issues, null, 2))
    return {
      status: 'CUSTOM_ERROR',
      error: error.schemaName + ' failed validation',
      data: error.issues
    } as FetchBaseQueryError
  }
 ```

First screenshot is my current Zod integration output, second screenshot is RTKQ test PR standard schema validation output

Screenshot 2025-04-08 at 3 13 33 PM

Screenshot 2025-04-08 at 3 14 50 PM

EskiMojo14:
re: type error, try adding `as const` to the status. it looks like Typescript isn't inferring the desired result type from context, which is a bit annoying. (alternatively use `satisfies` instead of `as` for better type safety)

i think it makes sense for onSchemaFailure to still be called, so I'll restructure that soon.

PR #4944, opened 2025-04-15
Chakra UI Updated from v1.0.0 to v2.10.7 to support after react18^ 

sriram-palanisamy-hat:
This is my first PR in Redux , so please let me know if I missed anything!

🛠 Fix: Chakra UI version upgrade for React 19 compatibility in examples

Chakra UI v1 only supports React up to v17. Since the project now uses React 19, some examples using Chakra UI v1 no longer work correctly in CodeSandbox 

To fix this, I’ve updated the Chakra UI version from v1 to v2 in the affected examples to ensure they work with React 19.

Here are a couple of examples where the issue was observed:

[Authentication Example](https://redux-toolkit.js.org/rtk-query/usage/examples#dispatching-an-action-to-set-the-user-state)

[Prefetching Example](https://redux-toolkit.js.org/rtk-query/usage/prefetching#prefetching-all-known-pages)

Let me know if anything else needs to be updated. Happy to make further changes!

### DIFF
--- a/examples/query/react/authentication/src/features/auth/Login.tsx
+++ b/examples/query/react/authentication/src/features/auth/Login.tsx
@@ -80,7 +80,7 @@ export const Login = () => {
           <PasswordInput onChange={handleChange} name="password" />
         </InputGroup>
         <Button
-          width={'100%'}
+          width="100%"
           onClick={async () => {
             try {
               const user = await login(formState).unwrap()


100pearlcent:
I have found pagination got the same issue as mentioned below.
- https://github.com/reduxjs/redux-toolkit/issues/4954

abishek-umasankar-hat:
LGTM

aryaemami59 (approved):

timdorr (approved):

aryaemami59:
@sriram-palanisamy-hat Thanks!

PR #4950, opened 2025-04-20
Use Canary for local search instead of Algolia

markerikson:
We've had some issues with our docs Algolia integrations over time, and in particular the React-Redux docs search isn't returning _any_ results right now.

Trying out Canary with Pagefind (which generates a "local" index on build), and it seems to work pretty well.  For comparison:

- Algolia (live):

image

- Canary + Pagefind (PR preview):

image

So not only does Canary + Pagefind seem to at least work and produce results, I think the display of search results with in-page context is actually an improvement.

### DIFF
--- a/website/docusaurus.config.ts
+++ b/website/docusaurus.config.ts
@@ -70,6 +70,7 @@ const config: Config = {
   customFields: {
     repoUrl: 'https://github.com/reduxjs/redux-toolkit',
   },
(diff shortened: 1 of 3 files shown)


ryota-murakami:
@markerikson Thanks update for this!
Now I feel RTK Docs working better than before, that shows important paragraph written at deep place of sitemap. 😃

Issue #4976, opened 2025-05-08
Expose codegen CLI from the package exports


markerikson:
> @markerikson we are calling the cli programmatically like this
> 
> ```ts
> const cliPath = require.resolve('@rtk-query/codegen-openapi/lib/bin/cli');
> 
> const runReduxCodegen = async (specUrl: string | undefined, configPath: string, outputPath: string): Promise<void> => (
>   new Promise((resolve, reject) => {
>     exec(
>       `tsx ${cliPath} ${configPath}`,
>       {
>         env: {
>           ...process.env,
>           SPEC_URL: specUrl,
>           STORE_ROOT_PATH: outputPath,
>         },
>       },
>       (error, stdout, stderr) => {
>         if (error) {
>           console.error(`exec error: ${error}`);
>           reject(new Error(`Execution failed with error: ${stderr || 'Unknown error'}`));
>         } else {
>           console.log(stdout);
> 
>           if (stderr) {
>             reject(new Error(`Execution produced errors: ${stderr}`));
>           } else {
>             resolve();
>           }
>         }
>       },
>     );
>   })
> );
> ```
> 
> when upgrading from 1.2.0 to 2.0.0, the error message is that /lib/bin/cli is not in exports of the package
> 
> nodeLinker is pnp 

 _Originally posted by @maddrag0n in [#4974](https://github.com/reduxjs/redux-toolkit/issues/4974#issuecomment-2863414379)_

PR #4977, opened 2025-05-09
feature: add cli to exports

maddrag0n:
closes https://github.com/reduxjs/redux-toolkit/issues/4976

exposes the cli in exports to be able to programmatically resolve the path at runtime and updates docs.

ie. *bin/openapi-codegen.ts*
```typescript
import { exec } from 'node:child_process'

const cliPath = require.resolve('@rtk-query/codegen-openapi/cli')

// you can also use esbuild-runner (esr) or ts-node instead of tsx
exec(`tsx ${cliPath} config.ts`)
```

### DIFF
--- a/.github/workflows/test-codegen.yml
+++ b/.github/workflows/test-codegen.yml
@@ -147,4 +147,4 @@ jobs:
           name: package
 
       - name: Run are-the-types-wrong
-        run: yarn dlx @arethetypeswrong/cli@latest ./package.tgz --format table
+        run: yarn dlx @arethetypeswrong/cli@latest ./package.tgz --format table --exclude-entrypoints cli
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,7 @@ typesversions
 *.tgz
 
 tsconfig.vitest-temp.json
+
+# node version manager files
+.node-version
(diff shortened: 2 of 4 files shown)


maddrag0n:
I had to expose both ESM and CJS entry points and types, because of the CI check failing otherwise. This doesn't really make any sense for the purpose of exposing the CLI path tho 🤷🏽.

maddrag0n:
Unfortunately the CI expects every export to be an ES/CJS module with types and everything. the cli itself does not export anything and no types are generated

markerikson:
If that's an issue, then per ATTW's docs we ought to be able to specify specific entry points to include or exclude:

- https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/packages/cli/README.md#entrypoints

maddrag0n:
@markerikson lmk if you still intend to merge this PR. otherwise feel free to close.

markerikson:
@maddrag0n Yeah, I'd like to get to it, but haven't had time or mental energy to do maintenance work lately.  Still on my list :)

markerikson:
Okay, yeah, seems reasonable. Thanks for your patience here!

PR #4996, opened 2025-05-25
Remove Serialisation from useStableQueryArgs

### LINKED ISSUE #4995: useStableQueryArgs + defaultSerializeQueryArgs + serializeQueryArgs - avoiding attempts to serialize huge parameters
I've got a use case where a query takes a bunch of data as input. Like a lot. So since I don't want to end up with a cache key that's huge, I implemented serializeQueryArgs and provided a suitable replacement that uniquely represents the data in a manageable size string identifier with no collisions, perfect for a cache key. Imagine my surprise when i'm seeing long running attempts at memoizing a json the size of Texas and when i trace it I find this:

```
      const stableArg = useStableQueryArgs(
        options.skip ? skipToken : arg,
        // Even if the user provided a per-endpoint `serializeQueryArgs` with
        // a consistent return value, _here_ we want to use the default behavior
        // so we can tell if _anything_ actually changed. Otherwise, we can end up
        // with a case where the query args did change but the serialization doesn't,
        // and then we never try to initiate a refetch.
        defaultSerializeQueryArgs,
        context.endpointDefinitions[endpointName],
        endpointName,
      )

```
The comment feels like it's treating the user as dumb. Is there a way to avoid this?
In short: I have a parameter that i need in queryFn and I don't want the library to ever try to serialize it.

riqts:
Fixes #4995 

I just added this code that @phryneas suggested in the discourse there, incase we wanted to run with that suggestion.
Can also read the thread there, but @EskiMojo14 raised a concern about large structures so I'll just leave this PR here incase we choose to go with it. Open to other suggestions I can spin up a PR for as well though.

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -39,11 +39,7 @@ import type {
   TSHelpersNoInfer,
   TSHelpersOverride,
 } from '@reduxjs/toolkit/query'
-import {
-  defaultSerializeQueryArgs,
-  QueryStatus,
-  skipToken,
-} from '@reduxjs/toolkit/query'
+import { QueryStatus, skipToken } from '@reduxjs/toolkit/query'
 import type { DependencyList } from 'react'
 import {
   useCallback,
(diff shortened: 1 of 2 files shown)


riqts:
> I'm not sure if I'm encouraged or disturbed that all the tests seem to have passed :)

Exactly the thought I had 😅 😅 😅 
Might take a closer look at the tests just to make sure

Issue #5014, opened 2025-06-18
The inferred type of 'X-Slice' cannot be named without a reference to '@reduxjs/toolkit/node_modules/immer/dist/internal'. This is likely not portable. A type annotation is necessary.


marinsokol5:
Hey folks, 

Opening as a duplicate of https://github.com/reduxjs/redux-toolkit/issues/1806 since I can't comment on that thread. 

It's indeed a small type issue, and I am opening it just because of the high bar for your existing types. If this was another library, I would not even bother.

But effectively, I agree with https://github.com/reduxjs/redux-toolkit/issues/1806#issuecomment-1925437633, and v2.2.7 does not solve the problem. 

[Link to TS playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgYygUwIYzQZQDbDJpwC+cAZlBCHAOQAC6AJgK4AeAVgM4D0MEEPAGtgMWgChQkWHBgBPMMSQAFDHLwQMTAILIYwCADtSFKjQbN23PgOGiJkw9ijkMROAGEILJ2ig4YLCVxODgANww8FjQALjhDFhAAIz9xEnFxZCMueGBDUWBIgKC4AF5EcMjouIAGEy4sYC5yYDQuT29ff0DsOAx2rx9nYuwMtDZpeCzDHJRO4YJ3ctRMbHxCNAAKBBD4jBBYuiyhv1oAGl28gqKetAvQyyIoLjid0NC8lYOnTZyggEpELt3nA-tgAHQRKJoADUMOBpHu7yYaC+aB+YLQgLeINBt0hVTQAFoiQiSEiPoY0U4AEJybQgea-W5nPp6AyGOKqdSaHTsowAHgSyT8AD5sQjQpiCdC4DDym59EZwWA1BotGSkeS0v8gA)

Copying the example code from https://redux-toolkit.js.org/api/createslice:
```ts
import { createSlice } from '@reduxjs/toolkit'
import type { PayloadAction } from '@reduxjs/toolkit'

interface CounterState {
  value: number
}

const initialState = { value: 0 } satisfies CounterState as CounterState

export const counterSlice = createSlice({
  name: 'counter',
  initialState,
  reducers: {
    increment(state) {
      state.value++
    },
    decrement(state) {
      state.value--
    },
    incrementByAmount(state, action: PayloadAction<number>) {
      state.value += action.payload
    },
  },
})
```

[...]

PR #5015, opened 2025-06-18
fix(types): re-exporting WritableDraft from immer

marinsokol5:
Context is in https://github.com/reduxjs/redux-toolkit/issues/5014


izhan:
Thanks for opening this PR – we are also hitting this issue. Is there any way we can get this merged?

Issue #4920, opened 2025-04-03
Infinite query always refetch all pages


Hos1g4k1:
Currently, I am implementing infinite scroll feature using your infinite query and I successfully integrated it. But, there is one issue. The desired behaviour is to refetch only the first page when user initiates the refetch action. This doesn't happen and all pages are refetched.

This is the query:

```
import { queryApi } from 'config/rtkQuery';
import { CatchesFeedItem } from './types';

const url = 'public_trips';

type GetCatchesFeedArgumentsType = {
  page?: number;
  per_page?: number;
};

type MetaType = {
  totalCount: number;
  pageCount: number;
  currentPage: number;
  perPage: number;
};

const catchesFeedApi = queryApi.injectEndpoints({
  endpoints: (build) => ({
    getCatchesFeed: build.infiniteQuery<
      { data: CatchesFeedItem[]; meta: MetaType },
      GetCatchesFeedArgumentsType,
      number
    >({
      infiniteQueryOptions: {
        initialPageParam: 1,
        getNextPageParam: (lastPage, allPages, lastPageParam) => {
          return lastPageParam + 1;
        },
      },
      query: ({ pageParam }) => ({
        url: url,
        method: 'get',
        params: {
          include: 'meta',
          page: pageParam || 1,
          per_page: 10,
        },
      }),
      transformResponse: (response: CatchesFeedItem[], meta: MetaType) => {
        return { data: response, meta: meta };
      },
      providesTags: ['CatchesFeed'],
    }),
  }),
  overrideExisting: true,
});

export const { useGetCatchesFeedInfiniteQuery } = catchesFeedApi;
```

In the documentation there is this sentence 

> If the cache entry is ever removed and then re-added, it will start with only fetching the initial page.

but I don't know what it means exactly and how to force this behaviour. Can you help me with this? I tried to invalidate the provided that and that doesn't work. Also, when I tried to reset the whole app state it worked but that's the overkill because I don't want to reset the whole state, just this small fraction.

markerikson:
Yes, refetching all pages is an intentional design decision:

- https://redux-toolkit.js.org/rtk-query/usage/infinite-queries#refetching

This behavior matches React Query, and we based our implementation's behavior on React Query.

The "if the cache entry is removed" line describes what happens if no more components subscribe to that data and the cache entry is garbage-collected after timeout.

[...]

PR #5016, opened 2025-06-18
Add refetchCachedPages option for infinite query

rossmartin:
This `refetchCachedPages` option allows an infinite query to NOT refetch all pages in the cache when a refetch happens. By default it is `true` and uses the existing behavior.

This is based on what was mentioned in the discussion here - https://github.com/reduxjs/redux-toolkit/discussions/5011. It's also based on this resolved issue here - https://github.com/reduxjs/redux-toolkit/issues/4920

We want the ability to opt out of this behavior at this level. We don't want to have to reset the `pages`/`pageParams` for a query when args change or during a refetch.

Here is an example of how it can be used:

```ts
    getPosts: builder.infiniteQuery<GetPostsResponse, GetPostsRequest, string | undefined>({
      infiniteQueryOptions: {
        // ...
        refetchCachedPages: false,
      },
      // ...
    }),  
```

Screenshot 2025-06-18 at 5 03 08 PM

### DIFF
--- a/docs/rtk-query/api/createApi.mdx
+++ b/docs/rtk-query/api/createApi.mdx
@@ -329,6 +329,13 @@ export type InfiniteQueryDefinition<
        * direction will be dropped from the cache.
        */
       maxPages?: number
+      /**
+       * Defaults to `true`. When this is `true` and an infinite query endpoint is refetched
+       * (due to tag invalidation, polling, arg change configuration, or manual refetching),
+       * RTK Query will try to sequentially refetch all pages currently in the cache.
+       * When `false` only the first page will be refetched.
+       */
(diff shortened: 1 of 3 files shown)


JacobJaffe:
Any update on this PR? This behavior is super helpful-- I've been patching this onto my projects (and can continue to, but it's tedious to build a new patch for each rtk version update), and refreshing has been working great with it.

markerikson:
I'll try to find some time to review it in the next few days. Good to know that A) it's useful, and B) the approach appears to be working - thanks for that feedback!

markerikson:
Out in https://github.com/reduxjs/redux-toolkit/releases/tag/v2.11.0 ! Thanks!

PR #5038, opened 2025-07-25
fix: allow setting `CaseReducerWithPrepare` action types

EnixCoda:
Hi, I noticed a type error when implementing a function that returns a `CaseReducerWithPrepare` and found a fix for it. Hope it looks good to you! I'll attach the original error I met in the file changes comment.

### DIFF
--- a/packages/toolkit/src/tests/createSlice.test-d.ts
+++ b/packages/toolkit/src/tests/createSlice.test-d.ts
@@ -8,6 +8,7 @@ import type {
   ActionReducerMapBuilder,
   AsyncThunk,
   CaseReducer,
+  CaseReducerWithPrepare,
   PayloadAction,
   PayloadActionCreator,
   Reducer,
@@ -15,7 +16,6 @@ import type {
   SerializedError,
   SliceCaseReducers,
   ThunkDispatch,
-  UnknownAction,
   ValidateSliceCaseReducers,
 } from '@reduxjs/toolkit'
 import {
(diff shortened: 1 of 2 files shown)


packages/toolkit/src/tests/createSlice.test-d.ts
```diff
@@ -992,4 +992,22 @@ describe('type tests', () => {
     })
     buildCreateSlice({ creators: { asyncThunk: asyncThunkCreator } })
   })
+
+  test('wrapping create reducer creators should be possible', () => {
+    const highOrderCreateCaseReducerWithPrepare = <S>(
+      creator: ReducerCreators<S>,
+    ): CaseReducerWithPrepare<S, PayloadAction<S, string, any, any>> =>
```
EnixCoda, line 999: Here used to have such error:

```
Type 'PayloadAction<S, string, any, any>' does not satisfy the constraint '{ payload: void; type: string; }'.
  Types of property 'payload' are incompatible.
    Type 'S' is not assignable to type 'void'.ts(2344)
createSlice.test-d.ts(997, 52): This type parameter might need an `extends void` constraint.
```

EnixCoda: BTW, ideally, I'd like to write `CaseReducerWithPrepare<S, PayloadAction<S, string>>` (no any) instead of `CaseReducerWithPrepare<S, PayloadAction<S, string, any, any>>` here.

EskiMojo14: Why annotate the return value here? Presumably it should infer from create.preparedReducer (which also actually doesn't use CaseReducerWithPrepare, it's more strongly typed)

EnixCoda: @EskiMojo14 Thanks, I did not try to `infer` because personally I prefer declaring return types and it work well in most cases.

This time I give `infer` a try (simply remove the return type) and it works.

I'm good with closing this PR if annotating return type not work was expected in this use case. :)

PR #5062, opened 2025-08-29
fix: incorrect position of keepUnusedDataFor for two scenarios

mattchannn:
As we can put keepUnusedDataFor in both `createApi` and `endpoints` level hence we need to correct examples for each of the use case.

Due to the [constraint from remark-typescript-tools](https://github.com/phryneas/remark-typescript-tools/issues/4), I am not able to have nested identifier in a Typescript type hence I put the example on the outermost level which I think no harm


markerikson:
Thanks!

PR #5064, opened 2025-09-01
Rewrite subscription handling and polling calculations for better perf

### LINKED ISSUE #5052: Performance issue with many subscriptions via `useQuery` vs. `useQueryState`
Hey team!

I've been diagnosing some performance issues in an application and I came across a somewhat unexpected culprit. We're using RTK Query for API data fetching & caching, as well as some global UI state management that is not related to API data.

I found that the auto-generated `useQuery` hooks add some additional overhead that seems related to subscription tracking. On the page in question, there were only 10-12 endpoints being used but some of the endpoints had 1000-4000 subscriptions (found by logging out the endpoint name + query args + keys in the `state.api.subscriptions` object)

We load most of this data when the page is rendered and don't often refetch it from sub-components. As an experiment, I swapped some of the generated `useGetXQuery` hook calls with the `useQueryState` hook that is _not_ auto-generated, i.e. `api.endpoints.getUser.useQueryState`, and found that the performance of mounting/unmounting components that called these hooks was much faster.

I wanted to raise the issue to see if there's any obvious reason for this extra overhead. All of the documentation seems to point to just using the generated query hooks and not reaching for the `useQueryState` / `useQuerySubscription` hooks available on each endpoint definition. 

I've put together a minimal reproduction repo to showcase this issue. The `UseQueryChild` and `UseQueryStateChild` components are intentionally calling the respective hook multiple times for the sake of simulating how many consuming components our real application might have at one time. You'll notice that the conditional render of the `UseQueryStateChild` component is _much_ faster than the `UseQueryChild`.

From my testing, it seems like `useQueryState` will still notify consuming components if the underlying data changes (i.e. something triggers an update to the cache), so is there any benefit to using `useQuery` over `useQueryState` if we prefetch the data at the top of the page tree and know that it will be cached by the time the consuming components use it (i.e. no additional API request needs to be fired)?

Reproduction repo: https://github.com/brandongregoryscott/rtk-query-subscription-performance

Reproduction demo:

https://github.com/user-attachments/assets/2371dfb7-5d81-496b-9a6f-008a017ca199

markerikson:
This PR significantly rewrites our existing internal RTKQ middleware subscription tracking and polling trigger systems to greatly improve perf, especially in the many-subscriptions-one-cache-entry case.

Fixes #5052 

## Background

This rewrite was the result of working on #5061 to abort pending requests when cache entries are removed, and #5052 that demonstrated slow perf with thousands of subscriptions to the same cache entry.

### Polling

In profiling the many-subscriptions case, I saw that `updatePollingInterval` was running after every single additional subscription (which we signify internally as a `/rejected` action with `condition: true`.  That's because any new subscription _could_ have included subscription options for a new polling value, and we need to find the lowest polling interval.

However, there's no reason to be doing that recalculation synchronously every time.    We don't even kick off the actual polling request immediately - we set a timer.  So, instead it's better to debounce this and do a single `updatePollingInterval` calculation on the next tick.  

Additionally, we were using plain JS `Record`s as the lookup table for polling items, and iterating keys in those objects to get the polling value entries.  Switching those to be `Map`s isn't worse and probably helps.

The only way I could figure out to test this was to hack in a test-only polling update counter and exfiltrate that via the middleware.  Stupidly ugly but worked.

### Subscriptions

Originally, our internal subscription system used plain object `Record`s as the key/value lookups for subscriptions by request ID, as a holdover from some of the earlier implementations. However, that meant that any time we needed to check for size or "is there at least 1 subscription", we had to do a `for...in` check to iterate keys and count them.  Based on perf profiling, this turned out to be relatively expensive.

This was made trickier because we added a pseudo-subscription for `${requestId}_running` in #3709 when we added cache entries for non-subscribed initiated queries, and needed to prevent cache collection of those.

When I reworked the logic in `cacheCollection.ts` in #3709 to add abort handling, I had to special-case the `_running` situation as we iterated keys.

This can all be avoided if we just move the `currentQueries` map from the `buildInitiate()` closure into the middleware `InternalState` object and make it available to all the middleware, then use that to check if there's a running query for this cache key.

Additionally, changing all of the subscription lookups from a `Record` to a `Map` lets us check `subscriptions.size` without any key iterating.  This did require touching all the places we accessed the subscriptions accordingly (including making sure we switched from `Object.keys(subscriptions)` to `subscriptions.keys()`).

## Changes

- Added a test to verify the existing `subscriptionsUpdated` serialization behavior
- Rewrote the subscription system:
  - Moved the middleware `InternalState` object to be initialized in `module.ts` so it can be passed to both `buildInitiate()` and `buildMiddleware()`
  - Moved the `runningQueries/runningMutations` maps to be in `InternalState`
  - Changed `subscriptions` to be nested `Map`s instead of `Records`, including all accesses to the subscription data
  - Changed the `subscriptionsUpdated` serialization to stringify the maps accordingly so we still get the right plain JS objects for use in the reducer
  - Removed the `${requestId}_running` handling and replaced it with a direct check against `runningQueries` for that request ID
- Rewrote the polling system:
  - Changed it to also use a `Map` instead of a `Record`
  - Debounced the polling updates with a `setTimeout(0)` so we only do one set of updates per tick
  - Added a hacky polling update counter in test environments
  - Restructured a couple tests to fix some shared state leakage I was seeing

## Results

Using the thousands-of-subscribed-components example from #5052 , and just unchecking and re-checking the "Render children" box to unmount and remount the components:

Perf before, dev:

image

Perf after, dev:

image

You can see we've eliminated the major perf bottlenecks in `polling.ts` and `cacheCollection.ts`, and we're left with just the React dev effect overhead.

Perf before, prod:

image

Perf after, prod

image

Similarly, total execution time dropped significantly and RTKQ is no longer at the top of the list.

### DIFF
--- a/.github/workflows/test-codegen.yml
+++ b/.github/workflows/test-codegen.yml
@@ -147,4 +147,4 @@ jobs:
           name: package
 
       - name: Run are-the-types-wrong
-        run: yarn dlx @arethetypeswrong/cli@latest ./package.tgz --format table
+        run: yarn dlx @arethetypeswrong/cli@latest ./package.tgz --format table --exclude-entrypoints cli
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -305,7 +305,7 @@ jobs:
       fail-fast: false
       matrix:
         node: ['22.x']
-        ts: ['5.3', '5.4', '5.5', '5.6', '5.7', '5.8', 'next']
+        ts: ['5.3', '5.4', '5.5', '5.6', '5.7', '5.8', '5.9', 'next']
         example:
           [
             { name: 'bundler', moduleResolution: 'Bundler' },
@@ -340,10 +340,11 @@ jobs:
         run: yarn workspace @examples-type-portability/${{ matrix.example.name }} run test
 
       - name: Test type portability with `moduleResolution Node10`
+        if: matrix.ts != 'next' && !startsWith(matrix.ts, '6.')
         run: yarn workspace @examples-type-portability/${{ matrix.example.name }} run test --module CommonJS --moduleResolution Node10 --preserveSymLinks --verbatimModuleSyntax false
 
       - name: Test type portability with `moduleResolution Node10` and `type module` in `package.json`
-        if: matrix.example.name == 'nodenext-esm' || matrix.example.name == 'bundler'
+        if: (matrix.example.name == 'nodenext-esm' || matrix.example.name == 'bundler') && matrix.ts != 'next' && !startsWith(matrix.ts, '6.')
         run: |
           npm --workspace=@examples-type-portability/${{ matrix.example.name }} pkg set type=module
           yarn workspace @examples-type-portability/${{ matrix.example.name }} run test --module ESNext --moduleResolution Node10 --preserveSymLinks --verbatimModuleSyntax false
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,7 @@ typesversions
 *.tgz
 
 tsconfig.vitest-temp.json
+
+# node version manager files
+.node-version
+.nvmrc
--- a/docs/rtk-query/usage/code-generation.mdx
+++ b/docs/rtk-query/usage/code-generation.mdx
@@ -84,6 +84,17 @@ const api = await generateEndpoints({
 })
 ```
 
+#### With Node.js Child process
+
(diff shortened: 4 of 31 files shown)


markerikson:
Tests are passing, we're just seeing PR failures due to `TS next` deprecating the `node10` module resolution algorithm

PR #5102, opened 2025-10-09
Allow executing effects that have become unsubscribed to be canceled by `listenerMiddleware.clearListeners`

chris-chambers:
Tasks created following the polling example given in [complex async workflows](https://redux-toolkit.js.org/api/createListenerMiddleware#complex-async-workflows) can't be canceled by `listenerMiddleware.clearListeners`.  Any time a listener is unsubscribed before it has finished running, this condition can occur.

This PR updates listener middlewares to explicitly track executing listeners (which may or may not be unsubscribed), so that cancelation can happen in all cases.

---

Here's a small reproduction, based on the example from the docs (there are a couple more scenarios in the accompanying tests):
```ts
const eventPollingStarted = createAction('serverPolling/started')

listenerMiddleware.startListening({
  actionCreator: eventPollingStarted,
  effect: async (action, listenerApi) => {
    // Stop listening for this action
    listenerApi.unsubscribe()

    const pollingTask = listenerApi.fork(async (forkApi) => {
      try {
        // Cancellation-aware indefinite pause
        await forkApi.pause(new Promise(() => {}))
      } catch (err) {
        if (err instanceof TaskAbortError) {
          // Never runs
          console.log('pollingTask aborted')
        }
      }
    })

    try {
      // Wait indefinitely
      await listenerApi.condition(() => false)
      pollingTask.cancel()
    } catch (err) {
      if (err instanceof TaskAbortError) {
        // Never runs
        console.log('top-level task aborted')
      }
    }
  },
})

store.dispatch(eventPollingStarted())

// Some time later, while the tasks are still running, call clearListeners.
setTimeout(() => {
  listenerMiddleware.clearListeners()
}, 500);
```

### DIFF
--- a/packages/toolkit/src/listenerMiddleware/index.ts
+++ b/packages/toolkit/src/listenerMiddleware/index.ts
@@ -262,10 +262,12 @@ const cancelActiveListeners = (
 
 const createClearListenerMiddleware = (
   listenerMap: Map<string, ListenerEntry>,
+  executingListeners: Map<ListenerEntry, number>,
 ) => {
   return () => {
-    listenerMap.forEach(cancelActiveListeners)
-
+    for (const listener of executingListeners.keys()) {
+      cancelActiveListeners(listener);
+    }
     listenerMap.clear()
   }
 }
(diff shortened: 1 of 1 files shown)


markerikson:
Looks good now, thanks!

markerikson:
Thanks, out in https://github.com/reduxjs/redux-toolkit/releases/tag/v2.9.1 !

PR #5124, opened 2025-10-31
fix: compare paths against Regex instances in ignoredPaths

### LINKED ISSUE #5123: Support regex ignorePaths in immutability check
This is somewhat related to https://github.com/reduxjs/redux-toolkit/issues/2553

I have an issue where a path I want to ignore in `immutableCheck` is somewhat dynamic, so I'd like to use a Regex to match the paths. 

However, this doesn't fully work, as `trackProperties` treats `ignorePaths` as if they are all string instances https://github.com/reduxjs/redux-toolkit/blob/6600116b407b86e18a3f8cecfa620788d44999aa/packages/toolkit/src/immutableStateInvariantMiddleware.ts#L49

johnste:
Fixes an issue where Regex instances in ignoredPaths weren't being properly evaluated.

Also updates some naming to better match the conventions in `detectMutations`

Fixes #5123

### DIFF
--- a/packages/toolkit/src/immutableStateInvariantMiddleware.ts
+++ b/packages/toolkit/src/immutableStateInvariantMiddleware.ts
@@ -15,13 +15,13 @@ export function isImmutableDefault(value: unknown): boolean {
 
 export function trackForMutations(
   isImmutable: IsImmutableFunc,
-  ignorePaths: IgnorePaths | undefined,
+  ignoredPaths: IgnorePaths | undefined,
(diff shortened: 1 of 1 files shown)


markerikson:
I'm a bit concerned our existing tests didn't catch this mismatch, but yeah, seems good. Thanks!

markerikson:
Out in https://github.com/reduxjs/redux-toolkit/releases/tag/v2.11.0 !

PR #5129, opened 2025-11-03
Byte-shave various chunks of code

markerikson:
This PR byte-shaves a variety of repeated bits of code I saw:

- Rewrote `setupListeners` to reuse event names and deduplicate `window.addEventListener`
- Found out that ESBuild auto-inlines TS enums, so that our checks like `status === QueryStatus.initialized` were getting rewritten to ` === "initialized"` every time.  The TS enum got runtime-compiled, but was unused.  Added equivalent string constants, so the enum is now used for type-checks, but the constants are used for runtime comparisons.
  - ESBuild ref: https://github.com/evanw/esbuild/releases/tag/v0.14.7
- Consolidated all third-party imports into reusable modules, as ESBuild does not deduplicate imports when it bundles, leading to multiple imports from the same package and renamed imports.  I had done this previously for RTK into RTKQ, but now done for other modules too.
- Added some wrappers for common lookup patterns (`context.endpointDefinitions[endpointName]`, some action meta fields, promise ref unsubscribes)

This looks like it knocks about 600 bytes minified off of the RTKQ core artifact, and another 200 bytes off of the RTKQ React artifact:

image

### DIFF
--- a/packages/toolkit/.size-limit.cjs
+++ b/packages/toolkit/.size-limit.cjs
@@ -1,8 +1,11 @@
 const webpack = require('webpack')
 let { join } = require('path')
 
-const esmSuffixes = ['modern.mjs' /*, 'browser.mjs', 'legacy-esm.js'*/]
-const cjsSuffixes = [/*'development.cjs',*/ 'production.min.cjs']
+const esmSuffixes = ['modern.mjs', 'browser.mjs' /*, 'legacy-esm.js'*/]
+const cjsSuffixes = [
+  /*'development.cjs',*/
+  /*'production.min.cjs'*/
+]
 
 function withRtkPath(suffix, cjs = false) {
   /**


markerikson:
Eh, messing with `size-limit` borked the comment comparisons for this PR. Ah well. I can see it's an improvement anyway :)

Confirmed using some Sonda analysis in an actual Vite app example that this knocks about 0.6K off the final app size.

packages/toolkit/src/query/core/setupListeners.ts
```diff
@@ -52,33 +78,42 @@ export function setupListeners(
...
     if (!initialized) {
-      if (typeof window !== 'undefined' && window.addEventListener) {
-        // Handle focus events
-        window.addEventListener(
-          'visibilitychange',
-          handleVisibilityChange,
-          false,
-        )
-        window.addEventListener('focus', handleFocus, false)
+      const WINDOW = window
```
glepretre: We lost the SSR support here in v2.10.0

```
UnhandledRejection ReferenceError: window is not defined
    at defaultHandler (file:///node_modules/@reduxjs/toolkit/dist/query/rtk-query.modern.mjs:417:22
```

PR #5135, opened 2025-11-13
feat(codegen): support explicit tag overrides without altering defaults

w3di:
## Summary
- avoid emitting empty `providesTags`/`invalidatesTags` when tag generation is disabled
- allow explicit `providesTags` and `invalidatesTags` overrides regardless of the inferred endpoint type
- add regression tests covering the new override behavior

### DIFF
--- a/packages/rtk-query-codegen-openapi/test/generateEndpoints.test.ts
+++ b/packages/rtk-query-codegen-openapi/test/generateEndpoints.test.ts
@@ -264,6 +264,26 @@ describe('endpoint overrides', () => {
     expect(api).not.toMatch(/providesTags: \["pet"\]/);
     expect(api).not.toMatch(/invalidatesTags: \["pet"\]/);
   });
+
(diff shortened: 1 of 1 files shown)


w3di:
I ran into this while wiring RTK Query codegen into my project whenever I wanted to tweak caching for just a couple of operations there was no good way to do it. The only lever was the global tag flag: turn it on and every endpoint gets the default tags; turn it off and I completely lost the ability to customise anything. The only workaround was to patch the generated files by hand, which is brittle and doesn’t scale.

This change removes that pain point. I can now declare per-endpoint providesTags and invalidatesTags right in the config, even if automatic tag generation is disabled. The generator keeps those overrides intact, so I no longer need any post-processing scripts. I also added tests to make sure the new behaviour is stable and doesn’t regress the existing output.

markerikson:
Out in https://github.com/reduxjs/redux-toolkit/releases/tag/%40rtk-query%2Fcodegen-openapi%402.2.0 !

PR #5141, opened 2025-11-23
Update Immer to v11 for better update performance

markerikson:
This PR:

- Bumps our Immer dep to https://github.com/immerjs/immer/releases/tag/v11.0.0 to pick up the architectural performance rewrites I've done there

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,8 @@ tsconfig.vitest-temp.json
 # node version manager files
 .node-version
 .nvmrc
+
+# Assorted local docs
+docs/dev-plans
+# Perf protobufs
+*.pb.gz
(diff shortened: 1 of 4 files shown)


antoinetissier:
@markerikson thank you for the great work on this (impressive talk at the React Paris conf !).
Do you know if redux-toolkit v1 could also use this improvement ?
If so, would it be possible to have a v1 release including it ?

markerikson:
@antoinetissier thanks!

No, no plans to backport any of these changes to RTK 1.x.  We shipped RTK 2.0 2.5 years ago and don't intend to do any more work on that line.  

Anything blocking you from upgrading to 2.x?

antoinetissier:
Thank you for your answer @markerikson 
We ship an application with a Redux store that can be enhanced.
So Redux is a peer dependency, and upgrading it to a new major version would be a breaking change.
It is Redux 4.2.0, hence why we're stuck on RTK 1.x

EskiMojo14:
do any of the breaking changes in our [migration guide](https://redux-toolkit.js.org/usage/migrating-rtk-2) actually affect you? if it's a peer dependency, it sounds like you could just adjust your range without being a breaking change.

Issue #5142, opened 2025-11-23
[Docs] Misleading description of paramsSerializer


denexapp:
Hi,

`fetchBaseQuery` accepts an optional `paramsSerializer`.

[This docs page](https://redux-toolkit.js.org/rtk-query/api/fetchBaseQuery#paramsserializer) says that the params will be passed to `URLSearchParams` directly:

> params will be given directly to new URLSearchParams()

But if we look at the [code](https://github.com/reduxjs/redux-toolkit/blob/0cd0fe9abb18e05bfd6868fbe3fa401c0bf63b4a/packages/toolkit/src/query/fetchBaseQuery.ts#L296), in fact there's a default transformation happening - an internal `stripUndefined` function is applied.

This transformation doesn't apply if a custom `paramsSerializer` is provided.

This can lead to an unexpected behavior. If user provides a basic custom `paramsSerializer` that takes the params and passes them to a new `URLSearchParams`, its behavior is gonna be different from the default `paramsSerializer`, contrary to what the docs say.

The description of `paramsSerializer` should be updated to mention this behavior

markerikson:
Could you file a docs PR to update that description? Thanks!

CO0Ki3:
@markerikson 

I just updated description of `paramsSerializer`'s behavior.
If you have any time to see it, please review it.
Thank you.

markerikson:
Thanks!

PR #5147, opened 2025-11-27
docs: explain stripping undefined params before URLSearchParams

CO0Ki3:
Add description of `paramsSerializer`'s behavior.

Refs #5142


denexapp:
Hey, not sure if it's worth mentioning in the docs, undefined fields get deleted only if a passed object is plain

PR #5168, opened 2025-12-18
Speed up type checking by explicitly annotating some type parameter variances

### LINKED ISSUE #3214: [RED-16] Slow intellisense in VSCode
Hey 👋 

I'm trying to figure out what slows our VSCode intellisense (TS-backed code auto-complete) down and may have found the main cause.

Not 100% sure it's RTK Query (`createApi`) or not, but figured it's a decent-enough thing I found that I may as well share it.

Where I work at we have a pretty hefty React/RTK app (not OSS 😞), we've been dealing with slow (not unbearable, but annoying) response rate from VSCode intellisense (feels like >1s until the suggestions list shows up, but looking at the TS Server logs it's probably ~800ms).

I tried a few things, eventually landed on this: 

If I any-fy the call to `createApi`, the TS Server logs report that `completionInfo` (which is in charge of computing the list of suggested items that show up in VSCode's autocomplete) drops from 840ms to 122ms.

Here's a video before the change (note how slow it takes from the time I hit `.` to when I see the suggestions:

https://user-images.githubusercontent.com/927310/221375943-1547b820-7f19-40b8-933a-0269d4983faa.mp4

Here it is when I make the following change:

```
```

To:

```
export const api = (createApi as any)({
```

https://user-images.githubusercontent.com/927310/221375968-ad185389-96b7-4146-98da-c4e072e283ca.mp4

<sub>[RED-16](https://linear.app/redux/issue/RED-16/slow-intellisense-in-vscode)</sub>

akaltar:
# Why
In certain cases, type inference could be slow and reach typescript depth limits. Eg https://github.com/reduxjs/redux-toolkit/issues/3214

# How
By using typescript performance tracing on some examples from the above issue, I've manged to track down the slowness to checking the variances of various type parameters.
I used TypeSlayer's Type search to find the types where the variance was checked, and using the result data from the performance trace, I filled in the variances.

Since not all types of variance can be marked explicitly, I skipped some of them or added a different specifier. Eg `in` for bivariant.
I'm not 100% sure if this is correct, but many combinations produced type errors in various places, so based on my current understanding a wrong variance would cause type errors elsewhere.
I'm also not sure if this catches all common occurrences of this issue as I was using very simple examples. I will test for robustness + effectiveness by running the changes through a proprietary project and reporting the results.

I'm also not sure if this is the ideal way to solve this problem, I'd assume that there are other simplifications one could make to the types that would result in the built-in variance checking to get faster, but I couldn't find those and don't know enough about what's slow in the current version to do that

I think this could be a breaking change for some users:
- in some cases the reduced type complexity will surface existing type issues that were too complex to discover before
- In some cases the variants might cause a different breakage in client code => I'm hoping that the tests sufficiently account for this

# What
- Change `EndpointBuilder` from a type alias to an interface
- Add `in` `out` or `in out` variance specifiers to some type properties

### DIFF
--- a/packages/toolkit/src/query/core/buildMiddleware/cacheLifecycle.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/cacheLifecycle.ts
@@ -135,13 +135,14 @@ export interface QueryCacheLifecycleApi<
 > extends QueryBaseLifecycleApi<QueryArg, BaseQuery, ResultType, ReducerPath>,
     CacheLifecyclePromises<ResultType, BaseQueryMeta<BaseQuery>> {}
 
-export type MutationCacheLifecycleApi<
-  QueryArg,
-  BaseQuery extends BaseQueryFn,
-  ResultType,
-  ReducerPath extends string = string,
-> = MutationBaseLifecycleApi<QueryArg, BaseQuery, ResultType, ReducerPath> &
+export interface MutationCacheLifecycleApi<
+  in QueryArg,
+  in BaseQuery extends BaseQueryFn,
+  out ResultType,
+  in ReducerPath extends string = string,
+> extends MutationBaseLifecycleApi<QueryArg, BaseQuery, ResultType, ReducerPath>,
   CacheLifecyclePromises<ResultType, BaseQueryMeta<BaseQuery>>
+  {}
 
 export type CacheLifecycleQueryExtraOptions<
   ResultType,
--- a/packages/toolkit/src/query/core/buildMiddleware/cacheLifecycle.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/cacheLifecycle.ts
@@ -51,12 +51,12 @@ export interface QueryBaseLifecycleApi<
   updateCachedData(updateRecipe: Recipe<ResultType>): PatchCollection
 }
 
-export type MutationBaseLifecycleApi<
+export interface MutationBaseLifecycleApi<
   QueryArg,
   BaseQuery extends BaseQueryFn,
   ResultType,
   ReducerPath extends string = string,
-> = LifecycleApi<ReducerPath> & {
+> extends LifecycleApi<ReducerPath> {
   /**
    * Gets the current value of this cache entry.
    */
@@ -70,7 +70,7 @@ export type MutationBaseLifecycleApi<
   >
 }
 
-type LifecycleApi<ReducerPath extends string = string> = {
+type LifecycleApi<in ReducerPath extends string = string> = {
   /**
    * The dispatch method for the store
    */
--- a/packages/toolkit/src/query/core/buildMiddleware/cacheLifecycle.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/cacheLifecycle.ts
@@ -52,10 +52,10 @@ export interface QueryBaseLifecycleApi<
 }
 
 export interface MutationBaseLifecycleApi<
-  QueryArg,
-  BaseQuery extends BaseQueryFn,
-  ResultType,
-  ReducerPath extends string = string,
+  in QueryArg,
+  in BaseQuery extends BaseQueryFn,
+  in ResultType,
+  in ReducerPath extends string = string,
 > extends LifecycleApi<ReducerPath> {
   /**
    * Gets the current value of this cache entry.
--- a/packages/toolkit/src/query/core/buildMiddleware/cacheLifecycle.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/cacheLifecycle.ts
@@ -89,7 +89,10 @@ type LifecycleApi<ReducerPath extends string = string> = {
   requestId: string
 }
 
-type CacheLifecyclePromises<ResultType = unknown, MetaType = unknown> = {
+type CacheLifecyclePromises<
+  out ResultType = unknown,
+  out MetaType = unknown,
+> = {
   /**
    * Promise that will resolve with the first value for this cache key.
    * This allows you to `await` until an actual value is in cache.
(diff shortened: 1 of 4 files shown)


markerikson:
.... huh. wow. 

this is already clearly over _my_ head :)

First question is what this does to our existing type-tests, then we can go from there.

Either way, though: regardless of whether this works or not and whether we merge it, _thank you_ for putting in serious investigation time, trying to optimize perf here, and filing this PR!

markerikson:
well nothing _obviously_ broke :)  that's either a very good sign, or a sign that we're not testing things sufficiently :)

@akaltar can you use the PR build to do some revised perf comparisons?
 
Also, I'd love to see before and after flamegraphs!

EskiMojo14:
regarding types vs interfaces, see #4467 - we found that usage of interfaces created type portability issues for the types listed.

akaltar:
Would it be against the project conventions to export a couple of these types? If not, I'll just convert them back to type aliases. Interfaces can have much better caching characteristics according to the TypeScript performance guide, hence why I've changed them, but they might not be strictly necessary for most of the optimizations here to work, so I'll check.

Also while initially developing I was manually keeping the type files in sync, and I was seeing much better results, so I'm doing a couple more passes to see how good I can get it, while also testing in a production app to see more real-world results

phryneas:
We try not to expose them to userland, as they are not guaranteed to be stable.
One thing that could work would be exposing them like this, maybe that will satisfy the portability requirement?
```ts
// index.ts
import {UnportableType} from "./foo"
export declare namespace InternalDoNotUse {
  export {UnportableType}
}
```

akaltar:
> We try not to expose them to userland, as they are not guaranteed to be stable. One thing that could work would be exposing them like this, maybe that will satisfy the portability requirement?
> 
> ```ts
> // index.ts
> import {UnportableType} from "./foo"
> export declare namespace InternalDoNotUse {
>   export {UnportableType}
> }
> ```

I'd rather try this if it's an acceptable solution, as for example `MutationBaseLifecycleApi` can't be optimized as a type alias, because it becomes an intersection, which doesn't support explicit variance specifiers.
I'll update my PR with this approach and check if that passes the tests

akaltar:
Simple example
before:
image

after:
image

Complex real-world app:
before:
image


after:
image

Sooo that's a bit sad, in the complex app it actually ends up slowing things down. I think my simple example might be too simple or maybe different changes are required. In any case I don't think this PR would be useful to merge in its current state. I'll see if I can find time to test more thoroughly on a real-world project with various changes to see if I can find optimizations that actually help

akaltar:
This was my first time even hearing about the type variance concept, and the available docs are quite limited, but I've finally managed to find some good documentation on how it's intended to be used in the ts release notes when it got added:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#optional-variance-annotations-for-type-parameters
And in the PR that added it: https://github.com/microsoft/TypeScript/pull/48240

Thanks to the github references, I've also found another example
of people using explicit variance as a perf improvement here:
https://github.com/wevm/viem/pull/2557

The TL;DR as far as I understand it is that:
- Only `in`, `out`, and `in out` can be specified, but internally there is also bivariant, which can't be marked explicitly and thus, it can't be optimized
- Only `in` and `out` are typechecked individually, `in out` can't be relied on in the same way, eg you might end up being more strict with what can be input into the type parameter, even if it's unnecessary.
- Specifying the variance explictly can change the type checking behaviour as sometimes typescript doesn't perform an exhaustive check, so as not to take too much time => This means any of these optimizations might uncover type issues in dependent projects

These extra docs were very reassuring, especially because it verified the idea that variance specifiers can and should be used for performance reasons.

With that in mind I've created a more advanced example that should roughly mimic type checking in our large repo, and added some more changes to optimize that use-case. I'll test that on the large code-base using the sandbox pr build.
<details>
<summary>comprehensive type example</summary>

```ts
import { createApi } from "@reduxjs/toolkit/query/react";

export const apiClientSplitApi = createApi({
  baseQuery: () => ({ data: null }),
  endpoints: (builder) => ({
    a: builder.query<null, void>({
      queryFn: async () => ({ data: null }),
    }),
  }),
});

apiClientSplitApi.enhanceEndpoints({
    addTagTypes: ['aaa'],
}).injectEndpoints(
    {endpoints: (builder) => ({
        b: builder.query({
            queryFn: async () => ({ data: "asdf"})
        }),
        c: builder.mutation({
            queryFn: async ({g}:{g:string}) => ({data: { resss: "h"}}),
        })
    })}
);

```

</details>


As for the type portability, it looks like a namespace didn't work, so I've opted to use renamed types with an `RTKInternalDoNotUse` prefix, please let me know if you find that an acceptable workaround or if I should look at other optimization options instead

akaltar:
The large code-base also shows consistent improvement this time. I've verified with many more runs this time to make sure run-to-run variance isn't the cause.
Before: 
image
After: 
image

With these latest changes this would be beneficial for us, but I do understand it does increase the complexity and the maintenance burden quite a bit, so no worries if you think it's not worth the effort and risk at this point in time

aryaemami59:
@akaltar Thank you so much for all the effort you're putting into this. I'll take a closer look at the changes as soon as I can.

Before moving forward, I want to double-check a couple of things:

1. That the [**variance annotations**](https://www.typescriptlang.org/docs/handbook/2/generics.html#variance-annotations) are correct. TypeScript generally discourages their use in production types (noting in the docs that they're primarily intended for "type debugging"), so we need to be confident they're being applied accurately and appropriately here.
2. The work done to resolve the [**type portability issues**](https://github.com/reduxjs/redux-toolkit/pull/4467) we previously ran into has inadvertently contributed to TypeScript performance regressions. To resolve the portability issues, we had to favor **`type`** aliases over **`interfaces`**, which as you probably know by now can cause [**performance issues**](https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections), especially given how heavily we rely on [**type intersections**](https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types). We need to make sure the performance benefits introduced by this PR don't push us too far back toward [**type portability issues**](https://github.com/reduxjs/redux-toolkit/pull/4467).

I'll follow up once I've had a chance to review everything in more detail.

akaltar:
This PR is likely built on wrong assumptions based on the responses here: https://github.com/microsoft/TypeScript/pull/62973
I think if any perf improvements are possible, it's by returning some of these types to be interfaces and exporting them with some private prefixes. If there is appetite for that kind of change, I'm happy to look into it separately.

markerikson:
@akaltar gotcha. yeah, if you've got any other ideas for speeding up types, we're definitely interested!

Issue #4322, opened 2024-04-03
remove implicit conversion to camel case of openapi operation id.


tombohub:
if open api schema states `operationId: house_inventory_items_retrieve` 

than this should work in RTK query code generation: `filterEndpoints: [/inventory_items/i],`

For a library that promotes no side effects this should not happen.

phryneas:
> For a library that promotes no side effects this should not happen.

You just earned yourself an issue close with a salty comment. We will definitely not help you unpaid in our free time with that behaviour.

tombohub:
@phryneas You are right, I apologize. I was salty at the moment and harshly opened issue instead of composing myself. Please forgive me! 

Can we work around it? Could we please just have a sentence in documentation that says operationId is converted to camelCase?

phryneas:
You're welcome to open a PR to the docs.

Issue #2675, opened 2022-09-05
[RTKQ Code Gen OpenAPI] feature request: convert case styles of request and response parameters


aidy1991:
When generating codes from OpenAPI schema, the parameters of request body and response use the names defined in the schema as they are. For example, if a parameter is defined as a snake case, it is necessary to exceptionally write snake case in the TypeScript code.
Since it is common to use camel case in TypeScript and JavaScript, we would like to use camel case in the code generated from OpenAPI schema.

phryneas:
If this would be a configurable option, I think we'd accept a PR for this.

blwinters:
I just started using the codegen functionality today and similar to above, we have snake_case on the Python backend and camelCase on the TypeScript clients.

I see that `generate.ts` does perform some camelCase conversion, but as I'm new to the package, the scope and intention of this conversion isn't clear to me. Which aspects of codegen does the camelCase conversion currently affect? That is, what does the existing conversion enable if the generated types are not being converted to camelCase?

I'd like to make a PR for this as a configurable option in the future. I'm just wondering what the best approach is in this mixed-case situation when using the codegen functionality as it stands.

blwinters:
I also read through [this issue](https://github.com/reduxjs/redux-toolkit/issues/2181), so it seems like some users are getting conversions (though undesired), so I'm not clear about the intended behavior and use.

issy:
I've opened a PR (#5174) to allow users to disable this behaviour if desired. Hopefully this is a satisfactory solution

Issue #2181, opened 2022-03-31
RTK-Query Codegen: Camel case applied incorrectly in some situations


sbward:
In my API I use a variation of camel casing that preserves uppercase in abbreviations. For example, `userId` becomes `userID`. Normally this is interpreted as camel case and I would not expect any transformations to be applied to the string.

The issue is that RTK-Query Codegen uses the lodash `camelCase` function, which converts `userID` to `userId`. Lodash's behavior in this regard has been controversial (see https://github.com/lodash/lodash/issues/5045) and I think it would be better to use an alternative (like [hump](https://www.npmjs.com/package/humps)) which has the expected behavior.

Relevant code:

https://github.com/reduxjs/redux-toolkit/blob/2bb44b97dc290b70e3ef2502c743c9e90520fe50/packages/rtk-query-codegen-openapi/src/generate.ts#L218-L221

jenniferrethi:
This is happening for us as well. Our generated types are all in camelCase but our models are in PascalCase so this is making it hard to actually use the generated types.

issy:
I have opened a PR (#5174) to allow users to "opt out" of the camelCase behaviour and use exact operation IDs instead. I hope this will be a satisfactory solution both for people affected by this issue and for the Redux maintainers

PR #5174, opened 2026-01-02
codegen: Add `exactOperationIds` option to `generateEndpoints` config

issy:
Currently, `@rtk-query/codegen-openapi` converts operation IDs to camelCase before generating API code. This is due to an implementation detail in `oazapfts` (see https://github.com/oazapfts/oazapfts/issues/591). For example, if your operation ID is `fetchMyJWTPlease` it will be renamed to `fetchMyJwtPlease`.
Although this is documented behaviour, it is not always desired behaviour for library consumers. I think it would be good to offer users an escape hatch from that behaviour. In this PR I am introducing the config property `exactOperationIds?: boolean`. When enabled, it will not transform any operation IDs. This comes with some additional checks to ensure that there are no missing/duplicate operation IDs, but I think it's acceptable with such an escape hatch that users should be expected to validate these things.

I have added a small section to the docs to reflect this:
image

I believe this should resolve the following issues:
https://github.com/reduxjs/redux-toolkit/issues/4322
https://github.com/reduxjs/redux-toolkit/issues/2181
https://github.com/reduxjs/redux-toolkit/issues/2675

Ideally this would be implemented in `oazapfts`. There are plans on the roadmap for a plugin system and a hook for naming/renaming operations. This would be ideal for a future refactor, but will have to wait until that work is completed on their side.


EskiMojo14:
closing as https://github.com/reduxjs/redux-toolkit/pull/5236 supersedes this and has now been merged - thank you!

PR #5182, opened 2026-01-15
fix(query): prevent `onQueryStarted` from triggering at end-of-list

### LINKED ISSUE #5167: `fetchNextPage`always  triggering `onQueryStarted`, even when `queryFulfilled` already occured.
It looks like the entire lifecycle of `onQueryStarted` occurs when `fetchNextPage` is called.

However, often `fetchNextPage` is expected to often indicate end-of-list. 

This seems like unexpected behavior, because no additional base-query usage is made, but any side effects of `onQueryStarted` DO occur.

Take the following example, with the code below. 

In this example, `patchCachedUser` syncs some metadata between the `query-notifications` endpoints with the `get-user` endpoint (the total notifications count, to prevent flickering while states reconcile).

The behavior that occurs is:

1. Hook mounts with page of data. `onQueryStarted` does *not* fire unless this is the first time the hook is mounted, as expected.
2. The end of list is reached, triggering `fetchNextPage`.
3. `fetchNextPage`, via `getNextPageParam`, knows we're at the end-of-list, returning undefined.
4. `onQueryStarted` is fired, ALWAYS. however, `queryFulfilled` is already resolved, and so it does the remaining work (e.g. pessimistic updates), using the old data. 

```ts

 getNotifications: build.infiniteQuery<
      {
        notifications: Notification[];
        user: Pick<User, "numUnreadNotifications">
      },
      {
        uid: string;
        limit: number;
        filter: "read" | "unread"
      },
      PaginationCursor | undefined
    >({
      onQueryStarted: async (args, { dispatch, queryFulfilled }) => {
        try {
          console.log("Fetching notifications for user:"); // THIS logs during a `fetchNextPage` call
          const { data } = await queryFulfilled; // Base query has its own logging -- this is already resolved during the calls.
          const lastPage = data.pages[data.pages.length - 1];
          const user = lastPage.user;
          dispatch(patchCachedUser(args, user)); // This now badly assumes that the data is the most-recent from the API, but its from a previous cached call.
        } catch (error) {
          void error;
        }
      },
     infiniteQueryOptions: {
        initialPageParam: undefined,
        getNextPageParam: ({
          notifications: lastPage,
        }): undefined | PaginationCursor => {
          const lastItem = lastPage?.at(-1);
          return lastItem && { createdAt: lastItem.createdAt };
        },
      },
```

---

This looks like unexpected behavior to me, and I don't see a clean way to get around this (to prevent side affects from occuring multiple times from the same resolved underlying `queryFulfilled`)

joseph0926:
Fixes #5167

## Summary

When `fetchNextPage()` or `fetchPreviousPage()` is called at end-of-list (when `getNextPageParam`/`getPreviousPageParam` returns `undefined`), the thunk was still executing and triggering `onQueryStarted` with stale cached data.

To solve the problem added an early bail-out in the thunk's `condition` function -> to check if the next/previous page actually exists before processing any action.

## Changes

- Added direction check in `buildThunks.ts` condition function
- Added test case for the fix
- Updated existing test expectation (removed `isFetchingPreviousPage: true` when no previous page exists)

## Test

- yarn test
- yarn type-tests
- yarn eslint

### DIFF
--- a/packages/toolkit/src/query/core/buildThunks.ts
+++ b/packages/toolkit/src/query/core/buildThunks.ts
@@ -971,6 +971,25 @@ In the case of an unhandled error, no tags will be "provided" or "invalidated".`
           return false
         }
 
+        if (direction && isInfiniteQueryDefinition(endpointDefinition)) {
+          const cachedData = requestState?.data as
+            | InfiniteData<unknown, unknown>
+            | undefined
+          if (cachedData?.pages.length) {
+            const { infiniteQueryOptions } = endpointDefinition
+            const pageParamFn =
+              direction === 'forward' ? getNextPageParam : getPreviousPageParam
+            const nextParam = pageParamFn(
+              infiniteQueryOptions,
+              cachedData,
+              currentArg,
+            )
+            if (nextParam == null) {
+              return false
+            }
+          }
+        }
+
         return true
       },
       dispatchConditionRejection: true,
--- a/packages/toolkit/src/query/tests/infiniteQueries.test.ts
+++ b/packages/toolkit/src/query/tests/infiniteQueries.test.ts
@@ -253,7 +253,6 @@ describe('Infinite queries', () => {
 
     checkEntryFlags('fire', {
       hasNextPage: true,
-      isFetchingPreviousPage: true,
     })
 
     const entry1PrevPageMissing = await res3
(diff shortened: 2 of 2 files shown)


joseph0926:
[475a8e3](https://github.com/reduxjs/redux-toolkit/pull/5182/commits/475a8e396c250dffe38f4a95f6085856a2d6d2ed)

The tests added in the previous commit seemed slightly insufficient, so I added additional test code for possible edge cases.

For example, while only forward testing existed before, this commit adds verification for backward cases, as well as handling for skip operations and null values....

themavik (commented):
Gating `initiate` when `getNextPageParam`/`getPreviousPageParam` returns `null` stops the extra `onQueryStarted` at the tail—matches the new tests. nit: the guard only runs when `cachedData.pages.length` is truthy; if an infinite query ever has a zero-page cached shape you might still dispatch—probably unreachable, but worth a glance.

PR #5206, opened 2026-01-22
chore(toolkit): migrate TypeScript setup to `"moduleResolution": "bundler"`

### LINKED ISSUE #5197: chore(toolkit): migrate TypeScript setup to `"moduleResolution": "bundler"`
- [X] Migrate TypeScript configuration to [**`"moduleResolution": "bundler"`**](https://www.typescriptlang.org/docs/handbook/modules/reference.html#bundler) now that [**`Node10`**](https://www.typescriptlang.org/docs/handbook/modules/reference.html#node10-formerly-known-as-node) resolution is [**deprecated**](https://github.com/microsoft/TypeScript/issues/62200).

aryaemami59:
# **This PR**:

- [X] Migrates TypeScript configuration to [**`"moduleResolution": "bundler"`**](https://www.typescriptlang.org/docs/handbook/modules/reference.html#bundler) now that [**`Node10`**](https://www.typescriptlang.org/docs/handbook/modules/reference.html#node10-formerly-known-as-node) resolution is [**deprecated**](https://github.com/microsoft/TypeScript/issues/62200).
- [X] Resolves #5197.

### DIFF
--- a/examples/type-portability/bundler/package.json
+++ b/examples/type-portability/bundler/package.json
@@ -7,36 +7,22 @@
   "main": "src/index.tsx",
   "dependencies": {
     "@reduxjs/toolkit": "workspace:^",
-    "react": "^19.0.0",
-    "react-dom": "^19.0.0",
-    "react-redux": "^9.1.2",
-    "react-router-dom": "^6.25.1",
-    "react-scripts": "5.0.1"
+    "react": "^19.2.4",
+    "react-dom": "^19.2.4",
+    "react-redux": "^9.2.0",
+    "react-router-dom": "^7.13.0"
   },
   "devDependencies": {
-    "@types/node": "^20.14.11",
-    "@types/react": "^19.0.1",
-    "@types/react-dom": "^19.0.1",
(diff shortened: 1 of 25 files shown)


packages/toolkit/vitest.setup.ts
```diff
@@ -41,6 +41,8 @@ const areErrorsEqual = (actualError: unknown, expectedError: unknown) => {
       actualError.message === expectedError.message
     )
   }
+
+  return undefined
```
EskiMojo14: is the explicit return required? it already returns undefined if there's no return statement

aryaemami59: Actually I just realized [**`vitest`**](https://vitest.dev) already internally uses an error-comparison helper, [**`isErrorEqual`**](https://github.com/vitest-dev/vitest/blob/89cbdaea3f86e699c8e46ee52b08204cbc8bf345/packages/expect/src/jest-utils.ts#L261C1-L290C2), which makes **`areErrorsEqual`** redundant.

I put up #5216 to remove it.

PR #5218, opened 2026-02-08
Export hook options types for RTK Query hooks

### LINKED ISSUE #5172: Export the options types for hooks such as `useQuery`, `useLazyQuery`, and `useMutation`.
Hello! I wrote a hook to unify calling all query and mutation APIs from the store in the current React context. My hook takes three parameters: the API endpoint path, the params, and the options. Since the options type is not exported directly, I have to use `Parameters` on `TypedUseQuery` to infer the `Options` type. It would be more convenient if the Options type could be exported directly.

```ts
type UseQuery<T extends EndpointDefinitions> = <
  K extends keyof T & string,
  Definition extends NonNullable<T[K]['Types']> = NonNullable<T[K]['Types']>,
>(
  path: K,
  arg?: Definition['QueryArg'],
  options?: Parameters<TypedUseQuery<Definition['ResultType'], Definition['QueryArg'], Definition['BaseQuery']>>[1],
) => TypedUseQueryHookResult<Definition['ResultType'], Definition['QueryArg'], Definition['BaseQuery']>;

const { data: data1 } = useQuery('api1.getUser', { id: 1 }, { refetchOnFocus: true });
const { data: data2 } = useQuery('api2.getStar', { id: 2 }, { refetchOnFocus: true });
```
If permitted, I could also submit a PR. Thank you!

veeceey:
## Summary

Exports `UseQuerySubscriptionOptions`, `UseQueryStateOptions`, and `UseMutationStateOptions` from `@reduxjs/toolkit/query/react` to allow users to directly reference these types without using `Parameters<T>` utility type workaround.

## Motivation

Currently, when building wrapper hooks that accept the same options as RTK Query hooks (`useQuery`, `useLazyQuery`, `useMutation`), developers must use the `Parameters` utility type to extract the options type, which is verbose and less discoverable:

```ts
// Current workaround
options?: Parameters<TypedUseQuery<...>>[1]
```

With these types exported, it becomes much cleaner:

```ts
// After this change
options?: UseQuerySubscriptionOptions & UseQueryStateOptions<...>
```

## Changes

- Added exports for `UseQuerySubscriptionOptions`, `UseQueryStateOptions`, and `UseMutationStateOptions` in `/packages/toolkit/src/query/react/index.ts`

## Test Plan

- TypeScript compilation succeeds
- Existing tests pass
- Types are properly exported and accessible to consumers

Fixes #5172

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -538,7 +538,7 @@ export type TypedUseQueryState<
 >
 
 /**
- * @public
+ * @internal
  */
 export type UseQueryStateOptions<
   D extends QueryDefinition<any, any, any, any>,
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -203,6 +203,19 @@ export type UseQuerySubscriptionOptions = SubscriptionOptions & {
 }
 
 /**
+ * Provides a way to reference the options accepted by the `useQuerySubscription`
+ * hook in userland code.
+ *
+ * Unlike other `Typed*` wrappers, this type has no generic parameters since
+ * {@linkcode UseQuerySubscriptionOptions} does not depend on a specific query
+ * definition.
+ *
+ * @since 2.2.8
+ * @public
+ */
+export type TypedUseQuerySubscriptionOptions = UseQuerySubscriptionOptions
+
+/**
  * A React hook that automatically triggers fetches of data from an endpoint, and 'subscribes' the component to the cached data.
  *
  * The query arg is used as a cache key. Changing the query arg will tell the hook to re-fetch the data if it does not exist in the cache already.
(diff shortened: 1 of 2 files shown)


aryaemami59:
We should probably do this as well

https://github.com/veeceey/redux-toolkit/blob/9aad60df2e3d74cd94e18d9bdf65b7e44649351e/packages/toolkit/src/query/react/buildHooks.ts#L527C1-L598C2

```diff
/**
- * @internal
+ * @public
 */
export type UseQueryStateOptions<
```

aryaemami59 (approved):

EskiMojo14:
instead of exporting these directly, we should have `Typed<name>` wrappers as with the others

veeceey:
Thanks for the feedback @EskiMojo14 and @aryaemami59! I've updated the PR in the latest commit (`97bc40f`) to use `Typed<name>` wrappers for all the exported hook option types, following the existing pattern:

- `TypedUseQuerySubscriptionOptions` (simple alias since the base type has no generics)
- `TypedUseQueryStateOptions` (already existed with proper generics, now re-exported)
- `TypedUseMutationStateOptions` (new, with generic parameters)

The `UseQueryStateOptions` was also changed from `@internal` to `@public` as suggested. Let me know if anything else is needed!

EskiMojo14 (approved):

EskiMojo14:
thanks!

PR #5226, opened 2026-02-17
docs: `selectInvalidedBy` for multiple cache updates across queries

bigcupcoffee:
This is a rough sample of what we use in our mobile app to update an entity across 10+ queries without invalidating them all

Switch cases are easier to extend IMO, while in our code we have some abstration layer for draft updaters and entity predicates (as we have secondary identifier plus nested entities), this is more for the devs to figure out if they feel like they need it

For some queries we even have updates outside of redux with an according custom undo return, so this also keeps the room for obscure custom cases (which I wish I didn't have to see & code)

Code definitely feels somewhat weird but I feel like we still need a mention of this usecase for `selectInvalidatedBy`, for simplicity sake the code could only cover updating the list queries, but I figured I'd present a more complex case at this point, what do you think?

Could probably use some touch up on the wording, feel free to comment 🙏


bigcupcoffee:
@markerikson thoughts?

docs/rtk-query/usage/manual-cache-updates.mdx
```diff
@@ -224,6 +224,104 @@ const api = createApi({
 })
 ```
 
+### Updates Across Queries
+
+The entity you want to update might be present in multiple differently shaped queries, such as
+(infinite) lists, or even inside another entity. If you don't want to invalidate those instead,
+you can avoid manually looping through every single cache entry by using [selectInvalidatedBy](https://redux-toolkit.js.org/rtk-query/api/created-api/api-slice-utils#selectinvalidatedby):
```
aryaemami59, line 231:
```suggestion
you can avoid manually looping through every single cache entry by using [`selectInvalidatedBy`](https://redux-toolkit.js.org/rtk-query/api/created-api/api-slice-utils#selectinvalidatedby):
```

PR #5236, opened 2026-02-25
feat(codegen): add `operationIdTransformer` option

### LINKED ISSUE #2181: RTK-Query Codegen: Camel case applied incorrectly in some situations

### LINKED ISSUE #2675: [RTKQ Code Gen OpenAPI] feature request: convert case styles of request and response parameters

### LINKED ISSUE #4322: remove implicit conversion to camel case of openapi operation id.

aryaemami59:
## Summary

- [x] Add an **`operationIdTransformer`** config option to **`@rtk-query/codegen-openapi`** that controls how OpenAPI **`operationId`** values are transformed into endpoint names.

## Motivation

- **`@rtk-query/codegen-openapi`** currently passes every **`operationId`** through **`oazapfts`**, which applies lodash's [**`camelCase`**](https://lodash.com/docs/4.17.23#camelCase).
  This mangles consecutive uppercase acronyms.
  For example, **`fetchMyJWTPlease`** becomes **`fetchMyJwtPlease`**, making it impossible to match the original **`operationId`** exactly.

This causes two categories of pain:

1. **`filterEndpoints` breaks on acronym-heavy names.** If your schema has **`operationId: fetchMyJWTPlease`**, you must filter by **`'fetchMyJwtPlease'`** (the mangled form) even though the schema says **`fetchMyJWTPlease`**. This is surprising and easy to get wrong. See #4322.
2. **Generated names silently diverge from the schema.** Teams that use the **`operationId`** as a stable identifier (e.g. for analytics, logging, or cross-service contracts) find the camelCase conversion breaks their assumptions. See #2181.

## Solution

- [x] Add **`operationIdTransformer?: 'camelCase' | 'none' | ((operationId: string) => string)`** to **`CommonOptions`**.

| Value                                 | Behavior                                                                                                                                         |
| ------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
| **`'camelCase'`** _(default)_         | Existing behavior. Applies lodash's [**`camelCase`**](https://lodash.com/docs/4.17.23#camelCase) via **`oazapfts`**. Fully backwards-compatible. |
| **`'none'`**                          | Uses the raw **`operationId`** verbatim, preserving exact casing from the schema.                                                                |
| **`(operationId: string) => string`** | Applies a custom transform for full control.                                                                                                     |

When **`'none'`** or a function is used, every operation **must** have an **`operationId`** defined. The codegen throws an informative error if any operation is missing one.

**`filterEndpoints`** is always matched against the **transformed** name, consistent regardless of which transformer is active.

### Usage

Preserve exact casing. **`fetchMyJWTPlease`** stays **`fetchMyJWTPlease`**:

```ts
import type { ConfigFile } from "@rtk-query/codegen-openapi";

const config = {
  schemaFile: "./openapi.json",
  apiFile: "./src/emptyApi.ts",
  outputFile: "./src/api.ts",
  operationIdTransformer: "none",
} satisfies ConfigFile;

export default config;
```

Custom transform:

```ts
import type { ConfigFile } from "@rtk-query/codegen-openapi";

const config = {
  schemaFile: "./openapi.json",
  apiFile: "./src/emptyApi.ts",
  outputFile: "./src/api.ts",
  operationIdTransformer: (operationId) => operationId.replace(/^get/, "fetch"),
} satisfies ConfigFile;

export default config;
```

### Related

- Resolves #4322.
- Related to #2181.
- Related to #2675.
- Supersedes #5174.

---

PR #5174 by @issy proposed a `boolean` **`exactOperationIds`** flag. A `boolean` is simpler to configure but can only toggle between two fixed behaviors. **`operationIdTransformer`** is a strict superset: passing **`'none'`** is equivalent to **`exactOperationIds: true`**, while the function form covers everything else. The implementation PR takes direct inspiration from #5174's approach and @issy will be added as a co-author of the commit.

### DIFF
--- a/packages/rtk-query-codegen-openapi/test/generateEndpoints.test.ts
+++ b/packages/rtk-query-codegen-openapi/test/generateEndpoints.test.ts
@@ -933,3 +933,191 @@ describe('esmExtensions option', () => {
     expect(content).toContain("import { api } from '../../fixtures/emptyApi'");
   });
 });
+
+describe('operationIdTransformer option', () => {
+  const schemaFile = resolve(__dirname, 'fixtures', 'operationIdTransformer.yaml');
+  const missingIdSchemaFile = resolve(__dirname, 'fixtures', 'operationIdTransformerMissingId.yaml');
+
+  const baseConfig = {
+    apiFile: './fixtures/emptyApi.ts',
+    hooks: true,
+    schemaFile,
+  };
+
+  const noneConfig = {
+    ...baseConfig,
+    operationIdTransformer: 'none' as const,
+  };
+
+  const prefixTransformer = <OperationIdType extends string>(operationId: OperationIdType) =>
+    `custom_${operationId}` as const;
+
+  describe('"camelCase" (default)', () => {
+    test('transforms consecutive uppercase letters', async () => {
+      const api = await generateEndpoints({
+        ...baseConfig,
+        filterEndpoints: ['fetchMyJwtPlease'],
+      });
+
+      expect.assert.isOk(api);
+
+      expect(api).toContain('fetchMyJwtPlease:');
+      expect(api).toContain('type FetchMyJwtPleaseApiResponse');
+      expect(api).toContain('type FetchMyJwtPleaseApiArg');
+      expect(api).toContain('useFetchMyJwtPleaseQuery');
+    });
+
+    test('does not change already-correct camelCase operationIds', async () => {
+      const api = await generateEndpoints({
+        ...baseConfig,
+        filterEndpoints: ['addPet'],
+      });
+
+      expect.assert.isOk(api);
+
+      expect(api).toContain('addPet:');
+      expect(api).toContain('type AddPetApiResponse');
+      expect(api).toContain('type AddPetApiArg');
+      expect(api).toContain('useAddPetMutation');
+    });
+
+    test('does not throw when operationId is missing (backward compat)', async () => {
+      await expect(
+        generateEndpoints({
+          apiFile: './fixtures/emptyApi.ts',
+          schemaFile: missingIdSchemaFile,
+        })
+      ).resolves.toBeTruthy();
+    });
+  });
+
+  describe('"none"', () => {
+    test('preserves consecutive uppercase letters in endpoint key', async () => {
+      const api = await generateEndpoints({
+        ...noneConfig,
+        filterEndpoints: ['fetchMyJWTPlease'],
+      });
+
+      expect.assert.isOk(api);
+
+      expect(api).toContain('fetchMyJWTPlease:');
+    });
+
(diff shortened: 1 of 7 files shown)


issy:
I'm glad this is gaining traction! Thanks for the acknowledgment. I like this idea

aryaemami59:
@issy

> I'm glad this is gaining traction! Thanks for the acknowledgment. I like this idea

Glad to hear it, and thanks for the original PR. It's what got the ball rolling on this!

The function form already lets you plug in any transform today (e.g. a `snake_case` or `PascalCase` library), but named shorthands like **`'PascalCase'`** or **`'snake_case'`** could be added as non-breaking additions to the union type down the line if there's enough demand. The architecture is there for it.

timdorr (approved):
I know less about this code than the rest of RTKQ, but it looks good to me in theory. I'd wait to get another approval before merging.

packages/rtk-query-codegen-openapi/test/generateEndpoints.test.ts
```diff
@@ -948,6 +948,194 @@ describe('esmExtensions option', () => {
...
+      expect(api).toContain('useAddPetMutation');
+    });
+
+    test('does not throw when operationId is missing (backward compat)', async () => {
+      await expect(
+        generateEndpoints({
+          apiFile: './fixtures/emptyApi.ts',
+          schemaFile: missingIdSchemaFile,
+        })
+      ).resolves.toBeTruthy();
```
EskiMojo14, line 1004: i wish vitest had a native `toResolve` instead of needing jest-extended for it

EskiMojo14 (approved):

PR #5237, opened 2026-02-25
fix(codegen): narrow `generateEndpoints` return type

aryaemami59:
## Summary

- [X] Follow-up to the [**suggestion**](https://github.com/reduxjs/redux-toolkit/pull/5174/changes#r2657784481) by @issy in #5174.
- [X] Adds function overloads to **`generateEndpoints`** so the return type narrows based on whether **`outputFile`** is provided:
  - Without **`outputFile`** → **`Promise<string>`**
  - With **`outputFile`** → **`Promise<void>`**

This eliminates the need for manual type assertions like **`await generateEndpoints(...) as string`** in consumer code.

### DIFF
--- a/packages/rtk-query-codegen-openapi/src/index.ts
+++ b/packages/rtk-query-codegen-openapi/src/index.ts
@@ -8,6 +8,14 @@ export type { ConfigFile } from './types';
 
 const require = createRequire(__filename);
 
+export async function generateEndpoints(
+  options: Omit<GenerationOptions, 'outputFile'> & { outputFile?: never }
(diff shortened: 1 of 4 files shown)


themavik (commented):
The overload split on `outputFile` matches how `generateEndpoints` returns `string` vs writes to disk. nit: callers who pass `outputFile: undefined` explicitly might still infer `Promise<string | void>` depending on `GenerationOptions`; worth sanity-checking against real consumer TS versions.

timdorr (approved):

PR #5268, opened 2026-04-01
Keep isSuccess: true when switching infinite query cache entries

riqts:
Fixes #5265

Updates `infiniteQueryStatePreSelector` to match the existing `queryStatePreSelector` logic for deriving `isSuccess`. When switching to a new arg that has an uninitialized cache entry, `isSuccess` now stays `true` if valid data exists from the previous entry.


markerikson:
Awesome, thanks!

PR #5273, opened 2026-04-05
fix: add setTimeout fallback to raf autoBatch strategy for background tabs

### LINKED ISSUE #5051: RTK Query does not trigger fetch in background tabs until document gains focus at least once
When using RTK Query in a React + Redux Toolkit application, queries are not executed immediately if the page is opened in a background tab (e.g., via Ctrl + Click / Cmd + Click). The fetch is deferred until the document receives focus for the first time.

This behavior breaks expected UX for users who open multiple links in background tabs, since no network requests are made until they manually switch to each tab. This can cause noticeable delays in rendering data when they finally open the tab.

Steps to reproduce:
1. Open any page that triggers an RTK Query fetch in a background tab using Ctrl + Click / Cmd + Click.
2. Do not switch focus to the new tab.
3. Observe that no network request is sent until the tab gains focus.

How I reproduced - 
Created a Query which will fetch an array of posts & I added a logic to set the document.title to posts.count.
Until I attained tab focus, the title is `undefined` only, as soon as focus is attained the api is fetched and title got populated correctly.

Expected behavior:
Queries should execute immediately on mount, even in background tabs, without requiring the document to gain focus.

Environment:
Redux Toolkit version: `"@reduxjs/toolkit": "^2.6.1",`

riqts:
Fixes #5051

The default `raf` autoBatch strategy now races `requestAnimationFrame` against a 100ms `setTimeout` fallback. Whichever fires first calls `notify()` and cancels the other.

Browsers don't fire `requestAnimationFrame` callbacks in background tabs (or heavily throttle them). This means batched subscriber notifications from RTK Query fetches completing in a background tab never arrive until the tab regains focus.

As identified by @phryneas in the issue thread, adding a `setTimeout` fallback ensures notifications still fire within 100ms even when RAF is suspended. In foreground tabs, RAF fires first (~16ms) and cancels the timer, so there's no behavior change.

- Added test: `setTimeout` fallback fires when RAF is suspended (simulated background tab)
- Added test: no double-notification when both RAF and `setTimeout` could fire

### DIFF
--- a/packages/toolkit/src/autoBatchEnhancer.ts
+++ b/packages/toolkit/src/autoBatchEnhancer.ts
@@ -15,6 +15,24 @@ const createQueueWithTimer = (timeout: number) => {
   }
 }
 
+const createRafWithFallbackTimer = (
+  raf: typeof requestAnimationFrame,
+  timeout: number,
+) => {
+  return (notify: () => void) => {
+    let called = false
+    const callback = () => {
+      if (called) return
+      called = true
+      cancelAnimationFrame(rafId)
+      clearTimeout(timerId)
+      notify()
+    }
+    const rafId = raf(callback)
+    const timerId = setTimeout(callback, timeout)
+  }
+}
+
 export type AutoBatchOptions =
   | { type: 'tick' }
   | { type: 'timer'; timeout: number }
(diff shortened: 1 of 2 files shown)


markerikson:
Oh neat, great idea!

PR #5301, opened 2026-05-22
Fix query loading state after skipped cache removal

### LINKED ISSUE #5300: [RTK Query] `removeQueryResult` deletes entry instead of resetting to `uninitialized`, breaking `isLoading` semantics
## Problem

When RTK Query removes a cache entry via `removeQueryResult` (such as when invalidating a tag with zero subscribers), it **completely deletes the query cache entry** from state. This has an unintended side-effect:

- When a component resubscribes to that query, the query slice has **no cache entry**, so the `isLoading` state does not correctly transition through `isUninitialized -> isLoading = true -> data`.
- Instead, it acts as if the query never existed; the state is not reset to `uninitialized`, it's just gone.

## Why it matters
This breaks the following:
- `isLoading` is meant to only be `true` when *no data exists and a fetch is in-flight* (page is loading from empty state)
- `isFetching` is meant for showing refetch/loading with already existing data

If a query entry is just deleted and not reset, subscribers cannot differentiate between a fresh initial loading state and a refresh/re-fetch. This can break loading-indicator logic and UI expectations.

## Code reference
```typescript
// buildSlice.ts
removeQueryResult: {
  reducer(
    draft,
    {
      payload: { queryCacheKey },
    }: PayloadAction<QuerySubstateIdentifier>,
  ) {
    delete draft[queryCacheKey]
  },
  ...
}
```

**What should happen instead:**
- reset the entry to status `STATUS_UNINITIALIZED` (not delete), so that the next subscription triggers the expected state flow

## Steps to reproduce
1. Create a query and subscribe to it (shows `isLoading=true` on initial mount)
2. Invalidate the query with no active subscribers (`removeQueryResult` fires)
3. Subscribe again. The entry is missing from state, so initial transitions do not occur and `isLoading` is not `true` as expected.

---

**Expected behavior:**
Deleting a query result when there are no subscribers should reset the entry to `uninitialized` (not delete), ensuring the correct isLoading/isFetching state transitions.

## Links/References
- Related discussion: https://github.com/reduxjs/redux-toolkit/discussions
- Originating user issue: https://github.com/reduxjs/redux-toolkit/discussions

## Labels
`RTK-Query`, `bug`

puneetdixit200:
## Summary

Fixes #5300.

When a hook is skipped, it keeps the last successful result so the skipped state can still expose previous `data`. If that cache entry is removed while skipped, unskipping should treat the next request as a fresh load instead of reusing the stale skipped result.

This resets the hook's last result when the current query state is uninitialized because the cache entry disappeared while skipped, while preserving the existing behavior for normal arg changes that intentionally keep previous data during a new fetch.

## Tests

- `yarn workspace @reduxjs/toolkit vitest --run src/query/tests/buildHooks.test.tsx -t 'unskipping a removed query cache entry restarts as an initial load'`
- `yarn workspace @reduxjs/toolkit vitest --run src/query/tests/buildHooks.test.tsx`
- `yarn workspace @reduxjs/toolkit vitest --run src/query/tests/buildThunks.test.tsx src/query/tests/cacheCollection.test.ts`
- `yarn workspace @reduxjs/toolkit test`
- `DISABLE_V8_COMPILE_CACHE=1 yarn workspace @reduxjs/toolkit eslint src/query/react/buildHooks.ts src/query/tests/buildHooks.test.tsx`
- `yarn prettier --check packages/toolkit/src/query/react/buildHooks.ts packages/toolkit/src/query/tests/buildHooks.test.tsx`
- `git diff --check`

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -1548,26 +1548,8 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
     lastResult: UseQueryStateDefaultResult<any> | undefined,
     queryArgs: any,
   ): UseQueryStateDefaultResult<any> {
-    // if we had a last result and the current result is uninitialized,
-    // we might have called `api.util.resetApiState`
-    // in this case, reset the hook
-    if (lastResult?.endpointName && currentState.isUninitialized) {
-      const { endpointName } = lastResult
-      const endpointDefinition = endpointDefinitions[endpointName]
-      if (
-        queryArgs !== skipToken &&
-        serializeQueryArgs({
-          queryArgs: lastResult.originalArgs,
-          endpointDefinition,
-          endpointName,
-        }) ===
-          serializeQueryArgs({
-            queryArgs,
-            endpointDefinition,
-            endpointName,
-          })
-      )
-        lastResult = undefined
+    if (shouldResetHookLastResult(currentState, lastResult, queryArgs)) {
+      lastResult = undefined
     }
 
     // data is the last known good request result we have tracked - or if none has been tracked yet the last good result for the current args
@@ -1608,26 +1590,8 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
     lastResult: UseInfiniteQueryStateDefaultResult<any> | undefined,
     queryArgs: any,
   ): UseInfiniteQueryStateDefaultResult<any> {
-    // if we had a last result and the current result is uninitialized,
-    // we might have called `api.util.resetApiState`
-    // in this case, reset the hook
-    if (lastResult?.endpointName && currentState.isUninitialized) {
-      const { endpointName } = lastResult
-      const endpointDefinition = endpointDefinitions[endpointName]
-      if (
-        queryArgs !== skipToken &&
-        serializeQueryArgs({
-          queryArgs: lastResult.originalArgs,
-          endpointDefinition,
-          endpointName,
-        }) ===
-          serializeQueryArgs({
-            queryArgs,
-            endpointDefinition,
-            endpointName,
-          })
-      )
-        lastResult = undefined
+    if (shouldResetHookLastResult(currentState, lastResult, queryArgs)) {
+      lastResult = undefined
     }
 
     // data is the last known good request result we have tracked - or if none has been tracked yet the last good result for the current args
@@ -1659,6 +1623,44 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
       isLoading,
       isSuccess,
     } as UseInfiniteQueryStateDefaultResult<any>
+  }
+
+  function shouldResetHookLastResult(
+    currentState: { isUninitialized: boolean },
+    lastResult:
+      | {
+          endpointName?: string
+          isUninitialized: boolean
+          originalArgs?: any
+        }
+      | undefined,
+    queryArgs: any,
+  ) {
+    if (!currentState.isUninitialized || queryArgs === skipToken || !lastResult)
+      return false
+
+    // If we had a last result and the current result is uninitialized,
+    // we might have called `api.util.resetApiState`, or the query cache entry
+    // might have been removed while a hook instance was skipped.
+    // In this case, reset the hook.
+    if (!lastResult.endpointName) {
+      return lastResult.isUninitialized
+    }
+
+    const { endpointName } = lastResult
+    const endpointDefinition = endpointDefinitions[endpointName]
+    return (
+      serializeQueryArgs({
+        queryArgs: lastResult.originalArgs,
+        endpointDefinition,
+        endpointName,
+      }) ===
+      serializeQueryArgs({
+        queryArgs,
+        endpointDefinition,
+        endpointName,
+      })
+    )
   }
 
   function usePrefetch<EndpointName extends QueryKeys<Definitions>>(
--- a/packages/toolkit/src/query/tests/buildHooks.test.tsx
+++ b/packages/toolkit/src/query/tests/buildHooks.test.tsx
@@ -4203,4 +4203,68 @@ describe('skip behavior', () => {
       currentData: undefined,
     })
   })
+
+  test('unskipping a removed query cache entry restarts as an initial load', async () => {
+    const queryCacheKey = 'getIncrementedAmount(undefined)'
+    const { result, rerender } = renderHook(
+      (skip: boolean) =>
+        api.endpoints.getIncrementedAmount.useQuery(undefined, { skip }),
+      {
+        wrapper: storeRef.wrapper,
+        initialProps: false,
+      },
+    )
+
+    await act(async () => {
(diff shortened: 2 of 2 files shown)


puneetdixit200:
Pushed 776af49 to make the published-artifact browser examples install Chromium explicitly before running Playwright.

The four failing jobs (`cra4`, `cra5`, `next`, and `vite`) all reached `yarn test` and failed at Playwright launch because the expected Chromium executable was missing from the runner cache. The node and native published-artifact examples already passed, so the install step is limited to the Playwright-based examples.

Local checks:
- `git diff --check`
- YAML parse for `.github/workflows/tests.yml`

The new GitHub Actions runs are currently waiting for fork-run approval before they can execute.

puneetdixit200:
Pushed 829995dc with a follow-up CI tweak.

The previous browser-install step reached Playwright, but `yarn playwright install --with-deps chromium` failed before downloading the browser because Ubuntu 24.04 no longer has a direct `libasound2` package candidate. The original failure was only a missing Chromium binary in the runner cache, so this now runs `yarn playwright install chromium` and avoids Playwright's apt dependency install path.

Checked locally:
- `git diff --check`
- YAML parse for `.github/workflows/tests.yml`

puneetdixit200:
Pushed 0db95f41 after checking the current-head logs.

The `next`/`cra5` jobs showed Chromium downloading during the initial `yarn install`, but the later Playwright test still failed because the expected executable under `~/.cache/ms-playwright/chromium-1048` was missing. The explicit install step only fetched ffmpeg, which suggests Playwright saw a stale or incomplete Chromium cache entry.

The workflow now clears `~/.cache/ms-playwright/chromium-*` immediately before `yarn playwright install chromium`, so the explicit step has to restore the Chromium browser used by the test runner.

Checked locally:
- `git diff --check`
- YAML parse for `.github/workflows/tests.yml`

puneetdixit200:
Pushed 2ee7cb12 with another published-artifact CI follow-up.

The current-head logs showed `yarn playwright install chromium` downloading Chromium, but the step returned before the expected executable existed under `~/.cache/ms-playwright/chromium-1048`, so the later Playwright launch still failed.

This now uses `PLAYWRIGHT_BROWSERS_PATH=0` for the published-artifact matrix so the browser install and test runner both use the example-local Playwright browser path instead of the shared runner cache. The install step also checks `chromium.executablePath()` immediately after install so any browser-install issue fails at that step with the resolved path.

Checked locally:
- `git diff --check origin/master...HEAD`
- YAML parse for `.github/workflows/tests.yml`

`actionlint` is not installed in my local environment, so I could not run that linter locally.

puneetdixit200:
Pushed 6f03da27 with one more Playwright install follow-up.

The current-head logs showed the example-local path from `PLAYWRIGHT_BROWSERS_PATH=0` was being used, but after `yarn remove` / `yarn add` mutated `node_modules`, `yarn playwright install chromium` still only fetched ffmpeg and left `node_modules/playwright-core/.local-browsers/chromium-1048/chrome-linux/chrome` missing.

The install step now uses `yarn playwright install --force chromium` so the explicit post-build install recreates Chromium in the same local browser path that the test runner resolves.

Checked locally:
- `git diff --check upstream/master...HEAD`
- YAML parse for `.github/workflows/tests.yml`

`actionlint` is not installed locally, so I could not run that workflow linter here.

puneetdixit200:
Pushed fc60c22e with another published-artifact CI follow-up.

The current-head logs showed that `PLAYWRIGHT_BROWSERS_PATH=0` was resolving to the example-local browser directory, but `yarn playwright install --force chromium` still only downloaded ffmpeg and then failed the immediate executable check because `node_modules/playwright-core/.local-browsers/chromium-1048/chrome-linux/chrome` was missing.

This update skips Playwright browser download during the initial example dependency install, then removes the resolved Chromium browser directory immediately before the explicit `yarn playwright install chromium` step. That avoids reusing a partial local browser directory after the workflow mutates `node_modules` with the packaged artifact install.

Checked locally:
- `git diff --check upstream/master...HEAD`
- YAML parse for `.github/workflows/tests.yml`

`actionlint` is not installed locally, so I could not run that workflow linter here.

puneetdixit200:
Pushed 829938cf after the approved run produced fresh logs.

The updated `vite` and `cra4` jobs confirmed that `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1` was applied during dependency install and that Chromium downloaded to 100% in the explicit install step, but the package-local `.local-browsers` executable was still missing immediately afterward.

This moves `PLAYWRIGHT_BROWSERS_PATH` out of `node_modules` to the runner temp directory for the published-artifact job, clears that Chromium cache before the explicit install, and keeps the immediate executable check before `yarn test`.

Checked locally:
- `git diff --check upstream/master...HEAD`
- YAML parse for `.github/workflows/tests.yml`

`actionlint` is not installed locally, so I could not run that workflow linter here.

puneetdixit200:
Pushed 7a68114a to fix the workflow parser issue from the previous commit.

The `runner` context is not available in that job-level `env` expression, so the browser path now uses a fixed Ubuntu temp directory instead. Local validation still passes:
- `git diff --check upstream/master...HEAD`
- YAML parse for `.github/workflows/tests.yml`

puneetdixit200:
Pushed c3aff796 after inspecting the current-head logs from the approved run.

The failed Playwright rows now show `PLAYWRIGHT_BROWSERS_PATH=/tmp/playwright-browsers`, the browser install command downloading Chromium to 100%, and then the immediate executable check failing before `yarn test`. This keeps the stable temp browser path and changes the check to wait up to 30 seconds for the resolved Chromium executable before failing with that path.

Checked locally:
- `git diff --check upstream/master...HEAD`
- YAML parse for `.github/workflows/tests.yml`

`actionlint` is not installed locally, so I could not run that workflow linter here.

puneetdixit200:
Pushed 80bd04ed after the current-head logs showed the 30-second wait still expired.

The failure is now clearly before `yarn test`: Playwright reports the Chromium download, but no executable is created at the resolved browser path even after waiting. This update derives the Chromium revision/path from Playwright, downloads the same Chromium archive directly, extracts it with Python into the resolved browser directory, chmods the executable, and keeps the immediate Playwright path check before tests.

Checked locally:
- `git diff --check upstream/master...HEAD`
- YAML parse for `.github/workflows/tests.yml`

`actionlint` is not installed locally, so I could not run that workflow linter here.

puneetdixit200:
The current head is green now.

On `80bd04ed`, the published-artifact browser rows that were failing before all completed successfully:
- `Test Published Artifact cra4`
- `Test Published Artifact cra5`
- `Test Published Artifact next`
- `Test Published Artifact vite`

The rest of the visible check set is success, skipped, or neutral, and the PR merge state is clean.

aryaemami59:
I've opened #5310 to fix the CI failures caused by the Playwright browser tests.

markerikson:
I'm not convinced #5300 is an actual bug and there's been no follow-up in that issue, so I'm going to close this.

PR #5302, opened 2026-05-25
fix(rtk-query): reset removeQueryResult entry to uninitialized instead of deleting

### LINKED ISSUE #5300: [RTK Query] `removeQueryResult` deletes entry instead of resetting to `uninitialized`, breaking `isLoading` semantics

xxiaoxiong:
Fixes #5300

Why
removeQueryResult currently does delete draft[queryCacheKey], which completely removes the cache entry from state. When a component resubscribes to the same query, no cache entry exists, so isLoading never transitions through the expected isUninitialized -> isLoading=true flow — breaking loading-indicator semantics.

Changes
Instead of deleting the entry, reset it to STATUS_UNINITIALIZED and clear the data, error, fulfilledTimeStamp, and startedTimeStamp fields.

Updated component: packages/toolkit/src/query/core/buildSlice.ts

Surface area
- RTK Query / Cache
- Bug fix

### DIFF
--- a/packages/toolkit/src/query/core/buildSlice.ts
+++ b/packages/toolkit/src/query/core/buildSlice.ts
@@ -291,7 +291,14 @@ export function buildSlice({
             payload: { queryCacheKey },
           }: PayloadAction<QuerySubstateIdentifier>,
         ) {
-          delete draft[queryCacheKey]
+          const entry = draft[queryCacheKey];
+          if (entry) {
+            entry.status = STATUS_UNINITIALIZED;
+            delete entry.data;
+            delete entry.error;
+            delete entry.fulfilledTimeStamp;
+            delete entry.startedTimeStamp;
+          }
         },
         prepare: prepareAutoBatched<QuerySubstateIdentifier>(),
       },


EskiMojo14:
are cache entries *ever* removed with this change?

markerikson:
Doesn't look like it. I don't have the issue loaded into my brain atm, and I get that there was something about "doesn't reset to unitialized" being discussed, but I do _not_ think this looks like the approach we'd want at first glance.

packages/toolkit/src/query/core/buildSlice.ts
```diff
@@ -291,7 +291,14 @@ export function buildSlice({
             payload: { queryCacheKey },
           }: PayloadAction<QuerySubstateIdentifier>,
         ) {
-          delete draft[queryCacheKey]
+          const entry = draft[queryCacheKey];
```
phryneas, line 294: The idea here is definitely to delete. It's called `removeQueryResult`, not `setQueryResultToUninitialized`.
From the back of my head, the store was never meant to contain `STATUS_UNINITIALIZED` in any way. That status is returned by selectors if the cache entry is not there.

xxiaoxiong:
Based on the feedback from @markerikson and @EskiMojo14, I understand that resetting to uninitialized instead of deleting is not the right approach for `removeQueryResult`. The concern about cache entries never being cleaned up is valid.

I'll close this PR. A proper fix for issue #5300 would require a more nuanced design around cache lifecycle management — perhaps tracking whether a removed-then-resubscribed query should show `isLoading` at all, rather than keeping entries around in the state.

Thank you for the review!

PR #5312, opened 2026-06-07
chore: update TypeScript to v6.0

### LINKED ISSUE #5292: chore: update TypeScript to v6.0
## Summary

- update TypeScript version to [**v6.0**](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0)
- raise the minimum supported TypeScript version to [**v5.5**](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5) (drop **`5.0`**-**`5.4`**)

aryaemami59:
## Summary

- Update TypeScript version to [**v6.0**](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0).
- Raise the minimum supported TypeScript version to [**v5.5**](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5) (drop **`5.0`**-**`5.4`**).
- Resolve [**#5292**](https://github.com/reduxjs/redux-toolkit/issues/5292).


EskiMojo14:
can we skip right to 7?

aryaemami59:
> can we skip right to 7?

~~good idea, I'll adjust it.~~

Looked into it, don't think we can yet. v7 doesn't expose the compiler API in its old form, it's landing incrementally behind **`typescript/unstable/*`**, so anything importing it breaks, including **`tsup`**'s dts generation. It also breaks **`@rtk-query/codegen-openapi`**, which has **`typescript`** as a runtime dependency, so that would break for consumers and not just for us.

PR #5318, opened 2026-06-18
docs: add Deno to install instructions

bartlomieju:
👋 Bartek from the Deno team here. Thanks for Redux Toolkit.

The "Getting Started" install section has npm and yarn tabs. This adds a Deno tab in parity (covering both `@reduxjs/toolkit` and the `react-redux` bindings), since Deno is a drop-in replacement for npm these days:

```bash
deno add @reduxjs/toolkit
```

Docs-only change. Happy to adjust wording or placement.

_Disclosure: I work on Deno, so I have an interest here. The goal is parity with the package managers you already list, not promotion. This PR was prepared with AI assistance under human supervision: I review every change myself and personally respond to any comments or review feedback._

### DIFF
--- a/docs/introduction/getting-started.md
+++ b/docs/introduction/getting-started.md
@@ -83,6 +83,19 @@ yarn add react-redux
 ```
 
   </TabItem>
+  <TabItem value="deno" label="deno">
+
+```bash
+deno add @reduxjs/toolkit
+```
+
+If you need React bindings:
+
+```bash
+deno add react-redux
+```
+
+  </TabItem>
 </Tabs>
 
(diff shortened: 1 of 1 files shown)


EskiMojo14:
not necessarily a bad change, but there are so many package managers now (npm, yarn, pnpm, bun, deno, aube, nub, vlt, vp) that it feels like a moving target if we were to try and cover them all. I honestly think more docs are going to adopt the approach of only including npm instructions - anybody using anything more bespoke should already know how to install npm packages using it.

aryaemami59 (approved):

markerikson:
Yeah, agreed. I have nothing against Deno here specifically, but we can't keep adding more entries indefinitely, and I don't expect that a user who's specifically chosen to use Deno as their preferred runtime is going to struggle with "how do I install this package?".

PR #5319, opened 2026-06-23
fix(combineSlices): scope stateProxyMap per instance to prevent cross-instance proxy collisions

JSap0914:
## Problem

`combineSlices.ts` declares `stateProxyMap` at module level:

```ts
const stateProxyMap = new WeakMap<object, object>()
```

This WeakMap is shared across **all** `combineSlices()` instances. The module-level `noopReducer` always returns the same `emptyObject` constant as initial state when no slices are given, so two empty `combineSlices()` calls both produce the **same object reference** as their initial state.

When a selector from instance A is called first with that shared state, a proxy is cached in the module-level `stateProxyMap` keyed by the state reference. A subsequent selector call from instance B with the same state reference retrieves the **cached proxy from instance A**, which closes over instance A's `reducerMap` and `initialStateCache` — so instance B's injected reducer's initial value is never consulted and the wrong value is returned.

```ts
const reducer1 = combineSlices()
const reducer2 = combineSlices()

// Both return the same  reference (noopReducer)
const preState = reducer1(undefined, { type: '@@INIT' })

reducer1.inject({ reducerPath: 'x', reducer: () => 'A' })
reducer2.inject({ reducerPath: 'x', reducer: () => 'B' })

const sel1 = reducer1.selector(s => s.x)
const sel2 = reducer2.selector(s => s.x)

sel1(preState) // 'A'  ✓
sel2(preState) // 'A'  ✗  — should be 'B'
```

## Fix

Move `stateProxyMap` inside `combineSlices` (one WeakMap per instance) and thread it as an explicit parameter through `createStateProxy`.  Each instance now has an isolated proxy cache, so selecting against a shared state object from a different instance creates a fresh proxy with the correct `reducerMap` and `initialStateCache`.

## Testing

A new unit test in `combineSlices.test.ts` reproduces the bug (fails without this fix, passes with it).  All 88 existing test files continue to pass.

> AI-assisted fix

### DIFF
--- a/packages/toolkit/src/combineSlices.ts
+++ b/packages/toolkit/src/combineSlices.ts
@@ -62,9 +62,8 @@ export type InjectConfig = {
 export interface CombinedSliceReducer<
   InitialState,
   DeclaredState extends InitialState = InitialState,
-  PreloadedState extends Partial<
-    Record<keyof PreloadedState, any>
-  > = Partial<DeclaredState>,
+  PreloadedState extends Partial<Record<keyof PreloadedState, any>> =
+    Partial<DeclaredState>,
 > extends Reducer<DeclaredState, UnknownAction, PreloadedState> {
   /**
    * Provide a type for slices that will be injected lazily.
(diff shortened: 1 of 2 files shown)


EskiMojo14:
seems edge case-y, but nonetheless a legitimate bug with a simple fix - thanks! (and thank you for disclosing that you're using AI for these, the honesty is appreciated :P)

PR #5326, opened 2026-07-01
fix(codegen): respect optional request bodies

### LINKED ISSUE #4824: `rtk-query-codegen-openapi` generates incorrect (required) types for optional request body
Example OpenAPI schema:
```json
{
  "openapi": "3.0.0",
  "info": {
    "title": "Example"
  },
  "paths": {
    "/Foo/{name}": {
      "delete": {
        "tags": [
          "Foo"
        ],
        "operationId": "deleteFoo",
        "parameters": [],
        "requestBody": {
          "content": {
            "*/*": {
              "schema": {
                "$ref": "#/components/schemas/deleteOptions"
              }
            }
          }
        }
      },
      "parameters": [
        {
          "name": "name",
          "in": "path",
          "description": "name of the Foo",
          "required": true,
          "schema": {
            "type": "string",
            "uniqueItems": true
          }
        }
      ]
    }
  },
  "components": {
    "schemas": {
      "deleteOptions": {
        "type": "object",
        "properties": {
          "apiVersion": {
            "type": "string"
          }
        }
      }
    }
  }
}
```

According to the [spec](https://swagger.io/docs/specification/v3_0/describing-request-body/describing-request-body/#requestbody-content-and-media-types), request bodies are optional by default. But the generated types look like:

```tsx
import { api } from './factory';
export const addTagTypes = ['Foo'] as const;
const injectedRtkApi = api
  .enhanceEndpoints({
    addTagTypes,
  })
  .injectEndpoints({
    endpoints: (build) => ({
      deleteFoo: build.mutation<DeleteFooApiResponse, DeleteFooApiArg>({
        query: (queryArg) => ({ url: `/Foo/${queryArg.name}`, method: 'DELETE', body: queryArg.deleteOptions }),
        invalidatesTags: ['Foo'],
      }),
    }),
    overrideExisting: false,
  });
export { injectedRtkApi as generatedApi };
export type DeleteFooApiResponse = unknown;
export type DeleteFooApiArg = {
  name: string;
  deleteOptions: DeleteOptions;
};
export type DeleteOptions = {
  apiVersion?: string;
};
```

Note how `deleteOptions` is required instead of optional in the `DeleteFooApiArg` type.

Apologies in advance if I've misread the spec/misunderstood something here! 🙇

MahinAnowar:
Fixes #4824.

## Problem

`rtk-query-codegen-openapi` hardcodes `required: true` for the request-body arg:

```ts
queryArg[name] = {
  origin: 'body',
  // ...
  required: true, // <- always required
  body,
};
```

As a result, **every** operation with a request body generates a non-optional arg (`body: T`), even when the OpenAPI document does not mark the body as required. Per the [OpenAPI 3.x spec](https://spec.openapis.org/oas/v3.0.3#request-body-object), `requestBody.required` **defaults to `false`**, so an un-required body should produce an optional arg (`body?: T`). Query and path params already respect this correctly via `param.required`; only the body branch was wrong.

## Fix

One line — use the resolved request body's own `required` flag, defaulting to the spec default:

```ts
required: body.required ?? false,
```

The `required` value flows into `createQuestionToken(!def.required)` (and the flat-arg `| undefined` branch), so the generated property becomes optional when appropriate.

## Tests

Added `test/fixtures/issue-4824.json` (a `deleteFoo` op with a non-required body and a `createBar` op with `required: true`) plus a regression test in `generateEndpoints.test.ts`:

- non-required body → `DeleteFooApiArg = { body?: … }`
- `required: true` body → `CreateBarApiArg = { body: … }` (stays required)

The test **fails on the current code** (the non-required body is emitted as `body:`) and **passes with the fix**.

## Snapshot updates

The existing snapshots captured the buggy output, so 15 updated. Every changed line is a request-body arg becoming optional (`user: User` → `user?: User`, `order: Order` → `order?: Order`, `body: Blob` → `body?: Blob`, etc.). Notably, petstore's `pet` body — which **is** marked `required: true` — stays `pet: Pet`, confirming the change only affects bodies the spec leaves optional.

The full package test suite (`yarn test`, incl. typecheck) passes and Prettier is clean.

### DIFF
--- a/packages/rtk-query-codegen-openapi/src/generate.ts
+++ b/packages/rtk-query-codegen-openapi/src/generate.ts
@@ -444,7 +444,8 @@ export async function generateApi(
         name,
         originalName: schemaName,
         type: getTypeFromSchema(withMode(ctx, 'writeOnly'), schema),
-        required: true,
+        // A request body is optional unless the spec explicitly marks it required (OpenAPI defaults `required` to `false`).
+        required: body.required ?? false,
         body,
       };
     }
--- a/packages/rtk-query-codegen-openapi/test/__snapshots__/cli.test.ts.snap
+++ b/packages/rtk-query-codegen-openapi/test/__snapshots__/cli.test.ts.snap
@@ -147,7 +147,7 @@ export type UploadFileApiArg = {
   petId: number;
   /** Additional Metadata */
   additionalMetadata?: string;
-  body: Blob;
+  body?: Blob;
 };
 export type GetInventoryApiResponse = /** status 200 successful operation */ {
   [key: string]: number;
@@ -155,7 +155,7 @@ export type GetInventoryApiResponse = /** status 200 successful operation */ {
 export type GetInventoryApiArg = void;
 export type PlaceOrderApiResponse = /** status 200 successful operation */ Order;
 export type PlaceOrderApiArg = {
-  order: Order;
+  order?: Order;
 };
 export type GetOrderByIdApiResponse = /** status 200 successful operation */ Order;
 export type GetOrderByIdApiArg = {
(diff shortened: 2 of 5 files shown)


EskiMojo14 (approved):
this seems potentially breaking, but in a bug-fix-y way?

do we anticipate anybody experiencing breaking changes in their codebase that might not expect it?

i guess the fix would just be to explicitly specify the body as required

EskiMojo14:
took a look at the linked issue, looks like this is expected - merged

PR #5334, opened 2026-07-13
fix(query): skip fetch when infinite query trigger is called with `skipToken`

### LINKED ISSUE #5028: `skipToken` is passed into my `infiniteQuery` as a `queryArg`
### Discussed in https://github.com/reduxjs/redux-toolkit/discussions/5027

<div type='discussions-op-text'>

<sup>Originally posted by **soggybeef** July  4, 2025</sup>
Language: Typescript
Version:  `"@reduxjs/toolkit": "^2.3.0",`

I was trying to conditionally run an infinite query to prevent the first request from being made upon mounting of the component. The `args` will be taken from `router` which is from `next/router`

I have the args for `queryArg` as `shouldSkip ? skipToken : args`. But when `shouldSkip === true` and `skipToken` is passed in, the first request is still being made, and this is what i got from my console logs when forming the query in the `services` page where the endpoint is defined:

image

Since the logic of forming my `queryParams` doesn't and shouldn't take into account the `skipToken`, the request is still being sent in to fetch data that i don't want to fetch. This data is returned, but not returned or populated to my parent component. But I see it under the Responses in my Network Requests.

image

Would love any tips and corrections on how I can better handle the `skipToken` in my `infiniteQuery`, thank you!

EDIT: I see similar behavior when using `skip: shouldSkip` as part of the second argument in my `useInfiniteQuery`; the original `args` that I had became the `queryArgs` and the request is sent and returned, but not poulated.</div>

xianjianlf2:
## Problem

When an infinite query hook is skipped (via `skipToken` or the `skip` option), calling `fetchNextPage()`/`fetchPreviousPage()` (or the lazy `trigger`) dispatched `initiate(skipToken)`, which fired a real request using `skipToken` as the query arg and created a bogus cache entry.

## Fix

Guard the shared infinite-query trigger so that when it is called with a skipped `arg` (`skipToken`), it unsubscribes any previous promise and bails out without dispatching, matching the auto-subscription behavior. This makes `fetchNextPage()`/`fetchPreviousPage()` on a skipped hook a no-op instead of firing a request with `skipToken` as the query arg.

```ts
if (arg === skipToken) {
  unsubscribePromiseRef(promiseRef)
  promiseRef.current = undefined
  return promiseRef.current as unknown as InfiniteQueryActionCreatorResult<any>
}
```

## Testing

Added a test in `packages/toolkit/src/query/tests/buildHooks.test.tsx` verifying that triggering `fetchNextPage()`/`fetchPreviousPage()` on a skipped infinite query does not dispatch a request or create a cache entry.

Closes #5028

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -2103,6 +2103,16 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
 
       const trigger: LazyInfiniteQueryTrigger<any> = useCallback(
         function (arg: unknown, direction: 'forward' | 'backward') {
+          // If the query is skipped (`arg` is `skipToken`), we must not start a
+          // request. Unsubscribe from any previous promise and bail out, so that
+          // e.g. calling `fetchNextPage()` on a skipped hook is a no-op instead
+          // of firing a request with `skipToken` as the query arg.
+          if (arg === skipToken) {
+            unsubscribePromiseRef(promiseRef)
+            promiseRef.current = undefined
+            return promiseRef.current as unknown as InfiniteQueryActionCreatorResult<any>
+          }
+
           let promise: InfiniteQueryActionCreatorResult<any>
 
           batch(() => {
--- a/packages/toolkit/src/query/tests/buildHooks.test.tsx
+++ b/packages/toolkit/src/query/tests/buildHooks.test.tsx
@@ -2543,6 +2543,50 @@ describe('hooks tests', () => {
       },
     )
 
+    test.each([
+      ['skip token', skipToken, undefined],
+      ['skip option', 'fire', true],
+    ])(
+      'useInfiniteQuery does not fetch when `fetchNextPage`/`fetchPreviousPage` are called while skipped via %s',
+      async (_, arg, skip) => {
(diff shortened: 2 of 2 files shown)


xianjianlf2:
Note on the two red **Test Types: TS next** checks: they fail identically on upstream master — run [29339252283](https://github.com/reduxjs/redux-toolkit/actions/runs/29339252283) (commit 58e8cfe, before this PR's CI run) shows the same 5 errors with `typescript@7.1.0-dev.20260714.1`: `buildHooks.ts(1859)` TS2769 and the `mapBuilders.test-d.ts` TS2578/TS2769 set. This PR doesn't touch either location (its only src change is around line 2103), so the failures are nightly-TypeScript breakage unrelated to this change.

xianjianlf2:
Synced with current `master` in `26f7a2e6` and applied the repository’s new oxfmt style in `bea1c344`. The merge includes upstream’s official TS 7 compatibility fix (`974a96f3`), which addresses the two unrelated TS-next failures from the prior CI run. The PR diff remains limited to the original two infinite-query files.

Validation on the current pnpm/oxlint/oxfmt toolchain:
- focused `buildHooks` suite: 71 tests passed, no type errors
- toolkit type tests
- oxlint
- full formatting check
- toolkit production build
- `git diff --check`

Fresh CI is now running against the current TypeScript toolchain.

PR #5335, opened 2026-07-15
docs: fix `selectFromResult` example data access

dfedoryshchev:
The `selectFromResult` JSDoc example destructures the query data as `const { data: posts } = api.useGetPostsQuery()`, so `posts` is already the endpoint's data. It then maps over `posts?.data?.map(...)`, dereferencing a `data` property that does not exist on the unwrapped value; the same example accesses it directly as an array with `data?.find(...)`.

Fixed the example to map `posts?.map(...)`.


EskiMojo14 (approved):

EskiMojo14:
thanks!

PR #5340, opened 2026-07-16
docs: document sending FormData / multipart requests with fetchBaseQuery

### LINKED ISSUE #5213: `formData` is required for `multipart/form-data` requests but this behaviour is undocumented
Is there a particular reason that `formData` is not documented anywhere? formData returns no results when searching on the RTK Query docs, I was only able to discover its existence from comments in threads like [this one](https://github.com/reduxjs/redux-toolkit/issues/2677#issuecomment-1619864459).

An example of use ([from here](https://notes.zachmanson.com/rtk-query-sending-form-data/)):

```typescript
    uploadSiteDocuments: build.mutation<{ file: string; uri: string }[], { files: File[] }>({
      query: (body) => {
        let formData = new FormData();
        body.files.forEach((file) => {
          formData.append(file.name, file);
        });

        return {
          url: "site/documents",
          method: "POST",
          body: formData,
          formData: true,
        };
      },
      transformResponse: (response: any) => response.data,
    }),
```

KAMRONBEK:
## Problem

`multipart/form-data` requests with `fetchBaseQuery` are completely undocumented — `FormData` and `multipart` return zero results across `docs/`. As a result, users copy a nonexistent `formData: true` request option from stray issue comments and blog posts (see #5213, where @phryneas confirmed the flag "doesn't seem to exist" — it appears nowhere in the source).

## Fix

Adds a `#### FormData` subsection to the `fetchBaseQuery` API page, under **Individual query options → Setting the body**, next to the existing `json` and `text` subsections. It shows the positive pattern:

- pass a `FormData` instance directly as `body` (no dedicated option needed — with a one-line note that there is no `formData` flag),
- `fetchBaseQuery` passes it through to `fetch` as-is and ensures no `Content-Type` header is set, so `fetch` generates the `multipart/form-data` content type with the correct boundary automatically,
- notes there is no need to set `Content-Type` manually — a manual value would be missing the boundary, so `fetchBaseQuery` removes it for `FormData` bodies.

This matches the actual implementation: `isJsonifiable` excludes `FormData`, and since #5112 `fetchBaseQuery` removes any `content-type` header for non-jsonifiable, non-string bodies (covered by the existing `FormData` tests in `packages/toolkit/src/query/tests/fetchBaseQuery.test.tsx`).

## Tests

Docs-only change. `yarn prettier --check docs/rtk-query/api/fetchBaseQuery.mdx` passes.

Fixes #5213

### DIFF
--- a/docs/rtk-query/api/fetchBaseQuery.mdx
+++ b/docs/rtk-query/api/fetchBaseQuery.mdx
@@ -287,6 +287,27 @@ By default, `fetchBaseQuery` assumes that every request you make will be `json`,
     }),
 ```
 
+#### FormData
+
+To send `multipart/form-data` requests, such as file uploads, pass a [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) instance directly as the `body`. No dedicated option is needed (there is no `formData` flag) &mdash; `fetchBaseQuery` passes the `FormData` through to `fetch` as-is and ensures no `Content-Type` header is set, so that `fetch` can generate the `multipart/form-data` content type with the correct boundary automatically. There is no need to set the `Content-Type` header manually for these requests &mdash; a manually set value would be missing the boundary, so `fetchBaseQuery` removes it for `FormData` bodies.
+
+```ts no-transpile
+ // omitted
+  endpoints: (build) => ({
+    uploadFile: build.mutation({
+      query: (file: File) => {
+        const formData = new FormData()
+        formData.append('file', file)
+        return {
+          url: `files`,
+          method: 'POST',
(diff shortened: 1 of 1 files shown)


KAMRONBEK:
@phryneas TL;DR for #5213 (you confirmed there's no `formData` flag): adds a `#### FormData` subsection to the fetchBaseQuery page next to the `json`/`text` ones showing the positive pattern — pass a `FormData` instance directly as `body`; fetch generates the multipart boundary; no need to set `Content-Type` manually (fetchBaseQuery removes it for FormData bodies, matching the #5112 behavior and the existing tests).

Issue #5093, opened 2025-09-30
Pessimistic Updates: isSuccess is always false


anotheri:
I'm using `"@reduxjs/toolkit": "^2.9.0"` and  follow the [pessimistic updates example](https://redux-toolkit.js.org/rtk-query/usage/manual-cache-updates#pessimistic-updates).

That's how my `updateItem` function defined in api:

```typescript
    updateItem: build.mutation<ItemResponse, ItemUpdatePayload>({
      query: (payload) => {
        const { id, ...body } = payload;
        return {
          url: `/items/${id}`
          method: 'PUT',
          body,
        };
      },

     async onQueryStarted ({ id }, { dispatch, queryFulfilled }) {
        try {
          const result = await queryFulfilled;

          // pessimistic update
          dispatch(
            api.util.updateQueryData('getItem', id, (draft) => {
              Object.assign(draft, result.data);
            }),
          );
        } catch (error) {
          console.error('Update Item error:', error);
        }
      },
      invalidatesTags: ['Items'], // i've tried do disable this but it doesn't matter, this instruction invalidates the cache of the item list
    }),
```

That's how i use the mutation hook in react component:

```typescript
const [updateItem, { isLoading, isSuccess }] = useUpdateItemMutation();
console.debug('isSuccess:', isSuccess); // ← this always logs `false` value
```

The cache of `getItem` is getting updated correctly, but the `isSuccess` flag doesn't change with pessimistic updates at all, it's always `false`, so I unfortunately can't rely on it to show visual feedback on successful update. The workaround is to track it manually in local component state after the `await updateItem(values).unwrap();` call, but i'd obviously prefer to avoid it.

On other hand, `isSuccess` changes as expected when i use optimistic updates with code from [another documentation example](https://redux-toolkit.js.org/rtk-query/usage/manual-cache-updates#optimistic-updates) like this:

```typescript
    updateItem: build.mutation<ItemResponse, ItemUpdatePayload>({
      query: (payload) => {
        const { id, ...body } = payload;
        return {
          url: `/items/${id}`
          method: 'PUT',
          body,
        };
      },
     async onQueryStarted ({ id, ...patch }, { dispatch, queryFulfilled }) {
        // Optimistic update
        const patchResult = dispatch(
          api.util.updateQueryData('getItem', id, (draft) => {
            Object.assign(draft, patch);
          }),
        );

        try {
          await queryFulfilled;
        } catch (error) {
          patchResult.undo();
[...]

PR #5341, opened 2026-07-16
test(query): add regression test for mutation isSuccess with pessimistic updates

KAMRONBEK:
## Summary

Investigation of #5093 ("Pessimistic Updates: `isSuccess` is always `false`"). **The reported behavior does not reproduce at the library level** — this PR adds a regression test that pins the correct behavior of the documented pessimistic-update pattern, and documents the investigation for triage.

## Investigation

1. **Hook-level repro (this PR's test):** a mutation whose `onQueryStarted` awaits `queryFulfilled` and then dispatches `api.util.updateQueryData(...)`, combined with `invalidatesTags` — exactly the [documented pessimistic-update pattern](https://redux-toolkit.js.org/rtk-query/usage/manual-cache-updates#pessimistic-updates). The `useMutation` hook's `isSuccess` becomes `true`, the cache patch is applied, and `isSuccess` stays `true` after the invalidation refetch settles. Passes on current master.
2. **Reporter's exact environment:** the same repro run against published `@reduxjs/toolkit@2.9.0` + `react@18.3.1` also passes, so this is not something that was fixed since 2.9.0.
3. **Store level:** the mutation substate provably reaches `status: 'fulfilled'`. Auditing `buildSlice.ts`, nothing in the pessimistic flow (`queryResultPatched`, `invalidateTags`, invalidation refetches) touches mutation substate — the only ways it is cleared are hook unmount (`promise.reset()`) and manual `reset()`.
4. **The in-thread `invalidateTags`-ordering "workaround":** instrumented action/render traces show the before/after-`updateQueryData` asymmetry appears or disappears purely with React flush timing (whether the `isSuccess: true` render commits before an invalidation-refetch-driven re-render/unmount). It has no mechanism inside RTK Query.

## Conclusion

"`isSuccess` always `false`" in an app means the `useMutation` hook instance was reset — i.e. the component holding it unmounted/remounted (for example, a parent that hides the form while the invalidated list query is refetching). Mutation hook state is per-instance by design and removed on unmount; sharing results across mounts is what [`fixedCacheKey`](https://redux-toolkit.js.org/rtk-query/usage/mutations#shared-mutation-results) is for. The new test guards the library-level behavior so future regressions of the documented pattern are caught.

Note: the test intentionally passes on unmodified master — there is no code fix because there is no library bug.

## Tests

- `yarn workspace @reduxjs/toolkit test src/query/tests/optimisticUpdates.test.tsx` — 10/10 pass, no type errors
- `yarn eslint packages/toolkit/src/query/tests/optimisticUpdates.test.tsx` / prettier — clean

Refs #5093

### DIFF
--- a/packages/toolkit/src/query/tests/optimisticUpdates.test.tsx
+++ b/packages/toolkit/src/query/tests/optimisticUpdates.test.tsx
@@ -397,6 +397,95 @@ describe('updateQueryData', () => {
   })
 })
 
+describe('pessimistic updates', () => {
+  // regression test for https://github.com/reduxjs/redux-toolkit/issues/5093
+  test('mutation hook reports isSuccess with pessimistic update and tag invalidation', async () => {
+    const extendedApi = api.injectEndpoints({
+      endpoints: (build) => ({
+        updatePostPessimistic: build.mutation<
+          Post,
+          Pick<Post, 'id'> & Partial<Post>
+        >({
+          query: ({ id, ...patch }) => ({
+            url: `post/${id}`,
+            method: 'PUT',
+            body: patch,
+          }),
+          async onQueryStarted({ id }, { dispatch, queryFulfilled }) {
+            const { data } = await queryFulfilled
+            dispatch(
+              api.util.updateQueryData('post', id, (draft) => {
+                Object.assign(draft, data)
+              }),
+            )
+          },
+          invalidatesTags: ['Post'],
+        }),
+      }),
+      overrideExisting: true,
+    })
+
+    baseQuery
+      // initial query
+      .mockResolvedValueOnce({
+        id: '3',
+        title: 'All about cheese.',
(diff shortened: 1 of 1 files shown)


KAMRONBEK:
@markerikson @phryneas TL;DR: investigated #5093 ("pessimistic updates: isSuccess always false") and it does NOT reproduce at the library level — the documented pattern works on master and on the reporter's exact published stack (RTK 2.9.0 + React 18.3.1). Store-level mutation substate provably reaches 'fulfilled'; the in-thread invalidateTags-ordering 'workaround' turns out to be a React flush-timing artifact; the realistic user-land cause is the mutation-holding component unmounting (per-instance state reset — `fixedCacheKey` territory). This PR pins the correct behavior with a regression test and documents the analysis for triage; equally happy if you'd rather just close #5093 as works-as-designed with this evidence.

PR #5343, opened 2026-07-20
fix(query): don't reset `keepUnusedDataFor` on condition-skipped queries

### LINKED ISSUE #4480: initiate with {subscribe: false} resets keepUnusedDataFor
Calling `initiate` with `{subscribe: false}` seems to be resetting `keepUnusedDataFor` and cached data never expires

```
// default keepUnusedDataFor is 60s

// Does the requests and caches the data as expected
store.dispatch(api.endpoints.myEndpoint.initiate({}, { subscribe: false }))

// Does not trigger a request and uses cached data as expected
setTimeout(() => store.dispatch(api.endpoints.myEndpoint.initiate({}, { subscribe: false })), 30 * 1000);

// Does not trigger a request and uses cached data which should be expired
setTimeout(() => store.dispatch(api.endpoints.myEndpoint.initiate({}, { subscribe: false })), 61 * 1000);
```

veksa:
Fixes #4480

If you keep dispatching `initiate` with `{ subscribe: false }` for an arg that's
already cached, the data never gets removed. Each call pushes the expiry back, so
`keepUnusedDataFor` effectively turns into "keep forever" as long as something
polls that endpoint without subscribing.

Repro from the issue:

```ts
// default keepUnusedDataFor is 60s

// fetches and caches the data
store.dispatch(api.endpoints.myEndpoint.initiate({}, { subscribe: false }))

// 30s later — served from cache, no request
setTimeout(() => store.dispatch(
  api.endpoints.myEndpoint.initiate({}, { subscribe: false })
), 30_000)

// 61s later — data should be gone, but it's still cached
setTimeout(() => store.dispatch(
  api.endpoints.myEndpoint.initiate({}, { subscribe: false })
), 61_000)
```

## Cause

The cache-collection middleware reschedules the removal timer whenever
`canTriggerUnsubscribe` matches, and it was matching `queryThunk.rejected` as a
whole.

When `initiate(..., { subscribe: false })` hits a fresh cache entry, the thunk's
`condition` returns `false` and bails out before running. But the thunk uses
`dispatchConditionRejection: true`, so it still dispatches a `rejected` action —
one tagged with `meta.condition === true`. That made a plain cache hit look like
a finished request: the pending removal timeout got cleared and a fresh one
started, even though nothing was fetched and no subscription changed.

## Fix

Ignore condition-based rejections when deciding whether to reschedule cleanup. A
query that bailed out via `condition` never ran and never touched a
subscription, so it shouldn't restart the removal timer. Real failures of
unsubscribed queries (`meta.condition === false`) still schedule removal as
before:

```ts
const isQueryThunkRejectedByError = (action: any) =>
  queryThunk.rejected.match(action) && !action.meta.condition
```

I also added a regression test that fires two `initiate({ subscribe: false })`
calls 30s apart and checks the entry is removed on the original deadline instead
of being pushed out.

### DIFF
--- a/packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/cacheCollection.ts
@@ -65,10 +65,21 @@ export const buildCacheCollectionHandler: InternalHandlerBuilder = ({
   const { removeQueryResult, unsubscribeQueryResult, cacheEntriesUpserted } =
     api.internalActions
 
+  // A `condition`-based rejection means the query bailed out before it ever ran
+  // (e.g. `initiate(arg, { subscribe: false })` served fresh cached data). Nothing
+  // was fetched and no subscription changed, so we must not treat it as an
+  // unsubscribe trigger: rescheduling the removal timer here resets
+  // `keepUnusedDataFor` on every such call and prevents unused data from ever
+  // being collected. See https://github.com/reduxjs/redux-toolkit/issues/4480
+  const isQueryThunkRejectedByError = (
+    action: any,
+  ): action is ReturnType<typeof queryThunk.rejected> =>
+    queryThunk.rejected.match(action) && !action.meta.condition
+
   const canTriggerUnsubscribe = isAnyOf(
     unsubscribeQueryResult.match,
     queryThunk.fulfilled,
-    queryThunk.rejected,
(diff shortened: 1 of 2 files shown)


EskiMojo14 (approved):

EskiMojo14:
the changes make sense to me, but i'd like a second pair of eyes on this

PR #5345, opened 2026-07-20
[3.0] Remove Action generic argument from `configureStore`

### LINKED ISSUE #5317: [3.0] remove `Action` generic argument from `configureStore`
We strongly discourage any other type than `UnknownAction` for a long time now.

This being there leads people in the wrong direction, e.g. see https://github.com/reduxjs/redux/issues/4887

veksa:
Closes #5317.

Drops the `Action` generic (`A extends Action = UnknownAction`) from `configureStore`, `ConfigureStoreOptions` and `EnhancedStore`.

Anything other than `UnknownAction` for the store's action type has been discouraged for ages, but having the generic there kept inviting the opposite. People assume that if an action type can be passed in, it should flow through `dispatch` and into middleware — then get surprised when it doesn't (reduxjs/redux#4887 is a good example). Removing the argument makes the contract obvious: the store always dispatches `UnknownAction`, and you narrow with type guards where needed.

### What changed

The three types now have one fewer generic — `configureStore<S, M, E, P>` and `EnhancedStore<S, E>` — and the store is always `Store<S, UnknownAction, …>`.

One thing worth calling out: the `reducer` option uses `any` in the action slot (`Reducer<S, any, P> | ReducersMapObject<S, any, P>`) rather than `UnknownAction`. `UnknownAction` there rejects perfectly good reducers typed for a specific action — including anything from `combineReducers`, since that keeps a concrete action union — because of parameter contravariance under `strictFunctionTypes`. `any` keeps accepting whatever reducer gets passed while the store type itself stays `UnknownAction`.

The API docs are updated to match, and the `middleware`/`enhancers` descriptions are fixed too — they still claimed an array could be passed, which hasn't been true since 2.x (a maintainer pointed this out in redux#4887 as well).

### Breaking change

Aimed at 3.0. Nothing changes if the action type was never written out explicitly — the default was already `UnknownAction`. It only bites something like `configureStore<State, MyAction>(…)` or an explicit `EnhancedStore<State, MyAction, …>` annotation, since dropping the parameter shifts the remaining generics along. The fix is to remove the action type argument.

### Testing

Type checks and the `configureStore` typecheck suite pass, and the reworked type test confirms a specifically-typed reducer is still accepted and just produces an `UnknownAction` store.

### DIFF
--- a/docs/api/configureStore.mdx
+++ b/docs/api/configureStore.mdx
@@ -40,7 +40,6 @@ Redux Toolkit's `configureStore` simplifies that setup process, by doing all tha
 
 interface ConfigureStoreOptions<
   S = any,
-  A extends Action = UnknownAction,
   M extends Tuple<Middlewares<S>> = Tuple<Middlewares<S>>
   E extends Tuple<Enhancers> = Tuple<Enhancers>,
   P = S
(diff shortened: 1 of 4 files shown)


EskiMojo14:
> One thing worth calling out: the reducer option uses any in the action slot (Reducer<S, any, P> | ReducersMapObject<S, any, P>) rather than UnknownAction. UnknownAction there rejects perfectly good reducers typed for a specific action — including anything from combineReducers, since that keeps a concrete action union — because of parameter contravariance under strictFunctionTypes. any keeps accepting whatever reducer gets passed while the store type itself stays UnknownAction.

Seems like the rejection would be correct? Reducers should not be typed to accept specific actions, that's the point of this change. Reducers can be called with *any* action (and will be), and should return state unchanged if the action is unrecognised.

veksa:
> Seems like the rejection would be correct? Reducers should not be typed to accept specific actions, that's the point of this change. Reducers can be called with any action (and will be), and should return state unchanged if the action is unrecognised.

Good point - you're right, I'll switch it to UnknownAction.

PR #5347, opened 2026-07-22
fix(toolkit): preserve unsubscribed results during invalidation

### LINKED ISSUE #5346: Query with `subscribed: false` returns undefined due to mutation race
Steps:

1. Create an API with a query and a mutation sharing the same cache tag
2. Start a request for the query with `subscribe: false`
3. Start a request for the mutation
4. If the mutation completes right before the query, the query returns `undefined`, breaking its API contract (types claim that `undefined` is never returned)

Here is a minimal reproducer: https://gist.github.com/emersion/1459bcabb66c5bb2c6c72f32f07fce3f

To run it:

```bash
git clone https://gist.github.com/1459bcabb66c5bb2c6c72f32f07fce3f.git rtk-query-unsubrace
cd rtk-query-unsubrace
npm install
npm start
```

The delay might need to be adjusted depending on the host machine. On my machine, the hardcoded delay is enough to hit the race consistently. If you can't reproduce with the hardcoded delay, use the random delay and let the script run for a while.

Setting `subscribe: true` and calling `unsubscribe()` after awaiting the query works around the bug. Calling `unsubscribe()` any earlier is not enough.

rajanpanth:
## Summary

- preserve the fulfilled result for an unsubscribed query when delayed invalidation runs during fulfillment
- defer cache removal until the running query settles
- add deterministic regression coverage for the invalidation race

Fixes #5346.

## Problem

A query initiated with `subscribe: false` can return an uninitialized result when a mutation triggers delayed invalidation while the query is still running.

During the fulfilled dispatch, invalidation sees the cache entry without an active subscription and removes it synchronously. `buildInitiate` then selects the public result after the cache entry has disappeared, so the initiating caller cannot observe the fulfilled data.

## Root cause

The invalidation middleware removes an unsubscribed query result before the running query promise and initiating result-selection flow have settled.

## Change

When the affected cache entry still has a running query, defer cleanup until that query settles.

Before cleanup, recheck:

- the current cache entry still belongs to the original `requestId`
- whether a newer request replaced it
- whether the entry gained an active subscription

An unchanged unsubscribed entry is removed. A newly subscribed entry is refetched, and a newer request is preserved.

## Validation

- focused issue regression — passed
- race-condition test file — 3/3 passed
- race-condition test file on Node 22.16 — 3/3 passed
- Redux Toolkit package suite — 88/88 files and 1,317 tests passed
- type tests and TypeScript validation — passed
- targeted ESLint and Prettier — passed
- production build — passed
- `git show --check` — passed

The broad package lint command references an absent `packages/toolkit/examples` directory. The optional size-limit command also remains blocked by Windows absolute-path resolution. The changed files pass targeted validation and the production build.

Rejected and aborted query results use RTK Query's non-rejecting result promise, so deferred cleanup does not introduce an unhandled rejection. No separate dedicated rejection or abort regression test is included in this change.

## Compatibility

No public API, dependency, lockfile, Node support, browser API, or TypeScript declaration change is included.

### DIFF
--- a/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/invalidationByTags.ts
@@ -122,11 +122,47 @@ export const buildInvalidationByTagsHandler: InternalHandlerBuilder = ({
 
         if (querySubState) {
           if (subscriptionSubState.size === 0) {
-            mwApi.dispatch(
-              removeQueryResult({
-                queryCacheKey: queryCacheKey as QueryCacheKey,
-              }),
-            )
+            const runningQuery = internalState.runningQueries.get(queryCacheKey)
+
+            if (runningQuery) {
+              const requestId = querySubState.requestId
+
+              void runningQuery.then(() => {
+                const currentState = mwApi.getState()[reducerPath]
+                const currentQuerySubState = currentState.queries[queryCacheKey]
+
+                if (
+                  !currentQuerySubState ||
+                  currentQuerySubState.requestId !== requestId
+                ) {
+                  return
+                }
+
+                const currentSubscriptionSubState = getOrInsertComputed(
+                  internalState.currentSubscriptions,
+                  queryCacheKey,
+                  createNewMap,
+                )
+
+                if (currentSubscriptionSubState.size === 0) {
+                  mwApi.dispatch(
+                    removeQueryResult({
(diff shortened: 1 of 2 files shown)


rajanpanth:
Friendly follow-up on this PR 👋

When you have a moment, could you please take a look? I’m happy to rebase, update tests, or make any requested changes.

Thanks!

PR #5348, opened 2026-07-23
docs: clarify `useLazyQuery` data is undefined until trigger

### LINKED ISSUE #4094: [RTKQ Query] Documentation for `useLazyQuery`'s `data` field
Hello.
As a newbie to rtk-query, I found the official documentation for `useLazyQuery` a bit confusing while trying to apply it to my project.

In the [description](https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#description-4), it says the following about `data`.
> Returns the latest request status and cached data from the Redux store

When I first read this description, I thought that `data` always tries to read the cached values from the store, regardless of whether the trigger is fired or not.

After running it myself, I realized that it was describing the behavior after the trigger.

English isn't my first language, so I'm not 100% sure, but is it just me or is this documentation confusing?
If so, is it okay to write a PR to clarify this?

Hashim1999164:
## Summary
- Clarify that `useLazyQuery` keeps `data` undefined and the result `uninitialized` until `trigger` is called, even when matching cache data already exists.
- Update the `UseLazyQuery` JSDoc and the hooks API page Returns section so they match that behavior.

Fixes #4094

## Test plan
- [ ] Review the `UseLazyQuery` summary rendered on the hooks API docs page
- [ ] Confirm the wording matches known lazy-query behavior (no store data until first trigger)

### DIFF
--- a/docs/rtk-query/api/created-api/hooks.mdx
+++ b/docs/rtk-query/api/created-api/hooks.mdx
@@ -630,7 +630,7 @@ type UseLazyQueryLastPromiseInfo = {
 
 - **Returns**: A tuple containing:
   - `trigger`: A function that fetches the corresponding data for the endpoint when called
-  - `result`: A query result object containing the current loading state, the actual data or error returned from the API call and metadata about the request. Can be customized with `selectFromResult`
+  - `result`: A query result object containing the current loading state, the actual data or error returned from the API call and metadata about the request. Can be customized with `selectFromResult`. Before the first `trigger` call, this stays `uninitialized` and `data` is `undefined` even if the cache already has matching data.
   - `lastPromiseInfo`: An object containing the last argument used to call the trigger function
 
 ### `usePrefetch`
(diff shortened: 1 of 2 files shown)


docs/rtk-query/api/created-api/hooks.mdx
```diff
@@ -630,7 +630,7 @@ type UseLazyQueryLastPromiseInfo = {
 
 - **Returns**: A tuple containing:
   - `trigger`: A function that fetches the corresponding data for the endpoint when called
-  - `result`: A query result object containing the current loading state, the actual data or error returned from the API call and metadata about the request. Can be customized with `selectFromResult`
+  - `result`: A query result object containing the current loading state, the actual data or error returned from the API call and metadata about the request. Can be customized with `selectFromResult`. Before the first `trigger` call, this stays `uninitialized` and `data` is `undefined` even if the cache already has matching data.
```
phryneas: > even if the cache already has matching data.

This sentence is misleading as it implies that matching cache data *could* exist.

But it cannot - not calling `trigger` means that `arg` has never been provided, and that means that a cache entry could never be determined.
There is nothing to match for.

Hashim1999164:
Removed the misleading cache-data clause from both the generated API docs source and the reference docs. The text now states only that the lazy query remains uninitialized with undefined data before the first trigger.

PR #5350, opened 2026-07-24
docs: fix `getOriginalState` example to use `Promise.resolve`

dfedoryshchev:
The `getOriginalState` `@example` calls `Promise().resolve()`, which throws at runtime (`Promise` must be called with `new`) and doesn't type-check (TS2348: `PromiseConstructor` is not callable). The intended microtask boundary is `Promise.resolve()`.


EskiMojo14:
thanks!

PR #5351, opened 2026-07-25
feat(utils): add truncateMiddle helper export to toolkit utils

mahenoorsalat:
## Description
Adds a type-safe \	runcateMiddle(str, maxLength)\ string formatting helper function to \packages/toolkit/src/utils.ts\ for cleanly formatting action types, devtools logger labels, and RTK Query endpoint keys.

## Features
- Middle string truncation with \...\ ellipsis.
- Exported from Redux Toolkit core utilities.


markerikson:
I don't see any reason to add this, and it's not even hooked up to anything.

PR #5356, opened 2026-08-04
chore(toolkit): remove redundant and unused type declarations

aryaemami59:
## Summary

- remove a duplicate overload from **`ConditionFunction`**
- return **`TypedCreateListenerEntry`** instead of **`TypedStopListening`** from **`TypedCreateListenerEntry.withTypes()`**
- fold the trailing **`& { effect }`** intersection into **`FallbackAddListenerOptions`**
- remove the redundant leading **`RefetchConfigOptions &`** from **`ConfigState`**, which **`ModifiableConfigState`** already intersects
- remove the unused generic type parameter from **`AnyQueryActionCreator`**

### DIFF
--- a/packages/toolkit/src/createReducer.ts
+++ b/packages/toolkit/src/createReducer.ts
@@ -191,7 +191,7 @@ export function createReducer<S extends NotFunction<any>>(
             if (previousState === null) {
               return previousState
             }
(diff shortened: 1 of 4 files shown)


EskiMojo14 (approved):
very nice

PR #5400, opened 2026-08-30
Fix entity adapters for Object.prototype-named IDs

### LINKED ISSUE #5399: Entity adapters mishandle Object.prototype-named IDs
### Reproduction

```ts
const adapter = createEntityAdapter<{ id: string }>()
const state = adapter.addMany(adapter.getInitialState(), [
  { id: "toString" },
  { id: "__proto__" },
  { id: "normal" },
])

console.log(state.ids) // ["normal"]
console.log(adapter.getSelectors().selectById(state, "constructor")) // inherited Function
```

This reproduces with both sorted and unsorted adapters on current `master` (`74200d87`).

### Cause

Entity dictionaries and temporary ID maps are ordinary `{}` objects, while existence checks use `in`. Prototype-named IDs are therefore mistaken for existing entities; assigning `__proto__` is also unsafe on ordinary objects. `selectById` directly exposes inherited properties.

### Expected

Any valid string `EntityId`, including names present on `Object.prototype`, should be stored, updated, removed, and selected like any other ID. Missing IDs should return `undefined`.

I have a focused fix and regression coverage for both adapter variants and can submit it immediately.

OskarEichler:
## Fixes

- Store, update, remove, and select valid entity IDs such as `toString`, `constructor`, and `__proto__` in both sorted and unsorted adapters.
- Use own-property checks instead of inherited-property checks for entity membership.
- Keep temporary ID dictionaries prototype-safe without changing the public entity-state object prototype.
- Return `undefined` from `selectById` for missing prototype-named IDs instead of exposing inherited functions.

Closes #5399.

## Verification

- The untouched baseline drops all prototype-named entities and returns an inherited function for a missing `valueOf` ID.
- Added full add/select/update/remove regressions for both adapter variants.
- `yarn workspace @reduxjs/toolkit test` — 88 files, 1,318 passing tests, two existing todos, zero type errors.
- Fixed standalone reproduction passes for sorted and unsorted adapters.
- `git diff --check`

## Compatibility

No public type or API-shape change. Ordinary entity dictionaries retain their existing prototype; only previously unusable valid string IDs change behavior.

The repository-wide formatter already reports the touched sorted-adapter source/test files on the exact base commit, before this patch; the new hunks follow the surrounding style. ESLint is currently blocked locally by the installed ESLint 7 dependency tree loading an ESM-only transitive module through CommonJS.

### DIFF
--- a/packages/toolkit/src/entities/sorted_state_adapter.ts
+++ b/packages/toolkit/src/entities/sorted_state_adapter.ts
@@ -13,6 +13,8 @@ import {
   ensureEntitiesArray,
   splitAddedUpdatedEntities,
   getCurrent,
+  hasEntity,
+  setEntity,
 } from './utils'
 
 // Borrowed from Replay
@@ -93,7 +95,7 @@ export function createSortedStateAdapter<T, Id extends EntityId>(
     newEntities: readonly T[] | Record<Id, T>,
     state: R,
   ): void {
-    let deduplicatedEntities = {} as Record<Id, T>;
+    let deduplicatedEntities = Object.create(null) as Record<Id, T>;
     newEntities = ensureEntitiesArray(newEntities)
     if (newEntities.length !== 0) {
       for (const item of newEntities) {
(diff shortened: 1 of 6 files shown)


markerikson:
This is _beyond_ an edge case and not something I see the need for us to handle.

markerikson:
Genuine question: how are you even coming up with these edge cases? _Why_ are you coming up with these edge cases?

I don't detect malicious intent in these PRs, but this is awfully close to spam.  No one has reported anything like this in the 7 years since RTK came out.  At some point it's on the user to not do stupid things.

PR #5402, opened 2026-08-30
Fix stateful RegExp matching in ignored paths

### LINKED ISSUE #5401: Global RegExp ignoredPaths produce stateful middleware results
### Reproduction

```ts
const ignored = /^ignored/g
findNonSerializableValue(
  { ignoredA: new Map(), ignoredB: new Map() },
  "",
  undefined,
  undefined,
  [ignored],
)
// => { keyPath: "ignoredB", value: Map(0) {} }
```

The same stateful `RegExp.test()` pattern is used while tracking and detecting mutations. On current `master` (`74200d87`), mutating a sibling covered by the same global ignore pattern is incorrectly reported. Sticky regexes have the same `lastIndex` behavior.

### Expected

Each path match should be independent of previous paths and should not change the caller’s RegExp state. All siblings matched by a configured ignored-path expression should be ignored consistently.

I have a small shared matcher fix and focused regression coverage for both invariant middleware implementations.

OskarEichler:
## Fixes

- Match every invariant-middleware path independently when `ignoredPaths` contains a global or sticky RegExp.
- Preserve the caller’s original `RegExp.lastIndex` after each match.
- Share the matching behavior between serializability traversal and both mutation-tracking phases.

Closes #5401.

## Verification

- Exact untouched-base reproductions show `/^ignored/g` alternately ignoring sibling paths: serializability reports `ignoredB`, while mutation tracking reports a mutation inside an ignored subtree.
- Added global and sticky RegExp regressions for both invariant middleware implementations, including `lastIndex` preservation.
- `yarn workspace @reduxjs/toolkit test` — 88 files, 1,320 passing tests, two existing todos, zero type errors.
- Fixed standalone serializability/mutation reproductions for both `/g` and `/y`
- `git diff --check`

## Compatibility

No public type or API-shape change. Non-stateful expressions and string paths retain their existing behavior; global/sticky expressions become deterministic and no longer leak matcher state.

### DIFF
--- a/packages/toolkit/src/immutableStateInvariantMiddleware.ts
+++ b/packages/toolkit/src/immutableStateInvariantMiddleware.ts
@@ -1,6 +1,6 @@
 import type { Middleware } from 'redux'
 import type { IgnoredPaths } from './serializableStateInvariantMiddleware'
-import { getTimeMeasureUtils } from './utils'
+import { getTimeMeasureUtils, isIgnoredPath } from './utils'
 
 type EntryProcessor = (key: string, value: any) => any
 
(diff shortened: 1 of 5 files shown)


timdorr:
This is a _really_ out there edge case.

PR #5404, opened 2026-08-30
Preserve falsy rejectWithValue payload semantics

### LINKED ISSUE #5403: Falsy rejectWithValue payloads lose rejectedWithValue semantics
### Reproduction

```ts
const thunk = createAsyncThunk("test", (_, api) => api.rejectWithValue(0))
const promise = store.dispatch(thunk())
const action = await promise

action.payload                    // 0
action.meta.rejectedWithValue     // false
await promise.unwrap()            // throws { message: "Rejected" }, not 0
```

Current `master` (`74200d87`) derives `rejectedWithValue` with `!!payload`, so `false`, `0`, `""`, `null`, and `undefined` all lose the fact that the payload came from `rejectWithValue`. This also changes `unwrapResult` / `.unwrap()` behavior.

### Expected

`meta.rejectedWithValue` should describe how the rejection was created, independent of payload truthiness, and unwrap should throw the exact supplied rejection value. The public type already models this as a discriminant.

I have a focused implementation that distinguishes an omitted payload argument from an explicitly supplied falsy value without changing the public action-creator signature, plus runtime coverage for all five falsy values.

OskarEichler:
## Fixes

- Derive `meta.rejectedWithValue` from whether the rejected action creator received a payload argument, not from payload truthiness.
- Preserve `.unwrap()` / `unwrapResult()` behavior for `false`, `0`, `""`, `null`, and `undefined` rejection payloads.
- Keep ordinary thrown errors and manually created rejections without a payload unchanged.

Closes #5403.

## Verification

- Exact untouched-base reproduction returns `meta.rejectedWithValue: false` and unwraps as `{ message: "Rejected" }` for every falsy payload.
- Added runtime coverage for all five falsy JavaScript values, asserting the rejected-action discriminant and exact unwrapped rejection identity.
- `yarn workspace @reduxjs/toolkit test` — 88 files, 1,321 passing tests, two existing todos, zero type errors.
- Fixed standalone controls for all five falsy values
- `git diff --check`

## Compatibility

No public type, function signature, or action shape change. This is an observable correction for values explicitly supplied through `rejectWithValue`; thrown errors and omitted payload arguments retain existing behavior.

### DIFF
--- a/packages/toolkit/src/createAsyncThunk.ts
+++ b/packages/toolkit/src/createAsyncThunk.ts
@@ -575,27 +575,29 @@ export const createAsyncThunk = /* @__PURE__ */ (() => {
     const rejected: AsyncThunkRejectedActionCreator<ThunkArg, ThunkApiConfig> =
       createAction(
         typePrefix + '/rejected',
-        (
+        function (
           error: Error | null,
           requestId: string,
           arg: ThunkArg,
           payload?: RejectedValue,
           meta?: RejectedMeta,
-        ) => ({
-          payload,
-          error: ((options && options.serializeError) || miniSerializeError)(
-            error || 'Rejected',
-          ) as GetSerializedErrorType<ThunkApiConfig>,
-          meta: {
-            ...((meta as any) || {}),
-            arg,
-            requestId,
-            rejectedWithValue: !!payload,
-            requestStatus: 'rejected' as const,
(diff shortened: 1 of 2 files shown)


timdorr (approved):

markerikson:
Dupe of #5395

PR #5408, opened 2026-08-30
Cancel pending subscription sync on API reset

### LINKED ISSUE #5407: resetApiState leaves the pending subscription sync timer running
### Problem

The batched subscription handler sets `updateSyncTimer = null` when handling `resetApiState`, but does not clear the existing 500ms timeout. The old callback still runs after reset, serializes current internal subscriptions, dispatches `subscriptionsUpdated`, and writes `updateSyncTimer = null` again.

If a new subscription schedules a new timer after reset, the old callback can also clear the handle reference for the newer timer, allowing overlapping sync timers.

### Expected

Reset should cancel the pending timeout before clearing the handle and subscription state, matching the polling/cache cleanup behavior elsewhere in the middleware.

I have a focused fake-timer regression that observes the post-reset internal action and a one-line cleanup fix.

OskarEichler:
## Fixes

- Clear the pending 500ms subscription-state synchronization timeout when `resetApiState` runs.
- Prevent a stale `subscriptionsUpdated` action from being emitted after reset.
- Prevent an old callback from clearing the handle of a newer post-reset sync timer and allowing overlapping timers.

Closes #5407.

## Verification

- Added a downstream middleware observer with fake timers. On the exact base it records one `subscriptionsUpdated` action 500ms after reset; after the fix it records none.
- Focused build-middleware suite: 11 tests pass with zero type errors.
- `yarn workspace @reduxjs/toolkit test` — 88 files, 1,317 passing tests, two existing todos, zero type errors.
- `git diff --check`

## Compatibility

No public type, API, action, or state-shape change. The only removed behavior is the obsolete deferred internal synchronization after an explicit API reset.

### DIFF
--- a/packages/toolkit/src/query/core/buildMiddleware/batchActions.ts
+++ b/packages/toolkit/src/query/core/buildMiddleware/batchActions.ts
@@ -134,6 +134,9 @@ export const buildBatchedActionsHandler: InternalHandlerBuilder<
     if (api.util.resetApiState.match(action)) {
       previousSubscriptions = {}
       internalState.currentSubscriptions.clear()
+      if (updateSyncTimer !== null) {
+        clearTimeout(updateSyncTimer)
+      }
       updateSyncTimer = null
       return [true, false]
     }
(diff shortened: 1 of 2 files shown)


timdorr:
`resetApiState` is something people only really run when logging out or cleaning up in tests. Letting that 500ms sync timer finish out doesn't break anything or mess with the cache, so we don't feel the need to add extra code to cancel it.

PR #5410, opened 2026-08-30
Release abort listeners after retry backoff

### LINKED ISSUE #5409: Completed retry backoffs retain abort listeners
### Problem

`defaultBackoff` adds a one-shot listener to the request signal for every retry delay. When the timer wins, that listener is never removed because `{ once: true }` only cleans it up if the abort event actually fires. A long-lived signal therefore retains one closure per completed backoff until it is eventually aborted or collected.

With repeated retries this unnecessarily retains timeout IDs, reject closures, and their surrounding Promise state.

### Expected

Whichever side settles the delay should clear the timer/listener owned by the other side. Aborting must still reject immediately, and already-aborted signals must retain their existing behavior.

I have a focused retry test that spies on the actual per-request signal before the first error resolves, plus symmetric cleanup in `defaultBackoff`.

OskarEichler:
## Fixes

- Remove a retry delay’s abort listener when its timer completes normally.
- Clear the timer and remove the same listener when abort wins.
- Avoid retaining one timeout/reject closure per completed backoff on long-lived request signals.

Closes #5409.

## Verification

- The regression spies on the real per-request signal before the first failed base-query attempt resolves, identifies the default-backoff listener, and requires the identical callback to be removed after the successful retry. It fails on the exact base and passes after the fix.
- Focused retry suite: 29 tests pass with zero type errors.
- `yarn workspace @reduxjs/toolkit test` — 88 files, 1,317 passing tests, two existing todos, zero type errors.
- `git diff --check`

## Compatibility

No public type, API, retry-count, timing, or error-shape change. Abort still rejects immediately; normal timer completion now releases its obsolete listener.

### DIFF
--- a/packages/toolkit/src/query/retry.ts
+++ b/packages/toolkit/src/query/retry.ts
@@ -33,12 +33,19 @@ async function defaultBackoff(
   const timeout = ~~((Math.random() + 0.4) * (300 << attempts)) // Force a positive int in the case we make this an option
 
   await new Promise<void>((resolve, reject) => {
-    const timeoutId = setTimeout(() => resolve(), timeout)
+    let abortHandler: (() => void) | undefined
+    const timeoutId = setTimeout(() => {
+      if (signal && abortHandler) {
+        signal.removeEventListener('abort', abortHandler)
+      }
+      resolve()
+    }, timeout)
 
     // If signal is provided and gets aborted, clear timeout and reject
     if (signal) {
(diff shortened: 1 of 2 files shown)


timdorr:
The abort signal here is created specifically for that single request. Once the request finishes, the whole signal gets cleaned up by JS anyway. People aren't reusing these signals across multiple requests, so this isn't causing memory leaks in practice.

PR #5414, opened 2026-08-30
Clear listener wait timers after early completion

OskarEichler:
## Fixes

- Clear listenerApi.delay timers immediately when the listener is cancelled.
- Clear take(predicate, timeout) timers when the predicate wins before the timeout.
- Remove the delay abort listener when its timer completes normally.

## Why

The existing races settle correctly, but the losing setTimeout remains registered until its full duration. Repeated cancelled delays or successful timed takes can therefore accumulate live timers and retained closures in long-running applications.

The new timer-count regressions fail on the exact master base: each path leaves one 60-second timer behind. Both report zero pending timers after this change.

## Verification

- Focused listener middleware suite: 51 tests pass, zero type errors.
- Full Toolkit suite: 88 files, 1,318 tests pass, two existing todos, zero type errors.
- Package build passes.

No public API, type, action, or listener semantics change; this only releases resources when a wait is already settled.

### DIFF
--- a/packages/toolkit/src/listenerMiddleware/index.ts
+++ b/packages/toolkit/src/listenerMiddleware/index.ts
@@ -167,10 +167,13 @@ const createTakePattern = <S>(
     })
 
     const promises: (Promise<null> | Promise<[Action, S, S]>)[] = [tuplePromise]
+    let timeoutId: ReturnType<typeof setTimeout> | undefined
 
     if (timeout != null) {
       promises.push(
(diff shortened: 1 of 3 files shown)


timdorr:
If `take` finds what it's looking for before the timeout hits, the code moves on right away. The leftover timer just fires and does nothing. We'd rather keep the listener code simple than add extra tracking just to clear a timer that's already harmless.

PR #5416, opened 2026-08-30
Release fetch timeout resources after completion

OskarEichler:
## Fixes

- Clear fetchBaseQuery timeout timers as soon as the request and response handling settle.
- Remove listeners from every source signal when the combined signal settles or is disposed.
- Apply cleanup through finally so success, handled errors, parsing failures, and unexpected throws all release resources.

## Why

A successful request with a configured timeout currently retains its timer, timeout controller, combined controller, and source-signal listeners until the entire timeout duration elapses. Repeated fast requests with long timeouts can accumulate those resources in long-running apps.

The exact built master baseline completes an immediate JSON request with a 60-second timeout while retaining one pending timer. The fixed build retains zero. The permanent test also verifies the API abort listener is removed.

## Verification

- Focused fetchBaseQuery suite: 65 tests pass, zero type errors.
- Package build passes.
- Existing global and per-request timeout-error coverage remains green.

No public API, type, timeout duration, error shape, response handling, or abort semantics change.

### DIFF
--- a/packages/toolkit/src/query/fetchBaseQuery.ts
+++ b/packages/toolkit/src/query/fetchBaseQuery.ts
@@ -234,133 +234,140 @@ export function fetchBaseQuery({
       ...rest
     } = typeof arg == 'string' ? { url: arg } : arg
 
+    const timeoutSignalResult = timeout ? timeoutSignal(timeout) : undefined
+    const combinedSignal = timeoutSignalResult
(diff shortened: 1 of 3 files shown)


timdorr:
This adds a fair amount of extra code and complexity to `fetchBaseQuery` just to manually clean up timers and signals that the browser already garbage-collects as soon as the request is done. We try to keep `fetchBaseQuery` as lean as possible, so we're going to pass on this.

PR #5417, opened 2026-08-30
Release external abort listeners after thunks settle

OskarEichler:
## Fix

Store the handler registered on a caller-provided AbortSignal and remove it when the async thunk settles.

## Why

The internal abort listener is already removed in finally, but the matching external listener is anonymous and remains attached until that external signal eventually aborts. Reusing a long-lived signal across completed thunks therefore accumulates retained listeners and thunk closures.

The focused regression spies on the actual external signal. Exact master registers its abort handler and performs zero removals after successful completion; this change removes the same handler. Already-aborted and in-flight abort behavior remain covered.

## Verification

- Focused createAsyncThunk suite: 44 tests pass, zero type errors.
- Package build passes.

No public API, type, action, result, abort reason, or timing semantics change.

### DIFF
--- a/packages/toolkit/src/createAsyncThunk.ts
+++ b/packages/toolkit/src/createAsyncThunk.ts
@@ -609,6 +609,7 @@ export const createAsyncThunk = /* @__PURE__ */ (() => {
 
         const abortController = new AbortController()
         let abortHandler: (() => void) | undefined
+        let externalAbortHandler: (() => void) | undefined
         let abortReason: string | undefined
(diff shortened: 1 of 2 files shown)


timdorr:
When people pass a custom abort signal to a thunk, it's almost always tied to a component that gets unmounted anyway. We haven't seen issues with signals being kept alive forever, so adding more listener tracking code to `createAsyncThunk` isn't something we want to take on right now.

PR #5418, opened 2026-08-30
Clear cache-removal timers on resubscription

OskarEichler:
## Fix

Cancel and remove a pending cache-removal timeout as soon as a subscription for that cache key becomes active again.

## Why

The timeout callback already rechecks subscriptions, so data is not removed incorrectly. However, the timer and its captured cache/API state remain alive for the entire keepUnusedDataFor duration after resubscription. With long retention periods and repeated subscribe/unsubscribe cycles, obsolete timers can accumulate unnecessarily.

The focused regression settles a query, unsubscribes, drains the subscription-sync timer, and observes one cache-removal timer. After resubscribing and draining the next sync timer, exact master still retains that timer; this change retains none.

## Verification

- Focused cache-collection suite: 9 tests pass, zero type errors.
- Package build passes.

No public API, type, cache retention duration, subscription, query, or eviction behavior changes.


timdorr:
This doesn't actually fix anything for users.

PR #5419, opened 2026-08-30
Keep duplicate listener cleanup from resetting setup state

OskarEichler:
## Fix

Make the cleanup returned by a duplicate default setupListeners call a true no-op.

## Why

Only the first default setup installs window listeners. Today, a second call installs nothing but returns a cleanup that sets the module-level initialized guard to false. Calling that cleanup makes a later setup install a duplicate set while the original handlers are still active, causing duplicate dispatches and mismatched teardown.

The focused regression performs first, duplicate, and third setup calls. After duplicate cleanup, exact master records eight window-listener registrations; this change remains at the original four and removes nothing until the real owner cleans up.

## Verification

- Focused setupListeners regression passes with zero type errors.
- Full Toolkit suite: 89 files, 1,317 tests pass, two existing todos, zero type errors.
- Package build passes.

No public API, type, event set, dispatch, custom-handler, or owner-cleanup semantics change.

### DIFF
--- a/packages/toolkit/src/query/core/setupListeners.ts
+++ b/packages/toolkit/src/query/core/setupListeners.ts
@@ -78,9 +78,7 @@ export function setupListeners(
       }
     }
 
-    let unsubscribe = () => {
-      initialized = false
-    }
+    let unsubscribe = () => {}
 
     if (!initialized) {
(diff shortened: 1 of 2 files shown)


timdorr:
This requires you to do some things that aren't realistic.

PR #5420, opened 2026-08-30
Support empty-string fixed mutation cache keys

OskarEichler:
## Fixes

- Treat an empty string as a present fixedCacheKey when tracking running mutations.
- Preserve empty-key mutation state across hook unmounts.
- Allow useMutation().reset() to remove shared state stored under an empty key.

## Why

fixedCacheKey is typed as string and mutation state/cache selection already uses nullish fallback semantics, but three runtime branches use truthiness. This makes the valid empty-string key behave inconsistently: shared data can be selected, yet reset does nothing, unmount removes it as if unkeyed, and getRunningMutationThunk cannot find it.

On exact master, the focused hook regression fails to reset the two shared empty-key results, and the running-mutation lookup returns undefined. Both pass with nullish presence checks; the hook case also verifies persistence across unmount/remount.

## Verification

- Focused hook/core suites: 20 tests pass, zero type errors.
- Full Toolkit suite: 88 files, 1,318 tests pass, two existing todos, zero type errors.
- Package build passes.

This is an observable bug fix for a previously inconsistent valid string value. No public type, API signature, action shape, or non-empty/undefined key behavior changes.

### DIFF
--- a/packages/toolkit/src/query/core/buildInitiate.ts
+++ b/packages/toolkit/src/query/core/buildInitiate.ts
@@ -580,7 +580,7 @@ You must add the middleware for RTK-Query to function correctly!`,
         ret.then(() => {
           runningMutations.delete(requestId)
         })
-        if (fixedCacheKey) {
+        if (fixedCacheKey != null) {
           runningMutations.set(fixedCacheKey, ret)
           ret.then(() => {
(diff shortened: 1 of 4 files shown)


markerikson:
I don't think we intend to support empty strings as valid cache keys.

PR #5422, opened 2026-08-30
Reset lazy queries with empty cache keys

OskarEichler:
## Summary

- treat an empty string as a present lazy-query cache key
- dispatch `removeQueryResult` when `serializeQueryArgs` intentionally returns `""`
- cover the public `reset` behavior with an empty custom cache key

## Why

`serializeQueryArgs` may return any string, including an empty string. The lazy-query `reset` callback currently tests `queryCacheKey` by truthiness, so the cache entry remains and the hook stays initialized for that valid key. Checking for `undefined` matches the actual optionality of the value.

## Verification

- `corepack yarn workspace @reduxjs/toolkit test buildHooks.test.tsx` (70 tests)
- `corepack yarn workspace @reduxjs/toolkit test` (88 files, 1,317 passed, 2 todo)
- `corepack yarn workspace @reduxjs/toolkit build`
- `corepack yarn workspace @reduxjs/toolkit type-tests`
- Prettier on changed files
- `git diff --check`

### DIFF
--- a/packages/toolkit/src/query/react/buildHooks.ts
+++ b/packages/toolkit/src/query/react/buildHooks.ts
@@ -2007,10 +2007,10 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
       )
 
       const reset = useCallback(() => {
-        if (promiseRef.current?.queryCacheKey) {
+        if (promiseRef.current?.queryCacheKey !== undefined) {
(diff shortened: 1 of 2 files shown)


markerikson:
I don't think we intend to support empty strings as valid cache keys.

PR #5423, opened 2026-08-30
Use native Vite tsconfig path resolution

OskarEichler:
## Summary

- replace `vite-tsconfig-paths` with Vite 8's native `resolve.tsconfigPaths` option
- update both Toolkit and OpenAPI codegen Vitest configs
- remove the two direct plugin dependencies and their now-unused lockfile graph

## Why

Vite 8 detects `vite-tsconfig-paths` and emits a migration warning on every test run because this behavior is now built in. Using the native option removes the extra plugin/configuration and drops `vite-tsconfig-paths` plus `tsconfck` from the dependency graph.

## Verification

- Toolkit test suite: 88 files, 1,316 passed, 2 todo, no type errors
- OpenAPI codegen test suite: 3 files, 100 passed, no type errors
- Toolkit build and declaration/type tests
- OpenAPI codegen build
- changed-file Prettier
- `git diff --check`

The prior Vite migration warning is absent from both test runs.

### DIFF
--- a/packages/rtk-query-codegen-openapi/package.json
+++ b/packages/rtk-query-codegen-openapi/package.json
@@ -74,7 +74,6 @@
     "rimraf": "^6.1.3",
     "ts-node": "^10.9.2",
     "tsup": "^8.4.0",
-    "vite-tsconfig-paths": "^5.0.1",
     "vitest": "^4",
     "yalc": "^1.0.0-pre.47"
   },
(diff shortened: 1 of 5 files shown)


timdorr:
We actually need to close this one because of our set of modern tooling PRs that's incoming. Those currently downgrade Vite and will clobber this change anyways. We can incorporate this change into those PRs.

PR #5428, opened 2026-08-31
fix(rtk-query): only treat a url as absolute when it starts with a scheme

MahinAnowar:
Found by probing RTK Query's url helpers rather than from a bug report, so there's no issue to link.

`isAbsoluteUrl` tests `(^|:)//` unanchored, so it answers "does this string contain `://` anywhere" rather than "does this url start with a scheme". A relative url that carries an absolute one in its query string matches, and `joinUrls` then returns it untouched — silently dropping `baseUrl`:

```js
joinUrls('https://api.example.com/', 'redirect?target=https://other.example')
// 'redirect?target=https://other.example'   <- baseUrl is gone
```

Same for anything else that embeds a url as a parameter, which isn't unusual: `img?src=http://cdn.example/a.png`, `login?next=https://...`, callback and return-to params. The request ends up resolved against the page instead of the API, and because nothing throws it just looks like a 404 from the wrong host.

The fix is to anchor the check and allow an optional scheme:

```js
/^([a-z][a-z\d+\-.]*:)?\/\//i
```

That keeps every case the existing `joinUrls` table already covers — `https://example.org` and the protocol-relative `//example.org` still count as absolute, `/banana` and `banana` still don't — and it also accepts custom schemes like `web+custom://`. Incidentally it's the same expression `msw` uses for its own `isAbsoluteUrl`, which is already in this repo's dev dependencies.

Worth noting the old behaviour isn't a deliberate loose check as far as I can tell: the comment above the function says "if either :// or // is present", which describes what the regex does rather than what an absolute url is, and there were no tests on `isAbsoluteUrl` itself — only the indirect coverage through `joinUrls`.

## Testing

Added a `describe('isAbsoluteUrl')` table covering schemes, protocol-relative, custom schemes, plain relative urls, and the embedded-url cases, plus two rows on the existing `joinUrls` table. Reverting only `isAbsoluteUrl.ts` fails exactly the five new assertions and leaves the 34 existing ones passing:

```
× https://example.com/api/ and redirect?target=https://example.org join to ...
× /api/ and img?src=http://cdn.example/a.png join to ...
× redirect?target=https://example.org is absolute: false
× img?src=http://cdn.example/a.png is absolute: false
× search?q=a://b is absolute: false
  Tests  5 failed | 34 passed (39)
```

With the fix, 39/39 pass, and the whole `src/query` suite is green at 31 files / 552 tests. Prettier clean.

### DIFF
--- a/packages/toolkit/src/query/tests/utils.test.ts
+++ b/packages/toolkit/src/query/tests/utils.test.ts
@@ -1,5 +1,10 @@
 import { vi } from 'vitest'
-import { isOnline, isDocumentVisible, joinUrls } from '@internal/query/utils'
+import {
+  isOnline,
+  isDocumentVisible,
+  joinUrls,
+  isAbsoluteUrl,
+} from '@internal/query/utils'
 
 afterAll(() => {
   vi.restoreAllMocks()
@@ -87,7 +92,39 @@ describe('joinUrls', () => {
     ['https://example.com/api/', '/banana', 'https://example.com/api/banana'],
     ['https://example.com/api/', 'https://example.org', 'https://example.org'],
     ['https://example.com/api/', '//example.org', '//example.org'],
+    // a relative url may carry an absolute one in its query string; the base still applies
+    [
+      'https://example.com/api/',
+      'redirect?target=https://example.org',
+      'https://example.com/api/redirect?target=https://example.org',
+    ],
+    [
+      '/api/',
+      'img?src=http://cdn.example/a.png',
+      '/api/img?src=http://cdn.example/a.png',
+    ],
   ])('%s and %s join to %s', (base, url, expected) => {
     expect(joinUrls(base, url)).toBe(expected)
+  })
+})
+
+describe('isAbsoluteUrl', () => {
+  test.each([
+    ['https://example.com', true],
+    ['http://example.com', true],
+    ['ftp://example.com', true],
+    ['web+custom://example.com', true],
+    ['//example.com', true],
+    ['banana', false],
+    ['/banana', false],
+    ['./banana', false],
+    ['banana/split', false],
+    ['?a=1', false],
+    // relative urls that merely contain an absolute one
+    ['redirect?target=https://example.org', false],
+    ['img?src=http://cdn.example/a.png', false],
+    ['search?q=a://b', false],
+  ])('%s is absolute: %s', (url, expected) => {
+    expect(isAbsoluteUrl(url)).toBe(expected)
   })
 })
--- a/packages/toolkit/src/query/utils/isAbsoluteUrl.ts
+++ b/packages/toolkit/src/query/utils/isAbsoluteUrl.ts
@@ -1,9 +1,14 @@
 /**
- * If either :// or // is present consider it to be an absolute url
+ * Determines whether a url is absolute: it either starts with a scheme
+ * (`https://example.com`) or is protocol-relative (`//example.com`).
+ *
+ * The test is anchored to the start of the string on purpose. A relative url can
+ * legitimately carry an absolute one inside its query string, as in
+ * `redirect?target=https://example.com`, and treating that as absolute would make
+ * `joinUrls` drop the `baseUrl`.
  *
  * @param url string
  */
-
 export function isAbsoluteUrl(url: string) {
-  return new RegExp(`(^|:)//`).test(url)
+  return /^([a-z][a-z\d+\-.]*:)?\/\//i.test(url)
 }


phryneas:
Wouldn't you need to urlescape that payload url? Both `:` and `/` are RFC3986 reserved characters.

The usage of `joinUrls` with including query parameters also looks like incorrect code - that's what you should use `params` for, not `url`.

phryneas:
As `joinUrls` and `isAbsoluteUrl` are internal helpers, this is only reachable if you use `baseFetchQuery` and 

* ignore `params`
* ignore `paramsSerializer`
* build your own querystring by hand

And while that is possible, at that point you are really responsible to correctly escape that querystring.

I'm not sure if we should add more bundle size for this.

MahinAnowar:
Fair on the escaping — you're right that the default path is safe, `new URLSearchParams` percent-encodes both characters.

But I don't think you need to ignore `params` to reach this. They get serialized and appended to `url` at fetchBaseQuery.ts:287-293, and `joinUrls` runs on the result at line 295, so both routes end up in the same call. And building the querystring by hand isn't really off the map either — it's option 2 under "Setting the query string" in the fetchBaseQuery docs, listed as a supported thing to do, with no note about escaping.

The other bit is that a raw `:` and `//` inside a query aren't malformed. RFC 3986 has `query = *( pchar / "/" / "?" )` and pchar includes `:`, so `?next=https://example.com` is a valid URI and `new URL()` parses it fine. Reserved doesn't mean forbidden everywhere, only that the character is protected where it acts as a delimiter. So a server handing back a link shaped like that isn't doing anything wrong, and the caller has no particular reason to think they need to encode it.

On size, it's 18 bytes minified and 16 after gzip. If that's the sticking point I can get it to 7 by just anchoring what's already there rather than spelling out the scheme grammar:

```js
return /^([^/?#]*:)?\/\//.test(url)
```

Same results on everything I threw at it, `chrome-extension://` included, since `[^/?#]*` can't run past a `?` or a `/`. Happy to switch to that, or to close this if you'd still rather not spend anything on it.

phryneas:
>  it's option 2 under "Setting the query string" in the fetchBaseQuery docs

Correct, and if read in context that documentation section describes an escape hatch if you want to do query string escaping by hand and not with the provided methods. Which means it becomes your responsibility, and if you fail to do so correctly, unexpected things might happen.

MahinAnowar:
That's fair, and I'll drop it if you're still not sold.

The one thing I'd push on is "if you fail to do so correctly" — there isn't an escaping that avoids this. Leaving `:` and `/` raw in a query is already correct under RFC 3986, so someone who escapes properly can still end up with `next?to=https://example.com` and silently lose the `baseUrl`. And the value doesn't have to be hand-built at all — it can come straight off a server response as a `next` link, with no encoding step for anyone to skip.

For what it's worth the anchored form is what the neighbours use. vite's copy of `@jridgewell/trace-mapping` is sitting in this repo's `node_modules` right now with `const schemeRegex = /^[\w+.-]+:\/\//` and a separate `startsWith("//")` for the protocol-relative case.

But it's your call on the bytes, and you know the tradeoffs here better than I do. Say the word and I'll close it.

phryneas:
I guess I'll leave that call to another maintainer.
 
At this point I'm far too allergic to PRs that fix issues that haven't been reported as "actually happening" by any of our users in the last 5 years - we get way too many of these theoretical issue fixes recently.

markerikson:
I absolutely agree with Lenz that we're getting a lot of PRs for bizarre edge cases that not a single person has ever reported, and it's getting on my nerves.

Can I ask how and why you even came up with this edge case?  Why were you looking at this in the first place?

I'll merge this one on the grounds that the change is tiny and comes with tests, but generally at this point I'm leaning towards "if it's not reported from in the wild we reserve the right to close it".

MahinAnowar:
Straight answer: it didn't come from something I hit in an app.

I go looking for one specific defect family across libraries — a predicate that checks whether a string *contains* something when what it means is *starts with*. Anything testing for `//` without a `^` is a candidate, which is how `isAbsoluteUrl` surfaced, and I traced it to the source and wrote the repro before believing it. I use AI tooling to help run those sweeps and I review and verify everything myself before it turns into a PR. Same pass looked at a few other RTK internals and dropped them, because they turned out to be deliberate.

Your point stands, though. A fix nobody asked for still costs you a review, and "technically wrong" isn't the same thing as "worth your time" — that's a distinction I should be making before I open something, not leaving to you. If the bar here is closer to "reported in the wild, or don't bother", that's genuinely useful to know and I'll work to it rather than guess.

Appreciate you taking this one regardless.

